PR #89, opened 2021-06-20
Check for Xorg running in tty

### LINKED ISSUE #87: Unsupported display server type tty when running without display manager
Hello! I'm starting my window manager with `startx` instead of with a display manager, and the display server type is incorrectly identified as `tty` instead of `x11`. I took a look at the method being used to detect the display server and the issue isn't in the implementation of the method you're using to check it, but with the result of the `loginctl` command. One alternative method might be checking the `WAYLAND_DISPLAY` and `DISPLAY` environment variables, as suggested here: https://unix.stackexchange.com/a/559950

I haven't tested it much at all, here's a mockup of what that might look like (in `src/platform/linux.rs`), I haven't been able to test it though cause it won't build on my system for other reasons even without these modifications.

```rust
pub fn get_display_server() -> String {
    match std::env::var("WAYLAND_DISPLAY") {
        Some(display) => "wayland",
        _ => match std::env::var("DISPLAY") {
            Some(display) => "x11",
            _ => "unknown",
        },
    }
}
```

Let me know if there's any other information I can provide, or testing I can do!

mtoohey31:
Closes #87

### DIFF
--- a/src/platform/linux.rs
+++ b/src/platform/linux.rs
@@ -353,11 +353,44 @@ fn get_display_server_of_session(session: &str) -> String {
     if let Ok(output) = std::process::Command::new("loginctl")
         .args(vec!["show-session", "-p", "Type", session])
         .output()
+    // Check session type of the session
     {
-        String::from_utf8_lossy(&output.stdout)
+        let display_server = String::from_utf8_lossy(&output.stdout)
             .replace("Type=", "")
             .trim_end()
-            .into()
+            .into();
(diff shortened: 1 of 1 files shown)


zzyandzzy:
in my gentoo linux, the 374 line replace
```rust
run_cmds(format!("ps -e | grep \"{}.\\\\+X\"", tty))
```
then run correct.

rustdesk:
How about if Xabc?

PR #97, opened 2021-06-21
Creating security guidelines, also known as SECURITY.md

vmfunc:
Hey, 
I was checking the security tab of the repo and saw that there were no guidelines related to security issues, so I started creating the `SECURITY.md` file. 
iirc, it will be shown on the security tab of the repo.
How about people (most importantly @rustdesk if you're not busy) help me finishing it and writing the guidelines ? I don't know what kind of guidelines are needed, but having them is really a good idea in case of really important security issues that are discovered.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,13 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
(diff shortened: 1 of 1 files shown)


rustdesk:
Great

PR #137, opened 2021-07-02
German README and correction of links in other MDs ("Deutsche" to "Deutsch")

ToasterUwU:
I was bored and browsed through the "Trending" section of GitHub and then I saw this project.

I love Rust, I love seeing cool stuff made with Rust, you didn't have a german README... So I thought, let's do it.

I tried to stick to the original wording, but I rephrased it if it sounded weird in German. I also changed the "We need your help to translate this README to your native language" to something like "We need your help to make this README better and keep it up to date".

The only "mature" thing I changed was the official server list. I converted it into a table since there is no reason not to and it looks better and is easier to read. ( If wanted I will do the same to the others, so it's the same on all of them )

Hope this helps someone out there and/or makes someone happy. 

Greetings from Germany.

### DIFF
--- a/README-DE.md
+++ b/README-DE.md
@@ -1,0 +1,148 @@
+<p align="center">
+  <img src="logo-header.svg" alt="RustDesk - Your remote desktop"><br>
+  <a href="#kostenlose-öffentliche-server">Server</a> •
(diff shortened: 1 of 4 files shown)


rustdesk:
Thanks

PR #170, opened 2021-07-15
Replace InnoSetup with improved NSIS Setup

### LINKED ISSUE #100: Windows Installer
Hi, it would be great to have an installer for silent installs like Teamviewer and [Anydesk](https://download.anydesk.com/AnyDesk.msi). I would like to create a [Chocolatey](http://chocolatey.org/) package of this but there doesn't seem to have an installer.

sitiom:
Thanks for putting the InnoSetup file. However, I have noticed a couple of problems with it:
- InnoSetup creates its own uninstaller and registry key. Rustdesk already has an uninstall parameter (`--uninstall`) and the InnoSetup uninstaller would just cause inconsistencies. The NSIS Setup uses the `--uninstall` parameter instead and installs in the correct reg key.
- Cannot overwrite RustDesk.exe and does not close process before installing (Fixed in NSIS)
- [Can be run in 32-bit](https://stackoverflow.com/questions/54335154/inno-setup-gets-launched-in-32-bit-even-after-specifying-architecturesinstallin6) (NSIS can still be run in 32-bit stops with an error message)

Here is the [setup file](https://github.com/rustdesk/rustdesk/files/6823235/rustdesk-1.1.6-setup.zip) for testing. If you want to compile it yourself, you need to have NSIS of course.

Closes #100.


rustdesk:
[UninstallRun]? are you sure we can uninstall from control penel and it calls --uninstall?

sitiom:
> are you sure we can uninstall from control penel and it calls --uninstall?

Yeah, check the NSIS code and try the setup above.

rustdesk:
thanks

PR #239, opened 2021-09-26
User archive resources and bundled DLLs for Win/Mac/Lin

TopsDesk:
This change implements archived/compressed resources in the executable. Changes are in man.rs and ui.rs. I have included the Win/DLL, Lin/so and Mac/dynlib in this branch. Have created a separate res directory and a bat file makerc.bat to create the compressed archive of html, tis, css resources. On windows a DLL is include_bytes to the executable and on linux, a .so file is merged to the executable. Upon execution, the DLL or .so file is written to the same path at the time of execution. If user chooses --install then the DLL or .SO is copied to the target locations. I still need to build on MacOS. I will update the code as needed for macOS


rustdesk:
I have check your changes, you have done great job. But you have too many your opinion biased changes, even including EOL.

PR #240, opened 2021-09-26
Target inline, use archived .html, .css and .tis files else use it from src dir

TopsDesk:
use the following command to create a compressed archive and use it with feature "inline" 
> packfolder ui resources.rc -i "*.html;*.css;*.tis" -v "resources" -binary

When feature inline is not defined then use resource files from "src" directory.  this helps in testing the tis, css and html files.


src/ui.rs
```diff
@@ -30,9 +28,6 @@ struct UI(
 );
 
 fn get_msgbox() -> String {
-    #[cfg(feature = "inline")]
```
rustdesk, line 33: Why do you remove inline?

src/ui.rs
```diff
@@ -1,6 +1,4 @@
 mod cm;
-#[cfg(feature = "inline")]
```
rustdesk, line 2: As we discussed, please do not modify inline feature.

Cargo.toml
```diff
@@ -106,7 +106,8 @@ osx_minimum_system_version = "10.14"
 #!!! rembember call "strip target/release/rustdesk"
 # which reduce binary size a lot
 [profile.release]
```
rustdesk, line 108: Please do not modify this, it is not necessary.

Cargo.toml
```diff
@@ -54,7 +54,7 @@ arboard = "2.0"
 
 [target.'cfg(target_os = "windows")'.dependencies]
 #systray = { git = "https://github.com/open-trade/systray-rs" }
-winapi = { version = "0.3", features = ["winuser"] }
+winapi = { version = "0.3", features = ["wincon","winuser"] }
```
rustdesk, line 57: comply with format please, there should be empty space.

src/ui.rs
```diff
@@ -64,6 +59,14 @@ pub fn start(args: &mut [String]) {
         ALLOW_FILE_IO as u8 | ALLOW_SOCKET_IO as u8 | ALLOW_EVAL as u8 | ALLOW_SYSINFO as u8
     )));
     let mut frame = sciter::WindowBuilder::main_window().create();
+    #[cfg(feature = "inline")]
```
rustdesk, line 129: As we discussed, do not modify inline feature, you should define another feature. Your inline seems better than mine, but my inline has the advantage which we can modify html dynamically.

src/ui.rs
```diff
@@ -115,19 +118,10 @@ pub fn start(args: &mut [String]) {
         log::error!("Wrong command: {:?}", args);
         return;
     }
+    /* fskhan 09-20-2021 with inline feature load from archive or load from directory */
     #[cfg(feature = "inline")]
```
rustdesk, line 189: Again the same problem. And you seems has format issue (indent). Please use cargo format.

src/platform/windows.rs
```diff
@@ -782,7 +782,13 @@ pub fn install_me(options: &str) -> ResultType<()> {
     if versions.len() > 2 {
         version_build = versions[2];
     }
-
```
rustdesk, line 785: format

src/ui.rs
```diff
@@ -11,6 +11,20 @@ use hbb_common::{
     tokio::{self, time},
 };
 use sciter::Value;
+
```
rustdesk, line 14: style problem, please use style like this "used std::{fs, path::...}", we need to keep the same style.

src/ui.rs
```diff
@@ -31,10 +45,63 @@ fn get_msgbox() -> String {
...
+#[cfg(target_os = "linux")]
+const DLL_FILE: &'static [u8] = include_bytes!("libsciter-gtk.so");
+#[cfg(target_os = "macos")]
+const DLL_FILE: &'static [u8] = include_bytes!("libsciter.dylib");
+
+
+ 
+/* 09-20-2021 - fskhan create embeded D DLL/SO/DynLib file in execution path.  
+   include desired version of the sciter.dll/so/dynlib file to eliminate incompatibility 
+*/
```
rustdesk, line 60: style

src/ui.rs
```diff
@@ -31,10 +45,63 @@ fn get_msgbox() -> String {
...
     // https://github.com/c-smile/sciter-sdk/blob/master/include/sciter-x-types.h
     // https://github.com/rustdesk/rustdesk/issues/132#issuecomment-886069737
-    #[cfg(windows)]
+
+
+/* 09-20-2021 - fskhan create DLL/SO/DynLib in execution path end */
+
+    let path = create_dll_target().expect("Couldn't");
+	
+	if !path.exists()
```
rustdesk, line 84: style "{"

src/ui.rs
```diff
@@ -31,10 +45,63 @@ fn get_msgbox() -> String {
...
+/* 09-20-2021 - fskhan create DLL/SO/DynLib in execution path */
+
 pub fn start(args: &mut [String]) {
     // https://github.com/c-smile/sciter-sdk/blob/master/include/sciter-x-types.h
     // https://github.com/rustdesk/rustdesk/issues/132#issuecomment-886069737
-    #[cfg(windows)]
+
+
+/* 09-20-2021 - fskhan create DLL/SO/DynLib in execution path end */
+
```
rustdesk, line 81: Ident

src/ui.rs
rustdesk: Writing dll every time looks very ugly.

src/ui.rs
TopsDesk: Check the current version. "Inline" has a conflict with the function, this code provides. So "inline" is oveer-ridden in 5 places. All other "inline" are still there.

PR #256, opened 2021-10-17
Transfer multiple files

### LINKED ISSUE #244: Transfer file - send multiple files
Hello !

I have tried with the shift key but i can't send more than one file. Is this possible to send multiple files ?
Thx

starccy:
To resolve #244.      

Select multiple files by clicking with `Ctrl` or `Shift`, then send/receive or delete them.     

### Attention
Multiple delete functions can not easily be implemented by looping, I have to change and add some ipcs, so it won't work on the lower version. For compatibility, a single selected file still can be deleted as usual, however.        

### Test Case
I have tested some cases manually to make sure it works as usual:           

| Description | Expected | Real |
| ------| ------| ------|
| Select a single file/folder, send/receive/delete it. | File be send/receive/deleted both in the list and disk. | Same as expected [ ✅ ] |
| Select some files/folders, send/receive/delete them. | A message box popups to show how many files will be deleted. Skip or confirm to delete. They should be handled with corresponding actions | Same as expected [ ✅ ] |
| Select some files/folders, **delete them on controlled end with lower version app**. | A job inserted into the job table, but it will be "waiting" forever. | Same as expected [ ✅ ] |
| Select a single file/folder, **delete them on controlled end with lower version app**.  | File will be deleted as usual. | Same as expected [ ✅ ] |

### DIFF
--- a/src/ui/common.tis
+++ b/src/ui/common.tis
@@ -257,6 +257,10 @@ handler.msgbox = function(type, title, text, callback=null, height=180, width=50
     self.timer(150ms, function() { msgbox(type, title, text, callback, height, width, retry); });
 }
 
+handler.block_msgbox = function(type, title, text, callback=null, height=180, width=500, retry=0) {
+    msgbox(type, title, text, callback, height, width, retry);
+}
+
 var reconnectTimeout = 1;
 handler.msgbox_retry = function(type, title, text, hasRetry, callback=null, height=180, width=500) {
     handler.msgbox(type, title, text, callback, height, width, hasRetry ? reconnectTimeout : 0);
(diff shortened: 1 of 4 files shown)


rustdesk:
Thank you so much

rustdesk:
Excellent job, have you ever thought about creating multiple jobs to solve this issue? It will be much less code change.

starccy:
Yes I tried. But I found only the first job would show the confirm box, the other jobs won't be executed. Then I didn't dig more deeper...

rustdesk:
I will dig on Tuesday and revert back to you.

PR #304, opened 2021-11-21
resolution changed, no need to reconnect

fufesou:
Images got broken if the connected peer changed it's resolution of current display.


rustdesk:
Thanks

PR #369, opened 2022-01-13
New launch daemon/agents and config sync service on macOS

chenbaiyu0414:
This pr includes new launch daemon/agents service and config sync service which copy configs between root and user domain, both two services aims to support better pre-login/after-login experience on macOS.


avaidhyajc:
@chenbaiyu0414  I am interested to know how RustDesk achieves screen-share in mac os .Can you please help me out with the following detail

- How does RustDesk detect login screen vs user context Desktop ? Does it have different services to capture them.

Request your inputs on this as i''m quiet new to Rust and find this PR challenging to understand. TIA

PR #408, opened 2022-02-19
Fix french translation

marie-bnl:
There are some typos or minor problems like translations that feel a bit weird in the french translation, so here's a fix


rustdesk:
Thank you so much.

PR #435, opened 2022-03-22
use vcpkg to find package

gengteng:
I think this issue is closed without being properly resolved, magnum-opus and RustDesk still fail to compile on my Mac(M1)：

* https://github.com/rustdesk/rustdesk/issues/165

This PR use [vcpkg](https://crates.io/crates/vcpkg) to find package.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3574,6 +3574,7 @@ dependencies = [
  "serde 1.0.136",
  "target_build_utils",
  "tracing",
+ "vcpkg",
  "webm",
  "winapi 0.3.9",
 ]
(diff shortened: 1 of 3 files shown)


rustdesk:
Have you ever tested it on Windows and Linux? I used vcpkg crate before, but it was very buggy.

gengteng:
> Have you ever tested it on Windows and Linux? I used vcpkg crate before, but it was very buggy.

Yes, I tested it on my Mac and Win10 manually, and used [CI](https://github.com/gengteng/rustdesk/actions/runs/2021087190) to test it on x86_64-unknown-linux-gnu.

rustdesk:
Thanks, I tested also

PR #441, opened 2022-04-05
Translate to Portuguese (Brazil)

mateusauler:
Translate UI and README to Portuguese (Brazil).

I wasn't able to test the translation by running the program, since I couldn't easily find a way to change its language. (Would be a nice feature to add.)


rustdesk:
Thx

PR #456, opened 2022-04-22
Optimize appbar for macos

xxrlzzz:
Add about/settings option for appbar on macos


rustdesk:
You screwed up this style.

image

and the menu should be only for running rustdesk without arguments, e.g. connection window should not have this menu.

rustdesk:
Thanks!

xxrlzzz:
Thanks for your reviewing, I have fixed the UI issue and made additional appbar items available only on main window.

PR #469, opened 2022-04-28
opt: more precise control method for transfering files

Kingtous:
It's necessary to skip identical files when transfering files. It will improve performance and user experience when using RustDesk. This PR brings features below:

- skip identical files (modified time and file size are identical in both remote and local fs).
- can make a choice by users when encountering an overwrite behaviors(same file name but not identical) during uploading/downloading files or folders. (can only be triggered when both running the specific version and above).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3586,7 +3586,7 @@ dependencies = [
 
 [[package]]
 name = "rustdesk"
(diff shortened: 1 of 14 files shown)


Kingtous:
the pr is combined to #510

PR #510, opened 2022-05-13
opt: file transfer can resume when encountering reconnecting or close&open window

Kingtous:
This PR brings more flexible network requirements when transfering. Improvements are listed below:

- Supports resuming file transfer by resending send request.
- Supports resuming folder transfer by sending offset of the uncompleted index to the remote.

Time to trigger the logic of resuming file-transfer progress:
- when reopen file transfer window if transfer lists was not fully complete in the prior file-transfer progress and the tranfer window was closed.
- when reconnection behavior is triggered (poor network, etc.) and successfully reconnect to the remote server.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3954,7 +3954,7 @@ dependencies = [
 
 [[package]]
 name = "rustdesk"
-version = "1.1.9"
+version = "1.2.0"
 dependencies = [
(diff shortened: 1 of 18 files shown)


Kingtous:
Note that this pr is developed based on #469. It contains both features.

PR #574, opened 2022-05-20
win7 shcore lib using fix

yukitteru:
To be honest, I don’t understand at all what this block of code should do:

```
#[cfg(all(windows, not(feature = "inline")))]
    unsafe {
        winapi::um::shellscalingapi::SetProcessDpiAwareness(2);
    }
    
```

if the dpi awareness setting is done in [manifest](https://github.com/rustdesk/rustdesk/blob/master/manifest.xml) Therefore, in the end, I simply deleted this block of code, as expected, this did not affect the performance of the application and its appearance on Win7, 8.1, 10. (When connecting to a remote desktop, too)

### DIFF
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -61,10 +61,7 @@ pub fn start(args: &mut [String]) {
     allow_err!(sciter::set_options(sciter::RuntimeOptions::GfxLayer(
         sciter::GFX_LAYER::WARP
     )));
-    #[cfg(all(windows, not(feature = "inline")))]
-    unsafe {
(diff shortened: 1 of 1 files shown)


rustdesk:
when your display is high decision  screen, you will know.

yukitteru:
> when your display is high resolution screen, you will know.


Why not just leave the default awareness setting with the application manifest? Or you can check the version of Windows OS

rustdesk:
> manifest

We can not use manefest when debugging.

yukitteru:
> 

Okay, I'll redo the PR

PR #590, opened 2022-05-22
Remove `remote.rs` static mut booleans unsafe code, by using AtomicBool

64kramsystem:
## Description

For supported types, static R/W globals unsafe code can be replaced by safe `Atomic*` types.

The pattern of usage is simple; example with `KEYBOARD_HOOKED`:

- AtomicBool#swap is used to fetch the old `KEYBOARD_HOOKED` value, while setting it to true;
- if the old value was true, there is effectively no change to `KEYBOARD_HOOKED`, and the flow exits from the enclosing function;
- if the old value was false, execute the function (the new `KEYBOARD_HOOKED` has been set to true by swap()).

All the other variables use simply `load`/`store`.

The most conservative ordering is used, as the context is not performance-sensitive (it doesn't matter, anyway, in this context).

Atomics are not supported on every platform, but the project assumes x86-64, which supports them.

## Testing

- connected (successfully) to a remote server, and observed the variable `KEYBOARD_HOOKED`/`IS_IN` values before/during/after the test
- connected to a remote server, and (successfully) enabled/disaled keyboard input (wrt `SERVER_KEYBOARD_ENABLED`)
- connected to a remote server, and (successfully) enabled/disaled clipboard paste (wrt `SERVER_CLIPBOARD_ENABLED`)

## Notes

- `SERVER_FILE_TRANSFER_ENABLED` seems not to have effect on the workflow I've tried; regardless of it being enabled/disabled, when the client initiates a file send, the server instantiates a new connection (window); this is the same in master

### DIFF
--- a/src/ui/remote.rs
+++ b/src/ui/remote.rs
@@ -68,7 +68,7 @@ fn get_key_state(key: enigo::Key) -> bool {
 
 static mut IS_IN: bool = false;
 static KEYBOARD_HOOKED: AtomicBool = AtomicBool::new(false);
-static mut SERVER_KEYBOARD_ENABLED: bool = true;
+static SERVER_KEYBOARD_ENABLED: AtomicBool = AtomicBool::new(true);
 static mut SERVER_FILE_TRANSFER_ENABLED: bool = true;
 static SERVER_CLIPBOARD_ENABLED: AtomicBool = AtomicBool::new(true);
 #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


rustdesk:
Thanks, go ahead please.

64kramsystem:
> Thanks, go ahead please.

Ok, the PR is ready; I've updated the description.

rustdesk:
Good Job!

PR #603, opened 2022-05-24
Strip release binary via Rust toolchain

64kramsystem:
As of Rust 1.59, full stripping support has been added (see https://blog.rust-lang.org/2022/02/24/Rust-1.59.0.html#creating-stripped-binaries).

If the project is intended to compiled with older Rust versions, feel free to close the PR straight away :smile:.


rustdesk:
Thanks!

PR #719, opened 2022-06-06
fix compile fail when use cli feature

chamgin:
there are compile error when use cli feature:
1. rpassword:: read_password_from_tty need change, https://docs.rs/rpassword/6.0.0/rpassword/index.html
2. no implement Interface send
3. can't use make_fd which implement on ui module

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,6 +1,6 @@
 use crate::client::*;
 use hbb_common::{
-    config::PeerConfig,
+    config::{PeerConfig, LocalConfig},
     log,
     message_proto::*,
     protobuf::Message as _,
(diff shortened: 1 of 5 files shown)


rustdesk:
Thanks, please do not use sciter in common.rs, because it should be independent of ui sdk. we are removing sciter.

chamgin:
邮件已收到，谢谢！

chamgin:
> Thanks, please do not use sciter in common.rs, because it should be independent of ui sdk. we are removing sciter.

yes i moved it from common to client/helper

rustdesk:
sorry, helper.rs should also be ui sdk independent.

rustdesk:
@Kingtous https://github.com/rustdesk/rustdesk/blob/7c2db5a83559ed9190abc5ceb717983d6fc15cff/src/client/file_trait.rs#L13 we do need refactor this part of code

rustdesk:
> test

Thanks

rustdesk:
@chamgin could you try out https://github.com/rustdesk/rustdesk/pull/724?

chamgin:
> @chamgin could you try out #724?

yes, it works, no need my commit. i just make some product test

chamgin:
邮件已收到，谢谢！

PR #752, opened 2022-06-13
add fastlane images

Heap-Hop:


rustdesk:
minimize it with tinypng.com

PR #757, opened 2022-06-13
add DisplayVersion regkey in installer

sitiom:
I'm planning to add RustDesk to Winget, but Winget relies on this info to detect the Version properly. This PR adds that change.
Before:
before

After:
after

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -1028,6 +1028,7 @@ copy /Y \"{ORIGIN_PROCESS_EXE}\" \"{path}\\{broker_exe}\"
 reg add {subkey} /f
 reg add {subkey} /f /v DisplayIcon /t REG_SZ /d \"{exe}\"
 reg add {subkey} /f /v DisplayName /t REG_SZ /d \"{app_name}\"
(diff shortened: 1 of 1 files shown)


rustdesk:
How do you sign it in Winget?

sitiom:
Sign? You mean add? Just make a PR in https://github.com/microsoft/winget-pkgs with the corresponding manifest. See [YamlCreate](https://github.com/microsoft/winget-pkgs/blob/master/doc/tools/YamlCreate.md) on how to create a manifest.

I would be waiting for this change to be incorporated in the next version before adding it to Winget.

rustdesk:
I meant https://stackoverflow.com/questions/252226/signing-a-windows-exe-file

sitiom:
I don't think you need to sign your program for winget.

rustdesk:
I guess Winget downloaded our exe from our GitHub release page.

sitiom:
Yes, that's how it works. Also, please see https://github.com/rustdesk/rustdesk/issues/758 for another concern regarding Winget.

PR #768, opened 2022-06-14
feat: hardware codec && abr

21pages:
H264/H265 hardware codec support for windows

used library: [hwcodec](https://github.com/21pages/hwcodec)

enable feature: `cargo run --features hwcodec`

under development


yangshujing:
when change prefer codec,it not works, in video_service.rs ,the current_hw_encoder_name is None

rustdesk:
> when change prefer codec,it not works, in video_service.rs ,the current_hw_encoder_name is None

Fix it then please.

PR #818, opened 2022-06-19
code enhancement

h1midi:
Signed-off-by: Hamidi Mohammed <mr.hammiddi@gmail.com>

lib\common.dart:236/246 - Unnecessary null comparison (The operand can't be null, so the condition is always true)
lib\models\file_model.dart:596 - Unused local variable
lib\models\native_model.dart:24 - Unused field
lib\pages\( connection_page.dart | settings_page.dart ) - 'canLaunch' & 'launch' is deprecated and shouldn't be used.
lib\pages\remote_page.dart:129 - Avoid empty statements
lib\pages\settings_page.dart:152/189/215/245 -  Avoid using braces in interpolation when not needed
lib\widgets\gestures.dart:597/598/599/646 - Unnecessary null comparison (The operand can't be null, so the condition is always true)

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -233,7 +233,7 @@ class AccessibilityListener extends StatelessWidget {
           }
         },
         onPointerUp: (evt) {
-          if (evt.size == 1 && GestureBinding.instance != null) {
+          if (evt.size == 1) {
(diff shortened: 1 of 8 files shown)


rustdesk:
@Heap-Hop

rustdesk:
Thanks

PR #911, opened 2022-07-04
fix rpm upgrade

fufesou:

Reference:
https://www.cnblogs.com/xingmuxin/p/8990255.html

image

Here's an example of `dnf upgrade <rustdesk-1.1.11>` from `rustdesk-1.1.10`.
75ca8b90294af55929fe6594a40df7a


itamarjp:
how about packaging rustdesk for fedora, centos and others ?

PR #928, opened 2022-07-06
password encryption

21pages:
1. Add security password, onetime password (a special case of random password).
* Default setting:
  * security password: enable
  * random password: enable
  * one-time password: disable
* Old version password -> security password.
* When will new random password automatically generated:
	* Program start
	* Enable random password
	* Activate one-time password
	* It is set to update after session
* The one-time password is invalid only after it is used.
* For operation to random password after session,  password type is not distinguished.

2. Crypt security/peer/peer-rdp/peer-os/socks5 password.
* Symmetric encryption, key:
   * PC: machine uid
   * Mobile: secret key
* The first two characters are used as version when store. 
* If encryption or decryption fails, the original string is returned.

3. Free password authentication for reconnection within 30 seconds and with same remote window
* 30s is from last-receive-time to new-connecion-time.
* Random password and it's enable state will not be recovered yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2087,6 +2087,7 @@ dependencies = [
  "lazy_static",
  "log",
  "mac_address",
+ "machine-uid",
  "protobuf",
  "protobuf-codegen",
  "quinn",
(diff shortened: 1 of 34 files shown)


K2M74:
能不能增加TOTP？在公共设备上连接托管设备，静态密码可能被有问题的设备窃取（千日防贼太累啊）。
如果能自定义长度就更好了，谷歌最多8位，其他软件有支持10位的。

21pages:
> 能不能增加TOTP？在公共设备上连接托管设备，静态密码可能被有问题的设备窃取（千日防贼太累啊）。 如果能自定义长度就更好了，谷歌最多8位，其他软件有支持10位的。

The length of the security password is greater than 6. We need to store a fixed password.

Issue #540, opened 2022-05-16
Some complaints from `systemd` and `namcap`


sukanka:
**Describe the bug you encountered:**
First, `systemd` complains about the unit service.
```
> systemctl status rustdesk.service
/usr/lib/systemd/system/rustdesk.service:9: PIDFile= references a path below legacy directory /var/run/, updating /var/run/rustdesk.pid → /run/rustdesk.pid; please update the unit file accordingly.
```
This line should be changed
https://github.com/rustdesk/rustdesk/blob/85a2a7cd4452502d16c10edf1ba81573f215ec7c/rustdesk.service#L9
to
```
PIDFile=/run/rustdesk.pid
```

Besides, `namcap` complains about the dependencies (**Only on arch and arch derived distributions**)
Please reconsider the dependencies.
```
rustdesk W: Dependency libxcb included but already satisfied
rustdesk W: Dependency libxfixes included but already satisfied
rustdesk W: Dependency included and not needed ('alsa-lib')
rustdesk W: Dependency included and not needed ('pulseaudio')
rustdesk W: Dependency included and not needed ('ttf-arphic-uming')
rustdesk W: Dependency included and not needed ('python-pip')
rustdesk W: Dependency included and not needed ('curl')
```

BTW, the file `/usr/share/rustdesk/files/pynput_service.py` should be put into `/usr/share/rustdesk/pynput_service.py`,
and `python-pynput` should be added as a dependency, instead of installing it via `sudo pip` in `postinstall`
...

**What did you expect to happen instead?**

...

**How did you install `RustDesk`?**
via aur package [rustdesk-bin](https://aur.archlinux.org/packages/rustdesk-bin), but these issues also occur in the officially released package.

---

**RustDesk version and environment**
1.1.9

rustdesk:
Thanks

ieugen:
Also encountered this on Debian:
```
[Vi iun 24 08:03:52 2022] systemd[1]: /etc/systemd/system/rustdesk.service:9: PIDFile= references a path below legacy directory /var/run/, updating /var/run/rustdesk.pid → /run/rustdesk.pid; please update the uni>
[Vi iun 24 08:03:52 2022] systemd[1]
```

rustdesk:
@fufesou take a note

fufesou:
> @fufesou take a note

Ok. I'll change the scripts on the next PR.

dolphinoracle:
[...]

Issue #56, opened 2021-06-04
Wayland support


botiapa:
I wanted to try out this application, as it seems very promising, however as I understand it doesn't currently support wayland. What would be needed for wayland support? Screen capture support, GUI, or something else?

rustdesk:
screen capture and mouse/keyboard control

rustdesk:
for mouse/keyboard, https://github.com/ReimuNotMoe/ydotool can help u

botiapa:
Screen capture is a bit trickier though, since gnome uses a whole different protocol than wlroots based compositors.

rustdesk:
Not sure if https://github.com/emersion/grim helps

botiapa:
I think that's for screenshots only (maybe not fast enough for streaming). I know wf-recorder works on wlroots based compositors.

rustdesk:
> maybe not fast

The bottleneck is the codec.

botiapa:
What are you using for encoding and decoding?

rustdesk:
vp9

h33p:
So it appears scrap would need to support wayland screen capture for this to work fully, but it doesn't seem to be a very active repo. Pipewire appears to be the way to go with screen capture there, it would most likely block until the user selects the screen to share, but it would end up being efficient. I know zoom takes screenshots for wayland capture, and it's awful. I could give a look at building pipewire into a scrap fork, but I've never done it so no promises.

rustdesk:
> I could give a look at building pipewire into a scrap fork

Looking forward to fork.

rustdesk:
clipboard etc：https://drewdevault.com/2019/02/10/Wayland-misconceptions-debunked.html

rustdesk:
KDE (krfb) and Gnome (gnome-remote-desktop)

rustdesk:
nomachine wayland support limitations: https://knowledgebase.nomachine.com/AR02P00969

rustdesk:
> I'm guessing they mean the ScreenCast component of xdg-desktop-portal, which AFAIK is supported by all three major players. Plasma, Sway, GNOME. You could probably take obs-xdg-desktop-portal or WayRec as examples on how to implement it.

xeechou:
Hi, I am the author of a Wayland compositor, so I can provide a little bit idea here, for it to work with Wayland, you basically need a server (for example VNC) as a Wayland client doing the screen capture (read the compositor) then send it over to your application over network.

[...]

PR #932, opened 2022-07-07
partial wayland support

fufesou:
#56 Wayland support -- [xdg-desktop-portal](https://flatpak.github.io/xdg-desktop-portal/)
#540 systemd and systemd(user) -- debian

1. The remote side should accept screen share every first connection in a server(--server) process lifetime..
image
2. KEY_CapsLock does not work well.
3. Mouse horizontal movements are not tested for now.
4. Self side cannot be wayland. Keyboard and mouse are not hooked to listen.
5. Unsupport.  Server need to be restarted as  xdg-desktop-portal pipewire need to be connect and accepted again.
    Not sure how to detect and reconnect pipewire for now.
    - [ ] Modify resolution.
    - [ ] Switch bettwen multiple screens.
    - [ ] Lock screen.
    - [ ] Unattended access


rustdesk:
You'd better reuse old get_display_server_of_session in platform/linux.rs which considered starting x11 from tty. We can move it to hbb_common.

libs/enigo/src/linux/pynput.rs
```diff
@@ -0,0 +1,280 @@
...
+            Key::Numpad0 => "0",
+            Key::Numpad1 => "1",
+            Key::Numpad2 => "2",
+            Key::Numpad3 => "3",
+            Key::Numpad4 => "4",
+            Key::Numpad5 => "5",
+            Key::Numpad6 => "6",
+            Key::Numpad7 => "7",
+            Key::Numpad8 => "8",
+            Key::Numpad9 => "9",
```
asur4s, line 79: Requires a specific numpad state.

Issue #958, opened 2022-07-11
upgrade protobuf to 3.1 and use with-bytes feature.


rustdesk:
https://crates.io/crates/protobuf 3.1 and use `with-bytes` feature.

``` Python
#!/usr/bin/env python3        

'''
auto change old 'Union::hello_world' to 'Union::HelloWorld"
'''
  
import sys    
import re    
    
for ln in sys.stdin:    
  m = re.search('Union::(\w+)', ln)    
  if m:    
    x = ''.join([x[0].upper() + x[1:] for x in m.groups()[0].split('_')])    
    ln = ln.replace('Union::' + m.groups()[0], 'Union::' + x)    
  print(ln, end='')  
```

PR #970, opened 2022-07-14
upgrade protobuf to 3.1.0 and use with-bytes feature, issues:958

twmsh:
issues #958


rustdesk:
Good Job!

rustdesk:
Seems you missed

> Note, codegen also need to be instructed to generate Bytes or Chars for bytes or string protobuf types instead of default Vec<u8> or String, just enabling option on this crate is not enough.

PR #1002, opened 2022-07-19
Create pl.rs

LeszekBober:
Polish language.


rustdesk:
To enable PL, you need to modify https://github.com/rustdesk/rustdesk/blob/master/src/lang.rs also.

PR #1013, opened 2022-07-21
Hwcodec: codec preference

21pages:
1. codec preference
2. linux compatible
3. Code optimization
* check hwcodec config every time server or client start, but it is not checked when the tray icon started the service.
* Simple function 'best' in hwcodec, always check hwcodec in another process.

### DIFF
--- a/libs/hbb_common/protos/message.proto
+++ b/libs/hbb_common/protos/message.proto
@@ -72,6 +72,11 @@ message Features {
   bool privacy_mode = 1;
 }
 
+message SupportedEncoding {
+  bool h264 = 1;
+  bool h265 = 2;
+}
+
 message PeerInfo {
   string username = 1;
   string hostname = 2;
@@ -82,6 +87,7 @@ message PeerInfo {
   string version = 7;
   int32 conn_id = 8;
   Features features = 9;
+  SupportedEncoding encoding = 10;
 }
 
 message LoginResponse {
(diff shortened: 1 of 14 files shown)


src/ui/header.tis
```diff
@@ -157,6 +160,13 @@ class Header: Reactor.Component {
...
                 <li #low type="image-quality"><span>{svg_checkmark}</span>{translate('Optimize reaction time')}</li> 
                 <li #custom type="image-quality"><span>{svg_checkmark}</span>{translate('Custom')}</li>
+                {show_codec ? <div>
+                <div .separator />
+                <li #auto type="codec-preference"><span>{svg_checkmark}</span>Auto</li>
+                <li #vp9 type="codec-preference"><span>{svg_checkmark}</span>VP9</li>
+                {codecs[0] ? <li #h264 type="codec-preference"><span>{svg_checkmark}</span>H264</li> : ""}
+                {codecs[1] ? <li #h265 type="codec-preference"><span>{svg_checkmark}</span>H265</li> : ""}
+                </div> : ""}
                 <div .separator />
```
rustdesk, line 170: test

libs/hbb_common/protos/message.proto
```diff
@@ -434,9 +440,17 @@ enum ImageQuality {
...
 message VideoCodecState {
-  int32 ScoreVpx = 1;
-  int32 ScoreH264 = 2;
-  int32 ScoreH265 = 3;
+  enum PerferCodec {
+    Auto = 0;
+    VPX = 1;
+    H264 = 2;
+    H265 = 3;
+  }
```
rustdesk, line 448: test again

rustdesk (changes requested):

src/ui/header.tis
21pages: Do not display any options when only VP9 is available?

rustdesk: Correct

21pages: using the installed version, a CMD window will flash by at the controlled end when connected or switching the codec. This needs to be modified.

PR #1032, opened 2022-07-24
Mouse input error on Mac, etc

sj6219:
correction 1.

When accessing the Korean version windows from the Mac, you cannot enter the Korean-English conversion key. The Microsoft Remote Desktop program uses the right Alt (option) key instead of the Korean-English conversion key. So, I modified it to work the same way.

~~correction 2.~~

~~When you click the mouse button in two places on the screen while pressing the Command key on the Mac, the mouse input will be incorrect. I modified it.~~


rustdesk:
@asur4s

libs/enigo/src/win/win_impl.rs
```diff
@@ -48,6 +49,11 @@ fn keybd_event(flags: u32, vk: u16, scan: u16) -> DWORD {
                     GetWindowThreadProcessId(GetForegroundWindow(), std::ptr::null_mut());
                 LAYOUT = GetKeyboardLayout(current_window_thread_id);
             }
+            if winapi::shared::minwindef::LOWORD(LAYOUT as usize as u32) == 0x0412 {
+                if vk == winapi::um::winuser::VK_RMENU as u16 {
+                    vk = winapi::um::winuser::VK_HANGUL as u16;
+                }
+            }
```
rustdesk, line 53: could you check `vk == winapi::um::winuser::VK_RMEN` first please? it is cheaper than  LOWORD, though a very little bit, :).
``` 
if vk == winapi::um::winuser::VK_RMENU as u16 && winapi::shared::minwindef::LOWORD(LAYOUT as usize 
                  vk = winapi::um::winuser::VK_HANGUL as u16;
            }
```

src/server/input_service.rs
```diff
@@ -379,8 +379,8 @@ fn handle_mouse_(evt: &MouseEvent, conn: i32) {
     let mut to_release = Vec::new();
     if evt_type == 1 {
         fix_modifiers(&evt.modifiers[..], &mut en, 0);
-        #[cfg(target_os = "macos")]
-        en.reset_flag();
+        // #[cfg(target_os = "macos")]
+        // en.reset_flag();
```
rustdesk: Can I know how reset_flag causes the problem? Just curious. :)

sj6219: The command modifier flag should not be zeroed when the mouse is pressed or raised. If it's an old code, when the second mouse click is pressed, the flag will be already zeroed.
This flag should be zeroed when the command key is raised.

rustdesk:
Your fix to mouse have problem in the case that no key up event. e.g. mobile -> mac, mac only receive mouse event with modifiers, but no key up events.

libs/enigo/src/win/win_impl.rs
sj6219: I modified it a little. 
`            if (LAYOUT as u16 == 0x0412) && (vk == winapi::um::winuser::VK_RMENU as u16) { 
                vk = winapi::um::winuser::VK_HANGUL as u16;
            }
`

src/server/input_service.rs
rustdesk: Yes, but it causes bug for controlling mac with mobile, mobile does not send key up event to mac.

sj6219:
image

Problems arise when multiple items need to be selected, as shown in the figure.

In the case of mobile, there will be other solutions.

rustdesk:
Thanks, there may be better solution, let's think.

sj6219:
Correction 3.

I corrected the error when the command + tab key is pressed.

sj6219:
correction 4.

I corrected the error that cannot be accessed when I log out of the Mac and log in again.

And when you do this, if you type the key, the program might lose connection or not be able to type, but it hasn't been fixed yet.
image
The figure shows that an error occurred when an exception occurred inside the TISGetInputSourceProperty function on line 611. I gave up because I couldn't access the inside of this function.

rustdesk:
> correction 4.
> TISGetInputSourceProperty

So we need to call isValidateInputSourceRef first before call TISGetInputSourceProperty?

src/rendezvous_mediator.rs
```diff
@@ -501,12 +501,13 @@ async fn direct_server(server: ServerPtr) {
...
-                        }
-                        sleep(1.).await;
-                    }
+                    // loop {
+                    //     if port != get_direct_port() {
+                    //         break;
+                    //     }
+                    //     sleep(1.).await;
+                    // }
+                    sleep(1.).await;
```
rustdesk, line 521: What do you mean here? sleep a moment then retry?

My original idea is to avoid repeating retry until the user changed the port.

rustdesk (changes requested):

src/rendezvous_mediator.rs
sj6219: the port was already allocated by another rustdesk process and could not be received. The process is about to be reallocated after that process dies.

rustdesk: I think we can have clever solution, e.g.
```
ntries++
for i in 0..ntries {
checkport change
sleep
}
```
ntries is reset if port changed or no errors. max of ntries can be set to 60

sj6219:
correction 5.
When I accessed my android phone, I added the ability to copy the client machine's clipboard to my android phone's.

asur4s:
> correction 2.
> 
> When you click the mouse button in two places on the screen while pressing the Command key on the Mac, the mouse input will be incorrect. I modified it.

Hey, I found an easier way to fix the mouse problem.

When the local Windows hits the win key, the remote macOS gets the command key. When you long press the win key, the command key will be continuously pressed and released. We only need to lock the state of the win key pressed to get the correct result.

sj6219:
I think the problem related to keyboard input has been fixed.

PR #1037, opened 2022-07-25
add: card, address book, fav ui, new AbModel

Kingtous:


Kingtous:
AbModel, which means `address book model`, is used to mantain runtime state for independent plugin `address book`.

Issue #1080, opened 2022-07-28
Can not transfer file from remote(Android) to local(Windows10)


tangxianzu-ming:
**Describe the bug you encountered:**
local file(Win10) can transfer to remote(Android) when press ">" button, but remote file (Android) can not transfer to local(Win10) when press "<" button.
...

**What did you expect to happen instead?**
remote file(Android) can transfer to local(Win10)
...

**How did you install `RustDesk`?**
download exe and apk file from https://rustdesk.com/zh
...

**RustDesk version and environment**
RustDesk version
local(Windows):1.1.9
remote(Android):1.1.10-1

Environment
local(Windows): win 10
remote(Android): pico g2 4k VR headset, Android 8
...

![2022728-93007](https://user-images.githubusercontent.com/11407632/181411027-f22273e8-69d0-4ea5-a967-b09a42118777.jpg)

rustdesk:
@Heap-Hop

rustdesk:
@Kingtous android is 1.1.10 version, windows is 1.1.9, seems that PR has compatibility issue.

JumperCzech:
both windows>android and android>windows dont work for me 
same rustdesk versions, android 11 MIUI 13

(using the android app)
android to windows gets stuck at 0%, windows to android says the file was successfully transferred but no file is present

PR #1102, opened 2022-07-29
fix: android overwrite detection for lower version server

Kingtous:
Related issue: #1080


tangxianzu-ming:
So you mean if i install RustDesk on Android(remote) and Windows(local) using the same version like 1.1.9, then i can transfer file from Android to Windows, right? but when i download apk file in "Releases v1.1.9" and install on android，it show up to be v1.1.10-1 on setting tab.

Issue #921, opened 2022-07-04
Rustdesk claiming MX Linux laptop is not using X11 when it definitely is? Preventing remote connections TO that laptop? (support boot not using Systemd)


dagleykd:
**Describe the bug you encountered:**

Installed RustDesk on the latest version of MX Linux (latest version of RustDesk, MX Linux up to date) on my HP Stream 13 laptop. When opening RustDesk on that laptop, I see the message 'Unsupported display server, x11 expected'. I can connect to and control clients fine from that laptop, but no computers can connect to said laptop as they are given the same message on their end. MX DOES use X11, so not sure why there is an error?

**What did you expect to happen instead?**

I should be able to have other computer connect to and remotely control this computer - RustDesk claims the display server is NOT X11 when it most definitely is?

**How did you install `RustDesk`?**

Latest GitHub 64-bit deb file Linux download from the Rustdesk website - installed through the GDebi package installer on MX Linux 21.1 64-bit XFCE.

**RustDesk version and environment**

I am using RustDesk for personal use to connect to my own PCs and my family's PCs remotely when necessary for support or remote access. Using the latest version of RustDesk.

rustdesk:
Why so many Linux distros? :(

dagleykd:
> Below is the function to check display server type. If you know a little about programming, you shoud be able to figure out some problem for us to fix it.
> 
> https://github.com/rustdesk/rustdesk/blob/de044db6d45e878fdd1077176675305e71375d6f/src/platform/linux.rs#L347

A bit of research seems to suggest it's because MX doesn't boot using Systemd by default....once selecting it to boot with Systemd, RustDesk works fine!

Is there any way to make this not necessary?

rustdesk:
> not necessary

Contribution is welcome.

sirtemporary:
Different (?) problem on different non-systemd distro.  Client running devuan 4 with xfce.  When another computer (Windows 10) connects to it, after authenticating, rustdesk immediately closes (terminal window says Illegal operation).  Will have some time to debug/track this down, hopefully within the next week.

sirtemporary:
Regarding my previous comment, so far I'm unable to duplicate this problem on my own hardware.  The problem occurred on another computer that is no longer in my possession (I did not have much time with it).  In the meantime, I'm looking at mx linux and the problem stems from the fact that within mx linux, loginctl outputs an error message if init is not systemd.  Will attempt a workaround.

rustdesk:
[...]

PR #1117, opened 2022-07-30
Update linux.rs

sirtemporary:
Fix for #921 
in mx linux, when started with non-systemd init, loginctl returns null string to stdout (and an error message to stderr).  this patch will use XDG_SESSION_TYPE and XDG_SESSION_ID environment variables if the loginctl code fails to determine these.


rustdesk:
Thanks

PR #1148, opened 2022-08-02
fix: hide restart menu item if version < 1.1.10

21pages:


rustdesk:
Do not forget flutter version.

PR #1152, opened 2022-08-02
Add Japanese translation of the app

MaySoMusician:
This PR will add the option for Japanese translation of the app

Note that `doc_mac_permission` should be updated after the document is translated. Maybe I will try it soon (TM)


rustdesk:
Thanks

PR #1388, opened 2022-08-28
Flutter desktop new remote menu

fufesou:
Add flutter menubar for desktop, tested on windows.

TODO:
1. Connection status icon.
2. Switch peer id and peer OS account.
3. Prevent frequent clicks on some switch options, eg privacy mode.
4. Floating menubar.


rustdesk:
Menu item veritical padding too big, connection-type(if secure or direct connection) icon missed, display map missed (can do it later).

PR #1402, opened 2022-08-30
fix: close all typed sessions when hide subwindow

Kingtous:


flutter/lib/desktop/pages/connection_tab_page.dart
```diff
@@ -63,15 +63,14 @@ class _ConnectionTabPageState extends State<ConnectionTabPage> {
             label: id,
             selectedIcon: selectedIcon,
             unselectedIcon: unselectedIcon,
-            closable: false,
```
Kingtous, line 66: @Heap-Hop revised

PR #1465, opened 2022-09-07
feat: support changing id via cmdline

XieJiSS:
This PR adds the following features:

+ Get current ID via command line
  + Note: `println!` does not work for no-console mode under Windows, i.e. `#![windows_subsystem = "windows"]`. See https://github.com/rust-lang/rust/issues/67159#issuecomment-563014833. This also explains why `./rustdesk.exe --version` shows nothing.
+ Set ID via command line

### Example Usage

```shell
$ ./rustdesk.exe --id 119206784
```

Result:

image

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -667,6 +667,11 @@ pub fn get_id() -> String {
     }
 }
 
+pub fn set_id(v: String) -> ResultType<()> {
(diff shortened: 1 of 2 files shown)


XieJiSS:
Closed due to inactivity & conflict

Issue #1456, opened 2022-09-06
Docker Build error "ffigen is not a command, or not executable"


topaz2002:
While building via docker-container i get error
```
error: failed to run custom build command for `rustdesk v1.1.10 (/home/user/rustdesk)`

Caused by:
  process didn't exit successfully: `/home/user/rustdesk/target/release/build/rustdesk-b236ee71c6208b6b/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=src/flutter_ffi.rs

  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ffigen is not a command, or not executable.', build.rs:99:56
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

...

**What did you expect to happen instead?**

successfully built

**How did you install `RustDesk`?**

```
git clone https://github.com/rustdesk/rustdesk
cd rustdesk
docker build -t "rustdesk-builder" .

```

**RustDesk version and environment**
last available

**In order to reproduce your issue, please add some information about the environment
in which you're running RustDesk.**
-->
docker --version
Docker version 20.10.17, build 100c701

 uname -a
Linux 4.14.35-2047.516.2.1.el7uek.x86_64 #2 SMP Wed Aug 17 16:30:09 PDT 2022 x86_64 x86_64 x86_64 GNU/Linux

Exactly follow the building instruction (Docker version)

rustdesk:
#763

PR #1492, opened 2022-09-11
Added flatpak feature

### LINKED ISSUE #459: Add this app to flatpak.
Hello! I think it would be cool to add this app to flatpak/flathub repository for easier cross linux distro installation.

sandr01d:
Added a flatpak feature that allows rustdesk to find the libsciter library in a flatpak environment as discussed with @Kingtous. I am currently unable to build the current state of the master branch with the same issues as #1456, but I've tested this on the state of the latest release and with this patch I am able to run RustDesk in a flatpak. If somebody can provide me with more information on how to build the master branch, I'll can test it out there aswell.
This does not fully close #459, there is some additional work to be done: `get_display_server_of_session` currently fails because systemd is not available in flatpak.


muelli:
note that it's possible to determine at runtime whether the app is executed in a flatpak environment by checking for `/.flatpak-info` or, less reliable, `FLATPAK_ID` environment variable.

rustdesk:
So flatpak can not be used as controlled side.

kolAflash:
> So flatpak can not be used as controlled side.

@rustdesk  
It might be possible to use RustDesk as controlled side via flatpak.  
At least this is possible for the AnyDesk flatpak (just tested it).  
https://flathub.org/apps/details/com.anydesk.Anydesk

So there might be a clue how to achieve this for RustDesk too.  
https://github.com/flathub/com.anydesk.Anydesk

Kingtous:
> > So flatpak can not be used as controlled side.
> 
> @rustdesk It might be possible to use RustDesk as controlled side via flatpak. At least this is possible for the AnyDesk flatpak (just tested it). https://flathub.org/apps/details/com.anydesk.Anydesk
> 
> So there might be a clue how to achieve this for RustDesk too. https://github.com/flathub/com.anydesk.Anydesk
> 
> Maybe this!? https://github.com/flathub/com.anydesk.Anydesk/blob/16172333ff605f92ffcd9b6d0703c7b2faec0209/com.anydesk.Anydesk.json#L16 https://docs.flatpak.org/en/latest/sandbox-permissions-reference.html

I've tested `Anydesk`, it does not have an daemon service, so it cannot be invoked when the `Anydesk` is closed. `RustDesk` aims to handle this feature, so we need a small and light daemon service to run forever in the background, which gives `RustDesk` ability to automatically open when a new connection is comming. We assume that flatpak does not support daemon service currently due to [this](https://flatpak.org/faq/#Can_Flatpak_be_used_on_servers_too_), it does not has [Service Management](https://snapcraft.io/docs/service-management) that `snap` does.

kolAflash:
@Kingtous  
> I've tested `Anydesk`, it does not have an daemon service, so it cannot be invoked when the `Anydesk` is closed. `RustDesk` aims to handle this feature, so we need a small and light daemon service to run forever in the background, which gives `RustDesk` ability to automatically open when a new connection is comming.

Sure, a background daemon is something else.

But in principal running RustDesk via FlatPack should should be possible for the controlled side. It's just necessary that the RustDesk application runs in "foreground" and not only as a daemon.

&nbsp;

> So flatpak can not be used as controlled side.

Maybe I got @rustdesk comment wrong. I just wanted to say, that it's possible to control a desktop from an application inside FlatPack in general.

Root-Core:
Other projects do this by staying in the tray, which seems to be a viable solution to this problem until daemon support lands on flatpack (if ever).

tristan-k:
Is there a way to run the nightly flatpak image in the background? For example like `flatpak run com.rustdesk.RustDesk --background`. I was able to create a autostart file doing this but now every time my computer starts the rustdesk windows pops up.

```
$ nano ~/.config/systemd/user/rustdesk.service

[Unit]
Description=Remote Desktop
StartLimitIntervalSec=500
StartLimitBurst=5

[Service]
ExecStart=flatpak run com.rustdesk.RustDesk
ExecStop=flatpak kill com.rustdesk.RustDesk
Restart=on-failure
RestartSec=5s

[Install]
WantedBy=graphical-session.target

$ systemctl --user enable rustdesk
```

PR #1510, opened 2022-09-13
Fix error of flutter deb

chiehw:
1. Fix packing deb of flutter.
2. Fix repeated typing when key press.
3. Rdev supports Andriod compilation.
4. Fix the error of staring the service from the wayland service command line.


rustdesk:
confliction

PR #1550, opened 2022-09-16
opt: main window save/restore offset/position

Kingtous:


rustdesk:
do not miss the other window

rustdesk:
no need for cm window

PR #1578, opened 2022-09-19
opt: binary name, macos frame, address book network catch

Kingtous:


rustdesk:
cm window?

Kingtous:
`DragToResize` is not used in cm window. Not it's fixed size on screen.

PR #1660, opened 2022-10-04
ua.rs Ukraine UI translate

kirill0ermakov:
I create Ukraine UI translate and have a hope You add this


rustdesk:
Thanks a lot, but you need to follow https://github.com/rustdesk/rustdesk/blob/master/src/lang/template.rs

kirill0ermakov:
Okay, i try again :)

Issue #863, opened 2022-06-24
fps adjustable for self-hosting


rustdesk:

bwoebi:
I would greatly appreciate this feature being added; while it's usable, noticeable lag due to the fps limitation is what's holding me off from actively using rustdesk.
I do appreciate the great effort invested into rustdesk though! It's nice and generally just works :-)

PR #1759, opened 2022-10-20
support adjust fps && fix statusbar ui

21pages:
1. Support fps adjustment for self-hosting or direct connection , range: 10~120,  #863 
2. Fix statusbar ui
3. Sync setting page service status
4. Add hwcodec dependency comment in buid.py

### DIFF
--- a/build.py
+++ b/build.py
@@ -71,7 +71,7 @@ def make_parser():
     parser.add_argument(
         '--hwcodec',
         action='store_true',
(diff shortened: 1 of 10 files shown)


rustdesk:
How about the connection status problem when switching custom server mentioned by @asur4s ?

rustdesk:
> windows

No workaround?

PR #1864, opened 2022-10-28
Command-Line Improvements

Zinal001:
Reworked the whole command-line handling in core_main.rs. It can now handle multiple arguments on the same call.

Also added two new commands: `--set-*` for setting most configurations in RustDesk2.toml _(Excluding nat_type, serial and local-ip-addr)_.

The `--set-*` commands verifies and saves the values into _RustDesk2.toml_.
To allow for the `--set-*` commands to be used with the `--install` command, it also saves both configuration files into cached files _(\_RustDesk.toml and \_RustDesk2.toml)_ which is read by the program on startup _(If they exists)_ and overwrites any configuration in the main files _(In the %appdata% folder on windows, for example)_.

These commands should be at least a partial solution to unattended installations.
Command-lines like this is now possible: `rustdesk --password PASS --set-relay-server RELAY-SERVER --set-custom-rendezvous-server ID-SERVER --set-verification-method use-permanent-password --set-allow-darktheme Y --install`

The other command is `--exit`, which exits the program. This is useful now since command such as `--password` will no longer automatically exit the program when done. _(Commands such as `--install` **will** exit them like before)_.


rustdesk:
Good job, but it seems too much code change.

Sympatron:
I think something like this is extremely useful for unattended installations. Please consider merging this or something similar.

knro:
would this make it possible to programatically retrieve ID + password as well for unattended devices?

vladimir-angelov-1337:
Please merge this, it's super useful.

SoftwareMagicIT:
@Zinal001 why if I implement in current source code this improvements, in debug working and in release mode empty output?

douardda:
why is this PR still not merged? either the PR is too big and ask for proper rework (split the big revision, etc.), accept the PR as is or reject it. "it seems too much code change" is not a good/clear feddback IMHO.

And TBH, a `+308/-27` diffstat on 3 files is not that big of a diff...

vladimir-angelov-1337:
Fo' real.

rustdesk:
@Zinal001 could you please resolve the conflict? and refactor hard coded available_options to more flexible way?

rustdesk:
Thanks @Zinal001, we will polish this part later.

PR #1900, opened 2022-11-01
hbb_common: Fix CONFIG locks unintentionally not acquired

64kramsystem:
Variables defined as `_` only are released immediately; see compile error:

    error: non-binding let on a synchronization lock
      --> libs/hbb_common/src/config.rs:798:13
        |
    798 |         let _ = CONFIG.read().unwrap(); // for lock
        |             ^   ^^^^^^^^^^^^^^^^^^^^^^ this binding will immediately drop the value assigned to it
        |             |
        |             this lock is not assigned to a binding and is immediately dropped
        |
        = note: `#[deny(let_underscore_lock)]` on by default

By appending a phony suffix to the variable name, the lock stays in scope.

### DIFF
--- a/libs/hbb_common/src/config.rs
+++ b/libs/hbb_common/src/config.rs
@@ -795,7 +795,7 @@ const PEERS: &str = "peers";
 
 impl PeerConfig {
     pub fn load(id: &str) -> PeerConfig {
-        let _ = CONFIG.read().unwrap(); // for lock
+        let _lock = CONFIG.read().unwrap();
         match confy::load_path(&Self::path(id)) {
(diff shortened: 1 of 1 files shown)


64kramsystem:
> Thanks a lot, could you help check the other files in our repo?

If a `cargo build` compiles all the files, then there's nothing to do, as I've caught those just by building.

If you didn't catch those error, for example in the build, I suspect that this is a former warning that has been converted to error in recent rustc versions. If that's the case, the build may be using an old version of rustc, and updating it will uncover new issues :smile:

PR #1949, opened 2022-11-03
Rename template.rs to es-ct.rs

flusheDData:
Template ready for catalonian language


flusheDData:
I meant to create a new file instead of renaming Template.rs

Issue #1474, opened 2022-09-08
Mouse scrolling speed too fast on Mac


iamGaryLiao:
**Describe the bug you encountered:**

Mouse scrolling speed was way too fast when controlling a Mac, even if adjusted the Mac's mouse scrolling speed to slowest.

**What did you expect to happen instead?**

I wished you can find out what was the cause of this and made the scrolling speed normal. (I used other remote control softwares too, and their mouse scrolling speed were quite similar with controlling locally)

**How did you install `RustDesk`?**

---GitHub release

**RustDesk version and environment**

 RustDesk 1.1.9, macOS Monterey 12.2.1

niehongbo:
Me too, I can't control the scrolling very well. View scrolls too fast when I remote control window on magic mouse

IsaacZD:
same issue here, quite annoying

rustdesk:
@asur4s

asur4s:
I didn't reproduce this bug, maybe it's because of the magic mouse?

ul:
It's reproducible for me when controlling another Mac with MacBook Pro's built-in trackpad.

rustdesk:
> trackpad

How about with mouse?

ul:
The scrolling speed is normal when I use a regular mouse.

rustdesk:
> The scrolling speed is normal when I use a regular mouse.

@xxrlzzz we need to test both mouse and trackpad for Flutter.

aiden0820:
Still happening on my MBP

Torrontes123:
I have the exact opposite problem. Controlling a Mac from Windows, scrolling using the wheel is painfully slow, regardless of the scroll settings on the Mac.

Throwaway24924:
Controlling Mac from Linux laptop
Scroling is fine most of the time like when using chrome 

But scrolling through a spreadsheet is so fast and inaccurate it's unusable i have to use arrow keys
There were other times where the speed was unberable but can't remember what I was doiing

PR #1987, opened 2022-11-06
 opt: mac scroll to fast

xxrlzzz:
for #1474


rustdesk:
Can this make scroll slow in some cases?

rustdesk:
Where do you get the "3"?

` (x, y) = (x * 3, y * 3);`

xxrlzzz:
https://user-images.githubusercontent.com/40893153/200231403-2446e25f-2cdb-4317-a5d4-fd00aef5f981.mp4

Previously we use line mode, but sciter emit a lot of events at one scrolling(fling), so the scroll speed will be too fast.

Sciter set direction into x,y (-1, 0, 1), and power into buttons. But the values of buttons vary, they didn't  this on documentations, so it will be unreliable. 

here is a code snippet I try to use event, it didn't work so will. 

```rust
#[inline]
#[cfg(target_os = "macos")]
fn fix_scroll_in_mac(mut x: i32, mut y: i32, buttons: i32) -> (i32, i32) {
    // mac can only have one direction.
    // Weird buttons value, it can be different signal with x,y
    use std::cmp::{max, min};
    let new_y = if y == -1 {
        y * (-buttons + 0xfffe) / 0xffff
    } else if y == 1 {
        y * buttons
    } else {
        y
    };

    let new_x = if x == 1 {
        x * (buttons + 0x1ffff) / 0x20000
    } else if x == -1 {
        x * (-buttons + 0x1ffff) / 0x20000
    } else {
        x
    };
    log::info!("old xy {x} {y}, new xy {new_x} {new_y} btn {buttons}");
    (x * 3, y * 3)
}

``` 

* 3 means to speed up the scrolling, and I tested in my mac(m1 and x86), it works.

rustdesk:
@xxrlzzz can you remove log::info! which will be dumped to log file, or change to log::trace?

rustdesk:
> Sciter

We are discarding Sciter, can you test with Flutter?

xxrlzzz:
> @xxrlzzz can you remove log::info! which will be dumped to log file, or change to log::trace?

My code snippet will not commit， and the file I modified seems only effect on sciter，but I will test on flutter and post result

xxrlzzz:
This scroll problem seems to a universal problem when using **track pad**, unfortunately I don't have a good device for windows develop, so I can't fix this problem for windows currently. But this patch is enough for fix mac itself.

Here is all test cases

- [x] Mac control mac (mouse, track pad)
- [x] Ios control mac (gesture) 
- [x] Win control mac (mouse)
- [x] Mac control win (mouse)

- [ ] Win control mac (track pad)
- [ ] Mac control win (track pad)

So there will need a follow patch to distinguish track pad input from mouse input, and send track pad input on win (linux may also needs support).

xxrlzzz:
This patch is ready for reviewing.

rustdesk:
Good job

rustdesk:
@asur4s could you help review?

src/client.rs
```diff
@@ -1529,6 +1529,25 @@ pub async fn handle_test_delay(t: TestDelay, peer: &mut Stream) {
     }
 }
 
+#[inline]
+#[cfg(target_os = "macos")]
+fn check_scroll_on_mac(mask: i32, x: i32, y: i32) -> bool {
```
chiehw, line 1578: I am not sure if this function will work in flutter version, Here is some relevant code.

https://github.com/rustdesk/rustdesk/blob/1b7950fc428e0c0e3f68d57efc66837c0a05d256/src/flutter_ffi.rs#L828

https://github.com/rustdesk/rustdesk/blob/1b7950fc428e0c0e3f68d57efc66837c0a05d256/flutter/lib/models/input_model.dart#L287

chiehw (changes requested):

src/client.rs
chiehw: Have you tested in flutter version? @xxrlzzz

xxrlzzz: I had test on iOS, download from App Store

xxrlzzz: You say flutter version, do your mean flutter for mac version? I only found flutter build for mobel.

chiehw: Yes, flutter for mac version. Windows, Linux, and Mac currently have flutter versions, They will all call the `check_scroll_on_mac` function

chiehw: I don't have a track pad to test if this function works properly in flutter

xxrlzzz: patch updated
flutter build keep crash with log, seems not my problem
```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: MissingPluginException(No implementation found for method getAll on channel plugins.flutter.io/shared_preferences_macos)
```

rustdesk: @fufesou removed share_pref dep in latest commit, you can remove it from  pub config, and cleanup cache and pod cache etc.

rustdesk:
@xxrlzzz have you test flutter verson on Mac?

xxrlzzz:
> @xxrlzzz have you tested flutter verson on Mac?

I have disabled this feature on flutter

rustdesk:
@xxrlzzz have you built flutter successfully on your machine?

xxrlzzz:
@rustdesk unable to connect client use flutter version
image

xxrlzzz:
https://user-images.githubusercontent.com/40893153/202911501-4487bf90-174b-43db-8acb-8f9d77a646b5.mov

Support for flutter finished, pls take a review. @rustdesk

rustdesk:
Good job.

rustdesk:
https://github.com/rustdesk/rustdesk/issues/2655

src/client.rs
```diff
@@ -1572,13 +1572,37 @@ pub async fn handle_test_delay(t: TestDelay, peer: &mut Stream) {
     }
 }
 
+/// Whether is track pad scrolling.
+#[inline]
+#[cfg(all(target_os = "macos"))]
+fn check_scroll_on_mac(mask: i32, x: i32, y: i32) -> bool {
+    // flutter version we set mask type bit to 4 when track pad scrolling.
+    if mask & 7 == 4 {
```
fufesou, line 1580: @xxrlzzz Hi, I'm testing trackpad on win->macOS.

I find trackpad will not work correctly because macOS check if `ControlKey::Scroll` is in the modifiers.

https://github.com/rustdesk/rustdesk/blob/1ededa3476aab868364f6853706e91b7b0ad4e4c/src/server/input_service.rs#LL841C32-L841C32

`check_scroll_on_mac` is only compiled on macOS.

So, I have some questions:

1. Why `check_scroll_on_mac` is needed ? `mask & 7 == 4` is not enough to know whether current event is trackpad event?
2. I cann't understand the logic after `let btn = mask >> 3;`  in `check_scroll_on_mac`. Can you please explain or give a link of what you refered?
   I've revert the code to Commit `260c924`, but I cannot find where `0xff88`, `0x78`, `0x780000` are from.

https://github.com/rustdesk/rustdesk/blob/260c924010361137dd67be145e9d5c0850804768/src/flutter_ffi.rs#L840


`handler.send_mouse` in 
https://github.com/rustdesk/rustdesk/blob/260c924010361137dd67be145e9d5c0850804768/src/ui/remote.tis

xxrlzzz: 2个原因
1. 没有其他的设备,所以只支持了mac 的 触控板
2. 当时还在兼容js做ui, hard code是js发出来的,我当时也没有找到这几个hard code的定义,用log打印出来的.

如果现在只有flutter,可以考虑重构一下这里的代码. 重点是识别出触控板和鼠标滚轮的区别, 并正确的发出scroll事件来模拟触控板. 希望能帮到你 :)

PR #2037, opened 2022-11-09
fix: sending file from local to remote

djex:
This fixes a bug where 0 byte files were being created when sending from local to remote.

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -1607,8 +1607,7 @@ async fn start_ipc(
                             file_num,
                             data,
                             compressed}) = data {
-                                stream.send(&Data::FS(ipc::FS::WriteBlock{id, file_num, data: Bytes::new(), compressed})).await?;
-                                stream.send_raw(data).await?;
+                                stream.send(&Data::FS(ipc::FS::WriteBlock{id, file_num, data, compressed})).await?;
                         } else {
                             stream.send(&data).await?;
                         }


rustdesk:
not accceptable， because sendraws is required for performance concern， but you can add a if

djex:
> not accceptable， because send_raw is required for performance concern when sending big data in our ipc， but you can add a `if`

Currently in master and nightly send_raw is not handled by the receiving end in this context. I spent awhile trying to figure this out. When I compared 1.1.9 source with nightly it looks like it's part of some refactored code that never got fully implemented. As of right now when sending files to remote it will create a 0 byte file. No data ever gets written to disk because there is nothing implemented to handle send_raw.

The original code is below: `stream.send` is handled in `handle_fs` but is creating an empty `Bytes` each frame. Then next` stream.send_raw` sends the data of the frame however there is nothing to handle this data in `handle_fs` or anywhere else that I could see.

`stream.send(&Data::FS(ipc::FS::WriteBlock{id, file_num, data: Bytes::new(), compressed})).await?;
 stream.send_raw(data).await?;`

Can you take a closer look at this to confirm what I found.

djex:
Turns out this is my mistake. I was building a nightly build and not master. This issue is already fixed in the latest master.

rustdesk:
@djex i misunderstood your issue, you are correct, your fix is fine, but it destroy our optimization, send_raw is added for performance optimization, wondering if you have interest to fix the problem without removing send_raw?

rustdesk:
@Heap-Hop

rustdesk:
#2099

rustdesk:
both sciter and flutter has the same issue， it may be caused by our code refactory when migrating from sciter to flutter about two month ago.

rustdesk:
@Heap-Hop take over please

rustdesk:
i will merge this as a workaround

Issue #70, opened 2021-06-09
Add way to remotely enter admin password


marekjedrzejewski:
When rustdesk is ran as standard user in Windows, when UAC prompt shows up, on the connecting side screen just freezes, and I don't see option to provide password without entering it on the other side. Would be very nice to have this option :)

rustdesk:
Sorry, not understand

rustdesk:
Ok, I got it. But I have no idea how to implement it yet for now. You'd better install it and let it run as service.

marekjedrzejewski:
AnyDesk has option to provide credentials to run its client with. I did some googling and looks like [CreateProcessWithLogonW](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createprocesswithlogonw) might be the function that it's using under the hood.

rustdesk:
Thanks, i will check

rustdesk:
> 企业windows用户都是user，DC管理下的用户运行rustdesk，远程进去后，rustdesk不能升级windows 特权，不会弹出windows管理员的验证窗口，我想问ruskdesk有没有解决这个问题进度。

schleyk:
> In teamviewer when asked for id and password you can insert windows credential instead. those credential are passed from the client to the remote machine server that once receive them restart itself with those credential.

![image](https://user-images.githubusercontent.com/17356113/176906417-b79e1a7a-0f8d-46a4-bcbc-c5353d80910f.png)

lasselin:
Hi,
Same issue here. It makes it useless when the remote user is not admin.
So it would be a nice feature for Rustdesk.
Thanks in advance.
Best regards,

bilogic:
I recall Teamviewer portable being able to control admin apps without asking for UAC, did I remember correctly? 
I meant to ask: how did they do it?

kintaro1981:
They are doing it in the way @schleyk said.

From the client you can choose to authenticate with the "generic" password generated from the server or with "windows authentication" in this way who give support insert local admin credential and click on log on. After this teamviewer server restart it self and launching it self  with the credential provided from the client (like a runas), at this point the client once connected to the server can see and interact with UAC.

bilogic:
Actually I just tested again, TeamviewerQS did request for UAC, that is probably why they are able to control all windows.
Perhaps RustDesk can do the same.

kintaro1981:
[...]

PR #2041, opened 2022-11-10
portable service

21pages:
Run seperate screen capture process(portable service) when click elevate on cm window for uac and elevated forground window.
1. Main processand separate process, exchange data through shared memory, synchronize data with read/write counters, there is four kind of data, 
* capture
* cursor-info
* key
* mouse
2. Whether or not to elevate is entirely up to the controlled end

# to-do
1. ~UAC confirmation timeout is uncertain, exchange ipc server/client~
2. ~Optimize sciter cm ui~
3. ~Flutter cm buttons displayed abnormally on release version~ (not return Expanded)
4. ~Check missing mouse_pos service~ (behavior is same between user and system)
5. ~better prompt~
6. Do more check on shared-memory, use same share memory file all the time if possible, figure out [allow_raw](https://github.com/elast0ny/shared_memory/blob/344a5c410eadb2018f1c4bf5f7c23b8ab2f88089/src/windows.rs#L18) option
7. Do more check on synchronization of state transitions
8. ~Figure out whether dxgi is usable, if usable, try dxgi before gdi~
9. Possibility of atomic operations between two processes
10. More test


rustdesk:
https://github.com/rustdesk/rustdesk/issues/70#issuecomment-860248231

rustdesk:
https://github.com/rustdesk/rustdesk/issues/70#issuecomment-1316445488

rustdesk:
https://github.com/rustdesk/rustdesk/discussions/1870

PR #2058, opened 2022-11-10
Use "_" instead of ":" in PeerConfig filenames

### LINKED ISSUE #1730: Remote is not saved in recent sessions when address contains port number
**Describe the bug you encountered:**

The remote is not saved in 'recent sessions' when connecting to a remote computer with non-standard RustDesk port.

For example: Rustdesk on remote computer is configured to accept IP connection at port 21234. Now when connecting to remote using the address 192.168.1.10:21234, the connection is successful but the peer is not saved in the recent sessions list.

I guess this is because RustDesk saves peers' addresses in filenames, and ':' is a disallowed filename character.

**What did you expect to happen instead?**

I expect the remote address with port number is also saved in recent session list as IP:PORT

**How did you install `RustDesk`?**

Windows Portable

**RustDesk version and environment**

Windows, version 1.1.9

biberino:
Windows doesn't allow ":" in filenames, thats why connections to peers with the portnumer (<ip>:<port>) dont remember the password nor appear in recent sessions.

PeerConfig files on *nix containing ":" will be renamed on the first run to stay backward compatible.

This fixes #1730


rustdesk:
How do you know it is not real _?

rustdesk:
You fix a bug, but introduce a new potential bug.

biberino:
Because the filenames contain only the ID's or IP's, so "_" will never be part of an filename and can be used as a special character.

biberino:
Are hostnames allowed as an ID? If yes then we need to choose another special character instead, something thats not a valid hostname.

rustdesk:
you can use some encoding，e.g. urlencode which handle special char / in url link

rustdesk:
Closed because this solution introduce potential bug, we need a better solution mentioned above.

rustdesk:
@Xerxes-2 could you help refactor this PR?

biberino:
I experiment with base64 encoding to support all future ID strings, because base64 is already a dependency. So far it works, whats missing is backwards compatibility. We need a way to tell if a filename is Base64 encoded or not (and if not, then convert it once). Maybe a special char at the start of the filename indicating that it's encoded?

rustdesk:
> I experiment with base64 encoding to support all future ID strings, because base64 is already a dependency. So far it works, whats missing is backwards compatibility. We need a way to tell if a filename is Base64 encoded or not (and if not, then convert it once). Maybe a special char at the start of the filename indicating that it's encoded?

Good job, you can add a long prefix something like "base64_". But for human friendly, please only encode if there is special char.

rustdesk:
Or you can decode first, if fail to decode, then it is not base64.

libs/hbb_common/src/config.rs
```diff
@@ -871,11 +884,22 @@ impl PeerConfig {
                             .map(|p| p.to_str().unwrap_or(""))
                             .unwrap_or("")
                             .to_owned();
-                        let c = PeerConfig::load(&id);
+
+                        let id_decoded_string: String;
+                        if id.starts_with("base64_") {
+                            let id_decoded = base64::decode(&id[7..], base64::Variant::Original)
```
rustdesk, line 935: potential crash here in extreme case, id == "base64_".

rustdesk:
Also please resolve confilicts first.

n1majne3:
hi, @rustdesk If the hbbs/hbbr uses customzied port, how to write it in filename? Thanks.

rustdesk:
https://github.com/rustdesk/rustdesk/blob/7371b9b66b5edf3e37a3fae3bb86dc6f26b23e37/src/license.rs#L22

PR #2064, opened 2022-11-11
fix mouse cannot be controled, when show remote cursor is on

fufesou:

Ignore intial cursor position when opt "show remote cursor" is on.

### DIFF
--- a/src/server/input_service.rs
+++ b/src/server/input_service.rs
@@ -13,6 +13,8 @@ use std::{
     time::Instant,
 };
 
+const INVALID_CURSOR_POS: i32 = i32::MIN;
+
 #[derive(Default)]
 struct StateCursor {
     hcursor: u64,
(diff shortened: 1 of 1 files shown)


src/server/input_service.rs
```diff
@@ -28,14 +30,27 @@ impl super::service::Reset for StateCursor {
...
 impl super::service::Reset for StatePos {
     fn reset(&mut self) {
-        self.cursor_pos = (0, 0);
+        self.cursor_pos = (INVALID_CURSOR_POS, INVALID_CURSOR_POS);
+    }
+}
+
+impl StatePos {
+    fn is_valid(&self) -> bool {
+        self.cursor_pos.0 != INVALID_CURSOR_POS && self.cursor_pos.1 != INVALID_CURSOR_POS
```
rustdesk: I think only checking `self.cursor_pos.0 != INVALID_CURSOR_POS` is enough, and this function can be inline `#[inline]`

src/server/input_service.rs
```diff
@@ -114,25 +129,27 @@ fn update_last_cursor_pos(x: i32, y: i32) {
...
-                let lock = LATEST_INPUT_CURSOR.lock().unwrap();
-                if now - lock.time < 300 {
-                    lock.conn
-                } else {
-                    0
-                }
-            };
-            sp.send_without(msg_out, exclude);
+        if state.is_valid() {
+            if state.cursor_pos.0 != x || state.cursor_pos.1 != y {
```
rustdesk: Changing to 
```
if state.is_valid() &&
            (if state.cursor_pos.0 != x || state.cursor_pos.1 != y)
```
is better, less code format change, easier for code review.

Issue #2147, opened 2022-11-16
Cursur hotspot not right


KoalaBear84:
**Describe the bug you encountered:**

The cursor hotspot is not right when the cursor is not the default pointer or hand. It just takes 0x0 (top right) of the cursor instead of like 15x15 or 16x16 when it's a crosshair. It was first mentioned in #2126 

Example:
![image](https://user-images.githubusercontent.com/5886900/202117626-e70a00f6-51a6-4c26-bc06-a3544935d040.png)

![image](https://user-images.githubusercontent.com/5886900/202117641-455f7e25-422a-4658-9ddc-ff00cc5f76c8.png)

**What did you expect to happen instead?**

Take the real hotspot of the cursor.

**How did you install `RustDesk`?**

@Kingtous [nightly release](https://github.com/Kingtous/rustdesk/releases/download/nightly/rustdesk-1.2.0-x86_64-pc-windows-msvc.exe), installed.

**RustDesk version and environment**

**Host**
Windows 10 Pro 21H2 (OS Build 19044.2251)
@Kingtous [nightly release](https://github.com/Kingtous/rustdesk/releases/download/nightly/rustdesk-1.2.0-x86_64-pc-windows-msvc.exe), installed.

**Client**
Windows 10 Enterprise 21H2 (OS Build 19044.2251)
Rustdesk nightly build https://github.com/rustdesk/rustdesk/commit/52ce3dd2c299e262a54af0b5350fa60f66215e60, installed

rustdesk:
@fufesou

fufesou:
https://github.com/rustdesk/rustdesk/pull/2154

KoalaBear84:
It looks like #2154 fixed it in the latest nightly! 👍 Thanks!

Amazing how this video is only 35 kB, but this one doesn't display in GitHub inline. 🤯
https://user-images.githubusercontent.com/5886900/202447808-d6429ae4-c345-4679-9537-d146c1043648.mp4

ForsakenRei:
> It looks like #2154 fixed it in the latest nightly! 👍 Thanks!
> 
> Amazing how this video is only 35 kB, but this one doesn't display in GitHub inline. 🤯 https://user-images.githubusercontent.com/5886900/202447808-d6429ae4-c345-4679-9537-d146c1043648.mp4

I think github only takes GIFs...

KoalaBear84:
It took another mp4, which I also recorded with ShareX in #2180

[...]

PR #2154, opened 2022-11-16
Fix cursor

fufesou:
1. Fix cursor (hotx, hoty).
2. Add option, zoom cursor.
3. Show menu when conn is established.


rustdesk:
#2147

Issue #2146, opened 2022-11-16
White outline (in fullscreen)


KoalaBear84:
**Describe the bug you encountered:**

A white outline appears around the whole 'screen' of the client. Sometimes enlarging on the bottom, which resets itself to 1 pixel after pressing "Reset canvas". It was first mentioned in #2126

1 pixel:
![image](https://user-images.githubusercontent.com/5886900/202116416-6eaae0f4-41f8-4a56-afa1-20ca50873f8a.png)

More outline at the bottom (open in new tab to see it if GitHub has white background):
![image](https://user-images.githubusercontent.com/5886900/202116457-d382c9e2-e78a-4947-b920-440eb364e54a.png)

**What did you expect to happen instead?**

No outline, just black like the rest of the 'fill'.

**How did you install `RustDesk`?**

@Kingtous [nightly release](https://github.com/Kingtous/rustdesk/releases/download/nightly/rustdesk-1.2.0-x86_64-pc-windows-msvc.exe), installed.

**RustDesk version and environment**

**Host**
Windows 10 Pro 21H2 (OS Build 19044.2251)
@Kingtous [nightly release](https://github.com/Kingtous/rustdesk/releases/download/nightly/rustdesk-1.2.0-x86_64-pc-windows-msvc.exe), installed.

**Client**
Windows 10 Enterprise 21H2 (OS Build 19044.2251)
Rustdesk nightly build https://github.com/rustdesk/rustdesk/commit/52ce3dd2c299e262a54af0b5350fa60f66215e60, installed

rustdesk:
Outline is expected if not full screen, but enlarged outline is unexpected.

rustdesk:
@fufesou

KoalaBear84:
Thank you for fixing! Much appreciated! 👍

fufesou:
>enlarging on the bottom
@KoalaBear84 Hi, I'm not sure what it means. Is it like this?
![image](https://user-images.githubusercontent.com/13586388/202196741-ce0c2001-620c-452e-907f-dfd867ac4f7b.png)

KoalaBear84:
Sorry, it wasn't fully clear. This is because my host monitor is 1920 x 1200, the laptop client is 1920 x 1080. But after a while, sometimes the white outline at the bottom grows to in this case 8 pixels (select the text around the image, or open it in a new tab):
[...]

PR #2160, opened 2022-11-16
fix window border width whenn fullscreen

fufesou:

https://github.com/rustdesk/rustdesk/issues/2146
Set edge width to 0 when fullscreen.
But [bottom outline growing](https://github.com/rustdesk/rustdesk/issues/2146#issuecomment-1317078953) problem is not fixed.


rustdesk:
#2146 mentioned two issues

PR #2183, opened 2022-11-17
feat: add build date on about page

Kingtous:
This PR adds a build date to show on about page, so users/devs can know which version is excatly in use.

image


giuliovale:
Grande

PR #2205, opened 2022-11-18
Feat dylib virtual display

fufesou:
1. Windows. Dynamic library for virtual display.


libs/clipboard/build.rs
```diff
@@ -18,7 +18,7 @@ fn build_c_impl() {
     if build.get_compiler().is_like_msvc() {
         build.define("WIN32", "");
         // build.define("_AMD64_", "");
-        build.flag("-Zi");
+        build.flag("-Z7");
```
fufesou, line 21: "Z7" flag may cause clipboard not work.

PR #2206, opened 2022-11-18
quick start portable ci

21pages:
1. Add quick start portable ci. QuickStart will request elevation at startup, and run portable service all the time.
2. Fix cleanup shared memory, at_exit() will not be called on flutter , clean it on window close callback.

to-do:
1. It will take more time to shutdown the app. If it is started immediately after closing, the previous process has not ended, and there will be problems such as not going online and random passwords not being displayed. It's happened when portable service process is running, maybe the main program ipc-server needs to be optimized.

### DIFF
--- a/.github/workflows/flutter-nightly.yml
+++ b/.github/workflows/flutter-nightly.yml
@@ -15,7 +15,7 @@ env:
 
 jobs:
   build-for-windows:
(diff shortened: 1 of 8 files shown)


rustdesk:
todo: replace feature with filename way like install

PR #2281, opened 2022-11-22
Support CIDR for IP whitelist

Xerxes-2:
allow IPv4 & IPv6 + CIDR in whitelist, will be checked as CIDR


rustdesk:
Could you also add IPv6 CIDR support? IPv6 is ready soon in RustDesk.

Xerxes-2:
IPv6 CIDR support add

MarleTangible:
Entering IPv4 CIDR block result an "Invalid IP" error. Am I missing something?
image

Xerxes-2:
> Entering IPv4 CIDR block result an "Invalid IP" error. Am I missing something? image

Could you please paste your IP address here, I don't encounter error with 192.168.0.0/24

MarleTangible:
I was using "10.100.0.0/16" with the latest stable release dated May 2, 2022. Looks like the change was merged on Nov 22, 2022. Could it be that the latest release does not currently have the CIDR support yet?

Issue #2180, opened 2022-11-17
Blurry screen [nightly]


KoalaBear84:
**Describe the bug you encountered:**

The 'remote' screen is blurry. It looks like it's depending on the odd or even height of the RustDesk window, and maybe also the width.

https://user-images.githubusercontent.com/5886900/202380107-1ac5954d-7be1-40a7-9173-e835f1e3d740.mp4

![image](https://user-images.githubusercontent.com/5886900/202380075-58c9a8dc-2885-4671-93b6-aa217a8e03e0.png)

**What did you expect to happen instead?**

Always clear screen when no scaling.

**How did you install `RustDesk`?**

**RustDesk version and environment**

Latest nightly on this repo, just run.

**Host**
Windows 10 Enterprise 22H2 (OS Build 19045.2251)
GitHub release, [latest nightly](https://github.com/rustdesk/rustdesk/releases/download/nightly/rustdesk-1.2.0-x86_64-pc-windows-msvc.exe) of 7 hours ago

**Client**
Windows 10 Pro 21H2 (OS Build 19044.2251)
@Kingtous nightly release, installed.

fufesou:
Thanks. @KoalaBear84  
I can't reproduce this problem. But I almost know where the problem is. I'll read the dev doc the make the picture better.

KoalaBear84:
Great. Not sure if it has anything to do with the screen sizes.

Host: 2560 x 1440
Client: 1920 x 1200

rustdesk:
If even number has better quality, we can always scale to even width/height, remaining one pixel as padding. let's test first, and also try painting argument adjustment first.

fufesou:
@KoalaBear84 Hi, would you please try to test if it is blured when you only change the height or width of the window?
The picture is kept aspect ratio. Changing height may not cause rescaling if window width is not big enough.
I just want to make sure if the problem exists without rescaling the picture.

KoalaBear84:
I hope this is the information you are looking first. At first the Rustdesk window is larger than the clients window, but it still blurs it. And also after the RustDesk window is smaller than the clients desktop, it keeps on blurring.

https://user-images.githubusercontent.com/5886900/202650257-a1f192a2-3a14-459a-8154-8ea58ac409c7.mp4

rustdesk:
bluring is expected if picture is scaled. Please compare with 1.1.9 with the same scale.

fufesou:
[...]

PR #2395, opened 2022-11-30
fix image blurry

fufesou:

https://github.com/rustdesk/rustdesk/issues/2180


fufesou:
(top,left) must be both integers, or Flutter cannot handle image painting correctly.
This problem persists until 3.6.10 pre.

PR #2397, opened 2022-11-30
Added Albanian (Shqip) Language

donald-2rm:
Soon all the phrases to be translate are to be added by us


rustdesk:
@donald-2rm you should follow https://github.com/rustdesk/rustdesk/blob/master/src/lang/template.rs?  Anyway, thanks for your job.

PR #2406, opened 2022-12-01
Update remote_menubar.dart

Agent-JY:
Change 'Close' to 'Close connection' due to conflicts with 'Close' in context of close window.

### DIFF
--- a/flutter/lib/desktop/widgets/remote_menubar.dart
+++ b/flutter/lib/desktop/widgets/remote_menubar.dart
@@ -498,7 +498,7 @@ class _RemoteMenubarState extends State<RemoteMenubar> {
 
   Widget _buildClose(BuildContext context) {
     return IconButton(
-      tooltip: translate('Close'),
+      tooltip: translate('Close connection'),
       onPressed: () {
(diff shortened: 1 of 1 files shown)


rustdesk:
"Close the connection"?

rustdesk:
Or "close this connection"?

Agent-JY:
Btw woudn't it be better to use specific names for all lables?
For example:

`translate("Connection.Toolbar.Close")`

and in [rustdesk/src/lang/en.rs](https://github.com/rustdesk/rustdesk/blob/master/src/lang/en.rs)
`("Connection.Toolbar.Close", "Close connection"),`

That wound make it more confortable to translate it.

rustdesk:
> translate("Connection.Toolbar.Close")

I know this style, but I did not like.

Agent-JY:
You didn't change 'Close' to 'Close connection'? Why?

rustdesk:
> Why

* This is not nessessary, I believe users are clever enough to know this "close" means "close the connection".
* I do not know if "Close connection" is correctly, I am not native English speaker either.
* Translate

Anyway, thanks for your effort.

PR #2517, opened 2022-12-11
refactor: migrate flutter_custom_cursor to the ported version of flutter engine

Kingtous:
Note: 
Devs need to upgrade the local flutter engine with https://github.com/Kingtous/engine/releases/tag/v3.0.5-rustdesk.2

### DIFF
--- a/flutter/lib/desktop/pages/remote_page.dart
+++ b/flutter/lib/desktop/pages/remote_page.dart
@@ -5,6 +5,8 @@ import 'dart:ui' as ui;
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
+import 'package:flutter_custom_cursor/cursor_manager.dart'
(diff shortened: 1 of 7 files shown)


flutter/lib/desktop/pages/remote_page.dart
```diff
@@ -109,17 +111,17 @@ class _RemotePageState extends State<RemotePage>
...
-          (String? lastKey, String? currentKey) async {
-        if (_firstEnterImage.value) {
-          _firstEnterImage.value = false;
-          return true;
-        }
-        return lastKey == null || lastKey != currentKey;
-      });
-      _isCustomCursorInited = true;
-    }
+    // if (!_isCustomCursorInited) {
```
Kingtous, line 114: @fufesou I don't know what this logic means. After migrating the API to the newest version of custom cursor api, it seems that it's no need to do with this logic. Instead of deleting it, I comment it temporarily.

PR #2556, opened 2022-12-15
Serbian language added

jboban:
Messages translated to Serbian language.


rustdesk:
incomplete, and you overwrote the other languages' changes.

PR #2586, opened 2022-12-17
add tab about translate

jimmyGALLAND:
Proposal to allow translations of 'About Tab'


jimmyGALLAND:
Sorry there is an error in the template.rs format

PR #2602, opened 2022-12-19
fix reconnect

21pages:
1. fix reconnect when reset by the peer
2. fix improper 10054 reconnect

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -892,6 +892,8 @@ pub struct LoginConfigHandler {
     pub supported_encoding: Option<(bool, bool)>,
     pub restarting_remote_device: bool,
     pub force_relay: bool,
+    pub direct: Option<bool>,
+    pub received: bool,
 }
 
 impl Deref for LoginConfigHandler {
@@ -929,6 +931,8 @@ impl LoginConfigHandler {
         self.supported_encoding = None;
         self.restarting_remote_device = false;
         self.force_relay = !self.get_option("force-always-relay").is_empty();
+        self.direct = None;
+        self.received = false;
     }
 
     /// Check if the client should auto login.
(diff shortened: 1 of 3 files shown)


fragtion:
Today I got the "Connection error" "An existing connection was forcibly closed by the remote host. (os error 10054)" on an already established session that was in viewer mode.  The client did not automatically try to reconnect the session, it had to be done manually. Is there any way to handle this reconnect retrying automatically?

21pages:
@fragtion when you try to start a connection?

fragtion:
> @fragtion It happened when you try to start a connection?
No, connection was already established for several days. Something caused an interruption (could have been the internet connection) and then the error popped up.  But when this specific error occurs, the client does not try to reconnect automatically. it would be nice to change this behavior if required

Client is the win32 sciter version in this case but I think the same applies for the other builds

rustdesk:
A video please if you want us consider your proposal seriouly. It will help us understand your scenario.

21pages:
In the current program, when encountering error code 10054, if it is a direct connection and no data is received, it will attempt to relay reconnect. If not, error code 10054 will not trigger a reconnection attempt.

fragtion:
You really need a video as proof that the "Connection error" "An existing connection was forcibly closed by the remote host. (os error 10054)" popup prevents auto-reconnection? C'mon guys.. How will that be different from just sharing a screenshot of the error window? Surely not even a screenshot is necessary for something like this. But fine, I will go through all the effort of reproducing this if that is really what you need

fragtion:
> In the current program, when encountering error code 10054, if it is a direct connection and no data is received, it will attempt to relay reconnect. If not, error code 10054 will not trigger a reconnection attempt.

For the record:
It was a direct IP connection.
It did not attempt to reconnect in any way.

rustdesk:
> different from just sharing a screenshot of the error window

A lot of difference.

Issue #2587, opened 2022-12-17
Android http API server problem


jboban:
Could not enter any http (not https) API server. The same server could be used on Windows platform. Have tried with stable 1.1.9 version and nightly build 1.2.0.

![Screenshot_20221217_165734](https://user-images.githubusercontent.com/11806707/208250679-4a981a03-a7ca-4a0a-98f5-2ad648a2fb49.png)
![Screenshot_20221217_170030](https://user-images.githubusercontent.com/11806707/208250699-14d924c3-7fc3-4601-8d35-d75aeceba663.png)

jboban:
What is wrong with port here, please? API server is up and running.

rustdesk:
@Heap-Hop

PR #2648, opened 2022-12-26
fix mobile api server validation bug

Heap-Hop:
https://github.com/rustdesk/rustdesk/issues/2587


flutter/lib/mobile/widgets/dialog.dart
```diff
@@ -264,7 +264,6 @@ void showServerSettingsWithValue(
         if (relayServerMsg != null) return false;
       }
       if (apiCtrl.text != oldCfg.apiServer) {
-        apiServerMsg = await validateAsync(apiCtrl.text);
```
rustdesk, line 267: give up validation?

Issue #981, opened 2022-07-12
Cannot wake up Mac Mini from sleep?


leonguyen52:
Hi, I'm using RustDesk on Android/iPhone app and tried to wake up my Mac Mini M1 which is running RustDesk for Mac v1.1.9 but I cannot wake it up to control from sleeping. I can use chrome remote desktop to wake it normally. Any suggestion?

drkvogel:
Same here with MacBook Pro 2018 host MacOS 10.14.6 and Linux Ubuntu client

tamiaO:
我使用 macBookPro 连接到 mac mini 时也存在此 bug，只要是 mac 睡眠或是锁定屏幕了，就无法成功连接到远程电脑。具体表现为（客户端提示）连接成功，等待画面传输，此时，用其它任何客户端都是连不上远程的

Seriels:
![EV1X42~%C}2DZL9981KNG(7](https://user-images.githubusercontent.com/33693501/187673838-57a83ac2-aefa-4c3e-be33-d7ef723019d8.png)


咋看日志。。。

Seriels:
>  
mac 是 kvm  安装的 系统    manjaro  是宿主机器
mac 连接 manjaro  是没有问题的   但是 连回到 mac 就 一直卡在 已连接 等待画面传输

mihauOne87:
Same here, Black Screen when connected to Sleeped iMac.

mihauOne87:
RustDesk is useless on MacBook After sleep Mode. TeamView working without Problems. Is the Team Plan some fixes for this Problem?

neteric:
Same issue, Black Screen when connected to Sleeped iMac.

neteric:
Same issue, can't wake up black screen when connected to sleeped macos.

ondohotola:
Same/Similar issue, I expect the login window (due to inactivity) but can ‘t get that.
On Windows it works very well, and when logged into the Mac it does as well.

Latest versions on all devices.

fisherwei:
我尝试了一下先用macos自带的vnc端口连接上，晃动一下鼠标点亮显示器，然后rustdesk就可以正常工作了。

能否考虑先简单的解决一下：服务端收到链接请求之后，让鼠标晃动1px，点亮睡眠的显示器？

rustdesk:
> vnc

@asur4s good clue，we can check out osxvnc.

KarryZhou:
[...]

PR #2678, opened 2022-12-29
Mac wakeup workaround

fazakis:
Introduced a call to "caffeinate" as suggested by @bellwether-softworks

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -96,8 +96,8 @@ pub fn new() -> ServerPtr {
 fn mac_wakeup(){
     #[cfg(target_os = "macos")]{
         use std::process::Command;
-        Command::new("/usr/bin/caffeinate").arg("-u").arg("-t 5").spawn().expect("failed to execute caffeinate");
-        println!("wake up macos");
+        Command::new("/usr/bin/caffeinate").arg("-u").arg("-t 5").spawn().ok();
+        log::info!("wake up macos");
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -93,6 +93,14 @@ pub fn new() -> ServerPtr {
     Arc::new(RwLock::new(server))
 }
 
+fn mac_wakeup(){
+    #[cfg(target_os = "macos")]{
+        use std::process::Command;
+        Command::new("/usr/bin/caffeinate").arg("-u").arg("-t 5").spawn().expect("failed to execute caffeinate");
```
rustdesk: `expect("")` may panic. please remove it. just use `ok()`;
https://doc.rust-lang.org/std/result/enum.Result.html#method.expect

src/server.rs
```diff
@@ -93,6 +93,14 @@ pub fn new() -> ServerPtr {
...
+        println!("wake up macos");
```
rustdesk: use `log::info!` instead of println! please.

src/server.rs
fazakis: Thanks

src/server.rs
fazakis: Thanks

src/server.rs
```diff
@@ -252,6 +261,7 @@ async fn create_relay_connection_(
         ..Default::default()
     });
     stream.send(&msg_out).await?;
+    mac_wakeup();
```
rustdesk: Why don't you put `mac_wakeup` in `create_tcp_connection`?

fazakis: sorry almost blind

rustdesk:
https://github.com/rustdesk/rustdesk/issues/981

PR #2701, opened 2023-01-02
ignore style warnings in libs/scrap

amyipdev:
Constant `is_cursor_embedded` does not follow the Rust standard stylistic convention of upper-case global variables and constants. This causes two warnings to be thrown when compiling (tested on Arch Linux, commit = 68fda34, Rust = 1.66.0), one each for the Wayland and X11 common modules. Since these variables are not new, their names should not be modified; to remove the warnings, this patch allows non-style-conforming names on these two constant declarations specifically, suppressing the warnings. It does not affect stylistic warnings on any other code within the project.

Signed-off-by: Amy Parker <apark0006@student.cerritos.edu>
Cc: fufesou <shuanglongchen@yeah.net>

### DIFF
--- a/libs/scrap/src/common/wayland.rs
+++ b/libs/scrap/src/common/wayland.rs
@@ -4,6 +4,7 @@ use std::{io, sync::RwLock, time::Duration};
 
 pub struct Capturer(Display, Box<dyn Recorder>, bool, Vec<u8>);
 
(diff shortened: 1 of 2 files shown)


rustdesk:
Could you help change is_cursor_embedded to upper-case?

PR #2703, opened 2023-01-03
Update README.md

if0else9:


if0else9:
May we thank you for your product in the news on our site?  nir.gov.ua.  I
would very much like the public server to work in Kyiv.  RustDesk works
very badly in Ukraine so far :-(

вт, 3 січ. 2023, 10:40 користувач RustDesk ***@***.***> пише:

PR #2709, opened 2023-01-03
Updated the macOS tray icons to vector files (SVG)

NoLooseEnds:
Updated the macOS tray icons to vector files (SVG). Just changed the files and renamed the paths.


rustdesk:
Have you tested? BTW, this solution only works for sciter version, for flutter version, Cargo.toml resource does not work, you need to put resource files somewhere in `flutter/`.

NoLooseEnds:
No, sorry, not properly tested. I have no idea how to. All the testing I've done is to manually change the file on the macos build to check size etc.

rakleed:
@rustdesk I could check by building the project, but not in the current 1.1.9 release version, nor in the nightly 1.2.0 version, I can't get the RustDesk icon to appear in the tray (menu bar) on macOS. When should it appear? On Windows, it appears after installation as a Windows Service, but on macOS, the application is immediately considered installed after moving it to `/Applications`, which I did.

сс @Kingtous

P.S. Looks like Flutter [doesn't support system tray out of the box](https://stackoverflow.com/q/64305453/11336855), third party libraries are required.

rustdesk:
Thanks @rakleed

rustdesk:
This PR does not work, the third party library we are using does not support svg.

Kingtous:
> 

on macOS, we should find a way from creating a `NSImage` object from a svg file, which accepts by macOS SDK. i.e., with a `NSImage`, we can easily create a tray icon from the svg. Maybe a `SVGKit` macos plugin (load and use `SVGKExporterNSImage` to get a OSX `NSImage`) or something needs to be added. And we need to render it first to NSImage ahead and then set the tray icon. I dont know whether the refactor is necessary, considering the more complexity of setting a macOS tray icon, and also the time cost of rendering svg.

> I could check by building the project, but not in the current 1.1.9 release version, nor in the nightly 1.2.0 version, I can't get the RustDesk icon to appear in the tray (menu bar) on macOS. 

Also, I've found [native svg support has been added in XCode 12](https://developer.apple.com/forums/thread/113015), which supports macOS 10.15+. But a valid NSImage seems cannot directly be loaded from a svg file.

rustdesk:
> the time cost

No time to do this for the time being.

NoLooseEnds:
Ah, too bad. You could consider updating the PNG that is there to a higher resolution one. It will make it look better at least. I've done that manually (just swapping out the png inside the .app) and it looks way better.

image

rustdesk:
@NoLooseEnds thanks, can you upload your new png files?

NoLooseEnds:
[mac_tray_48px.zip](https://github.com/rustdesk/rustdesk/files/10343107/mac_tray_48px.zip)

rustdesk:
@NoLooseEnds you can submit a PR to overwrite these two png files.

NoLooseEnds:
@rustdesk Ok, done. https://github.com/rustdesk/rustdesk/pull/2717

PR #2743, opened 2023-01-06
windows extended key

### LINKED ISSUE #2711: Shift + Arrow key didn't work properly on remote side
**Describe the bug you encountered:**

When holding Shift, arrow keys(left, right, up and down) didn't select text on remote side in a text editor.

**What did you expect to happen instead?**

Text selected as usual

**How did you install `RustDesk`?**

1.2.0 nightly build date 1/3/2022

**RustDesk version and environment**

1.2.0 nightly build on both host and remote, self hosted server.

sj6219:
fix #2711


rustdesk:
@fufesou please check if we need these changes in rdev? because our flutter version (map mode) is using rdev.

PR #2785, opened 2023-01-10
fix wayland input

fufesou:

1. Disable flutter key handle if rdev is enabled.
2. Use flutter to get lock modes, instead of read https://github.com/rustdesk/rustdesk/blob/367de991f6ddf9068c9b8d71ede2c18ac6892301/libs/enigo/src/linux/nix_impl.rs#L186


rustdesk:
https://github.com/rustdesk/rustdesk/issues/56

PR #2810, opened 2023-01-12
Allow setting custom server and key with env variables

botanicvelious:
1. Code edited to allow the setting of a custom server with env variables

RENDEZVOUS_SERVER1
RENDEZVOUS_SERVER2
RENDEZVOUS_SERVER3

2. Allows editing custom key with 

RS_PUB_KEY_VAL

3. Workflows updated to support adding these custom env variables via secrets.

4. Workflows updated to allow ignoring key signing if specific secrets are set.

5. Interface file checks hbb_common::config::RS_PUB_KEY to decide if to display "run your own server" message rather then the existence of custom-rendezvous-server

6. Added flutter-custom-build.yml so that people can use github workflows to generate there own files using github actions


botanicvelious:
Only part I am not super happy with is the RENDEZVOUS_SERVER1-3 values... couldn't figure out any other way to do it more cleanly however... if anyone has any ideas

.github/workflows/flutter-nightly.yml
```diff
@@ -15,6 +15,14 @@ env:
   # for multiarch gcc compatibility
   VCPKG_COMMIT_ID: "14e7bb4ae24616ec54ff6b2f6ef4e8659434ea44"
   VERSION: "1.2.0"
+  #To make a custom build with your own servers set the below secret values
```
rustdesk: space after `#` please, including below.

.github/workflows/flutter-nightly.yml
```diff
@@ -150,18 +158,21 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Import the codesign cert
+        if: ${{ env.NO_OSX_KEYS== 'true' }}
```
rustdesk: `env.NO_OSX_KEYS==` vs `env.NO_OSX_KEYS ==`?

why if `env.NO_OSX_KEYS`? should not be if `env.HAS_OSX_KEYS`?

.github/workflows/flutter-nightly.yml
```diff
@@ -15,6 +15,14 @@ env:
   # for multiarch gcc compatibility
   VCPKG_COMMIT_ID: "14e7bb4ae24616ec54ff6b2f6ef4e8659434ea44"
   VERSION: "1.2.0"
+  #To make a custom build with your own servers set the below secret values
+  RS_PUB_KEY_VAL: '${{ secrets.RS_PUB_KEY_VAL }}'
+  RENDEZVOUS_SERVER1: '${{ secrets.RENDEZVOUS_SERVER1 }}'
+  RENDEZVOUS_SERVER2: '${{ secrets.RENDEZVOUS_SERVER2 }}'
+  RENDEZVOUS_SERVER3: '${{ secrets.RENDEZVOUS_SERVER3 }}'
+  #ignore signing with key files if values below are set
+  NO_OSX_KEYS: ${{ secrets.NO_OSX_KEYS == 'true' }}
```
rustdesk: No need to add `secrets.NO_OSX_KEYS`, you can check available keys `MACOS_P12_BASE64`.

.github/workflows/flutter-nightly.yml
```diff
@@ -15,6 +15,14 @@ env:
...
+  NO_APP_KEYS: ${{ secrets.NO_APP_KEYS == 'true' }}
```
rustdesk: the same problem as above

libs/hbb_common/src/config.rs
```diff
@@ -77,12 +77,28 @@ const CHARS: &'static [char] = &[
...
+        Some(key) => key,
+        None => "rs-ny.rustdesk.com",
+    },
+    match option_env!("RENDEZVOUS_SERVER2") {
+        Some(key) => key,
+        None => "rs-sg.rustdesk.com",
+    },
+    match option_env!("RENDEZVOUS_SERVER3") {
+        Some(key) => key,
+        None => "rs-cn.rustdesk.com",
```
rustdesk: If `RENDEZVOUS_SERVER2 ` is not set, still use rustdesk server? looks wierd.

src/ui_interface.rs
```diff
@@ -243,7 +243,11 @@ pub fn set_peer_option(id: String, name: String, value: String) {
 
 #[inline]
 pub fn using_public_server() -> bool {
-    crate::get_custom_rendezvous_server(get_option_("custom-rendezvous-server")).is_empty()
+    if hbb_common::config::RS_PUB_KEY == "OeVuKk5nlHiXp+APNn0Y3pC1Iwpwn44JGqrQCsWqmBw=" {
+        return true
+    } else {
+        return false
+    }
```
rustdesk, line 250: I do not think. this is necessary.

.github/workflows/flutter-nightly.yml
botanicvelious: The issue with that key is if you do not have a key from apple to use it still tries to use no value and fails, this skips the signing process entirely

.github/workflows/flutter-nightly.yml
botanicvelious: It only skips the signing if that secret value is set, I can rename it if you want.

.github/workflows/flutter-nightly.yml
botanicvelious: Fixed

.github/workflows/flutter-nightly.yml
botanicvelious: So my issue here is that due to the super limited logic in global scope, I haven't been able to come up with a way that works for this, I am open to any ideas...

libs/hbb_common/src/config.rs
botanicvelious: That is correct see above

.github/workflows/flutter-nightly.yml
botanicvelious: Updated them to SKIP_APP_KEYS and SKIP_OSX_KEYS so its more clear what it does

.github/workflows/flutter-nightly.yml
botanicvelious: I have renamed them to SKIP_APP_KEYS and SKIP_OSX_KEYS so its more clear what it does

.github/workflows/flutter-nightly.yml
botanicvelious: Ah I understand what you want here, I will change it

src/ui_interface.rs
botanicvelious: So I just tested and verified, with the code the way it was before it put that message at the bottom when using github actions to build at least.

libs/hbb_common/src/config.rs
botanicvelious: This item the issue I am running into is there is limited logic i can do in global scope, I cant figure out another way to do it... I am open to ideas but I have tried a bunch of things and rust just limits the global scope too much

botanicvelious:
All items should be resolved except for the secrets.RENDEZVOUS_SERVER1-3 I cant come up with a way to do that in global scope, would be good to be able to make it take a string that it splits or otherwise but i cant seem to get that to work in rust global

rustdesk:
> All items should be resolved except for the secrets.RENDEZVOUS_SERVER1-3 I cant come up with a way to do that in global scope, would be good to be able to make it take a string that it splits or otherwise but i cant seem to get that to work in rust global

you can write your macro function to accept RENDEZVOUS_SERVERS string, split in macro to vec, maybe you can search if there is such macro, otherwise, pay some time to write

src/ui_interface.rs
```diff
@@ -243,7 +243,11 @@ pub fn set_peer_option(id: String, name: String, value: String) {
 
 #[inline]
 pub fn using_public_server() -> bool {
-    crate::get_custom_rendezvous_server(get_option_("custom-rendezvous-server")).is_empty()
+    if hbb_common::config::RS_PUB_KEY == "OeVuKk5nlHiXp+APNn0Y3pC1Iwpwn44JGqrQCsWqmBw=" {
```
rustdesk: Put "OeVuKk5nlHiXp+APNn0Y3pC1Iwpwn44JGqrQCsWqmBw=" here again is not good, we need a clever way (macro again) to avoid duplication.

rustdesk (changes requested):

src/ui_interface.rs
botanicvelious: so I did it via checking at compile time for the existence of the env variable as well as checking for the custom server as before, that should resolve this concern

rustdesk: There should be only one `OeVuKk5nlHiXp+APNn0Y3pC1Iwpwn44JGqrQCsWqmBw=` in the source code.

src/ui_interface.rs
```diff
@@ -243,7 +243,12 @@ pub fn set_peer_option(id: String, name: String, value: String) {
 
 #[inline]
 pub fn using_public_server() -> bool {
-    crate::get_custom_rendezvous_server(get_option_("custom-rendezvous-server")).is_empty()
+    let key_check: Option<&'static str> = option_env!("RS_PUB_KEY_VAL");
```
rustdesk: Honestly, `crate::get_custom_rendezvous_server(get_option_("custom-rendezvous-server")).is_empty()` has bug, you last change fixed the bug, but we need to get rid of duplication.

rustdesk (changes requested):

src/ui_interface.rs
botanicvelious: I believe this would work here, as the reason it seemed to me that it didn't work was because the configuration value wasn't set, this checks both the config value as well as the env var being set so that should cover every case? Could just check the env var i suppose but then the message would show up if someone put in a config value, i am still testing everything however to see if i missed anything

libs/hbb_common/src/config.rs
```diff
@@ -77,21 +77,13 @@ const CHARS: &'static [char] = &[
...
-    match option_env!("RENDEZVOUS_SERVER3") {
-        Some(key) => key,
-        None => "rs-cn.rustdesk.com",
-    },
-];
+//check for env variable RENDEZVOUS_SERVER if not use the default
+pub const RENDEZVOUS_SERVERS: [&'static str;3] = 
+    match option_env!("RENDEZVOUS_SERVER") {
+        Some(key) => [key,key,key],
+        None => ["rs-ny.rustdesk.com","rs-sg.rustdesk.com","rs-cn.rustdesk.com"],
```
rustdesk, line 84: Not following rust code format.

rustdesk: [key] is ok, no need to be 3 items.

src/ui_interface.rs
rustdesk: `crate::get_custom_rendezvous_server(get_option_("custom-rendezvous-server")).is_empty()` is buggy, because there is another way to define `cutom rendezvous server` in addtion to `custom-rendezvous-server` option. Comparing pub key is the most safest and simplest way. SInce you are modifing here, I hope you can fix this bug together. :)

rustdesk (changes requested):

libs/hbb_common/src/config.rs
botanicvelious: fixed

src/ui_interface.rs
botanicvelious: I added a default key option and used that, should resolve the issue

botanicvelious:
Just need to test the changes but i think that should probably do it, you mentioned that there was some rust formatting issues, what do you use to format your code?

antrys:
The docs on the website, or somewhere, really need to be updated to reflect the 1,2,3 being needed.
Based on online docs, I've spent the past week on & off trying to build nightly with RENDEZVOUS_SERVER env and wondering why i couldn't get it work.
I stumbled in here completely on accident.

botanicvelious:
The 1,2,3 are not needed, the code has been changed

Issue #126, opened 2021-06-27
Connect via domain name


initdc:
I tried to connect to a remote desktop using a public IP and it worked.

I have also deployed DDNS service, can you add the function of connecting via domain name? 
So that users can get higher upload rate and better experience.

rustdesk:
noted with thanks，but a small reminder, explosure to internet is not safe.

rustdesk:
Marked as wontfix before I can be convinced by anyone.

PR #2851, opened 2023-01-18
Allow direct connect to {hostname}:{port}

### LINKED ISSUE #2847: Allow direct connect to {hostname}:{port}
Currently, all the clients only supports direct connection to {ip}:{port} format, and lacks support for {hostname}:{port} format.

I have a pc with dynamic public ip, which will be bound to a hostname via ddns, and I want to connect to it with direct connection.
It will be really convenient to be able to connect with {hostname}:{port}.

AutumnSun1996:
Allow direct connect to {hostname}:{port}
Like the direct access by ip, it will now check the `peer` string with regex, and treat it as {hostname}:{port} if matched.
PS: Current regex is: `^[\-.0-9a-zA-Z]+:\d{1,5}$`

This PR should fix #2847 and #126

### DIFF
--- a/libs/hbb_common/src/lib.rs
+++ b/libs/hbb_common/src/lib.rs
@@ -347,19 +347,20 @@ mod test_lib {
         assert_eq!(is_ipv6_str("[1:2::0]:"), false);
         assert_eq!(is_ipv6_str("1:2::0]:1"), false);
     }
+
     #[test]
     fn test_hostname_port() {
-        assert_eq!(is_ipv6_str("a:12"), true);
-        assert_eq!(is_ipv6_str("a.b.c:12"), true);
-        assert_eq!(is_ipv6_str("test.com:12"), true);
-        assert_eq!(is_ipv6_str("1.2.3:12"), true);
-        assert_eq!(is_ipv6_str("a.b.c:123456"), false);
+        assert_eq!(is_hostname_port_str("a:12"), true);
+        assert_eq!(is_hostname_port_str("a.b.c:12"), true);
+        assert_eq!(is_hostname_port_str("test.com:12"), true);
+        assert_eq!(is_hostname_port_str("1.2.3:12"), true);
+        assert_eq!(is_hostname_port_str("a.b.c:123456"), false);
         // todo: should we also check for these edge case?
         // out-of-range port
-        assert_eq!(is_ipv6_str("test.com:0"), true);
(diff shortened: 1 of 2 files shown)


libs/hbb_common/src/lib.rs
```diff
@@ -321,6 +321,13 @@ pub fn is_ip_str(id: &str) -> bool {
     is_ipv4_str(id) || is_ipv6_str(id)
 }
 
+#[inline]
+pub fn is_hostname_port_str(id: &str) -> bool {
+    regex::Regex::new(r"^[\-.0-9a-zA-Z]+:\d{1,5}$")
```
rustdesk: No, this is not correct URL regex.

rustdesk (changes requested):

libs/hbb_common/src/lib.rs
AutumnSun1996: This regex is for hostname, not only url(or domain names).
Hostname like `B3F-W-12345`, `B3F-W-12345.sc` are also accepted here.

May you provide a better regex, or some other requirements?
e.g. `some-string` should be rejected, `some-other-string` should be accepted.

rustdesk: URL only please for the time being. for pure hostnane, we may add some prefix.

rustdesk: BTW, `1.1.1.1:90` also pass through `is_hostname_port_str`. Please use a serious URL regex.

AutumnSun1996: Regex updated.
And for the prefix, is there any advice?

rustdesk: I still think this PR can cause me some trouble in the future. Give me some time.

rustdesk: I mean about prefix for domain.

rustdesk: I have no good idea also, let's do it later.

rustdesk:
BTW, intranet hostname actually does not work, our code can not get ip from intranet hostname, we currently use dns lookup.

libs/hbb_common/src/lib.rs
rustdesk: BTW, intranet hostname actually does not work, our code can not get ip from intranet hostname, we currently use dns lookup. So we do not need to discuss prefix for now.

PR #2859, opened 2023-01-18
fix key check in nightly yaml

botanicvelious:
Looks like the yaml got overwritten so it broke when there was no keys

### DIFF
--- a/.github/workflows/flutter-nightly.yml
+++ b/.github/workflows/flutter-nightly.yml
@@ -592,7 +592,7 @@ jobs:
             ${{steps.sign-rustdesk.outputs.signedReleaseFile}}
 
       - name: Publish unsigned apk package
-        if: env.ANDROID_SIGNING_KEY != null
+        if: env.ANDROID_SIGNING_KEY == null
         uses: softprops/action-gh-release@v1
         with:
           prerelease: true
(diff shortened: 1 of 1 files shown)


rustdesk:
I do this intentionaly, because your env.MACOS_P12_BASE64 is not defined, and its value is not 'true"also.

rustdesk:
Without MACOS_P12_BASE64 etc, does not affect generating final files.

botanicvelious:
> Without MACOS_P12_BASE64 etc, does not affect generating unsigned final files.

It actually does, for example on my github actions you can see it fails without this https://github.com/botanicvelious/rustdesk/actions/runs/3946886853 If you look at the osx build you can see it fails with "Error: At least one of p12-filepath or p12-file-base64 must be provided"

but with it there is a success https://github.com/botanicvelious/rustdesk/actions/runs/3947551708

>I do this intentionaly, because your env.MACOS_P12_BASE64 is not defined, and its value is a string rather than "true".

The check is checking to see if it exists not checking the value of it

rustdesk:
where is it defined?

botanicvelious:
Ah I see your confusion, secrets.MACOS_P12_BASE64 maps to env.MACOS_P12_BASE64 automatically, ill clear up the pr so its consistant

rustdesk:
If I set `secrets.MACOS_P12_BASE64` to `"abcd"`, `${{ env.MACOS_P12_BASE64== 'true' }}` also works?

botanicvelious:
So I found the issue and fixed it, if you can re-open this pr so i can update it 👍

botanicvelious:
Basically there are 2 issues, first the "if" only allows env. values not secrets. values so i have to set the env. value to the secret value, then the second problem was my syntax was borked, it should be if: env.MACOS_P12_BASE64 != null

botanicvelious:
Looks like it synced, everything should work properly now when merged.

PR #2862, opened 2023-01-18
Significantly improved the translation

skycommand:
I significantly improved the translation. If you have any questions, please ask in either English or Persian. I speak both languages better than their native speakers.


rustdesk:
Thanks

PR #2873, opened 2023-01-20
opt: prevent duplicate window instance on windows

Kingtous:
This PR optimizes window management on Windows, which ensures only one active main window on Windows.

This PR also fixes an issue, which will open main window unexpectly when clicking connection shortcut on Desktop.


youhers:
Thank you

Issue #2891, opened 2023-01-21
The client cannot be installed on the PC


youhers:
**Describe the bug you encountered:**

...The nightly version cannot be installed ，and must carry the command，Clicking Install has no effect，The shortcut must be used with --install to install
**What did you expect to happen instead?**

...It should be clickable to install

**How did you install `RustDesk`?**

---

**RustDesk version and environment**

1.2.0Nightly version win10

youhers:
![截图_20230121094119](https://user-images.githubusercontent.com/106003909/213834884-9d323e3d-6c76-4dcc-91f8-9ea9b4aab301.gif)

rustdesk:
@21pages

rustdesk:
@Kingtous it may be related to your last commit

Kingtous:
> @Kingtous it may be related to your last commit

Yes, Let's figure it out.

youhers:
happy new year

youhers:
@Kingtous @rustdesk happy new year

rustdesk:
Happy new year!

PR #2892, opened 2023-01-21
fix: --install cannot be invoke caused by singleton optimization

Kingtous:
related: #2891

### DIFF
--- a/flutter/windows/runner/main.cpp
+++ b/flutter/windows/runner/main.cpp
@@ -1,20 +1,21 @@
 #include <flutter/dart_project.h>
 #include <flutter/flutter_view_controller.h>
-#include <windows.h>
 #include <tchar.h>
+#include <uni_links_desktop/uni_links_desktop_plugin.h>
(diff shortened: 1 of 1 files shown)


flutter/windows/runner/main.cpp
```diff
@@ -1,20 +1,21 @@
...
 #include "utils.h"
-// #include <bitsdojo_window_windows/bitsdojo_window_plugin.h>
-
-#include <uni_links_desktop/uni_links_desktop_plugin.h>
 
 typedef char** (*FUNC_RUSTDESK_CORE_MAIN)(int*);
 typedef void (*FUNC_RUSTDESK_FREE_ARGS)( char**, int);
 const char* uniLinksPrefix = "rustdesk://";
+/// Note: `--server`, `--service` are already handled in [core_main.rs].
+const std::vector<std::string> parameters_white_list = {"--install"};
```
Kingtous: If further arguments need to bypass, please add them here.

PR #2917, opened 2023-01-23
Bump LLVM ver and enable i686

botanicvelious:
Currently LLVM 10 is being used, it has some issues with rust compile times as there was a pretty big compile regression between LLVM 9 and 10. Bumped to the latest LLVM 15 as it is supported on rust as of about October, builds seem to work fine with it after testing. Build time is up to about 17% faster with LLVM 15.

Also enabled i686-pc-windows-msvc builds, the nightly release was missing i686 for windows.

Lastly removed the double install of Rust toolchain in the same process


rustdesk:
@Kingtous

botanicvelious:
so it appears that i686 doesnt work just enabling it, there are some other settings in the build that force it to 64 bit regardless so i removed that for now

Issue #188, opened 2021-08-03
Two-way transmission of sound


july008:
PC1 controls the PC. The sound from PC2 can be transmitted to PC2. Can the sound of PC1 be transmitted to PC2?
That is, the voice intercom function is realized.

rustdesk:
You can write

rustdesk:
I do not think it is a necessary feature, there are huge number of tools for voice intercom nowadays.

rustdesk:
https://www.reddit.com/r/rustdesk/comments/vl87h7/rustdesk_audio_appears_to_be_one_way_only/

rustdesk:
Need to confirm if the other vendors provide this.

vmsman:
I will tell you that I have had to conduct a parallel session in Jitsi Meet to have two way voice.  If Rustdesk supports voice, collaboration includes not only remote control, but voice.  As an example, consider TeamViewer which supports two way voice.  Screen sharing and two way voice exists in Jitsi Meet, but without remote control.  Rustdesk would be the first open source app to do two way voice and remote control in a single app

rustdesk:
@asur4s Check out how does the other vendors work first.

rustdesk:
@Kingtous, @asur4s is busy in looking for job, please follow up this.

Kingtous:
> @Kingtous, @asur4s is busy in looking for job, please follow up this.

get it.

Anaphylaxis:
> > Need to confirm if the other vendors provide this. The audio is mainly for system sound, not for intercom.
> 
> yes. I can Confirm that TV can make an audio call

And despite using TeamViewer for 4+ years I never used that feature once.

vmsman:
@Anaphylaxis EVERYTIME that I use Rustdesk, I launch Jitsi, Skype, or Jami in the background to speak with the person that I am supporting.   The problem with support these days is that too many people that are providing support do not communicate, cannot communicate or think they automatically know what someone needs.   Two way voice communication is an absolute requirement when providing support.

Anaphylaxis:
> @Anaphylaxis EVERYTIME that I use Rustdesk, I launch Jitsi, Skype, or Jami in the background to speak with the person that I am supporting. The problem with support these days is that too many people that are providing support do not communicate, cannot communicate or think they automatically know what someone needs. Two way voice communication is an absolute requirement when providing support.

[...]

PR #2981, opened 2023-01-29
feat: dual audio transmission support

Kingtous:
related: #188

This PR is working on make dual audio transmission possible.

- [x] multi-instance audio service for dual audio transmission.
- [x] add dual transmission protocol.
- [x] fit for audio playing on both of host and server side.
- [x] tested on the cases below
    - [x] Windows ---(controls)--> macOS 
    - [-] [Working on] macOS ---(controls)--> Linux, loop back is not working yet
    - [x] Linux ---(controls)--> macOS 
- [x] add ui selection
    - [x] sciter
    - [x] flutter 

This PR adds an option to enable dual audio transmission now. After switching to `dual way` of audio transmission, the input device will automatically set to the default speaker.

image


test2a:
So excited about this.....


Could this by any chance be modified to have video as well?

We can imagine like have a floating video widget showing your own and other side cameras...

rustdesk:
> So excited about this.....
> 
> Could this by any chance be modified to have video as well?
> 
> We can imagine like have a floating video widget showing your own and other side cameras...

no for short term

rustdesk:
@test2a please test and feedback.

test2a:
@rustdesk nightly seems to be 4 hours old so should i download that or wait for new build?

rustdesk:
Thanks, it will be ready tommorrow.

rustdesk:
I can trigger it now, will be ready soon.

PR #3002, opened 2023-01-30
Remove wayland fix for good

simonspa:
This pull completely removes the possibility of overriding the GDM3 settings to disable Wayland. Reasons:

* Wayland support is currently being added (see e.g. #932 #2360 and others)
* This possibility has previously been deactivated (see ad2210106722267130e94b1be2a234b59e8ba305) but has crept back in later
* System settings should not be altered from a 3rdparty program without clearly stating *what* is being changed.
* It is questionable if this should ever have been here in the first place...


rustdesk:
Agree, Thanks

rustdesk:
To be completely removed, we need to remove doc_fix_wayland etc in `https://github.com/rustdesk/rustdesk/tree/master/src/lang`. :)

simonspa:
Right, I can also remove those - I didn't know if you wanted to keep the warnings still in. Shall I remove that all?

rustdesk:
> Shall I remove that all?

Yes, please.

simonspa:
See #3003

PR #3076, opened 2023-02-04
feat: add url scheme handler for macos

Kingtous:
e.g. `rustdesk://connection/new/201109456`


bionemesis:
Does this do anything other than just open RustDesk at this point?

rustdesk:
Yes，but PR is welcome if you want more.

rustdesk:
> e.g. `rustdesk://connection/new/201109456`

This schema is ok as the formal one, but we should also support the short url, rustdesk://<id>, because this is the most common used one, need to make it the most convinient.

Issue #2365, opened 2022-11-29
Option to change modifier key mapping (mac <-> win)


drkvogel:
It would be good to be able to configure the mapping of modifier keys between different platforms.

For example, I am using RustDesk to remote from a Mac to a Windows host. On the Mac, I am used to using `cmd+tab` to change focus between apps, e.g. once I have finished doing something on the Windows host via RustDesk, I will tend to `cmd+tab` away to another application (e.g. to an editor like VS Code, particularly if I am doing something that involves copy-pasting text between the client and host). But RustDesk maps `cmd` on a Mac client to `Win` on a Windows host, and so `Win+tab` is output to the Windows host, which is the shortcut for Task View in Windows - so I keep accidentally opening Task View in Windows, which can be difficult to escape from sometimes - it is often not very responsive (which I think is a fault of Windows). I tried disabling the Windows key completely in the host, but although it disables physical keypresses from an attached keyboard, the host can still receive `Win` keypresses from RustDesk, so I have not found any way of working around this problem.

It would be good if at least there was an option to disable the mapping of `cmd` to `Win` for Mac hosts.

rustdesk:
https://github.com/rustdesk/rustdesk/discussions/3079

rustdesk:
#1547

rustdesk:
#1547

rustdesk:
@fufesou please review

rustdesk:
added in #3121

Issue #230, opened 2021-09-13
Win10 -> MacOS black screen with mouse cursor only (awake and keep the screen unlocked for all OS)


misaka10086:
**Describe the bug you encountered:**
When Win10(should be any version of client but only tested on Win10 yet) try to connect a MacOS machine with RustDesk pre-installed(internal Remote Desktop function also enabled which allow VNC client to connect). 
If the target Mac is auto locked due to long time no operation, it will show a black screen with mouse cursor only and unable to show login page. 

One way to resolve it is open a tunnel of port 5900(VNC) and run VNC viewer to connect to target machine and it will prompt login page normally, once unlocked, close VNC viewer and back to RustDesk, the login page is now prompted. 

**What did you expect to happen instead?**
When connected, a login page(same as login locally) should be displayed which allows unlock remote machine. 
...

**How did you install `RustDesk`?**
Github release

**RustDesk version and environment**
Windows: 1.1.8 
MacOS: 1.1.6
Mac mini m1(OS version: Big Sur 11.5) with no external display
This bug can only be recreated when target machine is locked(either manually or automatic due to timeout)

misaka10086:
**Screenshot of Win10 Rustdesk Window after connected:** 
![BAMIXMJ 49_AX3P}55256$F](https://user-images.githubusercontent.com/16052890/133121978-59963cca-4170-4db5-9829-ae67b79ad70b.png)

rustdesk:
A background service with privilege is required, should be duplicate to https://github.com/rustdesk/rustdesk/issues/77

Lighfer:
Any progress?

mrcblt:
I am also curious about any news regarding this issue. I can confirm that this issue persists with the latest version 1.1.9 on both Windows 10 and macOS. I have provided all permissions required by RuskDesk on my Mac.

In addition, this behavior is also reproducible from Android 12 (version 1.1.10-1) to macOS, again I only see a black screen with a cursor.

donthomas21:
Is there any update on this issue. I still cannot connect to any Mac that has gone to sleep.

Thanks.

rustdesk:
@asur4s

tan-wei:
I just met the similar issue. But I find that I can only see the previous screen (yesterday evening), now MacOS should be on lock screen to input password.

working-name:
Still an issue, lock screen does not render via rustdesk. If I connect via another remote software the screen will show and then everything will work in rustdesk.

[...]

PR #3121, opened 2023-02-08
allow swap key

sj6219:
I implemented #2365


rustdesk:
@fufesou please review.

fufesou:
@sj6219 Hi, I've briefly read the code. The implementation seems good.

1. Would please run cargo format to the files you've modified?
2. Please check remote peer OS. If MacOS --> MacOS, do not show the swap option.

fufesou:
It's not convenient for me to test right now, I will test it tomorrow.

sj6219:
> 2\. For connections. I think it is necessary to check remote peer OS . If MacOS --> MacOS, do not swap key.

If both computers use swap settings, there is no effect because key input is swapped  once on the client,  and the server swapped it again.

fufesou:
> > 2. For connections. I think it is necessary to check remote peer OS . If MacOS --> MacOS, do not swap key.
> 
> If both computers use swap settings, there is no effect because key input is swapped once on the client, and the server swapped it again.

Yes. But it will go strange if only one side use swap setting, which is possible.
Check if both MacOS and ignore swapping is what we can control, and the result will always be expected.

fufesou:
I think it may be better to allow swap key on the local side, whether it's MacOS or not.
The remote side may not swap keys.

It will be more convenient if the control side can decide whether or not to swap keys.

fufesou:
> To do so, We need to save settings for each server the client access. It will take time to change it because I am a beginner in flutter language.

Maybe no need to save settings for each connection.

The setting can be what you've designed.
1. It should checked and applied when each connection is established.
2. It should be applied to each establised connection when the setting is changed.

BTW, "Allow swap key" option may be easier to find on the remote page(remote menu). Like the [`kb_layout_type`](https://github.com/fufesou/rustdesk/blob/85dc3b0e586fa8ca35ac729fb03fd71019f9772a/libs/hbb_common/src/config.rs#L1074) option.

sj6219:
I tested it on windows and macos.

I didn't implement Linux because I didn't have a development environment.

And, there seems to be a bug in the old non flutter version.
The check box does not appear as shown below.
sciter

fufesou:
Hi, @sj6219  the option is under the keyboard button.

src/keyboard.rs
```diff
@@ -205,11 +205,63 @@ static mut IS_0X021D_DOWN: bool = false;
 pub fn start_grab_loop() {
     #[cfg(any(target_os = "windows", target_os = "macos"))]
     std::thread::spawn(move || {
-        let try_handle_keyboard = move |event: Event, key: Key, is_press: bool| -> Option<Event> {
+            let try_handle_keyboard = move |mut event: Event, key: Key, is_press: bool| -> Option<Event> {
             // fix #2211：CAPS LOCK don't work
             if key == Key::CapsLock || key == Key::NumLock {
                 return Some(event);
             }
+            {
```
rustdesk: Can we create a function for this part of code?

rustdesk (changes requested):

src/keyboard.rs
sj6219: I made a block to distinguish it from the changed part. You can get rid of it.

rustdesk: A new function is better, no need to put everything in one function.

rustdesk: Which is bad coding practise.

rustdesk:
Thanks for your change.
@fufesou  I have another concern, this can not work in Wayland AppInage / Flatpak version which get input from flutter rather than via grab system input.

sj6219:
The current implementation swaps data as soon as client receive the keystroke. Would it be simpler to swap data just before it's sent to the server?

https://github.com/rustdesk/rustdesk/blob/master/src/ui_session_interface.rs#:~:text=msg_out.set_key_event(evt.clone())%3B

rustdesk:
AnyDesk does not have the choice to allow swap key.

https://support.anydesk.com/knowledge/anydesk-on-macos#cross-plattform-shortcuts

So i start to hesitate.

rustdesk:
> Would it be simpler to swap data just before it's sent to the server?

Yes

rustdesk:
> AnyDesk does not have the choice to allow swap key.
> 
> https://support.anydesk.com/knowledge/anydesk-on-macos#cross-plattform-shortcuts
> 
> So i start to hesitate.

Appologize, I find where to modify swap on AnyDesk.

rustdesk:
image
It is a global setting on controlling side, effective for outgoing connections.

sj6219:
I'm fixing bug. Please wait.

sj6219:
Bug fixed

achimmihca:
Thanks for implementing this. I think this is what I am looking for.
I use a Windows host to connect to macOS and want to use for example `ctrl+c` to copy text.
On macOS I installed [Karabiner](https://karabiner-elements.pqrs.org/) for this, which works nicely but not when connected via RustDesk.

So again, thanks for implementing this option.

Cargo.toml
```diff
@@ -167,3 +167,8 @@ panic = 'abort'
 strip = true
 #opt-level = 'z' # only have smaller size after strip
 rpath = true
+
+[patch."https://github.com/fufesou/rdev"]
+#rdev = { path = "../rdev" }
+rdev = { git = "https://github.com/sj6219/rdev", branch = "sigma" }
+
```
rustdesk: Please submit a PR to fufesou/rdev.

flutter/lib/desktop/widgets/remote_menubar.dart
```diff
@@ -1580,6 +1580,9 @@ class _RemoteMenubarState extends State<RemoteMenubar> {
         ),
       );
     }
+    keyboardMenu.add(_createSwitchMenuEntry(
+        'Swap Control-Command  Key', 'allow_swap_key', EdgeInsets.zero, true));
+
     return keyboardMenu;
```
rustdesk: Please do this only if the peer os is mac.

src/ui/header.tis
```diff
@@ -156,6 +156,7 @@ class Header: Reactor.Component {
             <menu.context #keyboard-options>
                 <li #legacy><span>{svg_checkmark}</span>{translate('Legacy mode')}</li> 
                 <li #map><span>{svg_checkmark}</span>{translate('Map mode')}</li> 
+                <li #allow_swap_key><span>{svg_checkmark}</span>{translate('Swap Control-Command Key')}</li>
```
rustdesk: Only if peer os is mac.

src/ui_session_interface.rs
```diff
@@ -843,6 +920,22 @@ impl<T: InvokeUiSession> Interface for Session<T> {
             handle_test_delay(t, peer).await;
         }
     }
+    fn swap_modifier_mouse(&self, msg : &mut hbb_common::protos::message::MouseEvent) {
```
rustdesk, line 1015: miss new line above.

rustdesk (changes requested):

flutter/lib/desktop/widgets/remote_menubar.dart
rustdesk: and local side is not mac

sj6219:
I was hospitalized because I was sick. 
So I can't do coding work for 1-2 weeks.
 I want someone else to work or wait for me.

rustdesk:
Take care, this is not urgent.

IRZTAR (commented):
si.

Cargo.toml
rustdesk: Could you please solve this?

sj6219: I submitted a PR, but my PR has not yet been accepted.

flutter/lib/desktop/widgets/remote_menubar.dart
rustdesk: This seems not solved yet.

rustdesk (changes requested):

flutter/lib/desktop/widgets/remote_menubar.dart
rustdesk: Ignore this, solved.

Cargo.toml
fufesou: Sorry for late, the PR is accepted now.

rustdesk: Thanks, let me notify him.

rustdesk:
Do we need to update Cargo.lock for rdev?

rustdesk:
> Do we need to update Cargo.lock for rdev?

CI failed because of this i guess.

rustdesk:
Let me do this.

rustdesk:
@sj6219 do you have interest to work on this https://github.com/rustdesk/rustdesk/issues/230#issuecomment-1422615080?

ylluminate:
Any news here? Very problematic when trying to interact with Control Remote Desktop and cmd-a + cmd-z / cmd-shift-z don't work:
image

discoveryX64:
My issue is completely different from this, swap key is off, and even with it on beforehand it worked perfectly fine

linnv:
how about swap  Caps  Lock <--> Control?

linnv:
> > To do so, We need to save settings for each server the client access. It will take time to change it because I am a beginner in flutter language.
> 
> Maybe no need to save settings for each connection.
> 
> The setting can be what you've designed. It is an option that should be set once and applied to each connection.
> 
> 1. It should checked and applied when each connection is established.
> 2. It should be applied to each establised connection when the setting is changed.
> 
> BTW, "Allow swap key" option may be easier to find on the remote page(remote menu). Like the [`kb_layout_type`](https://github.com/fufesou/rustdesk/blob/85dc3b0e586fa8ca35ac729fb03fd71019f9772a/libs/hbb_common/src/config.rs#L1074) option.

Swap control--command key works on win->mac(windows control mac), nice feature!
how about swap Caps Lock <--> Control?

PR #3127, opened 2023-02-08
devcontainer for vscode

akhildevelops:
It will help to create a separate development environment in docker. It doesn't pollute the local system with tools and dependencies required for building rustdesktop. Even rustlang need not be installed.  Vscode can be connected to the dev environment container through dev containers extension.

The dockerfile for the dev environment is taken from [https://github.com/rustdesk/rustdesk/blob/master/Dockerfile](https://github.com/rustdesk/rustdesk/blob/master/Dockerfile) of the repo.

Pre-requisites:
- docker
- dev containers vscode extension

Open command pallete by pressing `ctrl+shift+p`and then goto `Dev Containers: Reopen in Container`. A new workspace will be opened that creates, installs required dependencies and runs dev environment. The vscode terminal points to container. It will also download and enable extensions such as lldb, rust-analyzer.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,19 @@
+FROM debian
+
+WORKDIR /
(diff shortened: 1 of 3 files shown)


rustdesk:
if the dicker file has no diff, please change to a symbol link.

rustdesk:
Let me do this

rustdesk:
Ok, i see the diff.

PR #3154, opened 2023-02-10
Addressbook login. Button instead of text

grummbeer:
Tab addressbook: Login button instead of plain text.

|before | after |
|--  | --  |
| rustdesk-adressbook-login-button-before | rustdesk-adressbook-login-button-after |


rustdesk:
Agree

PR #3169, opened 2023-02-11
Dialog. Adjust padding of the title.

grummbeer:
Using the defined padding also for the title for an even alignment. If no title is used, [its top padding](https://api.flutter.dev/flutter/material/SimpleDialog/titlePadding.html) will not apply

before > after
rustdesk-dialog-title-padding-before-after

Note:

The value for the contentPadding can be set by a constructor argument (there is [one call that makes use of it](https://github.com/rustdesk/rustdesk/blob/master/flutter/lib/mobile/widgets/dialog.dart#L156)). This value is apply to the content but not to the title nor action. This can lead to an not nice looking alignment, especially if a title is used and the value is similar or smaller than the default value.

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -653,6 +653,7 @@ class CustomAlertDialog extends StatelessWidget {
       child: AlertDialog(
         scrollable: true,
         title: title,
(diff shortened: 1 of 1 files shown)


rustdesk:
This works very well.

Issue #3157, opened 2023-02-10
[Bug] Unable to confirm password


giuliovale:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug Description

Unable to confirm password

### How to Reproduce

Try changing your password. Each time you type a character in the confirmation box, the cursor moves to password

### Expected Behavior

Confirm the password

### Operating system(s) on local side and remote side

Windows 11

### RustDesk Version(s) on local side and remote side

Nightly build (2023-02-09)

### Screenshots

![image](https://user-images.githubusercontent.com/6909743/218104525-afb1a72e-5564-4071-b0e3-a1e6ac36e74c.png)

rustdesk:
This is intended.

rustdesk:
I know your concern and what you wanna to explain. But we insist current design.

rustdesk:
I misunderstood.

rustdesk:
I do not see your problem. Even it is a problem on your side, i still think so https://github.com/rustdesk/rustdesk/issues/3147#issuecomment-1425570266.

rustdesk:
Next time, please recording a video, reading text is really hard to understand.

rustdesk:
Can you help fix?

youhers:
![截图_20230210220230](https://user-images.githubusercontent.com/106003909/218110914-e4869953-d41b-4f3f-86b6-f89ca2c9c7df.gif)
Let me provide gif

RudiOnTheAir:
Can confirm this.

rustdesk:
> let me provide gif

Wierd, I do not have this problem on my M1,  too bad I do not bring my windows today.

RudiOnTheAir:
I'm using not Windows. On my site it is debian!

ajstadlin:
I have this same issue in Ubuntu 22.0.4 LTS, Rustdesk nightly build 1.2.0, 2023-02-10 11:49

My work around is to Copy the password as text (e.g. from a Text Editor) then Paste the password text into the Password and Confirm fields.

To Paste into the Confirm textbox, I need to hold the mouse button down and Ctrl+V.  If the mouse button is not held down, the focus shifts back to the upper password textbox

//ajs

grummbeer:
https://github.com/rustdesk/rustdesk/blob/master/flutter/lib/desktop/pages/desktop_home_page.dart#L637

the focus is always forced to itself.

Zack-Heisnberg:
[...]

PR #3145, opened 2023-02-10
fix autofocus

grummbeer:
Fix autofocus for hostname in Settings | Network | proxy. 

Set cursor to "username" or "password", the focus goes back to "hostname"

PR #3171, opened 2023-02-11
fix autofocus

grummbeer:
Same behavior as in https://github.com/rustdesk/rustdesk/pull/3145. Using `autofocus: true` according to https://api.flutter.dev/flutter/material/TextField/autofocus.html and https://docs.flutter.dev/cookbook/forms/focus#focus-a-text-field-as-soon-as-its-visible

- dialog permanent password. see #3157,
- dialog RDP settings

rustdesk-rdp-settings

There are some one-field situations where this effect is not pop up.

- https://github.com/rustdesk/rustdesk/blob/master/flutter/lib/common/widgets/address_book.dart#L389
- https://github.com/rustdesk/rustdesk/blob/master/flutter/lib/common/widgets/dialog.dart#L57
- https://github.com/rustdesk/rustdesk/blob/master/flutter/lib/common/widgets/dialog.dart#L102
- https://github.com/rustdesk/rustdesk/blob/master/flutter/lib/common/widgets/dialog.dart#L189
- https://github.com/rustdesk/rustdesk/blob/master/flutter/lib/common/widgets/peer_card.dart#L644
- https://github.com/rustdesk/rustdesk/blob/master/flutter/lib/desktop/pages/file_manager_page.dart#L801

 The focus is also set right with "autofocus", should these fields also be changed (and included in this PR)?

### DIFF
--- a/flutter/lib/common/widgets/peer_card.dart
+++ b/flutter/lib/common/widgets/peer_card.dart
@@ -1013,7 +1013,7 @@ void _rdpDialog(String id) async {
                     decoration: const InputDecoration(
                         border: OutlineInputBorder(), hintText: '3389'),
                     controller: portController,
-                    focusNode: FocusNode()..requestFocus(),
+                    autofocus: true,
                   ),
                 ),
               ],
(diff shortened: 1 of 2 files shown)


rustdesk:
I changed all in https://github.com/rustdesk/rustdesk/commit/3d40569dee56c903b481f3ab27108f524bb74e6c

PR #3180, opened 2023-02-12
Update and rename template.rs to nl.rs

FastAct:
add Dutch translation


rustdesk:
please do not remove template，and you need to modify lang.rs also to make it effective. Thanks.

PR #3184, opened 2023-02-13
Change resolution

21pages:
- [X] win/x11/mac resolution api, win/x11 tested, mac only test native code.
- [X] allow only if client has keyboard permission.
- [X] cascading menu
- [X] auto adjust window


youhers:
If only you could automatically adjust the border like a sunflower without a dot

youhers:
If only you could adjust the border automatically like a sunflower

youhers:
Great Thank you

PR #3205, opened 2023-02-14
would like to add the dutch translation

FastAct:


rustdesk:
Many unnecessary code change in lang.rs

PR #3207, opened 2023-02-14
Modernized menu bar

NicKoehler:
after


rustdesk:
Good looking, but it is not eay to change hover / press any more in the code.

NicKoehler:
You are right, I'll try to fix that

rustdesk:
@NicKoehler could you also modernize our file transfer window? It looks super old school and ugly, including the task list on the right side.

Peek 2023-02-15 15-15

rustdesk:
and port forwad window.

https://github.com/rustdesk/rustdesk/blob/master/flutter/lib/desktop/pages/port_forward_page.dart

image

NicKoehler:
I can try after the menu bar, but now I have an issue that I'm not sure how to fix.. for the last 4 days when I lunch the app with `flutter run ` I receive this error at the top:

```
Launching lib/main.dart on Linux in debug mode...
main.dart:1
[2023-02-15T07:39:33Z INFO  librustdesk::server] DISPLAY=Ok(":1")
[2023-02-15T07:39:33Z INFO  librustdesk::server] XAUTHORITY=Ok("/run/user/1000/gdm/Xauthority")
Connecting to VM Service at ws://127.0.0.1:42043/QIugX25crgw=/ws
flutter: launch args: []
flutter: initializing FFI main
flutter: initialize failed: Invalid argument(s): Failed to lookup symbol 'session_get_rgba': /home/nic/Git-Packages/rustdesk/flutter/build/linux/x64/debug/bundle/lib/librustdesk.so: undefined symbol: session_get_rgba
```

if I launch with `flutter run --release` the error is gone.
It's so slow to build every time with the release flag, I need to fix this..

rustdesk:
Have you run `cargo build --features flutter`?

NicKoehler:
I remembered using it, but evidently not, now it works thank you

rustdesk:
BTW do not forget this button, it will be added when starting audio call.

https://github.com/rustdesk/rustdesk/pull/2981

image

NicKoehler:
Ok, if you could take a look and let me know if there's anything that you think needs to be changed or improved, it would be great

rustdesk:
The mergin of border seems a little small after this new look.

image

NicKoehler:
The margin now is 6 both vertically and horizontally between widgets.

rustdesk:
@NicKoehler the color change below from white to black in black theme seems a little wierd. Do you have any idea?

image

NicKoehler:
Do you mean that half pixel border?

rustdesk:
The border between titlebar and menubar seems invisible.

NicKoehler:
It looks like an optical illusion caused by the fact that the edge is only one pixel wide

image

rustdesk:
true

rustdesk:
Do you think below looks better? or add some decoration on top side of the menu bar.
image

NicKoehler:
Maybe with rounded corners on the top too

rustdesk:
Will try

Alintya:
What about showing the monitors immediately on hover?
That might als satisfy #3333

bilago:
> I did have this idea.

Immediate expanding on hover is a step in the wrong direction imo. The auto expand behavior happens currently in the nightly as I explained in a bug, and causes misclicks on other options.

A suggestion I would make would be:

Show all monitors as individual buttons. Require explicit clicks on menu options.

Add an additional button that you can click on that expands a menu showing the physical location of the connected monitors. This would satisfy the ask of this bug without compromising the new menu bar theme, and retain the feature found in 1.9.x

Pardon my crude mspaint skills:

image

rustdesk:
Auto expand behavior has been removed today.

Alintya:
> > I did have this idea.
> 
> Immediate expanding on hover is a step in the wrong direction imo. The auto expand behavior happens currently in the nightly as I explained in a bug, and causes misclicks on other options.
> 
> A suggestion I would make would be:
> 
> Show all monitors as individual buttons. Require explicit clicks on menu options.
> 
> Add an additional button that you can click on that expands a menu showing the physical location of the connected monitors. This would satisfy the ask of this bug without compromising the new menu bar theme, and retain the feature found in 1.9.x
> 
> Pardon my crude mspaint skills:
> 
> image

i still have the older build and fail to see how that was an actual problem to be honest.
That is like saying we need dedicated buttons to move the remote cursor in either direction because you could mess that up 🙈

PR #3241, opened 2023-02-16
Update logo in README

rakleed:
To make it look good in a dark theme, I removed the light bright background.


rustdesk:
But it looks ugly for me without light background in dark theme. Sorry, i can not accept your proposal.

rakleed:
@rustdesk what exactly is the problem? Looks good to me.

Screenshot 2023-02-16 at 18 14 26

rustdesk:
I still think old is better, with better contact.

rustdesk:
the blue gradient color looks strange on black background.

rustdesk:
it is my opinion. we can make a poll on the girhub discussions. let users decide.

rakleed:
The problem is that for a dark theme, a light background is a bad solution, as it is not pleasant for the eyes if you are in a dark room.

You can also just reduce its background to about the size of a square, but it seems that it will look strange.

rakleed:
> it is my opinion. we can make a poll on the github discussions. let users decide.

Sounds good.

rustdesk:
could you write a poll post?

rakleed:
Yes, sure. But I'll do it a little later.

rakleed:
@rustdesk and what do you think about such option: https://github.com/rakleed/rustdesk/tree/patch-1#readme?

rustdesk:
> @rustdesk and what do you think about such option: https://github.com/rakleed/rustdesk/tree/patch-1#readme?

Looks wierd.

rakleed:
@rustdesk if the last option is suitable, then you can reopen the PR and accept it.

rustdesk:
I still prefer our original one than your last option.

rustdesk:
@grummbeer @ NicKoehler how do you think about his proposals?

NicKoehler:
As a dark theme user I still like the old one more, the only thing that bother me is the font and colors..
I like something like this:

immagine

grummbeer:
haven't a bulletproof opinion for now.

i am also a dark mode user and yes, scrolling over the readme makes a bit a flash light

> the blue gradient color looks strange on black background.

can also see this, and the "Your remote desktop" loses contrast.

How about just reducing the brightness of the surrounding area? Maybe some transparency and the dark background will tone down the brightness.

check the alt tag: "RustDesk - Dit fjernskrivebord"?

rustdesk:
i am dark theme user on mac, and white theme user on win.

rustdesk:
Close since no consensus.

PR #3280, opened 2023-02-19
improve input of change ID

grummbeer:
- no ambiguous communication: reset error message on input
- unify error messages: add "Prompt" prefix
- unify text input: add label 

unifiy components behavior "Text input with constraints"

- Add visual feedback for the contraints
- validate local contraints first, before making an api call.   

|before|after|
|-- |-- |
|rustdesk-change-id-message-before|rustdesk-change-id-message-after|

- [x] doublecheck code
- [x] check wording

- "starts with a-z"  can be ambiguous. Is this meaning **only** lowercase? > "starts with a letter"
- "characters" > "allowed characters"?

longer names are more clear, but are in contrast to the very short messages in password dialog ("digit", "uppercase", "lowercase")

What do you prefer?


rustdesk:
Good job, go ahead.

grummbeer:
using the more descriptive version for wording. "starts with a letter" and "allowed characters"

PR #3282, opened 2023-02-19
Unify left labeled text input

grummbeer:
Unify components of same purpose. "Left labeled text input"

- unify alignment in horizontal and vertical direction
- unify space between input field and its label

@21pages  any suggestion on this? "Server area"? Second commit is for server. Just drop it, if you do not want it.

### before 

ui-unify-left-labeled-text-input-before

### after

ui-unify-left-labeled-text-input-after-horizontal 

ui-unify-left-labeled-text-input-after-vertical


21pages:
Great! It looks pleasing to my eyes!

Issue #3277, opened 2023-02-19
[Bug] Cannot automatically exit from full-screen connection


wakewon:
### Bug Description

In full-screen mode, the rustdesk window does not close automatically after disconnection, but leaves a black screen with a virtual desktop

### How to Reproduce

1. Connect to a remote computer.
2. Use the button on the menu at the top of the rustdesk to turn on full screen display.
3. Use the button on the menu at the top of the rustdesk to disconnect.
<img width="428" alt="image" src="https://user-images.githubusercontent.com/40191996/219936549-4d7317e1-4a6d-40c8-9250-b99a5ddd8f96.png">

4. The connection is interrupted, the window is black, but it will not close automatically.

5. If I use the cancel full screen button in the upper left corner, the next connection will be displayed in a strange little floating window.
<img width="420" alt="image" src="https://user-images.githubusercontent.com/40191996/219936628-cdc5daed-3cbf-4a5b-860a-324216c1fc01.png">
<img width="228" alt="image" src="https://user-images.githubusercontent.com/40191996/219936955-28d764f1-4a63-4360-ae46-458521dd41d7.png">

6. Whether it's a strange little window or a black screen, the only way to fix it is to exit rustdesk, otherwise the next connection is bound to have problems.
<img width="360" alt="image" src="https://user-images.githubusercontent.com/40191996/219936788-16bb6dbb-614f-4c5b-88b1-eff35591ff83.png">
<img width="373" alt="image" src="https://user-images.githubusercontent.com/40191996/219936816-e5ed8387-6b0e-4df3-abc9-eaa80abef956.png">

### Expected Behavior

Exit full screen and back to desktop.

### Operating system(s) on local side and remote side

macOS 13.2.1 -> Windows 10

### RustDesk Version(s) on local side and remote side

1.2.0 Build Date: 2023-02-19 00:17 -> 1.2.0 Build Date: 2023-02-18 00:32

### Screenshots

<img width="408" alt="image" src="https://user-images.githubusercontent.com/40191996/219936246-f9e597f8-e051-4dc9-910d-3b3f8d6c7472.png">

rustdesk:
@Kingtous

Kingtous:
Let's figure the problem out. Thanks for the feedback.

PR #3320, opened 2023-02-22
fix: orderout not working when fullscreen on macos

Kingtous:
temporary workaround for #3277

### DIFF
--- a/flutter/pubspec.yaml
+++ b/flutter/pubspec.yaml
@@ -59,7 +59,7 @@ dependencies:
     desktop_multi_window:
         git:
             url: https://github.com/Kingtous/rustdesk_desktop_multi_window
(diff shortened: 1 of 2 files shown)


flutter/pubspec.yaml
```diff
@@ -92,6 +92,7 @@ dependencies:
     password_strength: ^0.2.0
     flutter_launcher_icons: ^0.11.0
     flutter_keyboard_visibility: ^5.4.0
+    texture_rgba_renderer: ^0.0.8
```
Kingtous, line 95: The reason why the plugin is needed, is that the `rustdesk_desktop_multi_window` has injected this plugin for desktop platforms, we need to pass the compilation.

PR #3334, opened 2023-02-22
File manager redesign

NicKoehler:
This is the first implementation of the redesign, and while it's not identical to my previous mockup, it's still a big step forward. What do you think? Please let me know if you want me to make any changes.


rustdesk:
Wonderful, Let me take a look.

rustdesk:
@Heap-Hop please test and feedback.

rustdesk:
@NicKoehler It looks very good, but I have four concerns.

- [ ] the overflow you can see in the pic, need scrollable horizontally
- [ ] new line on datetime make the UI ugly, osx finder does not show new line.
- [ ] The empty looks wierd because there is no border. Should we bring all borders back?
- [ ] column is not adjustable, @Kingtous will follow this up.

image

Heap-Hop:
Thanks for your nice design!

And I also found something that need to improve with your help.

- In the light theme, the background color`s change make the theme color look grayer than before, and it make the inactive button unclear. (In dark theme it looks great)
- the back button has wrong direction.
shot1

grummbeer:
> The empty looks wierd because there is no border. Should we bring all borders back?

please not. no borders, but agree that it looks a bit weird. Take a look to heap-hops screenshot, similar situation on empty tab.

I don't know what to do, just a few thoughts.

- The simple `empty` is refer to itself, a "wide empty area". Maybe a nicer message in style of a transfer card, could help making the space not that empty: "[icon info] [\n] No file transfer in progress" or similar, my english is not that good.
- A heading like in the other two columns could help to make the third column more recognizable as the third column. "[icon transfer] File transfers" or similar.
- instead a border, the whole area could filled with background like an "empty card" (same radius, color, etc.). When the first transfer starts, the "empty card" "shrinks down" and become the first transfer card.

grummbeer:
i placed it first in https://github.com/rustdesk/rustdesk/discussions/3288, but here is the better place.

- Equalizing a bit the space?
    file-transfer-spacing


- Is there a flicker when hovering? Does it react twice?
  file-transfert-remove-flickering

NicKoehler:
A lot of things can be improved, with your feedback. This month is particularly busy for me, so I might not be able to work on it right away, but I'll do my best to find time and make the necessary improvements as soon as possible

rustdesk:
@NicKoehler take your time, :)

NicKoehler:
Using borders doesn't feel modern, so I think we need to create something like this (with padding adjustments, of course):

image

rustdesk:
Agree

NicKoehler:
Maybe the best approach is to show that space only when transferring and maintain the card as it is right now, what do you think?

rustdesk:
Like this?
image

NicKoehler:
Yes, and when you start transferring, the interface shrinks to make space for the transfer card, something like that. Alternatively, we can implement a button like browsers do that shows or hides transfers

rustdesk:
> Yes, and when you start transferring, the interface shrinks to make space for the transfer card, something like that.

I like this.

grummbeer:
Please consider the layout shift. 

The eye is focused on one specific item in the list and suddenly there is a move to the left. Focus lost unexpected. This is irritating. Users are "surprised" when "suddenly" something happens that "they didn't expect".

How long will this two column view be visible before the first file transfer is triggered? Afterwards the third column view is there anyhow.

Adding a toggle is a great idea, so the user has control over whether the transfers stack is visible or not. The layout shift is then also expected.

My suggestion is set the toggle default to "true" and replace this "empty" with a pleasing message.

In a further cycle the toggle could be turned into some sort of "notification" area, with loading indicator, short stats, expands (as overflow) and reveal the transfers stack, with option "make it sticky".

NicKoehler:
I think @grummbeer is right about the shrinking, this is what it looks like:

https://user-images.githubusercontent.com/53040044/221196042-cc59f011-43ea-477a-9063-c91bf57f55e3.mp4

rustdesk:
> replaceing this "empty" with a pleasing message.

Untitled

NicKoehler:
The only problem with this is that the color of the background in the job list should be the color of the card itself

grummbeer:
> The only problem with this is that the color of the background in the job list should be the color of the card itself

Is it? The inverted style doesn't look that bad in my opinion.

But if this a problem, How about seeing this background as a "emtpy-card". A nice message that makes it predictable for the user what this area is for.

empty-card-1-1

The first transfer replaces the empty card

empty-card-1-2

NicKoehler:
I'll try something like this

NicKoehler:
What do you guys think?

image

grummbeer:
- good: No layout shift
- good: the "wide empty area" is not that empty anymore
- good: descriptive message 
- so semi good: icon refers to "close" 

Do we have a "File Transfer" icon? It could be used also here:

file-transfer-control-remote-computer

Now it's just up to personal preference. mine are matched.

PS: Only the switch for show/hide job list is missing :-)

NicKoehler:
changed transfer icon:

image

grummbeer:
If you are satisfied with it, i am too. You have done a great job!

All concerns i have spotted out are about readability, commnunication, accessibility, consistency, usability, etc. These have been resolved from my perspective.

Having a two-column view by default is an absolutely valid wish, i can see it on the horizon, but it's the horizon and it will take a little more time to implement this in a sophisticated way.

You have already expressed that your time is limited. The fact that you still haven't stopped continuing here shows a lot about you. You call it a "big step", i would call it a giant leap and a great finish for this current cycle of designing the file transfer view.


Thats what i think.

rustdesk:
i am satisfied with all of these modified versions, all better than my expectation, great jobs.

rustdesk:
> Do we have a "File Transfer" icon?

👍

Issue #3349, opened 2023-02-24
Add dedicated request elevation button in rustdesk


radiandedy:
### Description

Currently if we need to request elevation to other user we need open application that have administrator and then select elevation.
 
in this case the application uac will appear and if we click no rustdesk uac will appear, this will make users have to click no to application administrator uac and then click yes on rustdesk elevation uac.

using dedicated elevation button like anydesk make user only click ok on rustdesk uac and dont need to open other applications

![image](https://user-images.githubusercontent.com/40633520/221086142-4ea41132-6d54-43bf-9b3c-3aded79b11ba.png)

### Impact

Using dedicated elevation button will make elevation remote user fast and other user that we remote dont need to open application in administrator.

rustdesk:
@21pages add this

rustdesk:
test tommorow please

PR #3353, opened 2023-02-24
peercard and request elevation menu

21pages:
1. Check existence before add to other tabs
2. Implement discovery delete, now just remove from config file.
3. Request elevation menu


rustdesk:
#3349

Issue #1947, opened 2022-11-03
excessive cpu usage in linux nightly


test2a:
**Describe the bug you encountered:**
cpu is up to 20.5% which is while rustdesk isn't even connected to a remote machine..

![Screenshot_20221103_222256](https://user-images.githubusercontent.com/39449563/199785182-afd0f5b7-fe38-42c9-9eb5-194e6e036b79.png)

...

**What did you expect to happen instead?**

...

**How did you install `RustDesk`?**
github nightly

---

**RustDesk version and environment**

rustdesk:
@Kingtous

Kingtous:
did u just start rustdesk and do nothing？

test2a:
@Kingtous 
![Screenshot_20221104_104005](https://user-images.githubusercontent.com/39449563/199893923-ce33826e-0a00-4880-84cf-964a620cdb6c.png)
![Screenshot_20221104_103913](https://user-images.githubusercontent.com/39449563/199893957-4c078d62-4877-47d8-a344-48a210ac85a3.png)


here is system monitor right now being connected to a windows machine. 58.8% even, i could not take a photo of that.
i am trying to recreate yesterday's after disconnection high CPU usage but this too looks very high. ok.  i am up to 61% now

test2a:
@Kingtous ok. so i in the above screenshot i was connected to a machine so i disconnected and let it sit while i was writing this and suddenly the CPU is up to now 25%.... from rustdesk.

can i send you some logs or something?

rustdesk:
@test2a Thanks for your report, we get some clue and will fix it soon.

rustdesk:
@test2a could you try new nightly again?

test2a:
@rustdesk ok. i do apt remove and install the nightly version and the cpu usage is now in cycles. 

up and down, 

i have left the remote connection "connected" and using other local windows and the graph shows just plain 35% all the time.
![Screenshot_20221105_105132](https://user-images.githubusercontent.com/39449563/200102593-27cf66cf-dd99-4bff-8532-733f4e32957d.png)


here is after i started directly using remote window for 1-2 minutes.
![Screenshot_20221105_105307](https://user-images.githubusercontent.com/39449563/200102872-ebf4c09d-59c2-4274-8434-13f5fa8a2c92.png)

so from this i can gather, keeping the rustdesk as top active window keeps the cpu usage to minimum but if it goes behind other active windows, the usage just keeps on growing

[...]

PR #3356, opened 2023-02-24
update options after login

fufesou:
Enable options in login message after login is succeeded.
Fix options not working, like "Lock after session end".


rustdesk:
https://github.com/rustdesk/rustdesk/issues/1947#issuecomment-1443176195

PR #3362, opened 2023-02-24
Added additional intent-filter

Orholt:
Hi, I saw that TeamViewer Host app uses this action in their intent filter in BootReceiver, so I think we should include it too.

About this intent:
This event is triggered when the device is booted up from a quick boot mode, which is a feature introduced in Android 5.0 Lollipop that allows devices to boot up faster by skipping certain system initialization steps.


rustdesk:
@Heap-Hop

PR #3374, opened 2023-02-25
fix/macos change resolution

fufesou:


grummbeer:
```dart
try {
      if (dw > size.width) {
        dxOffset = (x - dw * (x / size.width) - _x).toInt();
      }
      if (dh > size.height) {
        dyOffset = (y - dh * (y / size.height) - _y).toInt();
      }
    } catch (e) {
      // Unhandled Exception: Unsupported operation: Infinity or NaN toInt
      return;
    }
```

Thank you very much, very nice! You have overwritten my local change. I saw the black screen often in debug mode and made the same change, but wasn't sure if this is just a "my machine in debug mode"-thing, or if this is also possible in the release version. Next time i won't be so shy and report earlier … especially since I knew about the "black screen" issues :-(

rustdesk:
👍

Issue #3373, opened 2023-02-25
keyboard is wrong on latest nightly


test2a:
### Bug Description

for the last 3 days' nightlies i am facing wrong keyboard on remote side. I have attached screen recording from rustdesk itself which shows me attempting to type qwerty but it comes out wrong.

again, this was working properly 3-4 days ago but last 2 days it was intermittent but today i updated and it is wrong keyboard 100% of the time

### How to Reproduce

i don't know, just type

### Expected Behavior

correct keys should be transmitted

### Operating system(s) on local side and remote side

linux>windows

### RustDesk Version(s) on local side and remote side

1.20>1.1.8

### Screenshots

[c421231539_20230225114034.webm](https://user-images.githubusercontent.com/39449563/221341795-5fa3dd56-120e-4367-94ec-61de05e932dd.webm)

rustdesk:
@fufesou

fufesou:
@test2a Hi, thanks for your feedback.
I've tested 1.2.0(Linux) -> 1.1.8(Win), but failed to reproduce this issue.

Which input method do you use on local Linux PC ?

BTW, RustDesk before the ngihtly build support only the "Legacy Mode", which has some wrong characters in non-US input method. The ngihtly build has introduced "Map Mode", a much better way for keyboard input.

![1677324304187](https://user-images.githubusercontent.com/13586388/221354329-d30de3ef-8942-4325-8988-5678b0bb9483.png)

test2a:
![Screenshot_20230225_171942](https://user-images.githubusercontent.com/39449563/221355278-585ab7fd-8593-4650-99a7-1fd9dee2b8bc.png)


i use
````
 sudo apt-get remove rustdesk

````
 and then 
````
 sudo dpkg -i rustdesk-1.2.0-x86_64-unknown-linux-gnu-ubuntu-18.04.deb 

````

every morning on my local side. remote side is unchanged.

i don't have the keyboard icon, it did till some weeks ago but was gone recently.

the keyboard is "right now" working properly but sometimes it goes beserk. 

keyboard layout on remote side is "eng-us"

![Screenshot_20230225_172250](https://user-images.githubusercontent.com/39449563/221355406-53121eee-4081-45f1-8434-ebdcbe90c3b0.png)

fufesou:
It seems you're using wayland.

I modified the input logic a few days ago.
If local side is wayland or macOS (does not support input monitor), the input mode is set to "Map Mode".
[...]

PR #3375, opened 2023-02-25
Fix/flutter keyboard compatibility

fufesou:
1. Flutter keyboard grab, set legacy mode, if only legacy mode is supported.
2. Do not check keyboard permission on menu build. Or else the connection has to be closed and connect again, to make the keyboard menu show.


rustdesk:
#3373

PR #3378, opened 2023-02-25
opt: add android so lib in action artifacts

Kingtous:


rustdesk:
liblibrustdesk.so only?

rustdesk:
these files can also extracted from apk file.

rustdesk:
unzip the apk file, all under lib/arm64-v8a

Issue #1597, opened 2022-09-21
Connection error : Connection reset by peer OS error 104


clues4me:
Hi, just got twice issue get connected to client. I dont know what its, but i think related to security because its somehow blocked by router/firewall.
The day before its okay, nothing got wrong, but accidently its cant connect to a client, others clients is ok.

Im using debian on both side and installed correctly on x11 desktop.
Rustdesk version 1.1.9

![Cuplikan layar dari 2022-09-22 05-42-51](https://user-images.githubusercontent.com/28584711/191623917-4ee4a936-eeb3-48c0-b8f6-db0ffeae00c3.png)

21pages:
try this master [build](https://github.com/21pages/tmp)

21pages:
> Am also facing the exact problem.
> 
> > try this master [build](https://github.com/21pages/tmp)
> 
> Would you mind doing some explanation before I try that? BTW, I'm on Linux...

This problem has been fixed on master branch. I'll put  a linux build on that link later.

escape0707:
> > Am also facing the exact problem.
> > 
> > > try this master [build](https://github.com/21pages/tmp)
> > 
> > Would you mind doing some explanation before I try that? BTW, I'm on Linux...
> 
> This problem has been fixed on master branch. I'll put  a linux build on that link later.

Okay, thanks for the fix.

I'm using my Linux machine to setup a server and letting my windows enjoyer friend using mine, for a better direct connection. Does he need to update his client software or do I need to update my Linux server for the fix to work?

21pages:
Just update client

clues4me:
> try this master [build](https://github.com/21pages/tmp)

thanks, i'll try

rustdesk:
Can you try out nightly build? There is a fix, though not good enough, not sure if it works in your scenario.

weirdal3333:
The nightly build works better, but clicking or using the keyboard causes the os error 104

rustdesk:
> The nightly build works better, but clicking or using the keyboard causes the os error 104

What env on controlled side?

eMPee584:
Same thing here, with latest nightly .deb and APK.. this comes up in the system log:

```
Feb 12 14:24:49 base sudo[2193853]:     root : PWD=/ ; USER=empee584 ; ENV=XDG_RUNTIME_DIR=/run/user/1000 ; COMMAND=/usr/lib/rustdesk/rustdesk --cm 
[...]

PR #3381, opened 2023-02-26
mobile id suffix "\r" or "/r" for relay

21pages:
#1597


rustdesk:
should be added to client.rs

rustdesk:
at least in rs shared by all ui flutter/sciter

Issue #3332, opened 2023-02-22
Can't change path for recordings on macOS


rakleed:
### Bug Description

It is impossible to change the path for recordings in the macOS application.

### How to Reproduce

1. Open RustDesk settings
2. Go to `General`
3. On the `Recording` section, click `Change`

### Expected Behavior

A Finder window will appear with the option to select a new path.

### Operating system(s) on local side and remote side

macOS 13 → Windows 11

### RustDesk Version(s) on local side and remote side

1.2.0 nightly

### Screenshots

![image](https://user-images.githubusercontent.com/19418601/220759142-f2474a54-3239-4b55-a07b-5c6f9fcc0656.png)

### Additional Context

On Windows it works fine.

rustdesk:
close as fix merged

PR #3396, opened 2023-02-27
mac admin auth

21pages:
1. Add mac admin auth
2. Fix mac get resolutions, ensure that the set resolution is the same as the current bpp/refresh rate


rustdesk:
https://github.com/rustdesk/rustdesk/issues/3332

PR #3405, opened 2023-02-27
Modernize dialogs

NicKoehler:
For now only "new folder" in file manager and theme data.

| Before | After |
| -- | -- |
| image | image |


rustdesk:
👍

rakleed:
@NicKoehler maybe add a thin border and a small shadow to the dialog to make it easier to separate it from the background, as many apps in macOS do?

Screenshot 2023-02-27 at 23 22 31

rustdesk:
Can we keep the "cancel" button outlined? Let's see below two buttons on GitHub. I like the new background color of the button, but it is too rounded which looks a little cartoon for me.

image

rustdesk:
@grummbeer

grummbeer:
> Can we keep the "cancel" button outlined? Let's see below two buttons on GitHub.

Agree, github do it right. 

There are buttons on the right side that "confirm" the purpose of the dialog. "apply", "OK", "do", "go". Lefthand are buttons that represent the "safe harbor" with "cancel", "clear", "don't", "don't go". The colors should reflect the consequences.


I must think on it, is this a massive overhaul what affected the whole app, dark/light desktop/mobile and technical concepts as well? And some side effects, depending on the theming, will appear or are already introduced (so far).  


- very good: Indroducing theming.
- good: more and more components covered by theme and goes unified. 
- semi good: the theme changes reveals app wide the affected Elements, good to know but needs to be fixed now. 
- very bad: General theming is not implemented well at all (yet).
- good: Confirm destructive action (peer delete)
- good: Checkboxes are a bit "new rounded" and blue on dark, the checkColor can now be white again (perfect contrast, same color label).
- consider on: How to go rounded and flat but keep usability and accessibility (best UX, just rounded)
    - Flat in flat makes it hard to detect what or where an action is required or possible (button, input field) and where not. The typical design of a button/input does not have this problem (well known concept).
    - Color in color (weak contrasted) makes it hard to distinguish the elements.
- consider on: Beware of the differences between mobile and desktop and the shared parts. e.g. "CustomAlertDialog", "ThemeData".
- Buttons/actions in CustomDialog are bypassed. Why not create a "DesktopDialog" and keep the not fitting CustomDialog for the mobile one. In general i vote for stricter separation of mobile and desktop, components does not fit for both in every case. this Dialog is a good sample. Or make the shared components smaller.
- The approach looks twisted. Its bottom up "Component Dialog > App (theme)" and not top down "App (theme) > Compontent Dialog"

not a easy task … but how about changing focus from specific Component "Dialog" to the App theming first?

1. Implement a stable theming, create the foundation.
   - if possible keep the current look, just covered by the theme.
   - keep the difference between mobile/desktop 
      - create Theme extensions if required, eg. "mobileDark" or "mobileLight" (or similar no idea)
   - split or create dedicated components for mobile/desktop if required
   - organize the colors, currently is wilderness

2. Redesign Components, relies on a well implemented Theme.
   - predictable
   - no surprises
   - no unwanted side effects ("… oops, i wasn't aware of that")
   - easy in the best case, one change for all (per platform)

3. Envolve
   - add new features (e.g. "confirm delete")
   - you need a button for your feature? no problem, there is a quite nice collection of presets for your purpose.

NicKoehler:
@rustdesk do you mean something like this?

image

rustdesk:
@grummbeer Totally agree. This make me miss css / html which is so easy to define style / theme with much less code.

rustdesk:
> @rustdesk do you mean something like this?
> 
> image

It is better if we can see an outlined border.

rustdesk:
Yes， exactly

NicKoehler:
@grummbeer I think you are correct that the theme area requires improvement. Would you be willing to take a look at the issues you mentioned while I focus on the overall components, or perhaps we could collaborate to address the issues together?

grummbeer:
@NicKoehler you merged the master into your dev branch. Have you ever tried a rebase? Things will be much more smoother and keeping your commits clean.

NicKoehler:
@rustdesk I think the pr is ready. Please let me know if you require any additional changes.

rustdesk:
@NicKoehler seems something broken by this merge.

image

NicKoehler:
Sorry I don't see what's wrong in the screenshot, can you explain?

rustdesk:
Old is like this, bigger padding.

image

NicKoehler:
My build from yesterday before merge looks the same

immagine

immagine

rustdesk:
Let me check this.

rustdesk:
related to this change.

image

rustdesk:
Let me fix this, I will add margin to desktop page.

rustdesk:
and the lighttheme seems changed a lot.

rustdesk:
image

image

rustdesk:
@grummbeer how do you think about this background color change in light theme? I still prefer old one.

rustdesk:
Let me have poll for this.

grummbeer:
The colors are now inverted, based on the (needs of) file manager redesign.

|||
|--|--|
|text-overlow-modified-time-basename-in-transfer-cardignore the annotations|file-transfer-light|

The theming of background and card (transfer card) is what is now seen app wide.

- invert the colors (like ad), then the file transfer view is also inverted.
- stick with it an fix what is not looking good / harmonize components
- The sidebar is not a "card", so it can have a different gradation.

If you start the poll, the "Control remote desktop" will tip the scales in favor of "it was better before".

Because it is the most bad looking part, heap-hop stated this already in the merge request for the file manger, it is true. But I think this part looks strange in both variants. It kind of stands alone and gets lost in this "top/header" area. Now, with the same color for the background and the input, it's just a bit uglier than before. At the same time it is one of the hottest areas in the entire app, it could take up the entire header area instead of being a card. I bet changeing this area makes a lot better first impression.

Sorry for always elaborate so much (will try to reduce it) … I think the white background is easier to handle and elements that are not "carded" (eg. header "local"/"remote" in file manager view) look better on a neutral background.

rustdesk:
The new one wins https://github.com/rustdesk/rustdesk/discussions/3484

grummbeer:
:-) I was wrong, no one noticed the bad look of "control remote desktop"

rustdesk:
> At the same time it is one of the hottest areas in the entire app, it could take up the entire header area instead of being a card

I had thought about this also before.

rustdesk:
> no one noticed the bad look of "control remote desktop"

I noticed, :). Both of us are wrong.

rustdesk:
Maybe we can make some decoration with some other light color rather than current light gray color.

rustdesk:
Better?

image

grummbeer:
Better!

rustdesk:
i will do this change.

rustdesk:
Now old wins.

rustdesk:
@NicKoehler @grummbeer I think right aligned buttons look better. How do you think?

image

NicKoehler:
I think centered looks more balanced

grummbeer:
> I think centered looks more balanced

Not in all cases. 

|"simple confirm" nice but a bit low contrasted "cancel"-button | one-liner, small, all centered, nice | "information" nice|
|-- |-- |-- |
|information-confirm-delete|peer-rename-light|connecting-desktop-light|


Widder dialogs, no headline, icon in content, content not centered, but centered button. Not well align. The "Nicement" is reduced.

|widder, semi terrible| smaller, terrible |
|-- |-- |
|connection-error-desktop-light|dialog-information|


Mulitline Dialog, "Nicement" vanished completely. Headline left, buttons centered, no icons in buttons

|space to button differ | space to button |
|-- |-- |
|set-passwort-light|change-id-light|


|left labeled input | hard to figure out where the the eges of the dialog are|
|-- |-- |
modern-dialog-desktop-sideeffect|modern-dialog-desktop-sideeffect-color|


Mobile … 
|before | now | now | now | 
|-- |-- |-- |-- |
|dialog-mobile-before|mobile-now-dark|mobile-now-light-2|mobile-now-light|

It is "bottom up". The focus is on a small, specific component, but the entire application is "designed".


Possible solution:

- Think "top down" App (theme) > influence/designs > components (all the same, including dialogs)
- Do not share dialogs between mobile and desktop, create a component for each to avoid unwanted side effects or to avoid complexity for separating desktop/mobile in one component.
- Extend or replace "DialogButton" with "DialogCancelButton()", "DialogHeadline()", "DialogConfirmButton()", etc, which take care of the look. Always the same icon, spaces, etc.
- Get rid of all these "per situation" applied spaces, colors, icons, etc. 
- Create a Theme part for this dialog components, which defines spaces, colors, etc. per platform

grummbeer:
Let's fix them.

Mobile:

- buttons right or centered?
- headline left or centered?
- Input filled or underlinded? if filled, they need margin bottom.
- fix "OK" button in your own password dialog.

grummbeer:
shared dialogs

On mobile the password dialog behave a bit different, the validation message is show after typing, on desktop the pills react. No problem, because desktop and mobile have there own dialog for password. 

ChangeId is shared, and the behavior is the same as on desktop. [Now](https://github.com/rustdesk/rustdesk/pull/3513) without the pills the validation message is only visible after click. A good opportunity to have a "ChangeId" for both as well?

rustdesk:
> * buttons right or centered?
> * headline left or centered?

I jusy checked whatsapp, headline is left, buttons are right.

rustdesk:
> visible after click.

i think it is fine for mobile.

21pages:
Dark theme menu bar needs some optimization.

rustdesk:
> Dark theme menu bar needs some optimization. rustdesk_ik0Ub4jAnm

fixed https://github.com/rustdesk/rustdesk/pull/3526

grummbeer:
:-) 

- the background color of dialogs are inverted (not that bad, better contrast)
- OK button not white anymore (rely on `Brightness` what is removed)
- the horrible teal is back

colors-inverted-again
colors-teal-back-again

maybe more colors for specific usage are needed instead shareing one for all? one for menus, others for other usage eg. backgrounds?

rustdesk:
:(

grummbeer:
sorry, background not affected, still the same, was local change.

grummbeer:
i am willing to fix the dialogs. 

- rely on theme eg. input style, etc.
- remove all "per situation" applied colors, paddings, etc.
- keep distance between desktop/mobile, if requried eg. filled vs. underlined
- icon for all "confirm", like "OK" or "Apply"
- icon for all "cancel"
- healine left, buttons right,
- keep the possibilty to center some special dialogs like "connecting..." or some simple/small like "confirm delete", without the need to bypass "dialog.actions". eg. "connecting …" (button is placed via content instead of "actions", to solve this)
- headline with optional icon, on the left
- **Doing all this with care** :-)

rustdesk:
@grummbeer 👍 💯 🥇

grummbeer:
@21pages 

I do not intend to break anything again. What exactly was the problem before your change or what is your wish regarding the menu bar?

rustdesk:
> Dark theme menu bar needs some optimization.
> 
> image

Check above menu on dark theme before the fix.

21pages:
White menubar in dark theme

grummbeer:
Thanks. got it.

grummbeer:
@NicKoehler a bit a pitty, but the "modern dialogs" came to fast and do not fit in all situations, but we can't have different dialogs across the application, so i think reverting the rounded is the best for now.

The goal is not doing the best visual styling (i can't at all, i am not a designer), the goal is making a equal look by creating a foundation in the hope afterwards it will be possible forcing **all** dialogs to the same look by changing just a few vars.

All input fields/dialogs looks like this (ignore coloring for now, will be adjusted):

### Mobile

- Step back, close to the defaults, looks clean and pure, not buggy, as long as a better style fit for **all** 
- Inner padding will be adjusted.

| | | colors still too dark|compare/lineage default|
|--|--|--|--|
|Screenshot_20230316-190640_RustDesk|Screenshot_20230316-190718_RustDesk|Screenshot_20230316-172059_RustDesk|Screenshot_20230316-164202_Einstellungen|


### desktop

#### Buttons
   
- Adjust inner padding for better height/width radio
- Unify inner padding for outlined and icon vs. no icon.
- All buttons an desktop will have this look then

|now |after, equal spacing (ignore coloring)|
|-- |-- |
|buttons-unequal-paddings|buttons-after|

#### Input

- **All** Input fields rectangle with rounded corners and bordered. Radius taken from buttons


|before |after | after |
|-- |-- |-- |
|rename|rename-light-normal|input-rename-dark-focused|

|multiline left labeled | mulitline left labeled |
|-- |-- |
|left-labeled-light-after|input-multiline-dark|

@21pages, @rustdesk, @NicKoehler do you agree to implement this? Details can be discussed then in Merge Request.

rustdesk:
I still prefer old accent color, the old one is brighter.

> image

rustdesk:
Yes

> Color too dark
> image

NicKoehler:
> @21pages, @rustdesk, @NicKoehler do you agree to implement this? Details can be discussed then in Merge Request.

I think your changes are looking good, go ahead

grummbeer:
Cool, sounds like a consensus. I'll go ahead with a series of smaller patches instead of one growing and growing single merge request.

rustdesk:
👍

PR #3431, opened 2023-02-28
Update fa.rs

ghost:


grummbeer:
@mehdi-song I am curious, maybe you can help me understand the translations better.

### RTL

The headline of the "RDP Settings" (home | peers > Windows > three dot > RDP (pencil)) is a composition of "RDP + [Settings]", what results in the headline "RDP تنظیمات".

If and how does that affect your LTR flow?

If you could translate the headline in whole, what would it be?

("RDP Settings", "?"),

### context 

Do you always know where the translated text is show in the app? In which context it is used?

thanks.

grummbeer:
@mehdi-song a sample from german translation.

```
("Elevate", "Erheben"),
```

This is a direct word-to-word translation from the dictionary, but in this context it is completely wrong, because "elevate" does not mean "a mountain evelates from the ground" or "a button elevates from background" (which is the German "Erheben"). It is used in the context of "grant access" or "elevate user rights".

I suspect that translators may not know where a text is used, and therefore simply translate what is seen on the left.

What do you think, is something like this possible in fa.rs?

mehdi-song:
Hi @grummbeer 
Yes that right , some times I tested the Pre-release  to see what is wrong in translate and then correct them.
In translate `relay_hint_tip` I saw `en.rs` to find the text message but there is a problem in LTR RTL flow, `("Enable TCP Tunneling", "را فعال کنید TCP تانل")` if I write `TCP Tunneling را فعال کنید` the RTL not correct and I must write `را فعال کنید TCP تانل`.
In little text it's not hard to translate but in long text it is a problem

grummbeer:
Thank you

> `TCP Tunneling را فعال کنید ` the RTL not correct and I must write `را فعال کنید TCP تانل.`

That means that the headline "RDP settings" is not fixable by translation, because the direction is hardcoded in source. Its a bug.

>  I saw en.rs to find the text message

But it holds not all parts, but maybe should. 

Do you have an idea of what would be necessary for you to produce better quality and more contextual translations?

mehdi-song:
Thanks you
I thinks if add direction property `Right to left = true` to labels or TextBoxes with this is condition  `if (language == "Farsi" or language == "Persian") then xx.righttoleft = true` or any other idea.
When anyone add `Arabic` to languages then this problem exist.

grummbeer:
noted, thanks

Issue #3262, opened 2023-02-18
UI (mobile) Canvas overflow 


grummbeer:
### Bug Description

Depend on the translate button text, the button become very large and push the state indicator outside the canvas.

### How to Reproduce

use german language

### Expected Behavior

button and state indicator should be both visible in any case.

### Operating system(s) on local side and remote side

android

### RustDesk Version(s) on local side and remote side

build from source

### Screenshots

State indicator and button are visible at the same time

![stop-service-overflow](https://user-images.githubusercontent.com/67791701/219857866-bdeffdd5-884c-4677-9d41-ff821ddd0619.png)

With german translation

- the state indicator is not visible anymore
- the button itself does not fit

![stop-service-overflow-effect](https://user-images.githubusercontent.com/67791701/219857865-6c8d73c6-dfe3-4d5c-a45b-e90dc294ee1d.png)

### Additional Context

Elements needs to be rearranged

Idea:

Healine (Text what currently is button text)

below a row

- left: state indicator, with short text (ready, not ready)
- right: toggeling [Play][Stop] IconButton

rustdesk:
Thanks, please go ahead. You have been doing great job.

grummbeer:
need to build the apk locallly, but … https://github.com/rustdesk/rustdesk/discussions/763 its very inconvienient waiting for the apk build from workflow

rustdesk:
cargo run -d android?

rustdesk:
@21pages since you are working on android, please fix this.

grummbeer:
thanks :)

> cargo run -d android?

no success, "-d" is a unknown option. please don't waste time, i will get it working, just no time next days.

rustdesk:
@grummbeer you do not need to build rust for android, rust library can be extracted from the apk file directly (apk is zip format, `unzip rustdesk-xxx.apk`, under `lib/arm64-v8a/`).

grummbeer:
thank you for your help, [now I can build the lib locally](https://rustdesk.com/docs/en/dev/build/android/), but it was my device from the beginning … ABI, a bit embarrassing. Had to flash it with a current lineage. 

hopefully the last thing to solve. `flutter run -d [device]`
```
F/flutter (14144): [FATAL:flutter/runtime/dart_vm_initializer.cc(89)] 
Error while initializing the Dart VM: Wrong full snapshot version, 

expected   

   '501ef5cbd64ca70b6b42672346af6a8a' 

found 

[...]

PR #3453, opened 2023-03-01
Rework on Mobile (UI). Share Screen

grummbeer:
### Fixes

- https://github.com/rustdesk/rustdesk/issues/3262 
- https://github.com/rustdesk/rustdesk/issues/3535
- Wrong buttonstyle on "stop service" dialog

### Changes

- Remove hardcoded colors that prevent the nice look on dark mode
- Remove hardcoded fontfamily
- Move server status to the server info
- Move button "start service" near the "service not running" message that refers to it for quick access
- Add headline "your device" according to desktop "your desktop"
- Rename "Password" to "one-time password" waht it is and according to desktop
- Move connection card up into viewport
- Add "check" icon on "accept" button according to dialogs
- Move action button on connection card to right according to dialogs
- Add "info" icon to "no audio" information
- Reduce alpha for avatar circle in light mode for better contrast
- Remove layout shift, no element "hopping" on switch from connection page to server page.
- Change button text from "close" to "disconnect" for "disconnect" buttons
- Using "PaddingCard.title" also for "Service not running" headline
- Make "PaddingCard.title" follow theme (`textTheme.titleLarge`)
- Replace readonly textinput for ID/password, use "copy to clipboard" instead
- Use switches instead of text buttons for "Permissions" 
   - better look
   - well know concept for user
   - avoid confusing on wording. "open/close system settings" vs. "grant/revoke permission"
- Adjust tip text "or OPEN [Screen capture]" to "or enable [Screen capture]"
 
### service not running 

|before |after|
|-- |-- |
|service-not-running-audio-before|service-not-running-audio-after|

### dark

|before|after|
|--|--|
|mobile-server-page-dark-before-1|mobile-server-page-dark-after-2-2|

### light

|before|after|
|--|--|
|mobile-server-page-light-before-1|mobile-server-page-light-after-2-2)

### layout shift

Switch from connection page to server page.

|before |after|
|-- |-- |
|layout-shift-before|layout-shift-after|


grummbeer:
samples for message style

|||
|-- |-- |
|server-page-messages-light|server-page-messages-dark|

The boxed style is not used anywhere else and needs theming. Maybe an option for desktop.

Opinions?

rustdesk:
> 1. Icon. which is it?

ic_stat_logo.png, monochromatic required.

rustdesk:
> * is the `!` a good style in communication?

It looks wierd for me also.

rustdesk:
> * screencapture could be a switch?
> * "on/off" to icons check/cross?
> * theming

Go ahead, I trust your sense of design.

grummbeer:
Rearrange buttons und connection state

quick access to the hottst actions.

If the service is not running, the hottest move is start it. If the service is running, vice versa.

|service not running | service running | 
|--|--|
|server-page-no-service|server-page-connected|

Move the connection state up to the "server"-area, because it shows the server state

|connecting | fail | ready |
|-- |-- |-- |
|server-page-connected-state-1|server-page-connected-state-2|server-page-connected-state-3|

rustdesk:
> Move the connection state up to the "server"-area, because it shows the server state

Do we disable "Stop service" button while connecting?

grummbeer:
> Do we disable "Stop service" button while connecting?

will check it

grummbeer:
> Do we disable "Stop service" button while connecting?

service-stop

depend on `serverModel.mediaOk`
https://github.com/rustdesk/rustdesk/blob/master/flutter/lib/mobile/pages/server_page.dart#L291

should the button be unclickable while `serverModel.connectStatus == 0`?

grummbeer:
button style

button-stop-service

now TextButton

grummbeer:
Remove hardcoded colors from ID/Passwort, let the theme do.

rustdesk:
> should the button be unclickable while `serverModel.connectStatus == 0`?

Yes, i think.

image

grummbeer:
Move connection manager into scope

The connection manager is a first-class information and should therefore be directly accessible.

|before out of scope | after in scope |
|-- |-- |
|server-page-connection-manager-before|server-page-connection-manager-after|

sidenote: the server area has changed since https://github.com/rustdesk/rustdesk/pull/3405, (compare with screenhots above, the lines) also dialogs doesn't look consist anymore, not sure if this was wanted nor if follow ups planned to fix them. e.g. "set your own password" dark/light.

grummbeer:
> screencapture could be a switch?

https://github.com/rustdesk/rustdesk/issues/3535

i am now sure, the well known switches are the best solution.
mobile-server-page-settings-switches

rustdesk:
Looks much better.

grummbeer:
@rustdesk 

How to handle the translation for "android_start_service_tip"? The "OPEN" and "ClOSE" are obsolet now but tip refers to them.

```
"android_start_service_tip", 
"Tap [Start Service] or OPEN [Screen Capture] permission to start the screen sharing service."),
```

changing the english one, and reset the others to ""?


Note: Removing translatable keys is a bit risky because you might miss where this key is still used

grummbeer:
Connection card keep visible on "dismiss". It should removed from view, right?

rustdesk:
why not?

rustdesk:
@Heap-Hop

grummbeer:
> why not? 

The problem popup when the other side click "ok", the card still stay

rustdesk:
> > why not?
> 
> The problem popup when the other side click "ok" on "Closed manually by the peer", the card still stay

@Heap-Hop check this bug

Heap-Hop:
> > > why not?
> > 
> > The problem popup when the other side click "ok" on "Closed manually by the peer", the card still stay
> 
> @Heap-Hop check this bug

Fixed on #3544 .
Same event stream bug.

grummbeer:
todo

- increase size card headline back to 20
- hide "stop service" while in "connecting …" state
- serverinfo vertical align correction

@rustdesk, @NicKoehler please take a look. Screenshots in description.

Questions

### serverinfo

sever-info-alignment

- headline according to desktop, keep, remove?
- soften action buttons as on desktop?
- spacing?
- connection state: headline "connection", keep, rename or remove ([screenshot above](https://github.com/rustdesk/rustdesk/pull/3453#issuecomment-1451187470))?

### "stop service" buttons
 
Button placed under "permissions" but does the same as the switch below it. Is this button acting as some sort of a "emergency button"? like "something unwanted, stop all now!"? then it should placed next to the "ready" at serverinfo?

NicKoehler:
> do we have a "connect/disconnect" icon?

I don't think so, but I can make them if you want

grummbeer:
> I don't think so, but I can make them if you want

I think it would be nice to place them at all connect/disconnect buttons. What do you think?

Any suggestions or opinion on overall visual QA or the "serverinfo" card/"stop service" button?

grummbeer:
>> should the button be unclickable while serverModel.connectStatus == 0?

> Yes, i think.

A bit tricky. The service is independent from the state of the connection. Connection could fail, but the service can run independently. The switch for it is also permanently accessible. If is an emergency "stop everything", it could be visible only when clients are connected? no idea right now.

rustdesk:
connectStat is not client connection state, it is the connection to the RustDesk server.

rustdesk:
@grummbeer blue color here in dark theme looks wierd. light theme is fine.

image

grummbeer:
> connectStat is not client connection state, it is the connection to the RustDesk server.

I know, but the user may not. The connection state to rd server is independent from local screen capture service. So it makes no sense hideing the local "stop service" while connecting to rd severs. Expect there is a technical reason for it. It would be helpful to know what the indention was to show an good visibel button for it even the switch does the same. I an trying my best to figure out what was intended, just to respect and follow that.

grummbeer:
> blue color here in dark theme looks wierd. light theme is fine.

agree, i still would prefer picking a nice color from the logo an replace this dark blue on dark mode everywhere. But this is out of scope for this merge request.

grummbeer:
Please don't be surprised, the colors in the screenshots in the description don't show the whole truth. The colors currently look like this and as they are the current overall state.

colors

The colors from screenshot based on a local theme change (testing).

grummbeer:
@NicKoehler 
>> do we have a "connect/disconnect" icon?

> I don't think so, but I can make them if you want

Yes, please add them for later usage.

rustdesk:
> The connection state to rd server is independent from local screen capture service

It is dependent, without connected to rd server, screen capture service can not be triggered. But here is an exception, direct ip access can work even without connected to rd server, which can trigger screen capture service also.

rustdesk:
Could you resolve the conflicts? Translation files updated too frequently every day. :(

grummbeer:
sure. I am trained on that :-)

rustdesk:
Checkbox/button text color in night theme is expected to be fix in this PR? I just checked, it is black.

grummbeer:
> Checkbox/button text color in night theme is expected to be fixed in this PR? I just checked, it is black.

there black app wide, its up to the theme, which i changed locally but need more test/check for side effects before push it.

rustdesk:
> black app wide

Sorry, i forgot to mention it is on settings page for desktop in night theme

rustdesk:
image

NicKoehler:
@grummbeer What about something like this for connect/disconnect?

immagine

NicKoehler:
I forgot to post them

connect
disconnect

rustdesk:
:)

grummbeer:
Thanks, will try. 

In this size they look very similar, hard to distinguish.

Maybe we need to think about an alternative that is more focused than this long RJ45 cable. When was the last time the user even saw a cable like this? The cable takes up most of the space in this icon, but most of the space should be used for the essentials ("+", "-").

I'm thinking something like a plug or chain, what is faster recognizable as connect/disconnect.
plug

chain-sign-set-collection-icon-600w-1372971545


https://www.shutterstock.com/search/connect-disconnect-icon

NicKoehler:
Ok, I made these too:

connect
disconnect

grummbeer:
Very cool, will take them.

grummbeer:
sorry, one thing to check, do we have more outlined or more filled icons?

NicKoehler:
I think we have more filled

PR #3490, opened 2023-03-03
fix some incorrect capture magnifier logic

fufesou:
Fix some wrong logic with little impact.


rustdesk:
do not remove support message?

PR #3493, opened 2023-03-03
codec benchmark

21pages:
hareware codec is slow now.


rustdesk:
ridculous

21pages:
1920x1080 looks a little bit better

21pages:
master, without any optimization but the amf encoding time is shorter, it may be related to the computer state .

PR #3497, opened 2023-03-03
File Manager. Textoverflow for column file time

grummbeer:
Add Overflow for column file time

### Before
file-transfer-column-time-overflow-before

### After 

file-transfer-column-time-overflow-after


rustdesk:
Great Job.

PR #3498, opened 2023-03-03
File Manager. Show filename instead of path in tranfer list

grummbeer:
Items in tranfer list showing the full path, what is not very useful. Showing file name instaed. Full path is still visible via tooltip.

|before |after|
|-- |-- |
|file-manager-filebasename-before|file-manager-filebasename-after|

Please review code.


grummbeer:
Wait, Linux > windows has a problem with the paths. Maybe a DIRECTORY_SEPARATOR issue

|master| with this patch|
|-- |-- |
|file-manager-filename-windows-broken|file-manager-filename-windows-broken-after|

grummbeer:
just for the record, the bug from comment above is fixed.

linux-windows-paths-in-transfer-card-ok

PR #3555, opened 2023-03-08
Fix. Mobile. Remove left label in RDP dialog

grummbeer:
|mobile before| mobile after |desktop still as wanted|
|--|--|--|
|rdp-mobile-before|rdp-mobile-after|rdp-desktop-after|

### DIFF
--- a/flutter/lib/common/widgets/peer_card.dart
+++ b/flutter/lib/common/widgets/peer_card.dart
@@ -1126,56 +1126,67 @@ void _rdpDialog(String id) async {
             ),
             Row(
               children: [
-                ConstrainedBox(
-                    constraints: const BoxConstraints(minWidth: 140),
-                    child: Text(
-                      "${translate('Port')}:",
-                      textAlign: TextAlign.right,
-                    ).marginOnly(right: 10)),
+                isDesktop
+                    ? ConstrainedBox(
+                        constraints: const BoxConstraints(minWidth: 140),
+                        child: Text(
+                          "${translate('Port')}:",
+                          textAlign: TextAlign.right,
+                        ).marginOnly(right: 10))
+                    : SizedBox.shrink(),
                 Expanded(
                   child: TextField(
                     inputFormatters: [
                       FilteringTextInputFormatter.allow(RegExp(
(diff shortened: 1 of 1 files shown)


rustdesk:
No margin bettwen controls in "mobile after"?

grummbeer:
All multiline dialogs have this look now. :-(


Screenshot_20230303-162157_RustDesk
mobile-now-light

grummbeer:
Before they were underlined and this automaticlly make a visual margin bottom.

rustdesk:
Looks a little wierd without a margin.

image

grummbeer:
To keep the equal look I have added an extra condition for the bottom spacing … in hope of the ability to fix them all in one

```dart
.marginOnly(bottom: isDesktop ? 8 : 0),
```
… crazy but it is a bit of a mess at the moment.

grummbeer:
> Looks a little wierd without a margin.

Absolutly! But what to do? find them all and add margin manually on each?

rustdesk:
That's why I dislike flutter, it is so easy if it is html/css.

rustdesk:
It is also easier if we have a wraper like "EditorGroup", but it seems we do not have.

grummbeer:
> That's why I dislike flutter, it is so easy if it is html/css.

I have learned this also in the meanhwile. It is a shame, can not believe that.


> EditorGroup

maybe a container like this https://stackoverflow.com/questions/53884092/flutter-how-to-set-margin-for-particular-widget-in-a-form must investigate …

The filled came with the modern dialogs, so i guess its defined in the ThemeData. Maybe we could place a `isDesktop` an get them back underlined. A controlled change can then be made afterwards. 

As i statet my willing …

I've already started with a mock for the dialogs, including trying to theme it, but there's a steep learning curve. And a lot I just don't understand :-(

grummbeer:
:-) you know

rustdesk:
Yes, we need thorough test.

rustdesk:
Go ahead, i trust you.

grummbeer:
Merge this, the dialog looks at least then exactly like the others and better than left labeled.

Do you agree with bringing back the underlined version for mobile, which stay as long as a surprise-free design change is possible?

rustdesk:
underlined is ok for me

Issue #3561, opened 2023-03-08
Adding Client to Favorites makes them appear offline. (Nightly Build 1.2.0)


phoenixwarrior95:
### Bug Description

In the "Recent Connections" tab, the client appears online (same with the Discovered tab). Once clicking the three dot menu for that client and selecting "add to favorites", the client shows offline in all 3 tabs but when connecting, the session connects. When removed from favorites, client appears to be online again.

### How to Reproduce

1. Three Dot menu on a client in the "Recent Sessions" tab.
2. Select "Add to Favorites"
3. Client suddenly goes offline. 
4. To fix, remove client from favorites (three dot menu) and client appears to be online again.

### Expected Behavior

Client should be online (assuming that the client _is_ online)

### Operating system(s) on local side and remote side

Windows 11 -> Windows 11 and Android -> Windows 11

### RustDesk Version(s) on local side and remote side

1.2.0 -> 1.2.0

### Screenshots

![image](https://user-images.githubusercontent.com/43919854/223808665-5c7685bc-a6e7-4537-a0d6-f8be7a720bbb.png)
![image](https://user-images.githubusercontent.com/43919854/223808812-58772a28-9100-44c7-a4b5-c1c475e6573a.png)
![image](https://user-images.githubusercontent.com/43919854/223809873-6d59c337-be64-40e8-b11d-0d55642bfcba.png)

### Additional Context

I have also tested this on Android v1.2 Nightly Build and same results.

rustdesk:
@fufesou

fufesou:
@phoenixwarrior95 Thanks for your feedback. This issue should be fixed by PR 3567.

rustdesk:
close as fix merged

phoenixwarrior95:
No problem, @fufesou.

Glad I can help. Android apk seems to be having the same issue. Can you please take a look at it once you have the time? Thanks.

PR #3567, opened 2023-03-09
fix/online states 2

fufesou:
https://github.com/rustdesk/rustdesk/issues/3561


phoenixwarrior95:
Thank you. Has this been fixed for Android as well? @fufesou

fufesou:
> Thank you. Has this been fixed for Android as well? @fufesou
> 
> Android seems to be having this issue still, but it is already fixed for Windows. Thanks.

Thanks for your feedback, I'll try to fix it.

rustdesk:
APK is not updated today because a CI problem.

PR #3591, opened 2023-03-10
Translation missing in File Transfer

Mr-Update:
New Connection
Minimize
Maximize
Restore


grummbeer:
Can you please check this?

```dart
child: ActionIcon(
  message: 'Minimize',
```

A bit above in ActionIcon the given message will translate.

```dart
class ActionIcon extends StatefulWidget {
 // […]
 Widget build(BuildContext context) {
    return Tooltip(
      message: widget.message != null ? translate(widget.message!) : "",
```

Maybe only the strings in the translation file was missing?

PR #3596, opened 2023-03-10
Theme. Add secondary color, fix foreground color for buttons in dark mode

grummbeer:
Bring back the colored checkboxes/radios. Using secondary color to avoid unwanted side effects based on `colorScheme.fromSwatch`.

- CheckColor is now white again, match better now.
- White text for buttons/dark mode 

### Desktop
|||
|--|--|
|secondary-colors-dark|secondary-colors-light|

### Mobile

|||
|--|--|
|secondary-colors-mobile-dark|secondary-colors-mobile-light|

PR #3601, opened 2023-03-10
Design empty peers

NicKoehler:
Implemented what discussed in #3596


grummbeer:
@NicKoehler When switching between tabs, the "empty" message is immediately displayed, but also immediately replaced by existing peers. This is a kind of flash similar to a [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content). Is it possible to change this so that the message only appears when there are no peers returned from the async api call?

rustdesk:
👍

PR #3602, opened 2023-03-11
fix: upgrade ndk version to r23

Kingtous:
This PR fixes the compile for android ndk lib.


grummbeer:
Does this means the docs needs to be updated?

https://rustdesk.com/docs/en/dev/build/android/

> You could download it [here](https://dl.google.com/android/repository/android-ndk-r22b-linux-x86_64.zip), you’ll be required to install it manually or follow [this](https://developer.android.com/studio/projects/install-ndk) instruction.

"here" refers to https://dl.google.com/android/repository/android-ndk-r22b-linux-x86_64.zip

Is it now r23c?

- https://dl.google.com/android/repository/android-ndk-r23c-linux.zip
- https://github.com/android/ndk/wiki/Unsupported-Downloads#r23c

rustdesk:
yes

Issue #3585, opened 2023-03-10
The controlled person's mouse is forcibly dragged to the position where the controller's mouse was before


Seamain:
### Bug Description

When my friend don't control my computer, my mouse will force to move to the previous position of the mouse. It causes me to be unable to do anything

### How to Reproduce

When contrller connects the controlled person's computer, the controlled person's move his/her mouse, it will be reproduced

### Expected Behavior

I have free control of my mouse

### Operating system(s) on local side and remote side

Windows11 Pro for Workstations 22621.1344 -> Windows11 Pro for Workstations 22621.1344

### RustDesk Version(s) on local side and remote side

1.1.9 -> 1.1.9

### Screenshots

![](https://user-images.githubusercontent.com/92930091/224229342-40e8c541-d873-4dce-ab03-b60130b7df53.gif)

rustdesk:
try out nightly build please on both side.

Seamain:
This problem still exists in the nightly build

rustdesk:
@fufesou @chiehw

rustdesk:
Could you test again later? @Seamain

Seamain:
Ok, I will try again later

Seamain:
It's not work

rustdesk:
@fufesou

fufesou:
The mouse movement of the controlled side has a higher priority.

I'm not sure what your problem is now:
1. The mouse cannot be moved.
2. The mouse is moved without interference, but after a while, the mouse will move to the position of the control side.

rustdesk:
I guess the mouse is moving very slightly on controlling side (a poor mouse) though you can not sense.

Seamain:
> The mouse movement of the controlled side has a higher priority.
> 
> I'm not sure what your problem is now:
> 1. The mouse cannot be moved.
> 2. The mouse is moved without interference, but after a while, the mouse will move to the position of the control side.

My problem is 2

fufesou:
> The mouse is moved without interference, but after a while, the mouse will move to the position of the control side.

> I guess the mouse is moving very slightly on controlling side (a poor mouse) though you can not sense.

@Seamain You may need to check if your mouse on the controlling side.

Seamain:
The problem was solved when I and my friend all turned on the display of the remote cursor

rustdesk:
This PR https://github.com/rustdesk/rustdesk/pull/3600 is for you not turning on "remote cursor".

Issue #40, opened 2021-05-17
Give movement priority to host rather than client


jonatino:
Right now the client has priority over the host for controller the mouse. Teamviewer/anydesk use the opposite. Any movement the host does with the mouse will always override any movemented the client connected makes. 

I think rustdesk should follow the same logic.

EkanshBhatnagar:
I want to contribute to solve this issue, is there any contribution guide.

rustdesk:
> contribution guide

No yet. Please create pull request, then I will review. Thanks very much.

practical-mechanics:
I'm not sure why but it's still the same in my end, when I connect to remote PC, I will have a priority controlling the mouse which sometimes makes the guy on the other end go crazy.

jonatino:
@rustdesk @fufesou it would be nice if this was configurable on the client side. I think in 99% of cases you want the remote to always have input priority, but I think there's still a case for the current (client-side priority) behavior. Eg when I'm remote controlling my grandma's computer, she can't help but move the mouse constantly while I'm trying to do things. It'd be nice to allow for Client input priority in this specific case.

rustdesk:
reverted, https://github.com/rustdesk/rustdesk/commit/542d86b667a2da19f89591a03a39edd9ab4d406e, because causing much more trouble than benefits. @fufesou we need another clever solution.

rustdesk:
```
pub fn try_start_record_cursor_pos() -> Option<thread::JoinHandle<()>> {
     ....
    let track_peer = vec![];
    let track_sys = vec![];
    let handle = thread::spawn(|| {
        let interval = time::Duration::from_millis(33);
        loop {
            if !RECORD_CURSOR_POS_RUNNING.load(Ordering::SeqCst) {
                break;
            }

            let now = time::Instant::now();

          // update track_sys here
           // update track_peer here via LATEST_PEER_INPUT_CURSOR

          // compare the track
           
            let elapsed = now.elapsed();
            if elapsed < interval {
                thread::sleep(interval - elapsed);
            }
        }
        update_last_cursor_pos(INVALID_CURSOR_POS, INVALID_CURSOR_POS);
    });
    Some(handle)
}
```

rustdesk:
https://github.com/rustdesk/rustdesk/discussions/6178#discussioncomment-7390701 a good suggestion, i had done similar logic in Sciter version.

My1:
[...]

PR #3604, opened 2023-03-11
periodicly get cursor pos on conn

fufesou:
https://github.com/rustdesk/rustdesk/issues/3585

If "Show remote cursor" is not checked, "mouse_pos" service has no subscribers.
Then the controlled side cannot detect if local mouse is moved.
There's no way to set local mouse with higher prority.


rustdesk:
#40

PR #3608, opened 2023-03-11
fix remote menubar theme

21pages:
| before | after |
| ------ | ----- |
|  ac69c06fb26e26d823da115c59116c0   |   0becb7a518addcb314c537454f96e3e  |
|   8f7727cbec8900a09d1234bd2c8312d |  |

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -111,32 +111,39 @@ class ColorThemeExtension extends ThemeExtension<ColorThemeExtension> {
     required this.border,
     required this.border2,
     required this.highlight,
+    required this.drag_indicator,
   });
 
   final Color? border;
   final Color? border2;
   final Color? highlight;
+  final Color? drag_indicator;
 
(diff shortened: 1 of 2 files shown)


grummbeer:
What do you think of the radius on top?

224482909-43f3fc3a-c4cc-41ea-bdda-7b5f55f348c6

the dragger is also not rounded on top and it looks clear connected to the bar. The bar could look the same, because it is connected to the upper frame?

224482338-cd71d4c7-9078-4bdc-a1cd-08dc4abb0e99

- connected to bar 
- no small line above the bar … better!

21pages:
> bar connected to frame, like the drag

agree

rustdesk:
top radius was added after a dicussion with @NicKoehler

grummbeer:
btw @21pages you added a new named color `drag_indicator` to theme :+1: :+1: :+1: 

The right way to go i think, I will follow this approach also.

Issue #3472, opened 2023-03-02
Linux > Android, cursor says "input not allowed", but it is.


grummbeer:
### Bug Description

Ambiguous communication, the symbol on the mouse signals that the input is not allowed, but it is possible.

### How to Reproduce

Connect to android, "Input control" is "off", set it to "ON", the cursor symbol is still signals "not possible"

### Expected Behavior

Synbol should not be visible if input is allowed.

### Operating system(s) on local side and remote side

Linux > Android

### RustDesk Version(s) on local side and remote side

source

### Screenshots

![icon-on-mobile-ambiguos](https://user-images.githubusercontent.com/67791701/222460260-825b9549-f9b5-4004-be74-08ba066d6302.gif)

### Additional Context

debug mode

rustdesk:
@fufesou

fufesou:
@grummbeer Thanks for your feedback.
It seems android does not send permissions back to the control side.

rustdesk:
@Heap-Hop

grummbeer:
Something changed? Now i have always a blue one.
![icon-on-mobile-ambiguos-next](https://user-images.githubusercontent.com/67791701/222896147-bdc6f4da-2c40-432b-8016-dadf17bf541e.gif)

rustdesk:
nothing change yet

Heap-Hop:
#3544 only fix it partially.

There needs a feature on Android, switchs of controlling the permissions of each client, like desktop connection manager.
![cm_permission](https://user-images.githubusercontent.com/62206297/223389522-12579fd4-47e0-4a6b-bff2-594ffd76d822.png)

rustdesk:
so to avoid confusing, we should disable permission switching while there are connections, or pop up warn message to tell it will take effective with new connection.

PR #3627, opened 2023-03-13
Android server, add android_permission_may_not_change_tip for established connections.

Heap-Hop:
- Alert will show when permissions are changed while connections are established.

android_server_tip


Heap-Hop:
https://github.com/rustdesk/rustdesk/issues/3472

Issue #3508, opened 2023-03-04
File Manager. Dialog Confirm overwrite does not behave as expected.


grummbeer:
### Bug Description

The confirmation dialog for overwriting an existing file does not work properly.

1. sometimes it does not appear at all, files are silently skipped.
2. sometimes it appears only in one direction. Forgot which one it was, can't reproduce it at the moment because of "1."
3. when selecting multiple files, e.g. "select all", many dialogs are opened, but not as many as files were selected.
4. "Do this for all conflicts" does not work, behaves the same as "close this dialog".

### How to Reproduce

Open File Manger, select more than 5 files and click "receive" or "send", repeat it, play around.

### Expected Behavior

1. Dialog always displayed when a file already exists in goal.
2. "Do this for all conflicts" do so.

### Operating system(s) on local side and remote side

Linux > Android

### RustDesk Version(s) on local side and remote side

source

### Screenshots

Selected files are more than 5.

![confirm-overwrite-buggy](https://user-images.githubusercontent.com/67791701/222898488-3c7bc086-1138-48ca-9e6d-e40c2ae70f56.gif)

Dialogs disappear after a few clicks, but selected files are more.

Wanted to create a new video, but can't reproduce this anymore, instead i experienced that the files are skipped silently.

![confirm-overwrite-skipped-automaticlly](https://user-images.githubusercontent.com/67791701/222898541-bacff09b-c482-4a66-8938-2625089fa24a.gif)

rustdesk:
@Heap-Hop

rustdesk:
@grummbeer could you confirm if this is fixed?

rustdesk:
> @grummbeer could you confirm if this is fixed?

Ignore this please, not fix this yet.

Heap-Hop:
> 1. sometimes it does not appear at all, files are silently skipped.
> 2. sometimes it appears only in one direction. Forgot which one it was, can't reproduce it at the moment because of "1."
> 3. when selecting multiple files, e.g. "select all", many dialogs are opened, but not as many as files were selected.
> 4. "Do this for all conflicts" does not work, behaves the same as "close this dialog".

1,2,3 are expected. The current rule for handling same files is that **if the file name, creation date, and modification date all are the same at both ends, it will be skipped by default**. In other cases, the dialog will be showed.
 
4 is a bug. Need to fix this first. 
Also, `many dialogs are opened` (in 3.) which is the same problem.
[...]

PR #3658, opened 2023-03-15
feat: enqueue file overwrite confirm dialogs

Kingtous:
This PR implements an general event loop and a derived file confirm dialog loop, which fixes the remembered checkbox with multiple selection of files.

### DIFF
--- a/src/lang/template.rs
+++ b/src/lang/template.rs
@@ -477,6 +477,6 @@ pub static ref T: std::collections::HashMap<&'static str, &'static str> =
         ("Empty Username", ""),
         ("Empty Password", ""),
         ("Me", ""),
-        ("This file is identical with the peer's one", "")
+        ("identical_file_tip", "")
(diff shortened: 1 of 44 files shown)


src/lang/template.rs
```diff
@@ -477,5 +477,6 @@ pub static ref T: std::collections::HashMap<&'static str, &'static str> =
         ("Empty Username", ""),
         ("Empty Password", ""),
         ("Me", ""),
+        ("This file is identical with the peer's one", "")
```
rustdesk: change to identical_file_tip please?

rustdesk (changes requested):

src/lang/template.rs
Kingtous: Sure, fixed.

rustdesk:
https://github.com/rustdesk/rustdesk/issues/3508

PR #3676, opened 2023-03-16
fix remote DirectoryOptions isWindows initialization failure (not clickable issue on #3508)

Heap-Hop:
- https://github.com/rustdesk/rustdesk/issues/3508#issuecomment-1470051893


rustdesk:
This solves `not clickable` issue.

Issue #3653, opened 2023-03-14
Nightly support for arm 32bit?


fragtion:
### Bug Description

The nightly apk failed to install on my Galaxy S5 running LineageOS v19, but F-Droid version worked fine. Is it possible to add support for these devices to the build script as well?

### How to Reproduce

Try to install the nightly apk on an arm 32 bit Android device

### Expected Behavior

It should work just as the official release does

### Operating system(s) on local side and remote side

Android 12

### RustDesk Version(s) on local side and remote side

nightly

### Screenshots

none unless requested

rustdesk:
Low priority, but if you can not wait, you can do it yourself, https://github.com/rustdesk/rustdesk/blob/master/.github/workflows/flutter-nightly.yml, no hard.

fragtion:
Yeah I tried to fork and build myself but something went wrong with dependencies and then I thought "Why going through this struggle if it's a simple fix to add the flag to the nightly build so others don't face the same problem" which is why I decided it would be worth it to log the issue. But I understand it's a low priority. Thank you for your consideration :)

rustdesk:
added

PR #3692, opened 2023-03-17
feat: add armeabiv7 android build 

Kingtous:
This PR aims to uniform all dependencies on Android to one repo with prebuilt libs included, which reduces the complexity of compiling our RustDesk on Android. And also adds arm 32bit build of rustdesk for android.

related: #3653


fragtion:
This is awesome. Looking forward to i686's return now as well ;D

Issue #2549, opened 2022-12-14
problem with different kayboard language layout  (keyboard translation)


Darkmagister:
**Describe the bug you encountered:**

i'm using rustdesk from linux mint with us keyboard, i'm connecting to a windows machine with IT keyboard layout set, i get some really strange behavior with the key that have different function in the two layout
for example if i press the key **=** on my us keyboard i got **ì** in map mode or  **0** in mode legacy
(i've tried the 1.2.0 nightly also)

**What did you expect to happen instead?**

i was expecting to write = if i press = not to be converted in the same key position or similar

**How did you install `RustDesk`?**

from github release

**RustDesk version and environment**

1.1.9 and 1.2.0 nightly (even with the two keyboard mode it doen't work as i was expecting)

rustdesk:
#1547

fufesou:
@Darkmagister Hi, would you please try again with the latest ngihtly build. It may have been fixed now.
Choose **Map Mode**, the legacy mode has not been fixed.

Darkmagister:
i just tried again ... and same behavior as before,
in map mode i press on my english keyboard (linux)  -_ key and on my other pc with italian map (windows is mapped like that) i get '? that is the same key in the 2 different layout same physical key but i get the layout of the machine that i'm connected to instead of the one on my local machine

it's difficult to remember the layout of different language i think the best way is i to get the character from the local machine to be sent to the remote one ...

fufesou:
@Darkmagister Thanks for your test and replay.
But I don't think so send character to the remote is a good way.
When connected to a remote computer, input should behave like a local keyboard connected to the remote computer. The layout of the local keyboard and the system input of the remote computer should be used.

fufesou:
@Darkmagister Yes, you're right on some cases.
1. Send characters. The "Legacy mode" (But not work well for now).
2. Send code according to keyboard physical position. The "Map mode".

rustdesk:
translation mode is ready

glundby:
Remote keyboard 'norwegian' on target ubuntu shows like from US ASCI key positions. When target is a Windows machine the keys shows like 'norwegian' as they should. Using VNC Viewer the ubuntu machines shows 'norwegian' as they should! øæå....

PR #3719, opened 2023-03-20
translate mode: support x11 in server

chiehw:
#2549


libs/enigo/src/linux/nix_impl.rs
```diff
@@ -42,6 +42,37 @@ impl Enigo {
...
+                        Some(key) => key,
+                        None => {
+                            return Err("Failed to convert key to TFC_Key".into());
+                        }
+                    };
+                    tfc.key_click(tfc_key)
+                }
+            };
+            if res.is_err() {
+                Err("Failed to click char by tfc".to_string().into())
```
fufesou: `Err("Failed to click char by tfc".into())`

chiehw: `Err(format!("Failed to click {:?} by tfc", key).into())`

PR #3764, opened 2023-03-24
refactor dialog

fufesou:
Mainly move dialogs from `mobile` path to `common` path.

### DIFF
--- a/flutter/lib/common/widgets/dialog.dart
+++ b/flutter/lib/common/widgets/dialog.dart
@@ -517,9 +517,12 @@ _connectDialog(
         mainAxisAlignment: MainAxisAlignment.center,
         children: [
           Icon(Icons.password_rounded, color: MyTheme.accent),
-          Text(translate(usernameController == null
-                  ? 'Password Required'
-                  : 'Login Required'))
+          (usernameController == null
+                  ? Text(translate('Password Required'))
+                  : Tooltip(
+                      message: translate('login_linux_tooltip_tip'),
+                      child: Text(translate('login_linux_tip')),
+                    ))
               .paddingOnly(left: 10),
         ],
       ),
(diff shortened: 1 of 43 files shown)


rustdesk:
@grummbeer do you have any idea about this dialog design which is login to headless linux? Different from normal login, remote linux user/password is required for headless linux. I still think current dialog does explain clearly, the users can be confused by the additonal user name and password.

grummbeer:
https://github.com/rustdesk/rustdesk/blob/9ec0b4be56f04eb315ae1e8ada398a3e7eb49acb/flutter/lib/common/widgets/dialog.dart#L544


both sides are the same

rustdesk:
> https://github.com/rustdesk/rustdesk/blob/9ec0b4be56f04eb315ae1e8ada398a3e7eb49acb/flutter/lib/common/widgets/dialog.dart#L544
> 
> both sides are the same

@fufesou

rustdesk:
> On the first glace, the message from toolip could placed directly obove the system login to avoid question marks.

Yes, @fufesou place on the. dialog directly with small font compared with title, and change title to original one (before this PR). You even do not put a big question mark there, no guys know moving mouse there to show a tooltip.

rustdesk:
and this seperate is not eye-catching enough, when I say put a seperator there, you should have some thinking also, not just for finishing an assigned job.

image

grummbeer:
According to material https://m3.material.io/components/dialogs/guidelines#fb75c64c-81c1-45ea-b97b-eb0bcb4c5dc9

The headline should smaller and a kind of heading/description goes to both sections. Section 1. Login to system, section 2. Login to rustdesk.

rustdesk:
> Section 1. Login to system, section 2. Login to rustdesk.

Yes, @fufesou

rustdesk:
Seperator line can be remove if so, I think.

grummbeer:
> Seperator line can be remove if so, I think.

yes separation made by heading.

grummbeer:
sorry if this is a stupid question but …

https://github.com/rustdesk/rustdesk/blob/12a5a7cf8c277dabe02b197c227cf7c51e814ba1/flutter/lib/common/widgets/dialog.dart#L499-L509

`final username` and `final password` not used due commented out in login() arguments.

1. What happen with them on submit?

2. Can you clarify the situation with a comment? or similar
   It is irritating when stumple upon such situations because it is not clear what is intended. Just a forgotten test? Something unfinished? Wanted as hint for later work? Can it be saftly removed? Or going the save way and leaveing it untouched to prevent the author from being pissed?

fufesou:
> `final username` and `final password` not used due commented out in login() arguments.

They should be sent to the controlled side. These variables are keet here for the near future work.

grummbeer:
Thank you for the clarification!

PR #3791, opened 2023-03-26
Fix translate mode in win

chiehw:


rustdesk:
@fufesou

fufesou:
@cypherpunks03 Thanks for your feedback.

I can see the hotkey is broken by [this PR](https://github.com/rustdesk/rustdesk/pull/3719/files#diff-792b8d0bfd0c836c48cce2af6cc7d60ec434486ba56de63ffbaa1a6d09d17246).



If Win -> Win/macOS, the hotkeys may do not work on custom layouts.


But I'm not sure how hotkey is broken is **this PR**.

Can you please give more details about the hotkeys' broken?

1. What are the OS on both sides?
2. What are the key combinations? Did you have "Shift" in your key combinations?

fufesou:
This PR has fixed the issue above.
https://github.com/rustdesk/rustdesk/pull/3853

@cypherpunks03 But it may not fix your issue.

cypherpunks03:
Terrible, it would be better if I did not write this issue at all. Now even more broken than before.

I'm using a custom Colemak layout on top of QWERTY virtual keys so that when using hotkeys, they are interpreted as QWERTY-based. This allows me to use custom layouts from other languages while maintaining the same way of using hotkeys. I am using Win2Win connection in RustDesk. The remote machine does not have my layouts. I need to write on a remote machine on the Сolemak layout and use QWERTY-based hotkeys. The first implementation of the key translation mode worked perfectly for this task.


[colemak-dhm.zip](https://github.com/rustdesk/rustdesk/files/11124919/colemak-dhm.zip)

rustdesk:
@fufesou revert back first if it causes more trouble than goodness.

fufesou:
@cypherpunks03  Thanks for your feedbak. But I'm confused about your problem.

https://github.com/rustdesk/rustdesk/pull/3853 
This PR do only change the behavior of hotkeys on Win -> Win and Win -> macOS. It does not affect the letter translations.

I'll try to find the translation problem.

fufesou:
@cypherpunks03 I've tried your layout. 

For hotkeys in "Translate Mode", there are two ways to simulate:

1. Pass virtual keycode  
    Friendly to the original input layouts. But for custom input methods, the hotkeys keep original for the virtual key never changes.
2. Pass unicode characters  
    Friendly to the custom input layouts. But only for shortcut keys of ascii code. Cann't simulate some hotkeys that are not ascii.


I've also tested pressing 'e' and Ctrl + 'e', results:
```
Local:    'f', Ctrl + VK_E
TeamView: 'f', Ctrl + VK_E
AnyDesk:  'f', Ctrl + VK_F
RustDesk: 'f', Ctrl + VK_E
```


https://user-images.githubusercontent.com/13586388/229271657-79224ce9-832e-459a-a53e-585081306d62.mp4

cypherpunks03:
@fufesou Now try to write the letter 't' and use it to search for text in a Notepad by pressing Ctrl + VK_F in translation mode or to save a text by pressing Ctrl + VK_S ('r').

fufesou:
> Now try to write the letter 't' and use it to search for text in a Notepad by pressing Ctrl + VK_F in translation mode or to save a text by pressing Ctrl + VK_S ('r').

Yes.

I'm not very sure which of the above two ways is better.
But for me, keeping local operations seems better.

cypherpunks03:
@fufesou And use any other applications for tests, such as Notepad.exe or Firefox.exe, and not Visual Studio Code.

fufesou:
@cypherpunks03 Yes, you're right.

I've also tested notepad.exe. Pressing 'e' and Alt + 'e' results:
```
Local:         'f', Alt + f
TeamView: 'f', Alt + e
AnyDesk:   'f', Alt + f
RustDesk:  'f', Alt + e
```

fufesou:
I think vscode use [virtual keycode](https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes).
Some other applications use the literal value of the key.

Scancode is related to the position of the keyboard.

fufesou:
It may be better if both unicode and scancode are passed to the remote side.

1. If the unicode can be translated to a scancode in current layout. Simulate the scancode.
2. Or simulate the passed scancode.

But for now, the message does not allow to send the both codes.

fufesou:
I agree that the hotkey should follow the custom layout.
Using virtual keycode to simulate hotkey may be not a good choice.


> Can you please give more details about the hotkeys' broken?
What are the key combinations? Did you have "Shift" in your key combinations?

@cypherpunks03 Back to my previous quenstion. I still don't understand why releasing "Shift" break your hotkey.

cypherpunks03:
@fufesou Yes, all works now. Thanks! But now I have a problem with the Win + VK_ hotkeys. For example Win+(s) == Win+VK_D should minimize all windows. Check this, please.

fufesou:
> The Win + VK_ hotkeys. For example Win+'s' == Win+VK_D should minimize all windows. 

@cypherpunks03 This bug should be fixed by https://github.com/rustdesk/rustdesk/pull/3888.
The PR also try to simulate key combinations in the order of unicode and virtual keycode.
It is fit for both custom layout and all default layouts.

But, as mentioned in the PR, please use both the new build.

PR #3805, opened 2023-03-27
Delete duplicates tags from addressbook

djordans:


flutter/lib/models/ab_model.dart
```diff
@@ -39,6 +39,7 @@ class AbModel {
             peers.clear();
             if (data['tags'] is List) {
               tags.value = data['tags'];
+              tags.value = tags.toSet().toList();
```
rustdesk, line 42: order changed

PR #3806, opened 2023-03-27
add content-type to api/ab/get

djordans:


flutter/lib/models/ab_model.dart
```diff
@@ -27,7 +27,9 @@ class AbModel {
     abError.value = "";
     final api = "${await bind.mainGetApiServer()}/api/ab/get";
     try {
-      final resp = await http.post(Uri.parse(api), headers: getHttpHeaders());
+      var authHeaders = getHttpHeaders();
+      authHeaders['Content-Type'] = "application/json";
+      final resp = await http.post(Uri.parse(api), headers: authHeaders);
```
rustdesk, line 32: Since it is working without 'Content-Type', why do you bother to add it?

djordans: Just to make sure the api server responds in json and not in byte or xml or whatever, our api server only accepts requests that have a content-type to process responses correctly

Issue #65, opened 2021-06-07
Privacy Mode (Blacken the screen)


rustdesk:
Need your help, I have no idea how to implement it yet.

geshii:
You mean, clicking on a button to hide the screen ? But still keep the audio?

rustdesk:
> You mean, clicking on a button to hide the screen ? But still keep the audio?
> Sorry for bumping by the way

Blacken the physical screen, but remote display still works

rustdesk:
> > You mean, clicking on a button to hide the screen ? But still keep the audio?
> > Sorry for bumping by the way
> 
> Blacken the physical screen, but remote display still works

Need to write display card driver.

rustdesk:
https://www.wowza.com/community/t/blocking-screen-capture/38423/5    

> Actually blocking screen captures probably requires hooking into the video drivers to detect if anything else hooks into the video drivers.

geshii:
> 
> 
> > > You mean, clicking on a button to hide the screen ? But still keep the audio?
> > > Sorry for bumping by the way
> > 
> > 
> > Blacken the physical screen, but remote display still works
> 
> Need to write display card driver.

Oh I see ! But how will you re-enable it after ? If you want to enable it from the physical pc ? A keybind ?

geshii:
https://en.wikipedia.org/wiki/Hardware_overlay
Something like this ?

rustdesk:
> https://en.wikipedia.org/wiki/Hardware_overlay
> Something like this ?

Maybe, I am not sure if "hardware overlay window" affects remote display

rustdesk:
> re-enable

After connection closed

geshii:
> After connection closed

Ok

> Maybe, I am not sure if "hardware overlay window" affects remote display
I will try to learn more info about this ASAP

rustdesk:
@geshii Thanks

atomsos:
Software 向日葵(https://sunlogin.oray.com/download/) and teamviewer has the function. It seems involving a driver hacking.

e.g. 
![](https://pic2.zhimg.com/80/v2-bd0711fd67d0a9f9b5d0a7222b2ea1e9_720w.jpg)

![](https://pic3.zhimg.com/80/v2-9e93c9d17e8628af2033670856b2bb96_720w.jpg)

rustdesk:
[...]

PR #3815, opened 2023-03-28
Privacy Mode / Blacken Screen Off on Android remote side

sabxp:
I test privacy mode but it didn't work when Android is the remote host.
Screen off an android device while you are connected to this, it would be great to reduce battery consumption and increase privacity.

There is another github project here that have the ability to screen off the Android device screen while you interact with the device from a computer, it's here:
https://github.com/Genymobile/scrcpy

Maybe someone can take some code from this to apply here. If you call "scrcpy -Sw" it opens the Android Device Screen into the computer and blacken the Android's Device screen. You can interact with the android device from the computer and the Android device screen remains blacken or off.


sabxp:
I found this: Question about how can scrcpy tool can control a phone even its screen is turned off #2888
https://github.com/Genymobile/scrcpy/issues/2888

private API setDisplayPowerMode()

Sorry if I post this in bad place. I'm newby in github project request.

rustdesk:
You can dive into this, but we have no plan to support this yet, our time table is too full.

rustdesk:
https://github.com/rustdesk/rustdesk/issues/65

Issue #534, opened 2022-05-16
Audio latency will keep accumulating, causing lag and inconsistency


AutumnSun1996:
**Describe the bug you encountered:**

Found a large lag for Audio latency will keep accumulating, causing lag and inconsistency

**What did you expect to happen instead?**

Audio and video keep in sync.

**How did you install `RustDesk`?**

Both Windows portable version, from github releases.

---

**RustDesk version and environment**
Both RustDesk 1.1.8 on Windows10

The lag seems to keep accumulating, so after a long connection with audio on, there will be a noticeable lag.

I did a test with this video: https://www.bilibili.com/video/BV1m44y1t7Wh/
There is a lag for several seconds.

rustdesk:
Try out 1.1.9 please.

AutumnSun1996:
Will try 1.1.9 when I can access the remote machine again. just failed to install the new version remotely.
Is there any change for audio lag in 1.1.9?

rustdesk:
> Will try 1.1.9 when I can access the remote machine again. just failed to install the new version remotely. Is there any change for audio lag in 1.1.9?

A lot

AutumnSun1996:
The bug still exists in 1.1.9.
And I confirmed the steps to reproduce:
1. Connect to remote.
2. Turn audio on, and let something play on remote.
3. Wait for several minutes and the bug appears

Here is a video I recorded after 10 minutes wait, there is a lag of roughly 3 seconds.
[rustdesk-lag.zip](https://github.com/rustdesk/rustdesk/files/8699648/rustdesk-lag.zip)

rustdesk:
@Heap-Hop Have a look please.

Heap-Hop:
@AutumnSun1996 thx!

rustdesk:
@AutumnSun1996 could you please try out https://github.com/rustdesk/rustdesk/releases/download/tmp/rustdesk-1.1.9-putes.exe? need reinstall on both side.

AutumnSun1996:
Problem still exists.
Looked at the code #560, and seems u added some sort of timestamp at send time.
How about just increase playback rate if audio buffer at the receiver is large?

this is what i use for my websocket based remote desktop solution, and it works well

rustdesk:
close to favor https://github.com/rustdesk/rustdesk/issues/3748

rustdesk:
> this is what i use for my websocket based remote desktop solution, and it works well

[...]

PR #3816, opened 2023-03-28
audio latency

21pages:
#534 
There is a latency at the beginning in my test, and it does not increase over time.
The control end don't have enough time to initialize the audio device after receiving the AudioFormat,  so don't send AudioFrame util one second after sending AudioFormat, and it does helps in my test.
If video latency does increase over time, this PR will not help.

https://github.com/rustdesk/rustdesk/blob/81bc362b2b414414b58360f87040749dab8465ab/src/client.rs#L865
data_callback using old data may be the reason.
Input data callback maybe faster than output data callback, which cause increased delay.
https://github.com/RustAudio/cpal/issues/766

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4815,6 +4815,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ringbuf"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "93ca10b9c9e53ac855a2d6953bce34cef6edbac32c4b13047a4d59d67299420a"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "rpassword"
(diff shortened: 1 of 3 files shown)


rustdesk:
I prefer drop them on receiving side before initialization done, then no need to assume it is 1 seoncd.

rustdesk:
> Input data callback maybe faster than output data callback, which cause increased delay.

So we can use this https://github.com/rustdesk/rustdesk/issues/534#issuecomment-1144564127 method. By adjusting output sample_rate (higher a little than real data sample_rate), or find out why play callback can not follow up.

rustdesk:
The audio data flow can not be synced with playback, one is not stable, the other is fixed speed (playback rate). To avoid clipping, have to find way to dynamically change playback rate. Sadly, we can not find in cpal so far, let do it in the future.

PR #3829, opened 2023-03-29
Release modifiers when exit

chiehw:
Unexpected exit, modifiers are easy to stick. So release all modifiers on exit/breakdown.

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -618,6 +618,7 @@ impl Connection {
         }
         #[cfg(target_os = "linux")]
         clear_remapped_keycode();
+        release_modifiers();
         log::info!("Input thread exited");
     }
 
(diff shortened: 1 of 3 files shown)


src/server/connection.rs
```diff
@@ -618,6 +618,7 @@ impl Connection {
         }
         #[cfg(target_os = "linux")]
         clear_remapped_keycode();
+        release_modifiers();
```
rustdesk, line 621: Suppose a scenario, there are two connections, one is pressing shift, the other connection is closed.

grummbeer:
$ flutter/ndk_arm64.sh

> error[E0425]: cannot find function `release_modifiers` in module `crate::input_service`
  --> src/platform/mod.rs:39:27
   |
39 |     crate::input_service::release_modifiers();
   |   

https://github.com/rustdesk/rustdesk/actions/runs/4550478827/jobs/8023569658#step:11:55

PR #3854, opened 2023-03-31
Release modifier when multi conn

chiehw:
#3829


src/server/connection.rs
```diff
@@ -1366,6 +1368,28 @@ impl Connection {
                         } else {
                             me.press
                         };
+
+                        let key = match me.mode.unwrap() {
```
rustdesk: is this unwrap safe?

rustdesk (changes requested):

src/server/connection.rs
```diff
@@ -1366,6 +1368,28 @@ impl Connection {
...
+                            KeyboardMode::Map => Some(crate::keyboard::keycode_to_rdev_key(me.chr())),
+                            KeyboardMode::Translate => {
+                                if let Some(key_event::Union::Chr(code)) = me.union.clone() {
```
rustdesk: This clone seems unnessessary?

rustdesk (changes requested):

PR #3962, opened 2023-04-07
Update devcontainer.json

junbinhe:
need ndk 22 strip funciton


rustdesk:
do something really useful please.

PR #3968, opened 2023-04-08
Refact/led sync

fufesou:
1. Fix build
2. Move `LAYOUT` to the connrect place.
```rust
        unsafe {
            LAYOUT = std::ptr::null_mut();
        }
```
3. Check is numpad key when sync `NumLock`.
4. The code is very strange. Change this code to the logic of version 1.1.9.
```rust
#[cfg(target_os = "windows")]
fn has_numpad_key(key_event: &KeyEvent) -> bool {
    key_event
        .modifiers
        .iter()
        .filter(|&&ck| NUMPAD_KEY_MAP.get(&ck.value()).is_some())
        .count()
        != 0
}
...
fn is_numlock_disabled(key_event: &KeyEvent) -> bool {
    // disable numlock if press home etc when numlock is on,
    // because we will get numpad value (7,8,9 etc) if not
    if key_event.mode.enum_value_or(KeyboardMode::Legacy) == KeyboardMode::Legacy {
        has_numpad_key(key_event)
    } else {
        false
    }
}
```

The logic of 1.1.9 is 
```rust
    match evt.union {
        Some(key_event::Union::control_key(ck)) => {
            if let Some(key) = KEY_MAP.get(&ck.value()) {
                #[cfg(windows)]
                if let Some(_) = NUMPAD_KEY_MAP.get(&ck.value()) {
                    disable_numlock = en.get_key_state(Key::NumLock);
                    if disable_numlock {
                        en.key_down(Key::NumLock).ok();
                        en.key_up(Key::NumLock);
                    }
                }
...
```

### DIFF
--- a/libs/enigo/src/win/win_impl.rs
+++ b/libs/enigo/src/win/win_impl.rs
@@ -359,9 +359,6 @@ impl Enigo {
     }
 
     fn key_to_keycode(&self, key: Key) -> u16 {
-        unsafe {
-            LAYOUT = std::ptr::null_mut();
-        }
         // do not use the codes from crate winapi they're
         // wrongly typed with i32 instead of i16 use the
         // ones provided by win/keycodes.rs that are prefixed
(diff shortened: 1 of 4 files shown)


rustdesk:
maybe need to do a benchmark for match!, no idea its efficiency

https://kevinlynagh.com/notes/match-vs-lookup/

fufesou:
Considering efficiency here does not seem to make much sense.

If it is comparing `HashSet` and `match!`, there will be one more step to calculate the hash value.

PR #4031, opened 2023-04-12
feat: add a rust example for writing a custom plugin

Kingtous:
This PR changes a dummy C example to real rust examples, for the following reason.

- With a real example in `examples/custom_plugin`, devs can easily copy the examples to develop their own custom plugins with a standard project structure.
- With Rust, we can easily use the struct and APIs in the `librustdesk` without extra extern definitions like FFI.

Note that we should only keep FFI types for all APIs, so we can create SDK with more languages support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,5 @@ lib/generated_bridge.dart
 .vscode-server/
 .ssh
 .devcontainer/.*
+# build cache in examples
+examples/**/target/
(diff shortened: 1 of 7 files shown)


examples/custom_plugin/Cargo.lock
Kingtous, line 1: Maybe the Cargo.lock is too large for this PR, maybe we can delete it.

rustdesk:
Go ahead with https://github.com/rustdesk/rustdesk/discussions/878, making an example without a truely working plugin does not make sense too much.

Kingtous:
> Go ahead with https://github.com/rustdesk/rustdesk/discussions/878, making an example without a truely working plugin does not make sense too much.

Yes, I agree. More thoughts may be needed how to make api to be generic and reuseful.

PR #4036, opened 2023-04-13
Update Italian language

bovirus:


src/lang/it.rs
```diff
@@ -445,11 +445,11 @@ pub static ref T: std::collections::HashMap<&'static str, &'static str> =
         ("Display", "Visualizzazione"),
         ("Default View Style", "Stile visualizzazione predefinito"),
         ("Default Scroll Style", "Stile scorrimento predefinito"),
-        ("Default Image Quality", "Qualitài immagine predefinita"),
+        ("Default Image Quality", "Qualità immagine predefinita"),
         ("Default Codec", "Codec predefinito"),
         ("Bitrate", "Bitrate"),
         ("FPS", "FPS"),
-        ("Auto", "Automatico"),
+        ("Auto", "Codec automatico"),
```
rustdesk, line 452: But "Auto" may be used in the other context in the future, I think it is better to revert back.

bovirus: @rustdesk 

To avoid this kind of problem with not English language  it's better to use specific strings and not use same string for more than one context.

Because in Italian/French/German (etc) a strings can be male/female/neutral.
Then to have a unique string can create errors in not English language

Ex. the English string "Auto" (Automatic) in italian could be "Automatico" (male) or "Automatica" (female) depend by context.

To avoid this issue in language file for this context use the string "Automatic codec" (it's specific).

Thanks.

rustdesk: You are right, But for this case, codec is already in the context, no need to tell codec explicitly verbosely.

PR #4047, opened 2023-04-13
add deps for wayland

fufesou:
Not been adequately tested.
But no errors will be introduced.

I'll test later.

- [x] ArchLinux. https://github.com/rustdesk/rustdesk/discussions/4045#discussioncomment-5604638
- [ ] Fedora/CentOS (x86-64)
- [x] OpenSUSE (x86-64)
- [x] Ubuntu/Debian (x86-64)


fufesou:
`libgstreamer-plugins-base1.0-0` is already there.

https://github.com/rustdesk/rustdesk/blob/66f880cc805767ce49a23b40aebbf9fa0ae9c5f3/build.py#L267

PR #4050, opened 2023-04-14
Fix/archlinux deps

fufesou:
Fix archlinux PKGBUILD depends. 

Add depends in the workflow for now.

Depends in runtime should better be put into `package()`.
https://wiki.archlinux.org/title/PKGBUILD#:~:text=Dependencies%20defined%20inside,run%20the%20software.

However permission error occurs here on github workflow.
https://github.com/fufesou/rustdesk/actions/runs/4695949715/jobs/8325765588

I'm trying to add `sudo`.

https://github.com/fufesou/rustdesk/actions/runs/4696121406

I'll submit another PR if it works.

### DIFF
--- a/.github/workflows/flutter-build.yml
+++ b/.github/workflows/flutter-build.yml
@@ -1447,6 +1447,9 @@ jobs:
             python
             ttf-arphic-uming
             libappindicator-gtk3
+            pam
+            gst-plugins-base
+            gst-plugin-pipewire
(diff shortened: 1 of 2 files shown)


fufesou:
`sudo pacman -S --noconfirm package` build success.
But the packaged file will not install the required deps on install.

The package built through this PR will install the required deps on install.

PR #4063, opened 2023-04-15
flutter-build-sciter-64bit

deep-soft:
flutter-build.yml
build-for-windows-sciter:
add:
  target: x86_64-pc-windows-msvc
  arch: x64

For those who prefer 64bit version of sciter.


rustdesk:
No, because we are phasing out Sciter.

PR #4075, opened 2023-04-16
Add new translation string (System language)

bovirus:
@rustdesk 

Please check and merge. Thanks.

### DIFF
--- a/flutter/lib/desktop/pages/desktop_setting_page.dart
+++ b/flutter/lib/desktop/pages/desktop_setting_page.dart
@@ -415,7 +415,7 @@ class _GeneralState extends State<_General> {
       List<String> keys = langsMap.keys.toList();
       List<String> values = langsMap.values.toList();
       keys.insert(0, '');
-      values.insert(0, translate('Default'));
+      values.insert(0, translate('System Language'));
       String currentKey = data['lang']!;
       if (!keys.contains(currentKey)) {
         currentKey = '';
--- a/flutter/lib/mobile/pages/settings_page.dart
+++ b/flutter/lib/mobile/pages/settings_page.dart
@@ -504,7 +504,7 @@ void showLanguageSettings(OverlayDialogManager dialogManager) async {
       return CustomAlertDialog(
         content: Column(
           children: [
-                getRadio(Text(translate('Default')), '', lang, setLang),
+                getRadio(Text(translate('System Language')), '', lang, setLang),
                 Divider(color: MyTheme.border),
               ] +
               langs.map((e) {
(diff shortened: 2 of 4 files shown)


Mr-Update:
@bovirus Please remove the line 496 taken from de.rs in template.rs again.

I added this line yesterday 5 minutes after creating PR 4066 for a test in my fork only. That was probably too fast.
As a result, the line was subsequently added to PR 4066, which was not intentional on my part.

rustdesk:
Frankly, I do not agree with "default" to "System Language".

bovirus:
@rustdesk 

Please be collaborative.
Don't just close n issues without understqnd/solve the problem.

I use my spare time to enhnance your software translation and just to close without understand the problem is not polite.

As I explained (please remember what I explained) if you want to have a multilanguage interface you should understand that you cannot use generic terms without a main object in the string.

Because otherwise you cannot do a right translation in italian (but also in French and German).

String as 

Default
Other
Accept both

that has not a main object cannot be translated right in italian (can be male or female) and if you uwe in more context easily you make an error about male/female regarding context.

If you don't like "System language" (for me it's more clear then Default....) for "Default" please use "Default language".
The main object is "language" and you can translate in right male/female version. 

The same for

Other -> Other option
Accept both -> Accept both password.

Thanks.

rustdesk:
Yes, it is not polite, but it is efficient, maintaining a project is not to make all people happly, our time is not used to explain one thing again and again to every new comming. I still insist my original opinion.

> The users are clever enough to understand the default in the context.

bovirus:
@rustdesk

For me your consideration about "efficient" could be read in other way as "I do as less as I can".

I'm quite surprise that consider my professional request about male/female strings for not English language an "happy" consideration.

It means that you want to ignore that your application is not only English but is multl language.

English strings are not male/female.
Other language strings can be male/female.

Use "Default langauge" that it's specific it's better also if you have only English.

You have a strange and quite caotic approach about translation.

As the tips that is not include as complete strings (but only variable name) in template.rs and are instead available in en.rs.

Then to create/update a language the translator have to check two files (template.rs and en.rs).
What's the reason of this caotic and complicated approach.

Simple: probably less job to do.
For me is not professional and is not good for translator and for future improvements.

Please switch your mind to do a clear multi language approach and simple and clear situation about translation.

rename template.rs as _template,rs to see in top of list (and not in middle of other languages).
merge the strings of en.rs in template.rs and change in source code xxx_xxx_tip with full text strings.

Then the _template.rs will be a real template useful for translator to check/update language.

Thanks.

PR #4106, opened 2023-04-19
Feat/plugin framework

fufesou:
1. Suppress warns.
2. Change `GLOBAL_EVENT_STREAM` to private.
3. Add feature `plugin_framework`.

### DIFF
--- a/src/plugin/plugins.rs
+++ b/src/plugin/plugins.rs
@@ -1,19 +1,13 @@
 use std::{
     collections::HashMap,
-    ffi::{c_char, CStr},
-    path::PathBuf,
+    ffi::c_char,
+    path::Path,
     sync::{Arc, RwLock},
 };
 
 use super::{callback_msg, desc::Desc};
-use hbb_common::{
-    anyhow::Error,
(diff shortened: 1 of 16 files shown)


rustdesk:
@Kingtous if any input?

src/plugin/plugins.rs
```diff
@@ -0,0 +1,188 @@
...
+use crate::flutter;
+use hbb_common::{bail, dlopen::symbor::Library, lazy_static, libc, log, ResultType};
+
+lazy_static::lazy_static! {
+    pub static ref PLUGINS: Arc<RwLock<HashMap<String, Plugin>>> = Default::default();
+}
+
+/// Initialize the plugins.
+/// Return 0 if success.
+pub type PluginFuncInit = fn() -> i32;
```
Kingtous, line 18: If we need to init with some parameters, we could add a `void` pointer here.

deep-soft:
Need fix **bridge.yml** before build:
https://github.com/rustdesk/rustdesk/discussions/4099

PR #4124, opened 2023-04-20
Skip anything after an # symbol in the filename.

madpilot78:
Windows renames files adding (1), (2) etc. before the .exe in case of duplicates, which causes the host or key values to be garbled.

This allows to protect from this adding a final "#" symbol to the filename.

### DIFF
--- a/src/license.rs
+++ b/src/license.rs
@@ -37,6 +37,11 @@ pub fn get_license_from_string(s: &str) -> ResultType<License> {
         s
     };
     if s.contains("host=") {
+        let s = if s.contains("#") {
+            &s[0..s.find("#").unwrap_or(s.len())]
+        } else {
+            s
+        };
         let strs: Vec<&str> = s.split("host=").collect();
         if strs.len() == 2 {
             let strs2: Vec<&str> = strs[1].split(",key=").collect();


rustdesk:
#4162

madpilot78:
I'm not sure I understand why this was reverted. #4162 is exactly the issue I was trying to fix. Is there some other reason why this cannot be included that I fail to see?

Any improvement I can make to this patch to make it acceptable?

rustdesk:
Please communicate with the author of the issue.

rustdesk:
Recover this please then. Sorry for my misunderstanding.

rustdesk:
Not sure if the key will contain #. Here does not check #. https://github.com/rustdesk/rustdesk-server/blob/1f7d3fa05ce6c8b3198a29a4a5a8095419608fe3/src/common.rs#L132
You may need to submit a PR to rustdesk_server.

madpilot78:
I don;t think a valid key can contain a `#`. Anyway there are various symbols a key can contain that windows will not allow in a filename, and in that case the documentation says to just generate a new key. The same can be done for the `#` symbol.

The server does not need to know anything about this.

PR #4129, opened 2023-04-20
ios build

sj6219:
I'm posting it just in case you need it. 
I tested it on my iPad.
To build, you must change the code sign setting.
You can test the release build in xcode.

### DIFF
--- a/flutter/ios/Runner/Info.plist
+++ b/flutter/ios/Runner/Info.plist
@@ -2,8 +2,12 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>CADisableMinimumFrameDurationOnPhone</key>
+	<true/>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>$(DEVELOPMENT_LANGUAGE)</string>
+	<key>CFBundleDisplayName</key>
+	<string>Flutter Hbb</string>
 	<key>CFBundleExecutable</key>
 	<string>$(EXECUTABLE_NAME)</string>
 	<key>CFBundleIdentifier</key>
@@ -21,6 +25,8 @@
 	<key>CFBundleVersion</key>
 	<string>$(FLUTTER_BUILD_NUMBER)</string>
 	<key>LSRequiresIPhoneOS</key>
+	<true/>
+	<key>UIApplicationSupportsIndirectInputEvents</key>
 	<true/>
 	<key>UILaunchStoryboardName</key>
 	<string>LaunchScreen</string>
@@ -41,17 +47,5 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
-	<key>ITSAppUsesNonExemptEncryption</key>
-  <false/>
-  <key>io.flutter.embedded_views_preview</key>
-  <true/>
-  <key>NSCameraUsageDescription</key>
-  <string>This app needs camera access to scan QR codes</string>
-  <key>NSPhotoLibraryUsageDescription</key>
-  <string>This app needs photo library access to get QR codes from image</string>
-	<key>CADisableMinimumFrameDurationOnPhone</key>
-	<true/>
-	<key>UIApplicationSupportsIndirectInputEvents</key>
-	<true/>
 </dict>
 </plist>
--- a/flutter/ios/Runner/Info.plist
+++ b/flutter/ios/Runner/Info.plist
@@ -47,5 +47,17 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
+	<key>ITSAppUsesNonExemptEncryption</key>
+  <false/>
+  <key>io.flutter.embedded_views_preview</key>
+  <true/>
+  <key>NSCameraUsageDescription</key>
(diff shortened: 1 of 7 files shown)


rustdesk:
@Kingtous

rustdesk:
which parts are for fixing ios building? i guess `get_rgba` is one of them?

sj6219:
Previously, code signing was not possible during the build.
So erase the ./flutter/ios directory, and have recreated it with the following command.
flutter create --platforms=ios .

I struggled with a bug where the contents of liblibrustdesk.a were not added normally.
It would be helpful to refer to the xcode compilation option.

rustdesk:
except [flutter/ios/Runner/Info.plist](https://github.com/rustdesk/rustdesk/pull/4129/files#diff-db38101755fbfd274c6413f9b8bf7affe3b45a3c580c93661e45e1666ac9e15c), the other changes look fine.

sj6219:
info.plist was created in the process of recreating ./flutter/ios. I don't know the exact purpose of this file.

rustdesk:
Please revert the change of info.plist, I will merge the other files.

rustdesk:
You tested the debug version on iPad or the release version?

sj6219:
> Please revert the changes of `info.plist`, I will merge the other files.


The original contents were merged into info.plist.

sj6219:
> You tested the debug version on iPad or the release version?

Both versions were tested.
The debug version can only be executed in a debugging environment.

sj6219:
Apps created with the command "flutter buildios --release" will not be installed on the physical machine.
Must be built from xcode.

flutter/ios/Runner/Info.plist
```diff
@@ -2,8 +2,12 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>CADisableMinimumFrameDurationOnPhone</key>
+	<true/>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>$(DEVELOPMENT_LANGUAGE)</string>
+	<key>CFBundleDisplayName</key>
+	<string>Flutter Hbb</string>
```
rustdesk: ?

rustdesk: https://github.com/rustdesk/rustdesk/blob/d9cc3433d4e4e67449c4c9388336f0ebb6cd4bbe/flutter/ios/Runner/Info.plist#L18

sj6219: When recreating ./flutter/ios, it is automatically created. I don't know why it was added.

rustdesk: I removed it

rustdesk:
@sj6219 does it work well on your iPad?

sj6219:
> @sj6219 does it work well on your iPad?

yes.

rustdesk:
👍

sj6219:
@rustdesk 
There is a bug that the iOS version program crashes frequently.
In particular, if you watch a video, it crashes in 3 seconds.
If the screen doesn't change, it won't crash.
Debugging in vscode produces the following message.

Stop reason: EXC_RESOURCE (RESOURCE_TYPE_MEMORY: high watermark memory limit exceeded) (limit=2867 MB)

In my opinion, whenever the screen changes, it would be nice to call the garage collector of the flutter.
or return cached memory.

rustdesk:
@Kingtous I remember we have a PR fixing this for slow PC before, here seems different.

Kingtous:
> 

Any logs or some etc. btw, I added a x64-ios lib build which may help debug this problem using iOS simulator.
https://github.com/Kingtous/rustdesk_thirdparty_lib/commit/17f1f6bd245023236866726d51586d5cceedee58

Kingtous:
> @rustdesk There is a bug that the iOS version program crashes frequently. In particular, if you watch a video, it crashes in 3 seconds. If the screen doesn't change, it won't crash. Debugging in vscode produces the following message.
> 
> Stop reason: EXC_RESOURCE (RESOURCE_TYPE_MEMORY: high watermark memory limit exceeded) (limit=2867 MB)
> 
> In my opinion, whenever the screen changes, it would be nice to call the garbage collector of the flutter or return cached memory.

Which revision of RustDesk u used to trigger this bug? I have tried with revision eb96b0d4518f493cc6efaec7b3e30a4a1fcc9e24, it seems cannot reproduce.

https://user-images.githubusercontent.com/39793325/233903690-a3b7725a-9e69-448f-af48-50bd9251128c.mp4

sj6219:
@Kingtous 
Just tested with revision eb96b0d. The same error still occurs. Maybe it's because it's a simulator, not an actual device.

By the way, how do you show the program memory in the left part of the video?

Kingtous:
> 

Open your XCode, and press "Start" to run it directly. And the debugger is here:

image

sj6219:
https://user-images.githubusercontent.com/3592759/233911484-1c8c582a-33b0-4172-8f2f-77f1fbf870c1.mov

If you watch the video, you can see that the memory increases and decreases at first, and then the memory increases and crashes at the end.


At first, I didn't play the video on the connected screen, but at the end, I played, and then the program crashed.

sj6219:
The iPad has a bigger screen, so it seems to use more memory.
Use the iPad simulator to see how much memory you can increase up to.

sj6219:
I tried to correct the error,
in the flutter ios environment
It seems that there is no function to call the garbage collector.
I should give up.

Issue #4136, opened 2023-04-20
Connection refused after rustdesk password was entered | Linux <-> Linux


awalon:
### Bug Description

"Connection refused" running recent nightly version (both sides):

Log from remote side (linux as qemu guest):
`[2023-04-21 00:05:04.347353 +02:00] INFO [src/platform/linux.rs:211] DISPLAY: :0
[2023-04-21 00:05:04.347457 +02:00] INFO [src/platform/linux.rs:212] XAUTHORITY: /var/lib/sddm/.Xauthority`

File is missing: /var/lib/sddm/.Xauthority

### How to Reproduce

1. start connection to linux remote (running X11, plasma with sddm) from another linux instance
2. enter correct permanent password
3. Error: connection refused

### Expected Behavior

Connection is working after successful login.

### Operating system(s) on local side and remote side

Linux -> Linux

### RustDesk Version(s) on local side and remote side

1.2.0

### Screenshots

-

### Additional Context

Found:
`/var/run/sddm/{8ed08636-0dad-4fd7-9740-ea2d04831a3e}`
used by sddm process running on remote side by checking all processes:
`sudo cat /proc/*/environ|tr "\0" "\n" | grep -a "^XAUTHORITY=" | sed 's/XAUTHORITY=//g'`

Workaround
`sudo ln -s "/var/run/sddm/{8ed08636-0dad-4fd7-9740-ea2d04831a3e}" "/var/lib/sddm/.Xauthority"`
Restarted rustdesk service on remote side but not sure if needed.
As result: Connection to sddm Login screen works as expected

awalon:
In addtion to initial bug report "connection refused" occures after remote login (sddm)  as well.
```
sudo loginctl list-sessions:

SESSION       UID USER               SEAT  TTY
   2897 384700801 user@domain seat0
   4195 384700801 user@domain       pts/0
   4253 384700801 user@domain seat0
```
Rustdesk log:
```
[2023-04-21 00:23:56.810089 +02:00] INFO [src/platform/linux.rs:211] DISPLAY: :0
[2023-04-21 00:23:56.810198 +02:00] INFO [src/platform/linux.rs:212] XAUTHORITY: /var/lib/user@domain/.Xauthority
```
File not found as well

With:
`sudo ls -1 /proc/*/environ|xargs -I___ sudo grep -a -H -o -P "(?<=XAUTHORITY=)(.+?)(?=\x0)" "___"
`
[...]

Issue #440, opened 2022-04-03
X11 error: connection refused after login on gnome (arch/manjaro gnome)


NicKoehler:
## Describe the bug you encountered:
I enabled the service on my personal computer at home.
I usually turn on my computer using Wake on wan and connect from remote.
I received the following error on my client after I entered my password from the login screen:
`X11 error: connection refused`

## What did you expect to happen instead?
Login as normal.

##  How did you install `RustDesk`?
I installed it from the [AUR](https://aur.archlinux.org/packages/rustdesk-bin) repo.

---

## RustDesk version and environment:

Rustdesk 1.1.8

OS: Archlinux
DE: Gnome on XORG
Kernel: 5.17.1

---

## Workaround:

I can successfully connect if I restart the rustdesk service via SSH after I logged in with:
```bash
systemctl restart rustdesk
```

ceewanna:
My recent Manjaro/Gnome couldn't get rustdesk running even login with Gnome Xorg. I tried restarting rustdesk after login but it still couldn't run. However, if I activate rustdesk from command line, the login screen showed up but with warning that I couldn't login as it is on wayland (even I was actually on x11, checked from echo $XDG_SESSION_TYPE )

The above was observed from rustdesk 1.2.0 and 1.1.9.

On the same machine above, I had anydesk on. Anydesk could work properly, connecting to another Manjaro/XFCE. However, connection to itself from another machine isn't possible and was reported on screen that the remote is on wayland.

rustdesk:
root?

ceewanna:
You mean the user in the context above? No, just typical user which was created along linux setup.

I have been using rustdesk on Manjaro/XFCE for quite sometime without problem but recently there is a need to set up a certain package which requires wayland. Then I thought Gnome would be fine as it caters both x11 and wayland.

rustdesk:
Now it is x11 or wayland?

ceewanna:
Manjaro/Gnome has both Xorg (x11) and Wayland. I am talking about rustdesk on Manjaro/Gnome on Xorg that is not working.

ceewanna:
1) Activating rustdesk from icon not working.
2) Activating rustdesk from command line has this screen. 

![anydesk00001-1](https://user-images.githubusercontent.com/84608596/222043241-8a1626a8-7330-4250-a63b-18b8817708bb.png)

rustdesk:
> 1. Activating rustdesk from icon not working.
[...]

PR #4165, opened 2023-04-23
Fix for #4136 connection to linux refused by using wrong Xauthority file

awalon:
Solution:
1. Get file from sddm-greeter and startplasma-x11 (aka KDE) processes as well
2. Get users home dir from system instead of hard coded pattern /home/<user>
3. Prefer XAUTHORITY environment variable over Xorg -auth if available

### DIFF
--- a/libs/hbb_common/src/config.rs
+++ b/libs/hbb_common/src/config.rs
@@ -329,7 +329,6 @@ fn patch(path: PathBuf) -> PathBuf {
                                 .trim()
                                 .to_owned();
                             if !home_dir.is_empty() {
-                                log::info!("config::patch: got home dir from: {}", home_dir);
                                 return home_dir.into();
                             }
                         }
--- a/libs/hbb_common/src/config.rs
+++ b/libs/hbb_common/src/config.rs
@@ -328,13 +328,12 @@ fn patch(path: PathBuf) -> PathBuf {
                                 .to_string()
                                 .trim()
                                 .to_owned();
-                            if home_dir.is_empty() {
-                                return format!("/home/{user}").into();
-                            } else {
+                            if !home_dir.is_empty() {
                                 log::info!("config::patch: got home dir from: {}", home_dir);
                                 return home_dir.into();
                             }
                         }
+                        return format!("/home/{user}").into();
                     }
                 }
             }
--- a/libs/hbb_common/src/config.rs
+++ b/libs/hbb_common/src/config.rs
@@ -322,7 +322,19 @@ fn patch(path: PathBuf) -> PathBuf {
                         .trim()
                         .to_owned();
                     if user != "root" {
-                        return format!("/home/{user}").into();
+                        let cmd = format!("getent passwd '{}' | awk -F':' '{{print $6}}'", user);
+                        if let Ok(output) = std::process::Command::new(cmd).output() {
+                            let home_dir = String::from_utf8_lossy(&output.stdout)
+                                .to_string()
+                                .trim()
+                                .to_owned();
+                            if home_dir.is_empty() {
+                                return format!("/home/{user}").into();
+                            } else {
+                                log::info!("config::patch: got home dir from: {}", home_dir);
+                                return home_dir.into();
+                            }
+                        }
                     }
                 }
             }
--- a/src/platform/linux.rs
+++ b/src/platform/linux.rs
@@ -9,6 +9,7 @@ use hbb_common::{
     regex::{Captures, Regex},
 };
 use std::{
+    string::String,
     cell::RefCell,
     path::{Path, PathBuf},
     process::{Child, Command},
@@ -18,6 +19,7 @@ use std::{
     },
     time::{Duration, Instant},
 };
+use users::{get_user_by_name, os::unix::UserExt};
 
 type Xdo = *const c_void;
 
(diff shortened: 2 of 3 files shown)


libs/hbb_common/src/config.rs
```diff
@@ -322,7 +322,19 @@ fn patch(path: PathBuf) -> PathBuf {
                         .trim()
                         .to_owned();
                     if user != "root" {
-                        return format!("/home/{user}").into();
+                        let cmd = format!("getent passwd '{}' | awk -F':' '{{print $6}}'", user);
+                        if let Ok(output) = std::process::Command::new(cmd).output() {
```
rustdesk, line 326: How about if here is not OK?

src/platform/linux.rs
```diff
@@ -449,12 +451,37 @@ pub fn get_active_username() -> String {
     get_values_of_seat0(&[2])[0].clone()
 }
 
+pub fn get_user_home_by_name(username: &str) -> Option<PathBuf> {
+    return match get_user_by_name(username) {
+        None => {
+            log::debug!("no uid found for user: {}", username);
+            None }
```
rustdesk: format

src/platform/linux.rs
```diff
@@ -824,56 +862,107 @@ mod desktop {
...
-                        } else if auth_found {
-                            if std::path::Path::new(v).is_absolute() {
-                                self.xauth = v.to_string();
-                            } else {
-                                if let Some(pid) = line.split_whitespace().nth(1) {
-                                    let home_dir = get_env_from_pid("HOME", pid);
-                                    if home_dir.is_empty() {
-                                        self.xauth = format!("/home/{}/{}", self.username, v);
+
+                    if let Some(pid) = line.split_whitespace().nth(1) {
```
rustdesk: Why add one more `if` here?

src/platform/linux.rs
```diff
@@ -824,56 +862,107 @@ mod desktop {
...
+                                        } else {
+                                            base_dir = home_dir;
+                                        }
+                                        if Path::new(&base_dir).exists() {
+                                            self.xauth = format!("{}/{}", base_dir, v);
+                                            log::debug!("+ auth file with home path added: {}", self.xauth);
+                                        };
                                     } else {
-                                        self.xauth = format!("{}/{}", home_dir, v);
+                                        // unreachable!
```
rustdesk: Can you explain what changes here line870 -> line 895?

libs/hbb_common/src/config.rs
awalon: Good point missed that case, will be fixed by followup PR

src/platform/linux.rs
awalon: Missing line breake (after 2nd None) by updated PR as well

src/platform/linux.rs
awalon: Thanks for review, that's a left over for Xauth file fallback. Cann t test this here so this step will be  removed again with next PR.

src/platform/linux.rs
awalon: Highlevel: 
As mentioned by newly added debug messages auth- file will be extraxcted from Xorg process and for a relative path home path will be added.

If no HOME var was found the home folder for given user will be determined from system config (new function: get_user_home_by_name)  old logic just used hard coded  default pattern "/home/<username>" (now only if no system config was found)  which wasn't working for domain users (AD, samba, freeipa, ...) where an additional domain subfolder (/home/<domain>/<username>) needs to be added.

In addition detected file needs to exist, if not this loop won't detect any auth- file. So next step can detect a valid file.

Next step (Todo): Detect auth file by checking process tree (processes below Xorg) without need of predefined checks for Gname, XFCE, Plasma, ...). But as this was my first modifiaction using rust language, there was too less time for better solution.

libs/hbb_common/src/config.rs
```diff
@@ -322,6 +322,17 @@ fn patch(path: PathBuf) -> PathBuf {
...
                         .to_owned();
                     if user != "root" {
+                        let cmd = format!("getent passwd '{}' | awk -F':' '{{print $6}}'", user);
+                        if let Ok(output) = std::process::Command::new(cmd).output() {
+                            let home_dir = String::from_utf8_lossy(&output.stdout)
+                                .to_string()
+                                .trim()
+                                .to_owned();
+                            if !home_dir.is_empty() {
+                                log::info!("config::patch: got home dir from: {}", home_dir);
```
rustdesk: log::debug or remove please.

src/platform/linux.rs
```diff
@@ -824,22 +863,36 @@ mod desktop {
                 "ps -u {} -f | grep 'Xorg' | grep -v 'grep'",
                 &self.uid
             )) {
+                log::debug!("extracting xauth file for Xorg running with uid: {}", self.uid);
```
rustdesk: Too many log::debug! which it will be print to file, We'd better remove these log::debug before code merge.

rustdesk (changes requested):

src/platform/linux.rs
awalon: Debug messages was removed now

libs/hbb_common/src/config.rs
awalon: log message removed

rustdesk:
#4136

rustdesk:
I guess https://github.com/rustdesk/rustdesk/issues/440 has the same, no idea what process it is.

rustdesk:
#440

PR #4193, opened 2023-04-25
Feat/plugin framework

fufesou:
Change the function name `log` to `plugin_log`.

The `log` function name causes strange bug whey trying to trigger plugin functions.

The `log` function is called by `amp2Log2` 

image

image

### DIFF
--- a/src/plugin/callback_msg.rs
+++ b/src/plugin/callback_msg.rs
@@ -61,7 +61,7 @@ struct MsgToConfig {
 /// content: The content.
 /// len:     The length of the content.
 #[no_mangle]
-pub extern "C" fn cb_msg(
+pub(super) extern "C" fn cb_msg(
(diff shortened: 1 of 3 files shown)


fufesou:
Symbols are indeed shaded, here is an example.

[test_shade_log.tar.gz](https://github.com/rustdesk/rustdesk/files/11328454/test_shade_log.tar.gz)




## test normal `log` function
```
cd libs/mylib && cargo build --lib && cd - && cargo run
```

output:
```
main: test normal

static lib: foo log(3.0) = 1.10
```

##  test shading
```
cd libs/mylib && cargo build --lib && cd - && cargo run --features shade_log
```

output:
```
main: test shade function

main: log shades default log, with random params: _x: 140695338082616, _y: 0
static lib: foo log(3.0) = 0.00
```

PR #4215, opened 2023-04-27
Reimplement code extracting information from file name.

### LINKED ISSUE #4162: RustDesk for Windows does not read the correct key from the filename when Windows appends  version (#)
### Bug Description

When I set the -k _  optinon on the self hosted server the public key is checked. When on Windows client B the client application is downloaded for a second time, Windows appends (1) to the filename. The Rustdesk Windows client reads the key including the (1) from the filename. When i try to connect from Client A (with the correct key) to client B the connection between the two clients is not established because the keys are different.

### How to Reproduce

First of all on the local server set the -k _ option.
Start Rustdesk with the correct key on client A. Download the Rustdesk application on client B several times and run it from the last downloaded file (with the (#) appended).

### Expected Behavior

I think the RustDesk application should read the key from the filename without the appended (#) so the connection would be established. An extra delimiter in the filename could resolve the issue because then the Rustesk for Windows application knows where the key ends.

Both Windows 11

Both 1.2.0

### Screenshots

n/a

### Additional Context

n/a

madpilot78:
This code tokenizes the file name based on commas and extracts relevant parts sequentially.

`host=` is expected to be the first part.

Since Windows renames files adding (1), (2) etc. before the .exe in case of duplicates, which causes the host or key values to be garbled.

This allows using a `,` (comma) symbol as a final delimiter.

Fixes #4162

### DIFF
--- a/src/license.rs
+++ b/src/license.rs
@@ -82,3 +82,60 @@ pub fn get_license_from_string(s: &str) -> ResultType<License> {
     }
     bail!("Failed to parse");
 }
+
+#[cfg(test)]
+#[cfg(target_os = "windows")]
+mod test {
+    #[test]
+    fn test_filename_license_string() {
+        assert_eq!(
+            get_license_from_string("rustdesk.exe"),
+            Ok(License {
+                host: "".to_owned(),
+                key: "".to_owned(),
+                api: "".to_owned(),
+            })
+        );
+        assert_eq!(
+            get_license_from_string("rustdesk"),
+            Ok(License {
+                host: "".to_owned(),
+                key: "".to_owned(),
+                api: "".to_owned(),
+            })
+        );
+        assert_eq!(
+            get_license_from_string("rustdesk-host=server.example.net.exe"),
+            Ok(License {
+                host: "server.example.net".to_owned(),
+                key: "".to_owned(),
+                api: "".to_owned(),
+            })
+        );
+        assert_eq!(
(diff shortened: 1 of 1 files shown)


rustdesk:
add tesy please

rustdesk:
and you give evidence to convince me that key does not contain this character. i do not accept "i think".

madpilot78:
> and you need to give evidence to convince that `key` does not contain this character. Please not just "i think" or "it should".

@rustdesk This is easy. The key is base64 encoded, `,` and `#` are not part of the base64 alphabet, so they cannot appear in the encoded key.

> add test code please

I completely agree with this request but will require a little time, since I have no experience with rust testing.

rustdesk:
https://github.com/rustdesk/rustdesk/blob/2b8a98da0680cf94aae46154091247f856a42e87/src/ipc.rs#L875

madpilot78:
@rustdesk I added tests for the filename data extraction code.

I hope I have done it correctly, CI looks happy.

If any other change is required please give me directions, I'll be happy to comply!

rustdesk:
> This code tokenizes the file name based on commas and extracts relevant parts sequentially.
> 
> `host=` is expected to be the first part.
> 
> Since Windows renames files adding (1), (2) etc. before the .exe in case of duplicates, which causes the host or key values to be garbled.
> 
> This allows using a `,` (comma) symbol as a final delimiter.

Add this comment to code please.

PR #4231, opened 2023-04-29
fix file transfer always connecting

21pages:
#4230
it is because FingerPrintState not initialized on file manager page, show fingerprint on remote tab only first


rustdesk:
How about non-secure connection? will work?

21pages:
non-secure connection works, it will show empty fingerprint

rustdesk:
👍

Issue #4240, opened 2023-04-30
Screensaver is not activated when a Rustdesk session is open.


MagnumGR:
### Bug Description

When a rustdesk session is open, even is minimized or not, and there is inactivity on the local PC the screensaver of Windows 10 has not activated in appropriated time.
When i close the rustdesk session, the screensaver activated at the specific time.

### How to Reproduce

1. Set screensaver at 5' 
2. Open rustdesk session and minimize.
3. Inactivity of more 5' on PC.

### Expected Behavior

When there is inactivity of the specific time example 5' (setting on screensaver in Windows 10 ) must activate the screensaver.

### Operating system(s) on local side and remote side

Windows 10

### RustDesk Version(s) on local side and remote side

1.20 Local 1.19 Remote

### Screenshots

None

rustdesk:
@21pages we do have this behavior,  

https://github.com/rustdesk/rustdesk/blob/c7a68283073ccc8429b8ade204102156d5b9bcc3/flutter/lib/desktop/pages/remote_page.dart#L110

please turn off when minimized and turn on again when non minimized.

rustdesk:
close as  fix merged

4C3T:
Can't make heads or tails of what's been done about this as all I see is bug reports being merged 3 weeks ago.
I'm affected by this as I at times forget to close the remote window and I lock the computer before i get to bed, and this is preventing the lock screen since it won't go into screensaver mode.

tonjansenwehl:
17:14 29-12-2024

2024 12 29 - Screensaver does not work when connected from remote pc

Hello,

I notice that my screensaver does not work when a remote pc is connected with rustdesk to my laptop.

Can you help me?

Regards,

tonjansenwehl@proton.me

PR #4247, opened 2023-05-01
fix last commit

21pages:
When the window is unminimized, onWindowMaximize  or onWindowRestore can be called when the old state was maximized or not.


rustdesk:
#4240

Issue #4069, opened 2023-04-15
macOS URL handler weird behaviour


SupersonicWaffle:
### Bug Description

When opening a URL like rustdesk://connection/new/xxxxxxxxx, it only works to open a new connection when RustDesk is closed.
If RustDesk is already running it will just activate the window.

### How to Reproduce

Use rustdesk:// URL when RustDesk is running

### Expected Behavior

Open a new connection as it does on windows

### Operating system(s) on local side and remote side

macOS -> any

### RustDesk Version(s) on local side and remote side

1.2.0 -> any

### Screenshots

none

rustdesk:
@kingtous

Kingtous:
Sorry for being late due to being quite busy these days. I've tested the latest macOS nightly build but I cannot reproduce your problem, I can open a new connection via `rustdesk:://` schema, no matter whether the RustDesk window is running or not.

Please make sure u are using the latest macOS nightly build from our repo, and install the .app into your `Application` folder.

If the problem still exists, please provide the log from your terminal by opening RustDesk with this command:

```
/Applications/RustDesk.app/Contents/MacOS/RustDesk
```

an example of the log after your invoke your schema:

```
flutter: uni links called:
flutter: handled by uni links: false
flutter: [MultiWindowHandler] active window changed: [0]
```

and a more question: have you installed the service by clicking the installation button and entering your password?

rustdesk:
I have confirmed, it works when service is both installed and not installed. 

@Kingtous but I see another issue, when service is installed and the main window (or connection windows) is not active, it can not activated by clicking on RustDesk icon in lauchpad, please take a look.

PR #4255, opened 2023-05-02
fix: window on top when url scheme received from ipc

Kingtous:
This PR moves the uri empty check into `on_url_scheme_received `, which is received from IPC.

Place the empty check directly in `parseRustdeskUri` will cause the following command fails:

- `rustdesk rustdesk://xxx`, which is required from cli and macos uni links fallback.

Tested on macOS Ventura 13.3.1.


rustdesk:
https://github.com/rustdesk/rustdesk/issues/4069

PR #4283, opened 2023-05-06
disable desktop_drop for now

fufesou:
it causes crash while closing client on fufesou's new installed win11.

and https://github.com/rustdesk/rustdesk/pull/4289 https://github.com/rustdesk/rustdesk/pull/4290


rustdesk:
@Kingtous FYI

PR #4301, opened 2023-05-08
feat: add entry in left panel and dialog callback

Kingtous:


src/flutter_ffi.rs
```diff
@@ -751,6 +751,25 @@ pub fn main_load_recent_peers() {
...
+            (
+                "peers",
+                serde_json::ser::to_string(&peers).unwrap_or("".to_owned()),
+            ),
+        ]);
+        return SyncReturn(serde_json::ser::to_string(&data).unwrap_or("".to_owned()));
+    }
+    SyncReturn("".to_string())
+}
+
```
rustdesk, line 772: If added this, main_load_recent_peers can be removed, no need two.

rustdesk: old
```
         let _res = flutter::push_global_event(
             flutter::APP_TYPE_MAIN,
             serde_json::ser::to_string(&data).unwrap_or("".to_owned()),
         );
```
 looks ugly.

Kingtous: current the recent peers, etc are loading with this global events by @fufesou . Maybe we need to change the `PeerCard` from events driven to direct-call driven @fufesou ?

fufesou: I think direct replacement is problematic.
Because there are some places where main_load_recent_peers is used as a trigger event to re-render the ui.

PR #4344, opened 2023-05-12
Feat/linux exec privileged

fufesou:
wrap pkexec


SergeyMy:
thanks for taking this on
.icp* sock will be overwritten if no rights
.pid is only meaningful for deamon and should not be overwritten by the user
Снимок экрана от 2023-05-12 21-49-53
maybe something like that? :)

Issue #4351, opened 2023-05-13
The latest nightly version is unable to display the picture when connected to Android.


zhangbo8418:
### Bug Description

The latest nightly version is unable to display the picture when connected to Android.

### How to Reproduce

Android -> Android
Windows -> Android
MacOS -> Android
None of them are able to display.

### Expected Behavior

Display normally.

### Operating system(s) on local side and remote side

Windows -> Android

### RustDesk Version(s) on local side and remote side

1.20

### Screenshots

Connect to the TV.
![6E{SBA(~0G @U)@` NWX0@S](https://github.com/rustdesk/rustdesk/assets/46797568/bde92d08-5d8b-4ac0-8c18-ef7785d6a654)

Connect to Xiao Ai Smart Speaker.

![@O36@TG @PDYYR9AJ@1SH$X](https://github.com/rustdesk/rustdesk/assets/46797568/4625f82e-7479-4661-a128-3e48fe89f798)

rustdesk:
@21pages should be related to yesterday jni change

rustdesk:
The new nightly build will be ready soon.

Issue #4118, opened 2023-04-20
[Crash] app often crashes (Android 6.0 and Android 7.1 , armv7)


areful:
### Bug Description

Often crash on Android (as client).

### How to Reproduce

When android client connected with remote server, it works well.

But when I disconnect from remote server manually by click "Disconnect" button, it's often crash. Not every time, but too often.

### Expected Behavior

Works well when click disconnect.

### Operating system(s) on local side and remote side

Windows

### RustDesk Version(s) on local side and remote side

1.20 -> 1.1.9

### Screenshots

![image](https://user-images.githubusercontent.com/2671987/233244896-a541e117-cc76-451b-a4d7-0efb16bb040b.png)
![image](https://user-images.githubusercontent.com/2671987/233245019-daae07c2-5040-46e7-88c8-83a2855b0473.png)

### Additional Context

part of crash log:
```
_2023-04-20 10:34:34.568 32675-814/com.carriez.flutter_hbb D/LOG_SERVICE: Start Capture
2023-04-20 10:34:34.569 32675-814/com.carriez.flutter_hbb D/LOG_SERVICE: ImageReader.newInstance:INFO:Info(width=540, height=960, scale=2, dpi=80)
2023-04-20 10:34:34.571 32675-814/com.carriez.flutter_hbb D/LOG_SERVICE: ImageReader.setOnImageAvailableListener done
2023-04-20 10:34:34.571 32675-814/com.carriez.flutter_hbb D/LOG_SERVICE: startRawVideoRecorder,screen info:Info(width=540, height=960, scale=2, dpi=80)
2023-04-20 10:34:34.575 633-1221/system_process I/DisplayManagerService: Display device added: DisplayDeviceInfo{"RustDeskVD": uniqueId="virtual:com.carriez.flutter_hbb,11150,RustDeskVD,0", 540 x 960, modeId 27, defaultModeId 27, supportedModes [{id=27, width=540, height=960, fps=60.0}], colorTransformId 0, defaultColorTransformId 0, supportedColorTransforms [], density 80, 80.0 x 80.0 dpi, appVsyncOff 0, presDeadline 16666666, touch NONE, rotation 0, type VIRTUAL, state ON, owner com.carriez.flutter_hbb (uid 11150), FLAG_PRIVATE, FLAG_PRESENTATION}
2023-04-20 10:34:34.578 32675-490/com.carriez.flutter_hbb I/flutter: flutter got android msg,on_state_changed,{name: media, value: true}
[...]

PR #4353, opened 2023-05-13
fix andriod jni update

21pages:
#4351

### DIFF
--- a/libs/scrap/src/android/ffi.rs
+++ b/libs/scrap/src/android/ffi.rs
@@ -49,10 +49,11 @@ impl FrameRaw {
         self.enable = value;
     }
 
-    fn update(&mut self, data: *mut u8) {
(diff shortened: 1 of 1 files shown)


libs/scrap/src/android/ffi.rs
```diff
@@ -49,10 +49,11 @@ impl FrameRaw {
         self.enable = value;
     }
 
-    fn update(&mut self, data: *mut u8) {
+    fn update(&mut self, data: *mut u8, len: usize) {
         if self.enable.not() {
             return;
         }
+        self.len = len;
```
rustdesk, line 56: Big threading problem here, len and data are not synced. And also no guarentee of the lifetime of data. I will note this in https://github.com/rustdesk/rustdesk/issues/4118

PR #4393, opened 2023-05-16
plugin_framework, add uuid on plugin signature verification

fufesou:


src/plugin/callback_msg.rs
```diff
@@ -80,6 +80,7 @@ pub(super) struct MsgToExtSupport {
 
 #[derive(Debug, Serialize)]
 struct PluginSignReq {
+    uuid: Vec<u8>,
     peer_id: String,
```
rustdesk: Remove peer_id then?

rustdesk (changes requested):

src/plugin/callback_msg.rs
fufesou: I thought about it, the `uuid` needs to be included in the msg to ensure that the uuid is valid and not filled in arbitrarily.

PR #4434, opened 2023-05-20
allow building with pkg-config libraries

selfisekai:
closes #4420


paulmenzel:
See merge/pull request https://github.com/rustdesk-org/magnum-opus/pull/6 as a follow-up.

Issue #4336, opened 2023-05-11
Left of chat is too distracting


bilogic:
### Bug Description

![image](https://github.com/rustdesk/rustdesk/assets/946010/f16cd577-07d5-491d-b8df-4fb3999c289e)

Left side of the chat windows is too attention grabbing, many times, users don't know it is a chat window.

### How to Reproduce

See image

### Expected Behavior

All the red boxes can be shrinked into 1 row and moved to the top of the chat.
Conversation part can become full width.

If need be, I can follow up with a mock wire frame

### Operating system(s) on local side and remote side

Windows 10 -> Windows 11

### RustDesk Version(s) on local side and remote side

1.1.9 -> 1.1.9

### Screenshots

Already provided

rustdesk:
mock up please

rustdesk:
@grummbeer

grummbeer:
I need to think more about it. If I follow the whole chat behavior from the controlling side (good, well implemented) to the controlled side (has issues), then there are different points that could be reconsidered. It also makes a difference if the controlled side wants to send a message(icon not obvious enough? toolip/label missing?) or if it receives a message (how to get clear notice of it?).

> Left side of the chat windows is too attention grabbing

Yes, there is a visual unbalance. The light mode has less contrast also. 

With 1.2 the messages has a colored background, what makes the messages a bit more obvious than in the screenshot.

Haveing a second column with the chatbox is not that bad for me. So the important actions are always reachable and you can follow the chat at the same time. Maybe its all about the presentation, a better visual balance and contrasting?

> many times, users don't know it is a chat window.

This is something other than distracted. 

- They do not know a chat is available at all? When they want to send a message
- They miss a incoming message due the visually unbalance? incoming messages needs a "attention seeker"?

> If need be, I can follow up with a mock wire frame

Sounds good. I think at least the "disconnect" button must be always reachable with one click and without searching for it.

michalwitek1232:
@rustdesk  I'm not a graphic designer but maybe something like this?

![image](https://github.com/rustdesk/rustdesk/assets/42500203/611637cb-f204-46f8-96f4-1118f16bc67e)

michalwitek1232:
Also dark version

![image](https://github.com/rustdesk/rustdesk/assets/42500203/1855d8c0-78a6-44fa-9e11-f2974e29d60a)

michalwitek1232:
I love the part about
```
[...]

PR #4438, opened 2023-05-20
Connection/Chat

NicKoehler:
As discussed in #4336 

Here's the changes:

- [x] Larger Chat
- [x] Grid layout
- [x] Fluent icons
- [x] Chat moved to left
- [x] Removed name from received messages
- [x] Other minor adjustments (Colors, Padding, Margins)

| Light | Dark |
| -- | -- |
| immagine | immagine |


rustdesk:
👍

flutter/lib/common.dart
```diff
@@ -196,6 +183,8 @@ class MyTheme {
     ),
   );
 
+  static const SwitchThemeData switchTheme = SwitchThemeData(splashRadius: 0);
```
grummbeer, line 186: No switch used. SwitchTheme can be removed?

grummbeer: Switches are used at android server/settings page, they might affected by this change. not sure.

rustdesk:
Can we add off/on status to tooltip?

grummbeer:
> how do you think about this?

The labeled switch/checkbox is the solution to exactly this problem. 

Even with e.g. red/green colored, it is a riddle whether green means "I am safe" or if permission was granted (possibly "I am not safe"). Same with grey/blue

rustdesk:
I still think it can be solved visually. For example, add some mark on the icon as below. @michalwitek1232 do you have any idea for such design?

image

grummbeer:
> I still think it can be solved visually. For example, add some mark on the icon as below 

:-) yeah good idea!

NicKoehler:
Do you like something like this?

immagine

rustdesk:
How about?

image

NicKoehler:
What do you think?

immagine

rustdesk:
> What do you think?

How about if removing ON?

rustdesk:
> How about if removing ON?

Ignorethis, both "ON/OFF" looks better.

michalwitek1232:
> I still think it can be solved visually. For example, add some mark on the icon as below (stack layout can be used to add these mark). @michalwitek1232 do you have any idea for such design?
> 
> image image



> What do you think?
> 
> immagine

I'm thinking about something like this maybe?

image

What this design tells user ?
- This could be enabled but right now is disabled

I mean, the problem with buttons as checkboxes/switches is that it's kind hard to signalize user that this could be flipped ON/OFF

rustdesk:
Thanks @michalwitek1232

rustdesk:
Since fluent icon pub increase the final exe file size a lot, I will remove fluent icon pub and use svg files.

michalwitek1232:
> Since fluent icon pub increases the final exe file size a lot (1.3 M), I will remove fluent icon pub and use svg files.
> 
> image

That makes a lot of sense. Good move 👍

PR #4476, opened 2023-05-24
change resolution, macos needs more tests

fufesou:
This PR may fix the fuzzy controlled end(macOS) when connection is established.

But it is still not sure if the way to set the resolution on macOS is correct.


rustdesk:
I think we should avoid calling change_resolution in connection.rs if there is no behavior of changing resolution.

Issue #4542, opened 2023-06-04
"Scale original" is blurry depending on scroll mode and rustdesk window size. [Nightly]


po0p:
### Bug Description

When "Scale original" is selected:
Selecting scrollbar gives clean image (at least for text)
Selecting ScrollAuto makes text a little bit blurry.
Autoscroll works without blur on AnyDesk, so I'm sure there is a way to fix it :)
making Rustdesk window more than remote resolution size makes image blurry, making is smaller makes image be okay.
Host desktop resolution: 3840x2160.
Remote screen resolution: 1920x1080.

### How to Reproduce

>Select "Scale Original"
>Alternate between different scroll modes and windows sizes
>Change rustdesk window size with scrollmode = scrollbar and "scale original", from very small window to window larger than remote screen resolution

### Expected Behavior

Expecting image to be clear and non blurry at all times 👍🏻 

### Operating system(s) on local side and remote side

Windows 10 -> Windows 10

### RustDesk Version(s) on local side and remote side

1.2.0 -> 1.2.0 (both nightly)

### Screenshots

1. Window size on host is ~1920x1080. So not **whole** remote window fits (as it is 1920x1080 resolution).
1.0 Scale original + **Scrollbar** -> https://imgur.com/a/SJwT6GH -> text is perfectly fine
1.1 Scale original + **ScrollAuto** -> https://imgur.com/a/bJwtwca (text becomes blurry)

2. Changing rustdesk window size on host (from very small to very big), using **Scrollbar** and **Scale Original**
2.0 Very small window -> text is fine https://imgur.com/a/b34Jv1C
2.1 Window matching remote screen size + a little bit more space at top (basically i resized rustdesk window to 1080p, then made it taller) -> text is fine https://imgur.com/a/9iK1HTk
2.2 Rustdesk window ~1-10px wider than host resolution (any amount of making it wider than remote resolution produces this for me) -> text is blurry https://imgur.com/a/6g9fZTp

po0p:
Edit: fitting whole remote screen into rustdesk window (using "Adjust Window" under screen settings) make text and image clear and non blurry.

rustdesk:
@fufesou

rustdesk:
I just tested, both scrollbar/auto can have blurry text, even it is clear, it can also become blurry after moving mouse.

rustdesk:
[...]

PR #4543, opened 2023-06-04
fix default peer options

fufesou:
https://github.com/rustdesk/rustdesk/issues/4542#issuecomment-1575499792


rustdesk:
fixed https://github.com/fufesou/rustdesk/commit/8c0fa989b65ac21e78ff6b5406cabb1d497848e1

Issue #4359, opened 2023-05-13
Virtual display driver resets to 1024x768 when remote machine is restarted


FCrane:
### Bug Description

Hi!

When I use the virtual display driver provided by Rustdesk (and no display is attached to the remote machine), there a a few issues:

1) The resolution can not be changed in the "Display Settings" of Windows 11 directly. I can select any resolution from the drop list, Windows asks me if I want to keep the change, but no change has occurred. It always stays at 1024x768.

2) I can successfully change the resolution in "Advanced Display" -> "Display adapter properties for Display 1". However, when the remote OS is restarted, it always reverts back to 1024x768.

Regards!

### How to Reproduce

See description

### Expected Behavior

Resolution change possible from regular "Display Settings" and should be kept after reboot.

### Operating system(s) on local side and remote side

Windows 11 x64

### RustDesk Version(s) on local side and remote side

Latest nightly

### Screenshots

n/a

rustdesk:
@fufesou this is nessessary.

fufesou:
We can't detect the changes of the virtual display.

But we can provide the ability to change the resolution through Rustdesk. And remember the resolution.

rustdesk:
Yes, remember on controlling side.

FCrane:
But the remote OS itself has to remember (and display) the resolution correctly! Otherwise it would always revert to 1024x768 after a reboot (messing up icon positions, window positions of programs that are started with Windows and that are already open before RustDesk, etc.) and then change to whatever the user sets in RustDesk. This means 2 screen resolution changes (taking system resources, time for changing the screen, etc.) each reboot, which is not good...

fufesou:
@FCrane Hi, there a two features added:
1. The resolutions can be remembered.
2. The resolutions of the virtual display can be set to any value.

They may solve your problems.

Can you please test and give some feedbak, thanks.

rustdesk:
close as fix merged

FCrane:
Sorry, still not working! I changed the resolution of the VIRTUAL monitor in the Registry to 1920x1200 and indeed Windows 11 then uses this resolution when starting without a monitor attached. However, Windows itself is still unusable! Dialogs do not work, layouts are wrong etc. Completely unusable! Only attaching a monitor works. 

Sorry, but your "Virtual Display Driver" is unusable, At least on a freshly installed Windows 11, there is no way to get it running correctly without a physical monitor attached.

fufesou:
[...]

PR #4560, opened 2023-06-05
Refact/virtual display dummy display

fufesou:
https://github.com/rustdesk/rustdesk/issues/4359#issuecomment-1572621649

This PR is only partially tested, by setting `dummy_display_side_max_size` to 2000.

### DIFF
--- a/src/server/video_service.rs
+++ b/src/server/video_service.rs
@@ -340,16 +340,6 @@ fn create_capturer(
     };
 }
 
-// to-do: do not close if in privacy mode.
-#[cfg(all(windows, feature = "virtual_display_driver"))]
-fn ensure_close_virtual_device() -> ResultType<()> {
-    let num_displays = Display::all()?.len();
-    if num_displays > 1 {
-        let _res = virtual_display_manager::plug_out_headless();
-    }
-    Ok(())
-}
-
 // This function works on privacy mode. Windows only for now.
 pub fn test_create_capturer(privacy_mode_id: i32, timeout_millis: u64) -> bool {
     let test_begin = Instant::now();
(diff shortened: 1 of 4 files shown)


src/server/video_service.rs
```diff
@@ -993,21 +986,38 @@ fn try_get_displays() -> ResultType<Vec<Display>> {
...
+#[cfg(windows)]
+fn no_displays(displays: &Vec<Display>) -> bool {
+    let display_len = displays.len();
+    if display_len == 0 {
+        true
+    } else if display_len == 1 {
+        let display = &displays[0];
+        let dummy_display_side_max_size = 800;
+        display.width() < dummy_display_side_max_size
+            && display.height() < dummy_display_side_max_size
```
rustdesk: <=

rustdesk:
you can test this by adjust your display to low resolution.

rustdesk:
or test with a bigger dummy_display_side_max_size

fufesou:
I've tested setting `dummy_display_side_max_size` to 1200.

New virtual display is plug in if the resolution is `1024x768` or `800x600`.

fufesou:
There's a side effect that the resolution will be reset to original when plug in virtual display.

`dummy_display_side_max_size` is set to 1200.

I mannually set the remote resolution to `800x600`.
The resolution is set `2160x1440` automatically when the connection is established and the virtual display is plug in.
There's no record of resolution restore.

I'll try to find the reason.

fufesou:
It works fine.

The resolution of display 0 is set to `2160x1440` when the virtual display is plug in.

Because the setting of two displays is set by previous operations.
When new monitor is plug in, the setting is applied.

PR #4569, opened 2023-06-07
src/lang/en.rs: improve tip about Nouveau driver

xalt7x:
Nouveau driver is Linux specific.
Also it can't be installed because it's built into mainline Linux kernel.
Related to  [commit 5f6a264](https://github.com/rustdesk/rustdesk/commit/5f6a2642781b85b753d029db7cab547a3454bc9d) and conversation #2523


21pages:
https://github.com/flutter/flutter/issues/85460#issuecomment-968728034
Looks good, but still need a small fix. I think this is a GPU problem, not a nouveau problem, so the tooltip needs to point out that if the official Nvidia driver is installed, uninstall and enable the Nouveau driver.

21pages:
👍

PR #4605, opened 2023-06-09
macos cursor bugfix

sj6219:
After updating the mac os, there is a bug where the cursor is not visible. So I fixed the bug.


rustdesk:
👍

PR #3670, opened 2023-03-15
Multi display in menu bar (toolbar)

NicKoehler:
Implemented what discussed in #3333

| Enabled | Disabled |
| --- | --- |
| image | image |
| image | image |

PR #4614, opened 2023-06-11
option collapse toolbar

21pages:
#4424


rustdesk:
https://github.com/rustdesk/rustdesk/pull/3670#issuecomment-1470438228

PR #4641, opened 2023-06-14
WIP Trusted clients

grummbeer:
https://github.com/rustdesk/rustdesk/discussions/3988

So far …

- Need feedback and suggestions.
- Need help: How to open a dialog outside the window from which it was called? How do I get it to the center of the screen when I call the dialog from server_page?

### Dialog after "accept"

confirm-dialog-dev

### settings
settings-dev

### settings delete item
settings-delete-dev

- [ ] dialog outside server_page
- [ ] implement at sever_page::handleAccept

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -1911,11 +1911,13 @@ Widget dialogButton(String text,
           ? OutlinedButton(
               onPressed: onPressed,
               child: Text(translate(text), style: style),
+              style: OutlinedButton.styleFrom().merge(buttonStyle),
             )
           : OutlinedButton.icon(
               icon: icon,
               onPressed: onPressed,
               label: Text(translate(text), style: style),
+              style: OutlinedButton.styleFrom().merge(buttonStyle),
             );
     } else {
       return icon == null
--- a/flutter/lib/common/widgets/dialogs/widgets/buttons.dart
+++ b/flutter/lib/common/widgets/dialogs/widgets/buttons.dart
@@ -1,0 +1,29 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_hbb/common.dart';
+
+/// Default "OK" button with "check" icon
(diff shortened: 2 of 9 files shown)


rustdesk:
@dinger1986

rustdesk:
The user probabaly read nothing and click ok again. :(

rustdesk:
> * How to open a dialog outside the window from which it was called? How do I get it to the center of the screen when I call the dialog from server_page?

[AlertDialog](https://api.flutter.dev/flutter/material/AlertDialog-class.html)

grummbeer:
> The users probabaly read nothing and click ok again. :(

… yes … maybe something like this?

two-buttons
two-columns

Is this stuff still a wanted at all?

dinger1986:
So would this show up after giving the other user your ID and one-time password? Or just when they are logging in?

grummbeer:
> So would this show up after giving the other user your ID and one-time password? Or just when they are logging in?

When the other wants to connect, after "accept". Green or blue "accept" (both do the same).

cm

rustdesk:
> When the other wants to connect, after "accept". Green or blue "accept" (both do the same).

Green one will elevate priviledge (windows portable version only).

dinger1986:
I think it should show up but its enough hassle downloading and getting remote software like this installed for genuine clients when doing support we dont want to make it harder, or too much harder.

grummbeer:
> I think it should show up but its enough hassle downloading and getting remote software like this installed for genuine clients when doing support we dont want to make it harder, or too much harder.

Yes. Closing this?

btw, how do your customers know which of the two accept buttons they should click? green or blue?.

rustdesk:
> green or blue?.

Yes, that's a problem, can we call the green one "Accept and Elevate"?

dinger1986:
> > green or blue?.
> 
> Yes, that's a problem, can we rename the green one "Accept and Elevate"? and add tooltip to tell why elevate.

Yeah makes sense

PR #4684, opened 2023-06-18
decode utf8 json

21pages:
1. Decoded json using utf8 characters for username/groupname
2. Show user info if login.


rustdesk:
Remove administrator, it is useless, but confuse users.

PR #4689, opened 2023-06-18
Update README.md

SergeyMy:
tested on freshly installed RockyLinux8 (Centos8/RHEL8) //build flutter version


rustdesk:
Thanks, but we prefer original vcpkg based one.

SergeyMy:
my addons basically duplicate the line for ubuntu.
without them, rustdesk won't compile. or something I don't know?

Issue #2283, opened 2022-11-22
File copy and paste function not working


thgadelha8bits:
**Describe the bug you encountered:**

File copy and paste function not working in Windows 10

...

**What did you expect to happen instead?**
Is it possible to copy a file from my computer and paste it into the session user's computer?

...

**How did you install `RustDesk`?**

Rustdesk installed on windows 10 64bit

nightly Release 1.2.0 installed on both computers

---

**RustDesk version and environment**

rustdesk:
enabled?

fufesou:
@thgadelha8bits @zkteco-home Can you please check if this option is enabled?
![image](https://user-images.githubusercontent.com/13586388/203526055-0820125d-2a43-4c1f-af15-e0e11ec72059.png)

thgadelha8bits:
> @thgadelha8bits @zkteco-home Can you please check if this option is enabled? ![image](https://user-images.githubusercontent.com/13586388/203526055-0820125d-2a43-4c1f-af15-e0e11ec72059.png)

Yes, its enable. but not working

rustdesk:
1.1.9 works?

thgadelha8bits:
> 1.1.9 works?

no

rustdesk:
@fufesou it seems our file copy/past is not robust enough, though we can not reproduce it. Let me reopen this and keep watch.

wazerstar:
@rustdesk 

I can confirm with latest nighties.

The copy/move will sometimes crash the target end machine explorer or just freeze rustdesk connection to the controlling unit, while it will be stuck for either a random amount of time or refresh within a few seconds. 

Trying to copy/move once again will never work until you deactivate rustdesk connection down in tray on target machine and enable it again, not very convenient asking the user to do this to continue to operation.

But that's how it is with windows machines atm.

rustdesk:
We may have to rewrite our file copy&paste, current implementation is ugly and buggy. That's why we do not enable it by default.

test2a:
@fufesou hey.... i just updated to the latest nightly on my linux but i dont see the option, is this windows only?

test2a:
i updated to the latest version as i said and clipboard  "from local to remote"  is not working while "remote to local" is working.

fufesou:
@test2a "File copy&pates" only works with connection between Windows.

test2a:
@fufesou got it. the clipboard should work, maybe something broke recently?

test2a:
[...]

PR #4700, opened 2023-06-20
Fix/clipboard flutter

fufesou:
Refactor windows clipboard (file).

image

image

image

A small part of the logic is modified in the `C` code.
It is mainly to modify the control logic of the `Rust` code.

Current changes may break clipboard in sciter version.  I'll modify the code in sciter version based on this code.

1. When passing messages, if the clipboard message is `MonitorReady`, `FormatList`, or `FormatDataRequest`, the message should be blocked according to the configuration. Blocking the message should notify the clipboard to avoid waiting for the message until timeout.
2. When disconnecting, both the client and server will clear the clipboard. However, only the clipboard contents created in their own session will be cleared.
3. For the Flutter version, more accurate clipboard control is provided.

TODOs (I'll fix with another PR):
1. The following system error.  `A --> B`. If `A` copies `a.txt`, B want to copy and paste it's `a.txt` to `A`, or vice versa.

https://github.com/rustdesk/rustdesk/assets/13586388/d677a506-b3d1-44e1-9ac8-ba75212fcd8c

2. Sciter version.
3. Check again whether the related issues are still there.
4. Pass the file information of the clipboard to `Rust`.

### DIFF
--- a/libs/clipboard/src/cliprdr.h
+++ b/libs/clipboard/src/cliprdr.h
@@ -191,8 +191,6 @@ extern "C"
     typedef UINT (*pcCliprdrServerFileContentsResponse)(
         CliprdrClientContext *context, const CLIPRDR_FILE_CONTENTS_RESPONSE *fileContentsResponse);
 
-    typedef BOOL (*pcCheckEnabled)(UINT32 connID);
-
     // TODO: hide more members of clipboard context
     struct _cliprdr_client_context
     {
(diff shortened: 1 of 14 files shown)


rustdesk:
Make sure sciter not broken first.

rustdesk:
https://github.com/rustdesk/rustdesk/issues/2283#issuecomment-1612465795

rustdesk:
let's stop wasting time on this shit/failed implementation.

rustdesk:
A new implementation should not rely on OS too much.

PR #4758, opened 2023-06-24
Fix/tmp config

dignow:
Simple fix temporary password "-" on mobile, when temporary password is enabled.

======

There may still be potential problems.

https://github.com/rustdesk/rustdesk/blob/fbf31f1f1dd64eff8786fa0755d3e8171429bc52/src/ui_interface.rs#L39

`OPTIONS` makes copy of `Config::get_options()`.

It may cause data inconsistency between the two places.

### DIFF
--- a/src/flutter_ffi.rs
+++ b/src/flutter_ffi.rs
@@ -1666,7 +1666,7 @@ pub mod server_side {
         _class: JClass,
     ) {
         log::debug!("startService from jvm");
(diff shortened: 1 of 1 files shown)


rustdesk:
https://github.com/rustdesk/rustdesk/commit/233e9d80e6d6a58d7adc475b58b0dfb2c561f7ea

PR #4770, opened 2023-06-25
Feat/mediacodec

Kayuii:
Enable the hardware decoding of vp9 and vp8 in mediacodec, and extract the decoding thread.


rustdesk:
Yes, hardware does not have better performance regarding speed, depends on gpu/cpu memory copy performance. Can you furtherly improve this by send GPU decoded frame to the frontend texture directly?

21pages:
Great job! Here is a project you can refer to to display the data directly through the gpu without going through the cpu. https://github.com/alexmercerind/media_kit

rustdesk:
Only work on `api-level-30`?

Kayuii:
android > api-level-26 
I haven't found a way to do GPU rendering in the container yet. Alvr proposed a method of using Vulkan API to utilize GPU rendering, but I still don't understand that approach.

Kayuii:
> Only work on `api-level-30`?

It can reach API level 26, some features have been tested previously.

21pages:
https://github.com/irondash/irondash/blob/277f621a78b3121e0f3ad31e720117ef8fc3fbf7/texture/src/lib.rs#L20
https://github.com/irondash/irondash/blob/277f621a78b3121e0f3ad31e720117ef8fc3fbf7/texture/src/lib.rs#L118
I think there should be a way to convert yuv to bgra in gpu, chatgpt tells that OpenGL ES can do that, or decode vp8/vp9 into surface.

rustdesk:
> > Only work on `api-level-30`?
> 
> It can reach API level 26, some features have been tested previously.

The apk can be installed devices under API level 26 ?

Kayuii:
> > > Only work on `api-level-30`?
> > 
> > 
> > It can reach API level 26, some features have been tested previously.
> 
> Can the apk be installed on devices under API level 26 ?

It cannot be used below API level 26, because it uses the ImageReader class, which requires API level 26 support.

rustdesk:
> > > > Only work on `api-level-30`?
> > > 
> > > 
> > > It can reach API level 26, some features have been tested previously.
> > 
> > 
> > Can the apk be installed on devices under API level 26 ?
> 
> It cannot be used below API level 26, because it uses the ImageReader class, which requires API level 26 support.

Thanks, got it.

rustdesk:
Google play starts to require higher api level now.

> App must target Android 13 (API level 33) or higher

rustdesk:
@Kayuii do you have interest continue working on this? directly rendering the textures.

Though we do not release this as default apk, but we can release addtional apk with this feature on.

PR #4865, opened 2023-07-04
+public free server(nt-vps)

NT-VPS:


rustdesk:
put this in readme.md will be very political sensitive, though i am neutral, but......

NT-VPS:
Yes, modify only readme_ru.md
Thanks.

PR #4868, opened 2023-07-04
Added audit and clippy

dinger1986:
For this to work, need to change Settings -> Actions -> General -> Workflow Permissions to allow read and write.

Also It is currently setup to run on PR and push, it might be better to be scheduled?

Have created one for both rustdesk and rustdesk-server but dont want to PR both until the above are answered.


dinger1986:
failing cause its designed to fail if it finds issues

PR #4869, opened 2023-07-05
remove cm use peer info as title

21pages:
#4866 

confy::load_path() will create a file with default config if not exist.


rustdesk:
reverted， let fix this by fixing confy

PR #4930, opened 2023-07-09
wrong verification code

dignow:
If the user has enabled email verification and the verification code is wrong when logging in. The client will get the error "Verification code error or timeout".


rustdesk:
only translation？

PR #4934, opened 2023-07-10
opt mobile chat page

21pages:
1. add border to chatbox
2. add clear button to mobile id input
3. opt mobile chat page menu
   * remove "me"
   * same peer same direction merge chat message
   * same peer different direction show different menu
4. add mobile chat message unread
5. make chat page readOnly if not connected for mobile.
6. mobile not show chat icon overlay on main page 

Test on mobile/desktop, do more test later.

87fddf3e08e2c40feb9919b6523baeb

b0e9cff3fd04cd20fa16b7f68563589

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -2189,19 +2189,21 @@ Widget buildRemoteBlock({required Widget child, WhetherUseRemoteBlock? use}) {
       ));
 }
 
-Widget unreadMessageCountBuilder(RxInt? count) {
+Widget unreadMessageCountBuilder(RxInt? count,
+    {double? size, double? fontSize, double? marginLeft}) {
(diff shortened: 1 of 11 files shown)


sahilyeole:
Hi @21pages, any insights on why this condition was added?
https://github.com/rustdesk/rustdesk/blob/7356b7a104e8a530ede40d805688c66e18923067/flutter/lib/common/widgets/overlay.dart#L35-L37
If I remove the condition the chat window is opened on ios or else not.
However, I have to say it is not draggable and these icons do not work on ios.

Screenshot 2023-08-30 at 9 47 16 PM

21pages:
> If I remove the condition the chat window is opened on ios or else not.

Remove it if that works.

PR #4976, opened 2023-07-12
Allow to explicitly increase the bitrate massively

bwoebi:
This has the benefit - over direct connections - to be able to send a lot more data. In particular fast moving frames may avoid fps drops (or short stutter) with a higher target bitrate.

See also: #4682

I'm currently hiding the functionality well behind an extra checkbox you need to tick to be able to select up to 40x the bitrate. Not sure on the proper UI, but I'm open to any suggestions, maybe also some hint in a tooltip that this should only be used on peer to peer connections or dedicated servers?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5146,7 +5146,7 @@ dependencies = [
 
 [[package]]
 name = "rustdesk"
-version = "1.2.1"
+version = "1.2.2"
 dependencies = [
  "android_logger",
  "arboard",
--- a/flutter/lib/common/widgets/dialog.dart
+++ b/flutter/lib/common/widgets/dialog.dart
@@ -1220,7 +1220,8 @@ customImageQualityDialog(SessionID sessionId, String id, FFI ffi) async {
   qualityInitValue =
       quality != null && quality.isNotEmpty ? quality[0].toDouble() : 50.0;
   const qualityMinValue = 10.0;
-  const qualityMaxValue = 100.0;
+  const qualityMoreThresholdValue = 100.0;
+  const qualityMaxValue = 4000.0;
   if (qualityInitValue < qualityMinValue) {
     qualityInitValue = qualityMinValue;
   }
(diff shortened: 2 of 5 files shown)


rustdesk:
Can you show me the video about the UI change?

rustdesk:
BTW, this TIS is being discarding, our new version has been switched to flutter.

bwoebi:
Here's a short video: https://streamable.com/srwxxm

I can also update the flutter UI, once I get approval on the UI :-)

bwoebi:
@rustdesk I would appreciate if the flutter install & build steps would be called out somewhere in the README or if cargo run would just take care of it.
Now I tried to follow the steps from the github actions:
https://github.com/rustdesk/rustdesk/blob/fbfad5f86d62fded83891840c3e1bc9af30300c4/.github/workflows/flutter-build.yml#L85-L87
https://github.com/rustdesk/rustdesk/blob/fbfad5f86d62fded83891840c3e1bc9af30300c4/.github/workflows/flutter-build.yml#L97
Re-running the python line fails with `FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'flutter/build/windows/runner/Release/RustDeskIddDriver'` - removing the file works again.

Overall, it feels a bit like a brittle experience, building sciter is much more straightforward :-D
(But working with flutter is much more intuitive, took me much less time. So I appreciate flutter being used :-))

Added flutter support now: https://streamable.com/1jlgew

rustdesk:
@bwoebi we have changed qos a lot lately.

bwoebi:
@rustdesk Thanks, I've rebased the branch.

Fwiw: from my own testing the perfect bitrate is around _screen width * screen height * fps_ bit/s (around 40 times less than if you'd send every frame as bitmap). While the fps seem to be mostly CPU bound.
For me that's about a bitrate of 1000%.

I'll try the build with the modified qos over the next few days.

bwoebi:
@rustdesk I like the qos changes, I had noticeably less spurious stutter, in particular with increased bitrate. Thanks for the improvements!

However, with the standard bitrate heavily animated content would still just degrade too much. (Even though it's better than before.)

rustdesk:
Let's do this ourself if needed.

rustdesk:
close as we will handle ourself.

rustdesk:
@21pages Let's do this ourself if needed.

PR #5046, opened 2023-07-18
multi user video qos and fix quality change

21pages:
1. store multi user's setting in video qos
2. allow client to increase fps 
3. client transfer full speed decode fps to server
4. revert vpx use webrtc parameter, which makes the dynamic picture better, but the bitrate adjustment does not work.
5. codec set quality seperately. Make sure the distinction between different image quality is obvious, and the highest image quality behaves well for dynamic scenes, Still not good for relay connection. 
6. Automatically adjust q_max/q_min according to custom bitrate
7. uses the remaining 50%(win: cpu usage, unix: loadavg) of cpu as codec thread.
8. Even in direct connections, lower bitrate can get lower fps because of frame drop and  failed encoding.

### DIFF
--- a/libs/scrap/examples/benchmark.rs
+++ b/libs/scrap/examples/benchmark.rs
@@ -2,7 +2,7 @@ use docopt::Docopt;
 use hbb_common::env_logger::{init_from_env, Env, DEFAULT_FILTER_ENV};
 use scrap::{
     aom::{AomDecoder, AomDecoderConfig, AomEncoder, AomEncoderConfig},
-    codec::{EncoderApi, EncoderCfg},
+    codec::{EncoderApi, EncoderCfg, Quality as Q},
     Capturer, Display, TraitCapturer, VpxDecoder, VpxDecoderConfig, VpxEncoder, VpxEncoderConfig,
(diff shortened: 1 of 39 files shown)


SergeyMy:
no desire to move hbb_common to the repository? tired of transferring it to the Rendezvous server 
изображение

rustdesk:
Should we expose "Quantizer" again? I believe nobody can understand it.

libs/scrap/examples/benchmark.rs
```diff
@@ -129,11 +143,7 @@ fn test_vpx(
         size / yuv_count
     );
 
-    let mut decoder = VpxDecoder::new(VpxDecoderConfig {
-        codec: codec_id,
-        num_threads: (num_cpus::get() / 2) as _,
-    })
-    .unwrap();
+    let mut decoder = VpxDecoder::new(VpxDecoderConfig { codec: codec_id }).unwrap();
```
rustdesk, line 146: Let's remove this unwrap, though it is from old code.

rustdesk: Also please check the other unwrap, expect. etc.

PR #5124, opened 2023-07-25
Fix sciter build on MacOS

bwoebi:
c86a8fff034c7dd3b3ddf4f33d69eaee5c3538b0 accidentally duplicated that part of the code from send_mouse (the function just above). Removing it.


rustdesk:
@fufesou

Issue #5109, opened 2023-07-24
Rustdesk does not retain Monitor locations/positions after program close/reopen


godzfire:
### Bug Description

The main Rustdesk window nor the connection window remember/retain which monitor they are on and where after closing and reopening the app.

### How to Reproduce

1- Open Rustdesk
2- Connect to a remote session
3- Move either windows to a different monitor
4- Quit the program and reopen

The windows will always revert back to the default main monitor and screen location.

### Expected Behavior

The windows will retain their monitor and location placement information and reappear back in those spots after closing and reopening.

### Operating system(s) on local side and remote side

Windows 10 -> Mac

### RustDesk Version(s) on local side and remote side

1.2.0 build 2023 06-28

### Screenshots

N/A

rustdesk:
@fufesou

rustdesk:
close as fix merged, please try out the latest nightly build.

RichardBurt65:
Confirmed fixed on 1.2.2 nightly.

PR #5125, opened 2023-07-26
Fix/window save pos

dignow:
https://github.com/rustdesk/rustdesk/issues/5109

Save the position when the window is moved or resized.

Please merge https://github.com/rustdesk-org/rustdesk_desktop_multi_window/pull/2 first.

Notes:
1. For Windows and MacOS, save the position on moved and resized is enought.
2. For Linux, the main window does not listen the moved and resized events. It should also save the position on close.

Windows will get negative positions if the window is minimized.
While Linux does not.

https://github.com/rustdesk/rustdesk/assets/136106582/cef180fe-679a-4a71-ad7e-8db1df70dc7c

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -1383,12 +1383,17 @@ Future<void> saveWindowPosition(WindowType type, {int? windowId}) async {
         return;
       }
       final position = frame.topLeft;
+      if (position.dx < 0 || position.dy < 0) {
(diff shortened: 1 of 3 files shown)


rustdesk:
pub merged， modify lock file then

rustdesk:
how about mac？

fufesou:
Mac already emit "moved" and "resized" events in the multi window pub.

Issue #4529, opened 2023-06-02
The number will be misplaced when entering the remote ID


SamGenTLEManKaka:
### Bug Description

The number will be misplaced when entering the remote ID

### How to Reproduce

Just enter the remote ID and it will reappear

### Expected Behavior

Just enter the remote ID and it will reappear

### Operating system(s) on local side and remote side

Windows server2016 && Windows server2019

### RustDesk Version(s) on local side and remote side

nightly

### Screenshots

https://github.com/rustdesk/rustdesk/assets/24671601/e3cb3483-c771-4cfa-8214-2ebdedd01e38

Apashh:
Same problem with nightly on Windows 11 ! ;)

rustdesk:
Interesting, but not on my windows. Your font also has problem.

rustdesk:
Which day's nightly?

SamGenTLEManKaka:
> Which day's nightly?

There's a problem with the latest one

rustdesk:
I jsut reproduced this issue when turning on IME, no problem without IME.

rustdesk:
close as fix merged

PR #5187, opened 2023-07-29
fix IDTextEditingController extra character

21pages:
#4529

This can fix repeat characters when have non-english ime.


danhduykhan-code (changes requested):
Duykhan

PR #5198, opened 2023-07-30
Fix/win file clipboard

dignow:
This PR do not fix the common "file copy&patest" issues.

It fix the following "File copy&paste" cases.

image

image

The only change is `libs/clipboard/src/windows/wf_cliprdr.c`.
The others are debug log lines.

```C
	if (!IsClipboardFormatAvailable(CF_HDROP))
	{
		return ERROR_SUCCESS;
	}
```


ClSlaid:
Will this cut off the capability to common formats like `CF_TEXT`?

PR #5208, opened 2023-07-31
fix, win file clipboard, when authorized by the remote

dignow:
https://github.com/rustdesk/rustdesk/issues/2283

Fix "File copy and paste function not working" when the remote side manually click authorize.

But I'm not sure if it is exactly the same issue.

The dart code is simple a refactor. No code logic is changed.

### DIFF
--- a/flutter/lib/models/server_model.dart
+++ b/flutter/lib/models/server_model.dart
@@ -410,18 +410,25 @@ class ServerModel with ChangeNotifier {
   updateClientState([String? json]) async {
     if (isTest) return;
     var res = await bind.cmGetClientsState();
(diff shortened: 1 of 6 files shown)


rustdesk:
Impossibly it is the same. I do not think users need to use copy&paste when connection is established via accept button.

dignow:
https://github.com/rustdesk/rustdesk/assets/136106582/407a8e71-6e14-45cf-a6d1-cae1c391ff96


https://github.com/rustdesk/rustdesk/assets/136106582/03b683e9-91ce-4be3-8d37-1923d94a7ec6

PR #5256, opened 2023-08-05
Refact/separate remote window

dignow:
https://github.com/rustdesk/rustdesk/assets/136106582/6d2db328-07ba-4f44-bb53-25b232e8265c

TODOs:

- [x] If the pictures of the remote side are not changed, it will not send the picture. The result is the new window keeps "Connected, waiting for image...". Need to cache the last image or notify the remote to send image data.

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -1408,8 +1408,24 @@ Future<void> saveWindowPosition(WindowType type, {int? windowId}) async {
       sz.width, sz.height, position.dx, position.dy, isMaximized);
   debugPrint(
       "Saving frame: $windowId: ${pos.width}/${pos.height}, offset:${pos.offsetWidth}/${pos.offsetHeight}");
+
   await bind.setLocalFlutterConfig(
       k: kWindowPrefix + type.name, v: pos.toString());
+
+  if (type == WindowType.RemoteDesktop && windowId != null) {
+    await _saveSessionWindowPosition(windowId, pos);
+  }
+}
+
+Future _saveSessionWindowPosition(int windowId, LastWindowPosition pos) async {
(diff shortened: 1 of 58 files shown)


SergeyMy:
@dignow

I think I found an ERROR :(
1. launch the first connection window in normal mode
2. launch the second connection window in a "separate window"

all - one rustdesk process consumes 100% cpu - one core
**this won't stop until rustdesk is closed completely** :(

dignow:
@SergeyMy Thanks for your feedback. I'll try fix it.

rustdesk:
Anoter report, https://github.com/rustdesk/rustdesk/discussions/2889#discussioncomment-6648679

SergeyMy:
oh and a bonus...
(happened several times)
two connections are open, on the first I call "move tab to new window" and get a black screen or a frozen picture (stream break)
_I didn’t understand the repeatability, did not look for where the bug is it in connection with the detection of process leaks._

dignow:
> all - one rustdesk process consumes 100% cpu - one core

I've tested win, linux, macos. Only linux has this issue.

I need some more time to find out the reason.

SergeyMy:
in Linux vs. Windows, the system of processes / threads is fundamentally different. something that did not happen on ios is somewhat unusual

SergeyMy:
@dignow
https://github.com/rustdesk/rustdesk/pull/5269#issuecomment-1666917610

add:
the process "leaks" strictly to one client, you seem to mark them by ID, re-opening a new window with the same ID already works (does not create new threads)
IMHO there is just a deaf cycle from the loss of messages

PS
I am ready to endure it, but if this is disabled in the master branch, I will have difficulties with the tests.

PR #5269, opened 2023-08-06
tmp, remove separate remote window on linux

dignow:
Temp commit, for cpu and other issues.

https://github.com/rustdesk/rustdesk/pull/5256#issuecomment-1666726529

https://github.com/rustdesk/rustdesk/pull/5256#issuecomment-1666742122

I'll add this feature back to linux once the issue is fixed.

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -1285,6 +1285,9 @@ mainSetLocalBoolOption(String key, bool value) async {
 }
 
 bool mainGetLocalBoolOptionSync(String key) {
+  if (key == kOptionSeparateRemoteWindow && Platform.isLinux) {
+    return false;
+  }
   return option2bool(key, bind.mainGetLocalOption(key: key));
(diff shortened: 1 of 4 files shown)


SergeyMy:
> I'll add this feature back to linux once the issue is fixed.
sorry, I'm willing to tolerate thread leaks just for the sake of the multi-windows feature.

PS
PS noticed two unusual messages in the logs:
(when opening a new window with a leak)
```
WARN  librustdesk::flutter] Global event stream of type remote is started before, but now removed
...
ERROR rdev::linux::grab] Failed to receive event, receiving on a closed channel
```
//build with debug

dignow:
> ```
> WARN  librustdesk::flutter] Global event stream of type remote is started before, but now removed
> ...
> ERROR rdev::linux::grab] Failed to receive event, receiving on a closed channel
> ```

Yes, that's the reason.

PR #5273, opened 2023-08-07
fix, separate window, 100%CPU & event stream leak

dignow:
1. Fix repeatedly grabbing keyboard, which causes 100% CPU.
2. Distinguish between the global event streams.

Fix https://github.com/rustdesk/rustdesk/pull/5256#issuecomment-1666726529

May also fix https://github.com/rustdesk/rustdesk/pull/5256#issuecomment-1666742122

@SergeyMy Thanks, you really helps a lot.


SergeyMy:
yes it seems to work :)

PR #5298, opened 2023-08-08
opt peercard order

21pages:
1. update connect id when create new connection (android not work, I'll figure it out lately)
2. opt delete dialog title
3. skip PeerConfig saving when not receiving peerinfo to keep peercard order
.


rustdesk:
> * update connect id when create new connection (android not work, I'll figure it out lately)

Some one reported this.

PR #5440, opened 2023-08-19
peer tab icon hover decoration

21pages:
hover decoration of right peer tab icon, add tile card checkbox  margin

https://github.com/rustdesk/rustdesk/assets/14891774/77623c69-5a53-449f-aff1-0d809227fdba


SergeyMy:
@21pages 
How can you change the color of these lights?
and they crawl a little on top of each other
изображение

PR #5483, opened 2023-08-22
Using github latest release url to check for software update

beelchester:
https://github.com/rustdesk/rustdesk/discussions/5475#discussioncomment-6790057

https://github.com/rustdesk/rustdesk/assets/73148455/3b85268b-afe5-42fe-9f4b-12810d4c66d3

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -831,30 +831,19 @@ pub fn check_software_update() {
 
 #[tokio::main(flavor = "current_thread")]
 async fn check_software_update_() -> hbb_common::ResultType<()> {
-    sleep(3.).await;
+    let url = "https://github.com/rustdesk/rustdesk/releases/latest";
+    let latest_release_response = reqwest::get(url).await?;
+    let latest_release_version = latest_release_response
+        .url()
+        .path()
+        .rsplit('/')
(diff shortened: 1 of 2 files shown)


src/common.rs
```diff
@@ -831,30 +831,19 @@ pub fn check_software_update() {
...
-            }
-        }
+    let url = "https://github.com/rustdesk/rustdesk/releases/latest";
+    let latest_release_response = reqwest::get(url).await?;
+    let latest_release_version = latest_release_response
+        .url()
+        .path()
+        .rsplit('/')
+        .next()
+        .unwrap();
```
rustdesk, line 841: potential crash here (unwrap)

rustdesk: Never use unwrap except have to use, e.g. mutex.

beelchester: Ok, got it.
Looks like you have fixed it https://github.com/rustdesk/rustdesk/commit/28ddf6cf079a5cad14e75b1e311596c026fef326 
Thanks.

rustdesk: one-day effort we used to position this crash since it happened in very specific case only.

PR #5491, opened 2023-08-23
Update and improve tw translation

PeterDaveHello:
Looks like some added translation seemed to be Simplified Chinese, not the Traditional Chinese that Taiwanese will use, maybe should try to prevent that situation causing those strings need to be translated twice?


rustdesk:
Got it. @fufesou @21pages

Issue #5513, opened 2023-08-25
Connection to locked computer freezes on wait for image


papaia:
### Bug Description

When trying to connect to a locked computer (with Rustdesk open on it), connection gets stuck on "Connected, waiting for image" and does not proceed any further. None of the buttons are clickable and there is no way to proceed.

### How to Reproduce

1. Connect to a remote (Windows 10) computer
2. Lock it using the Start menu
3. Disconnect
4. Connect to the same computer

### Expected Behavior

Should be able to enter OS password or see the lock screen.
Alternatively be able to click buttons to enable auto login for the next session.

### Operating system(s) on local side and remote side

Windows 10 professional -> Windows 10 professional

### RustDesk Version(s) on local side and remote side

1.2.1 -> 1.2.1 and 1.2.2 -> 1.2.1

### Screenshots

![screenshot of waiting for image](https://github.com/rustdesk/rustdesk/assets/43409674/3e87786b-aa66-4816-84ea-d852194ac5f5)

rustdesk:
https://github.com/rustdesk/rustdesk/issues/5488#issuecomment-1689936063

rustdesk:
https://github.com/rustdesk/rustdesk/pull/5489 available in nightly build, let me know it works for you.

rustdesk:
@fufesou follow up this if 1.2.3 does not work and reproduce his lock scenario. need to find a way for this case.

papaia:
I downloaded 1.2.3 and used the menu to gain elevation which fixed the problem and I am now logged in.
However, I was not able to see the "Send ctrl-alt-del" option in the menu.
In any case, thanks for the support!

fufesou:
@papaia Thanks for your feedback.

I cannot see the "Lock" action on the start menu.
But I can reproduct this issue by click "Sleep" and then click the mouse the get awake. Then connect through RustDesk.

> I was not able to see the "Send ctrl-alt-del" option in the menu.

Did you hava rustdesk installed on the controlled side?

> I downloaded 1.2.3 and used the menu to gain elevation which fixed the problem and I am now logged in.

Which menu did you use to gain elevation?

fufesou:
@papaia Hi, can you please help confirm whether the controlled side can wake up.

I cannot reproduce this problem every time. 
So I'm not sure about it.

rustdesk:
Please try out the nightly build 1.2.3. we put some fix.

PR #5523, opened 2023-08-27
win, try awake the os by mouse movement and right click

fufesou:
https://github.com/rustdesk/rustdesk/issues/5513

Try send right click to awake the OS.


rustdesk:
need to confirm if it can awake linux os password after lock

fufesou:
This PR does not affect the os password box.

The left mouse button is only sent if the password entered is not empty.

PR #5539, opened 2023-08-29
Adding support MacPorts and /usr/local/ opus library

xavetar:
I didn't end up getting rustdesk to work with sciter. I saw in the documentation that now for macOS you need to build together with Flutter. These commits will work on macOS if sciter works, which doesn't work on macOS 13.5.1 and possibly some versions below. Sciter definitely supports OSX, but I get:

[Log](https://pastebin.com/ZkmHKsMj)

For this commit to work, you need to add this [commit](https://github.com/rustdesk-org/magnum-opus/pull/4).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,37 +1,45 @@
-/build
-/target
-.vscode
-.idea
-.DS_Store
-libsciter-gtk.so
-src/ui/inline.rs
-extractor
-__pycache__
+# Rust & Cargo
+debug/
+target/
+Cargo.lock
+**/*.rs.bk
 src/version.rs
-*dmg
-*exe
-*tgz
-cert.pfx
-*.bak
-*png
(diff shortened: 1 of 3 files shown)


xavetar:
Hmm, funny, I didn't know to copy src/ui to target/release for this to work. Now everything works fine, but an error occurs when you click on the ellipsis:



`2023-08-29 10:02:56.094 rustdesk[11353:3131834] *** Assertion failure in void _NSWindowSetFrameIvar(NSWindow *, NSRect)(), NSWindow.m:1029`

xavetar:
Now I have copied the entire src folder and no error occurs. Yes it works.

xavetar:
Yes, everything works, but the keyboard does not work. Tried to enable Input Monitoring, nothing happens.

PR #5562, opened 2023-08-31
version information in self-extracted executable, year 2023 in Copyright

deep-soft:
change year in Copyright, 2022 to 2023


rustdesk:
Please modify rc/rs file directly rather than using sed.

PR #5570, opened 2023-09-01

deep-soft:
https://github.com/rustdesk/rustdesk/discussions/5560

### DIFF
--- a/libs/portable/build.rs
+++ b/libs/portable/build.rs
@@ -1,5 +1,5 @@
 extern crate embed_resource;
 
 fn main() {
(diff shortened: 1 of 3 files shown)


libs/portable/build.rs
```diff
@@ -1,5 +1,10 @@
 extern crate embed_resource;
+use std::fs;
 
 fn main() {
-    embed_resource::compile("icon.rc", embed_resource::NONE);
+    let runner_res_path = "./Runner.res";
```
rustdesk, line 5: Where is Runner.res? How was it produced?

PR #5574, opened 2023-09-01

deep-soft:
https://github.com/rustdesk/rustdesk/discussions/5560

### DIFF
--- a/.github/workflows/flutter-build.yml
+++ b/.github/workflows/flutter-build.yml
@@ -102,14 +102,14 @@ jobs:
         continue-on-error: true
         shell: bash
         run: |
-          runner_res=$(find . -name "Runner.res")
+          runner_res=$(find . -name "Runner.res");
           if [ "$runner_res" == "" ]; then
-            echo "Runner.res: not found"
+            echo "Runner.res: not found";
           else
-            echo "Runner.res: $runner_res"
-            cp $runner_res ./Runner.res
-            echo "list ./Runner.res"
-            ls -l ./Runner.res
+            echo "Runner.res: $runner_res";
+            cp $runner_res ./libs/portable/Runner.res;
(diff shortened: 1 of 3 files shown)


deep-soft:
> not answer my question in your last pr yet.

PR deleted because failed.

Runner.res is "produced" only on flutter version, from Runner.rc
In sciter version there is no Runner.res because there is no Runner.rc

That's why Windows flutter version compiled ok only by modifying flutter-build.yml
and sciter version remained as before (with no version information included).

On the other hand "Runner.rc" does not contain the current version of the program but "Runner.res" does contain the current version on compilation.

deep-soft:
Now i'm trying to fix this PR, to work for flutter and sciter version.

deep-soft:
can't fix this error:
cp: cannot stat '.res': No such file or directory

in flutter-build.yml
i have this step:
```
      - name: find Runner.res
        # Windows: find Runner.res (compiled from ./flutter/windows/runner/Runner.rc), copy to ./Runner.res
        # Runner.rc does not contain actual version, but Runner.res does
        continue-on-error: true
        shell: bash
        run: |
          runner_res=$(find . -name "Runner.res")
          if [ runner_res != '' ]; then
            echo "Runner.res: $runner_res"
            cp $runner.res ./Runner.res
          fi

```
and the result is:
cp: cannot stat '.res': No such file or directory

instead of:
cp: cannot stat 'Runner.res': No such file or directory

I don't know why the string Runner is replaced with nothing even if I use double quote !!!

deep-soft:
@rustdesk 
This PR works for flutter version and include FileVersion information in self-extracted executable (same as for rustdesk.exe). 
For sciter version Runner.rc must be generated (with corect FileVersion) and compiled to Runner.res
So, for sciter version, of self-extracted executable, only the icon is included (as before this PR).

PR #5623, opened 2023-09-07
add file log page to cm

21pages:
* Only send and receive logs are shown
* For older version, client send to server doesn't have size information, because server side doesn't know the total_size.
* Not switch tabs automatically when new files are transferred
* If cm side page is open, not pop up automatically when new files are transferred 
* Show unread message count
* The cm tab remains open when closed if a file transfer has previously occurred, which follows anydesk

https://github.com/rustdesk/rustdesk/assets/14891774/e69497f7-1bad-4cbb-b2cd-a5b93e87b381


rustdesk:
> * The cm tab remains open when closed if a file transfer has previously occurred, which follows anydesk

keep open for ever?

PR #5645, opened 2023-09-10
feat, mouse wheel and touchpad scroll mode, default or reverse

fufesou:
https://github.com/rustdesk/rustdesk/discussions/5369

Add global default option and session option for "Reverse mouse wheel".

TODO:
There may be a potential bug for the default (bool) options, which use 'Y' as true and empty as false.

For example:

If one set the session option "Reverse mouse wheel" to be false, the session option is removed(set as empty).
But if the default option is then changed.

The session does not know whether to use the changed default option or the false option.

May be it is better to set "Reverse mouse wheel" to "N" if one want to disable the option.

### DIFF
--- a/flutter/lib/consts.dart
+++ b/flutter/lib/consts.dart
@@ -138,6 +138,11 @@ const kRemoteScrollStyleAuto = 'scrollauto';
 /// [kRemoteScrollStyleBar] Scroll image with scroll bar.
 const kRemoteScrollStyleBar = 'scrollbar';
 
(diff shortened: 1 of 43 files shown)


rustdesk:
whithout any background, do you think the user can understand you in the picture?

rustdesk:
Move to default optinons in Display tab.

rustdesk:
change to checkbox, "Reverse mouse wheel"

rustdesk:
> There may be a potential bug for the default (bool) options, which use 'Y' as true and empty as false.

We knew this drawbacks, no need to over consider about this.

PR #5673, opened 2023-09-12
Update cn.rs “或 IP”

mokeyish:
记得之前是支持 IP 直连的，但是一直没找到输入的地方。最后发现，首页输入框原来是可以直接填 IP。

把翻译加上 “或 IP” 就直观了。


mokeyish:
@rustdesk  In an era when IPv6 will be prevalent, it would be a pity not to support IP direct connection. If it only supports relay mode, wouldn't RustDesk become another TeamViewer?

PR #5674, opened 2023-09-12
auto disconnect

21pages:
#5540
basically same as Anydesk
* default not enabled, 10 minutes
* not used in file transfer, port forward
* the elapsed time will be updated when receive key/mouse event (even if no permission), chat message, refresh screen, elevation request
* will not reconnect automatically

48964da8457c896633d79d90878cd98


rustdesk:
TV has countdown, may add countdown in the future, no decided yet.

PR #5720, opened 2023-09-17
enable group

21pages:
1. Support modify peer tab visibility, right click for desktop, long press for mobile
2. Enable group, display all accessible users and devices:
* enabled users and devices of  own group
* enabled users and devices of accessible groups , which consistent with the web console setting
* enabled unbinded devices

Notice: This will work in the next pro version, current pro version will prompt to upgrade to 1.1.10

image

b5f40b8748ad1b15b2ba75c8e03f3b2


flutter/lib/common/widgets/peer_tab_page.dart
```diff
@@ -182,14 +201,17 @@ class _PeerTabPageState extends State<PeerTabPage>
...
     final model = Provider.of<PeerTabModel>(context);
     Widget child;
-    if (model.indexs.isEmpty) {
-      child = Center(
-        child: Text(translate('Right click to select tabs')),
-      );
+    if (model.visibleIndexs.isEmpty) {
+      child = visibleContextMenuListener(Center(
+        child: Text(translate(isMobile
+            ? 'Long press to select tabs'
```
rustdesk: Where does this display on mobile? Show me a pic.

21pages: 1acb24961aea54c2469ca9043b611b2

PR #5752, opened 2023-09-19
fix, set texture widget offset to int on linux

dignow:
Linux requires the `Texture` widget offset to be integers.

TODOs:

I'll push other PRs.

1. Resizing window is not smooth. Recent commits.

https://github.com/rustdesk/rustdesk/assets/136106582/37e0c593-7757-4be2-a250-a29e1add3cab

2. "Scale adaptive" blurry image, `Texture` widget.

https://github.com/rustdesk/rustdesk/assets/136106582/c59a2bb1-f814-4891-9c65-4c1d09071d76


dignow:
> Resizing window is not smooth. 

The smooth issue on resizing window disappear. I just checkout some commit and rebuild.
Maybe it's an issue that is hard to reproduce. I'll try find the reason when I see it again.

PR #5773, opened 2023-09-21
Fix wayland cursor mismatch on multiple screens

### LINKED ISSUE #4288: Cursor does not match the actual position on secondary screen (wayland, cursor controlling always on the primary screen)
### Bug Description

I am using two screens:

only screen 2 is shared (which is fine, while I am not sure how to select the screen at all)

The mouse courser on the controlling mobile is on the second screen, but on the controlled side it is on the primary screen.

### How to Reproduce

1. Set up two screens
2. connect and share the 2nd one
3. allow mouse cursor control
4. mouse coursers don't match

### Expected Behavior

The blue and the green arrow should match.

But in my case I see the green cursor on Android while on my desktop I see it and click on the blue position.

So controlling is actually not possible.

Ubuntu 22.04 with Wayland > Android Phone (phone is the controllnig side, wayland is the controlled side)

local: nightly as of today (md5sum 2fc23126e0dd6eadfb3d70cf7a83a85d  rustdesk-1.2.0-x86_64.deb)
remote: downloaded from appstore

Should not be identical with https://github.com/rustdesk/rustdesk/issues/4261

### Screenshots

image

### Additional Context

Looks like the primary screen offset is not taken into account?

I also switched positions of the screens in the Display config - this did not seem to help.

beelchester:
Fixes #4288
Fixed mouse resolution.
Tested Ubuntu wayland (virtual secondary display).
Works perfectly with Xwayland (enabled by default on most wayland compositors), even on different display resolutions.
But if someone doesn't have Xwayland (which I think is quite unlikely) this fix will still work in 2nd and 3rd display position cases as shown in the video below, as long as both displays have same resolutions.

Following video shows various cases of display positions (sharing display no. 2):

https://github.com/rustdesk/rustdesk/assets/73148455/d9f1ce8e-66b8-4645-b37a-24b141998a75

### DIFF
--- a/src/server/wayland.rs
+++ b/src/server/wayland.rs
@@ -2,6 +2,7 @@ use super::*;
 use hbb_common::{allow_err, platform::linux::DISTRO};
 use scrap::{is_cursor_embedded, set_map_err, Capturer, Display, Frame, TraitCapturer};
 use std::io;
+use std::process::{Command, Output};
 
 use crate::client::{
     SCRAP_OTHER_VERSION_OR_X11_REQUIRED, SCRAP_UBUNTU_HIGHER_REQUIRED, SCRAP_X11_REQUIRED,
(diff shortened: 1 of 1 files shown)


rustdesk:
@fufesou review this please.

PR #5798, opened 2023-09-24
mobile still use no scrollable widgets

21pages:


rustdesk:
merged as workaround, fix listview problem on mobile in next PR please.

Issue #5872, opened 2023-10-01
No option to disable screen cursor leading to double typing of letters (ios->android)


yogeshbenz01:
### Bug Description

There is no option to disable the on screen cursor. Even though i turn it off on display setting it still shows up. While typing every letters are typed twice.

Currently using in iphone.

### How to Reproduce

Type letter in iphone.

on screen cursor is showing which cant be removed.

### Expected Behavior

Letters are supposed to be clicked once.

### Operating system(s) on local side and remote side

Ios 17

### RustDesk Version(s) on local side and remote side

1.2.2>1.1.10

### Screenshots

`
![IMG_8987](https://github.com/rustdesk/rustdesk/assets/38241265/4a881cc0-61a4-45f5-91e5-d077245238c2)
![IMG_8986](https://github.com/rustdesk/rustdesk/assets/38241265/5f4c5378-ac9f-4863-b6ce-9778d1a083f1)
![IMG_8985](https://github.com/rustdesk/rustdesk/assets/38241265/08ec7aee-b895-4789-9578-7f9963fc7c21)
![IMG_8984](https://github.com/rustdesk/rustdesk/assets/38241265/68f8d43f-a99e-4c95-a9e2-5a2ff3cc9c6c)
`
![image](https://github.com/rustdesk/rustdesk/assets/71636191/0e802aeb-4156-4225-8170-0973d0c6a491)

![image](https://github.com/rustdesk/rustdesk/assets/71636191/323ddec6-5722-43df-a418-6a4eb8877205)
![image](https://github.com/rustdesk/rustdesk/assets/71636191/046b04d7-132d-4507-ac17-cf2dda4db373)

![image](https://github.com/rustdesk/rustdesk/assets/71636191/929c5674-e43e-4c18-9c7d-58391a7b7208)

rustdesk:
How about both 1.2.2? Frankly, we never test mobile -> mobile.

rustdesk:
@fufesou let's remove cursor pointer for mobile -> mobile, it is useless.

yogeshbenz01:
> How about both 1.2.2? if double letter. Frankly, we never test mobile -> mobile.

It was working fine before the update i believe like the below comment said its better to remove the cursor pointor for mobile to mobile.

rustdesk:
@fufesou is "double letter" related to the "cursor"?

fufesou:
I don't know, but I think they should be unrelated. I'll try fix the issue later.

[...]

PR #5874, opened 2023-10-01
refact, remove cursor if mobile -- mobile

fufesou:
https://github.com/rustdesk/rustdesk/issues/5872#issuecomment-1741926108

Do not show `CursorPaint()` if mobile --> mobile.

Only peer Android is checked.

TODO:

Check peer ios.


rustdesk:
iOS can not be controlled, there is no peer ios.

PR #5947, opened 2023-10-08
feat(translation): add and fix Dutch strings

AnonymousWP:


AnonymousWP:
Related question now I'm at it: https://github.com/rustdesk/rustdesk/discussions/5948

PR #6000, opened 2023-10-12
Update "Key" translation

Kleofass:


rustdesk:
Thank you very much, I do not think `key` is appropriate for translation.

Issue #5986, opened 2023-10-11
RustDesk 1.2.3 window does not show on Ubuntu MATE 22.04 unless I run it with an argument (shown on task bar)


wodin:
### Bug Description

I'm doing this on an `Ubuntu MATE 22.04` machine running Xorg.

I had version 1.2.1 installed, but it was crashing when trying to connect to a remote machine. So I downloaded rustdesk-1.2.3-x86_64.deb and installed it.

Now, when I run it, it shows on the task bar, but the window itself does not display. If I hover over it on the task bar, I see the thumbnail of the window correctly. If I click on it, I see the outline of the window minimizing, but do not see the actual contents of the window. Also, while the window is unminimized, it seems to accept mouse input. Or at least, if I click in the area where it would be, the window "behind" the RustDesk window does not respond to the click. So it is as if the RustDesk window is there, but just invisible.

If I run rustdesk from the command line and give it any argument except `--version` (e.g. `-h`, or `--help`, or `-x` or `x`) then the window shows correctly!

After that I downloaded the `AppImage` and it had the same behaviour.

### How to Reproduce

Run `rustdesk` from the command line with no arguments -> invisible window

Run `rustdesk somearg` from the command line -> window shows correctly

### Expected Behavior

When I run RustDesk from the command line with no arguments, or from the Ubuntu MATE menu, I expect the window to be visible.

### Operating system(s) on local side and remote side

Ubuntu MATE 22.04 -> N/A

### RustDesk Version(s) on local side and remote side

1.2.3 -> N/A

### Screenshots

![RustDesk showing thumbnail, but invisible window](https://github.com/rustdesk/rustdesk/assets/4579020/2cf018e0-c14c-4c91-b6eb-c4817138ce6f)

As you can see in the screenshot, the thumbnail shows correctly when hovering over RustDesk on the task bar, but the window is not showing. The window is *not* minimized. It is just invisible. If I click on it in the task bar, I can see the minimize animation. If I click on it again, to unminimize it, it is still invisible.

rustdesk:
How about Flatpak?

rustdesk:
> Run `rustdesk` from the command line with no arguments -> invisible window
> 
> Run `rustdesk somearg` from the command line -> window shows correctly

@fufesou seems some stupid bug related to window hide/show etc

wodin:
> How about Flatpak?

Same behaviour

fufesou:
[...]

PR #6006, opened 2023-10-12
Fix window hide sometimes

fufesou:
https://github.com/rustdesk/rustdesk/issues/5986

The error: `handleUriLink()` returns `true`, which causes `hide()`.

But "async/await" usually causes the window to show up after `hide()`.


rustdesk:
@21pages please test this more on different OS.

PR #6021, opened 2023-10-13
added null check

kaushal1717:
Added null check on pos parameter

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -1701,7 +1701,7 @@ Future<bool> restoreWindowPosition(WindowType type,
   }
   pos ??= bind.getLocalFlutterOption(k: kWindowPrefix + type.name);
 
-  var lpos = LastWindowPosition.loadFromString(pos);
+  var lpos = LastWindowPosition.loadFromString(pos!);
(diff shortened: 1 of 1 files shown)


flutter/lib/common.dart
```diff
@@ -1701,7 +1701,7 @@ Future<bool> restoreWindowPosition(WindowType type,
   }
   pos ??= bind.getLocalFlutterOption(k: kWindowPrefix + type.name);
 
-  var lpos = LastWindowPosition.loadFromString(pos);
+  var lpos = LastWindowPosition.loadFromString(pos!);
```
rustdesk, line 1704: This is not safe, because exception will be thrown if pos is null.

kaushal1717: so can i make chenge the data type of pos from String to String?

rustdesk: My ide does not report error since `pos ??= bind.getLocalFlutterOption(k: kWindowPrefix + type.name);` making sure it is not null. Changing to `pos!` will cause error in my ide.

kaushal1717: I can use try catch on that

PR #5985, opened 2023-10-11
add option filter ab by intersection

21pages:
https://github.com/rustdesk/rustdesk/issues/5981
https://github.com/rustdesk/rustdesk/discussions/5984

default unchecked

https://github.com/rustdesk/rustdesk/assets/14891774/be3ce1b7-ed0a-45a8-8e3a-2b6f21080c45

PR #6024, opened 2023-10-13
Update de.rs

Mr-Update:


grummbeer:
@Mr-Update "Schnittmenge" would match better? 

sample https://github.com/rustdesk/rustdesk/pull/5985

Issue #486, opened 2022-05-06
Dual screen remote dual screen


chenjianhua:
I use a computer with two screens to remotely control a computer with two screens. It is hoped that two screens can be displayed at the same time. This is an awesome idea.

rustdesk:
Yes, it is awesome, but this involves a lot of code change, already in the to-do list many month ago.

techflashYT:
Any update on this?  Would love to use rustdesk as a software-based dual monitor solution over my LAN.

ForsakenRei:
I hope we can do this as well, currently I think we can have two connections to one PC(though in my case I need reopen rustdesk windows to start a new connection) but the problem is the two different sessions follows the remote focus, which means I cannot use the two screen on the remote side as both sessions will always display the same focusd screen.

rustdesk:
https://github.com/rustdesk/rustdesk/discussions/918

daTony16822:
@ForsakenRei 
How could you get 2 connections to same computer? I am trying but failing.

thank you

daTony16822:
@ForsakenRei 
I am getting the same problem. Did you have a chance to look into that?
Do you need help?

wizzyto12:
Is there any plans for implementing this soon? Its kind of a bummer for us...

oodonnell:
was looking to change to rustdesk after my teamviewer gave issues

this is stopping me from being able to, I usually sign in twice to the same desktop and use monitor 1 on one client and monitor 2 on the other.. but both connections will follow what i select on the other nullifying the objective

I can't continue to use this software while this isnt a functionality

Lordvader81:
This application works great for single monitor remote login sessions and while you really can't complain considering that it's free, however if the Devs did add in the ability to display two different screens at the same time somehow i'm sure there are plenty of people that would be willing to pay a small but fair amount ...1$ Per Month or somthing like that.
with emphasis on small & fair .... 
A reminder to the Dev team that this is the reason why millions and millions of users are users are moving away from application applications that were supposed to be free or have very low user fees such as anydesk ..... I understand that a lot of these companies moved in this direction because well they are companies and they need to make money However most of them are just plain greedy with their charge per user per PC mentality...

think about it .... there are billions of active PC users that require remote login capabilities especially now but times are changing with the workforce ... if you're charging an average of$1.00 a month with no other restrictions you're making a billion dollars a month just in subscription fees....... Why get greedy ??.. I mean it really makes no sense to me to be honest.

[...]

PR #6064, opened 2023-10-17
Feat/open multi windows

dignow:
Add an option "Use all my displays when connecting", to open multiple windows and set fullscreen.

> Note: It only shows remote available displays, we do not create virtual display like RDP for you.

https://github.com/rustdesk/rustdesk/assets/136106582/f6ab4881-ef21-4b00-b015-dcd3c537ab80


rustdesk:
https://github.com/rustdesk/rustdesk/issues/486

PR #6093, opened 2023-10-18
Fix remote home button in file transfer

### LINKED ISSUE #3508: File Manager. Dialog Confirm overwrite does not behave as expected.

beelchester:
fixes https://github.com/rustdesk/rustdesk/issues/3508#issuecomment-1758746599
The original issue of #3508 has been already fixed (confirmed by OP and me).

~~Storing and fetching remote home directory for remote home button.~~
Setting the directory as empty is opening home directory.
Rest implementation is unchanged. Local remote home button is working the same as before.

https://github.com/rustdesk/rustdesk/assets/73148455/15103ad5-a035-4ead-9773-bb163c2ce3db

### DIFF
--- a/flutter/lib/models/file_model.dart
+++ b/flutter/lib/models/file_model.dart
@@ -261,6 +261,7 @@ class FileController {
       required this.getOtherSideDirectoryData});
 
   String get homePath => options.value.home;
+  void set homePath(String path) => options.value.home = path;
   OverlayDialogManager? get dialogManager => rootState.target?.dialogManager;
 
   String get shortPath {
@@ -376,6 +377,11 @@ class FileController {
   }
 
   void goToHomeDirectory() {
+    if (isLocal) {
+      openDirectory(homePath);
+      return;
+    }
+    homePath = "";
     openDirectory(homePath);
   }
 
--- a/flutter/lib/models/file_model.dart
+++ b/flutter/lib/models/file_model.dart
@@ -375,14 +375,8 @@ class FileController {
     history.add(directory.value.path);
   }
 
-  void goToHomeDirectory() async {
-    if (isLocal) {
-      openDirectory(homePath);
-      return;
-    }
-    final homeDir = (await bind.sessionGetPeerOption(
-        sessionId: sessionId, name: "remote_home_dir"));
-        openDirectory(homeDir);
+  void goToHomeDirectory() {
+    openDirectory(homePath);
   }
 
   void goBack() {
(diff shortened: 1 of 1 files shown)


flutter/lib/models/file_model.dart
```diff
@@ -423,6 +429,14 @@ class FileController {
         }
       } else if (options.value.home.isEmpty) {
         options.value.home = fd.path;
+
+        final homeDir = ( await bind.sessionGetPeerOption(
+          sessionId: sessionId, name: "remote_home_dir"));
+
+        if (homeDir.isEmpty){
```
rustdesk: I guess you should run format.

flutter/lib/models/file_model.dart
```diff
@@ -423,6 +429,14 @@ class FileController {
...
+          bind.sessionPeerOption(
+            sessionId: sessionId, name: "remote_home_dir", value: fd.path);
+        }
```
rustdesk: Do we really need store it as peer option? it will write disk and pollute peer options.

> I could not find any direct implementation for the remote home directory yet but since the home path is correct in the first connection we can just store it according to the session ID and fetch it whenever needed.

beelchester: Ok, I will try to find a better way to get home directory. Or do you mean to store it somewhere else?

sahilyeole:
New changes:

- Removed the need for storing.
- Setting the directory as empty is opening home directory.
- Tested win, android, linux as remote side. Remote button is working the same as the video above.

PR #6094, opened 2023-10-18
open individual window, add offset

dignow:
https://github.com/rustdesk/rustdesk/assets/136106582/82cebf1b-35bf-43dc-981b-6408c217239f


rustdesk:
Follow OS convention please.

Issue #3556, opened 2023-03-08
physical keyboard desktop -> mobile


michalwitek1232:
### Bug Description

Hi, keyboard used by client does not send keys into remote device.

Also, to be sure, I tested every key mapping from this list:
![obraz](https://user-images.githubusercontent.com/42500203/223693470-50f90710-8528-450e-999a-94abca1177d6.png)

None worked

### How to Reproduce

When I'm pressing any keys on my keyboard, for ex. when trying to type: google.com, when I'm pressing `g o o g l e . c o m`, My android remote does not put those keys into a text field ( in my example, web browser search bar ).

### Expected Behavior

I should be able send my keyboard keys into remote with ease, so I could just type on my keyboard instead of remote virtual keyboard (ex. google keyboard, swiftkeys etc)

### Operating system(s) on local side and remote side

Windows 11 -> Android

### RustDesk Version(s) on local side and remote side

nightly -> nightly (52ce3dd)

### Screenshots

Permissions on remote:
![obraz](https://user-images.githubusercontent.com/42500203/223693003-7b39a340-bb3e-4c09-8720-837150f66c1e.png)

### Additional Context

Android device system info:
POCO F2 Pro, MIUI Global 13.0.3 (Stable)
Android 12 SKQ1.211006.001

Connection by self-hosted relay

rustdesk:
Yes, soft keyboard only for android host.
Close as wontfix

rustdesk:
We can reopen this, if some one can figure out a solution.

michalwitek1232:
btw, any idea why this is not working? I mean any clue?

michalwitek1232:
I read about something called [wifi keyboard](https://f-droid.org/zh_Hans/packages/com.volosyukivan/), maybe it's possible to integrate it with rustdesk. I'll try to do it. Wish me luck @rustdesk

rustdesk:
Interesting

rustdesk:
@Heap-Hop

DonRichie:
> Yes, soft keyboard only for android host. Close as wontfix

I am sorry, but I can't understand this answer. Is there a way to forward keyboard input to a controlled android device?
On my Android 13 device RustDesk is activated as accessibility app, but I can't enter keys, only mouse.
Do I miss something?

Linux -> Android 13

Zhurama:
[...]

PR #6097, opened 2023-10-19
Physical keyboard to android support

mcfans:
# Description
Support KeyEvent input in android side. Fix issue(https://github.com/rustdesk/rustdesk/issues/3556)
/claim #3556
# Detailed Design
~~Android doesn't provide an API for send key event to the whole system, since most scenario that needs keyboard input are text input. I used android accessibility service to find current focused text input widget and set text to it. I create an invisible text input widget in InputService, and set current focusing accessibility node info's text and selection to it, send the key event to it then retrieve it's new state and set the new text back to the current focusing node.In this way, we don't need to handle what different key will do to text on our side, just use system default implementation.~~
Use [getInputMethod()](https://developer.android.com/reference/android/accessibilityservice/AccessibilityService#getInputMethod) API, then send key event to it.
# Other Approach
Register an IME to system then use that to send key stroke to system. [WIFIKeyboard](https://github.com/IvanVolosyuk/wifikeyboard) use this approach. But Android system can only have one IME at a time, receive side will not able to use their on screen IME while using RustDesk, that would make features like chatting not work.
# Changes
- Modified build.gradle to generate java code from message.proto for usage of KeyEvent in Kotlin code. KeyEvent is a struct with many field, pass its' fields through ffi is complicated, so I pass serialized data through ffi.
- Updated [rdev](https://github.com/mcfans/rdev), add android key code, make keyboard mode(map) work.
- Added canRetrieveWindowContent to accessibility_service_config.xml
# Test
Tested on:
iOS -> Android,
macOS -> Android(map mode),
windows -> Android(legacy mode)
# Compatibility
Android receive side needs update to this version to work.
iOS control side needs update to this version to show a soft keyboard.
Desktop version doesn't need update for legacy keyboard mode, need update for map keyboard mode.
# A Video
https://github.com/rustdesk/rustdesk/assets/15027395/9464a26f-dfb0-40be-a849-698dce4f1b71


Cargo.toml
```diff
@@ -64,7 +64,7 @@ default-net = "0.14"
 wol-rs = "1.0"
 flutter_rust_bridge = { version = "1.75", features = ["uuid"], optional = true}
 errno = "0.3"
-rdev = { git = "https://github.com/fufesou/rdev" }
+rdev = { git = "https://github.com/mcfans/rdev" }
```
rustdesk: Please merge to https://github.com/fufesou/rdev
@fufesou  please review.

flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
```diff
@@ -252,9 +261,88 @@ class InputService : AccessibilityService() {
         }
     }
 
+    @RequiresApi(Build.VERSION_CODES.N)
+    fun onKeyEvent(data: ByteArray) {
+        val keyEvent = KeyEvent.parseFrom(data);
+        val handler = Handler(Looper.getMainLooper())
```
rustdesk: can not parse it in rust?

rustdesk:
@fufesou please have more review.

flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
mcfans: It's serialized on rust side by purpose, because I want to use generated KeyEvent class on Kotlin side instead of pass the  fields of it on rust side through ffi that would be a lot of parameters and less convenient.

Cargo.toml
mcfans: I will open a PR for that.

rustdesk:
Have you tested different language? e.g. German, French.

flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
rustdesk: Agree

mcfans:
> Have you tested different languages? e.g. German, French. And combination key, e.g. alt + key, will generate special character for different language.

Not yet. I will try it later.

fufesou:
Win -> Android , en->en, the charaters are all correct for both "Map mode" and "Translate mode".

https://github.com/rustdesk/rustdesk/assets/13586388/052514b6-58c3-429c-a664-5ef4cef13ee6


## needs improvement

1. Input events will have a side effect that will set the cursor to the end. So the arrow control doesn't work, the cursor always moves to the end.
2. The input does not work on the  input box on the web page. Tested chrome browser.

image

3. Mouse click event issue. The specified position of the text cannot be clicked and range selection cannot be made. This has nothing to do with keyboard input, but I always want to click the mouse to a specified location.

## avc permission issues

I'm not familiar with avc issues in android. 

They're not related to the keyboard issue. But if you're familiar with it, it's good.

image

mcfans:
I found a new way to do this. Accessibility service has a method called getInputMethod, we could then send key event directly to it.This could solve previous problem.
> Input events will have a side effect that will set the cursor to the end. So the arrow control doesn't work, the cursor always moves to the end.
Old way can only set text and text selection separately, if application doesn't do some special text handling, this is fine.But if app would change text after text set, then our text selection data would be wrong. So I ignored text selection data for correctness. In this new way I leave this to system to handle, arrow key works fine.
> The input does not work on the input box on the web page. Tested chrome browser.
This is because findFocus gave a wrong input widget, send event to IME also fix this.

mcfans:
> Have you tested different languages? e.g. German, French. And combination key, e.g. alt + key, will generate special character for different language.

I tested combination key, I could enter exclamation mark(shift + 1). I couldn't test different languages for now because I found out my android device doesn't support other languages. Need some help on that.

flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
```diff
@@ -252,9 +261,27 @@ class InputService : AccessibilityService() {
         }
     }
 
+    @RequiresApi(Build.VERSION_CODES.N)
+    fun onKeyEvent(data: ByteArray) {
+        val keyEvent = KeyEvent.parseFrom(data);
+        getInputMethod()?.let { inputMethod ->
```
YangCheung: getInputMethod added in [API level 33](https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels)

rustdesk: Good point

rustdesk: You may have some modification to make it compatible with lower level target device. @mcfans

mcfans: I checked the wrong field, I thought we were using API 33. My bad. So I just keep this for higher version and add the old code back for lower API?

rustdesk: I am not familiar with kotlin about how to call higher API on high target device but still can run on low target device.

fufesou: My android device may be the lower version. The input does not work now, and it reports:
```
D/ffi     ( 3011): librustdesk::server::connection: call_main_service_key_event fail:Java exception was thrown
```

mcfans: Could you try it again? I added old code back with a version check.

fufesou: The keyboard input works now.

fufesou:
> Have you tested different languages? e.g. German, French. And combination key, e.g. alt + key, will generate special character for different language.

The "Translate mode" works fine.

For the "Map mode" (fr -> fr), the the characters are different. It seems the default en-use keyboard layout is used.

image

Maybe it's better hide the "Legacy mode" and "Map mode" if local is desktop and remote is Android.

rustdesk:
> "Map mode" means using the "position code" (scancode) to control the key inputs.

👍

fufesou:
@mcfans 👍 

Do you have plan to implement the "Map mode"?
It's an important keyboard mode.

mcfans:
I am trying to implement Map mode these days.But there are two problems. First we can only send a keycode that matches US keyboard layout, Android keycode only has those keys. Set scancode of a keyevent is ineffective base on my experiments. Second I only found ways to send key events to application, not send to input method. On Android a hardware keyboard event is sent to input method, then the input method would figure out correct text to insert into application. I failed to find a way to do the same thing. However I have not tested what text would be inserted if the OS language matches input method, because my android device couldn't do that. If that could work, we could add map mode back.

mcfans:
I fixed translated mode and legacy mode in those commits.

fufesou:
> what text would be inserted if the OS language matches input method

The scancode should be handled by the input method. It should have nothing to do with the OS language.

> I only found ways to send key events to application, not send to input method.

I'm not familiar with Android. But it seems you are right, the event should be sent to the system or input method.

mcfans:
If we can send key event to the system, everything would be perfect.

mcfans:
So should we merge this?

rustdesk:
AnyDesk's map mode work?

mcfans:
I tested it. It doesn't work correctly with 1:1 mode on and local input method is fr, its output is US layout keys. Same behavior with my implementation.

rustdesk:
Our default mode is translation or legacy?

Cargo.toml
rustdesk: Finish this first

rustdesk:
> Same behavior

What behavior?

mcfans:
I disabled map mode if peer is android. But is there some system disabled other two? If so, that might be a issue.

mcfans:
> What behavior?

Output is always US layout keycode represented text, Android input method keyboard layout has been ignored.

Cargo.toml
mcfans: @fufesou Could you take a look at [this](https://github.com/fufesou/rdev/pull/9)?

rustdesk:
> > What behavior?
> 
> Output is always US layout keycode represented text, Android input method keyboard layout has been ignored.

I think we need to have more test. I do not think AnyDesk will allow a useless mode there.

mcfans:
Could you be more specific about more tests? I've already tested windows(fr input method) -> android(fr input method). I won't say it's useless because English input method users could use this mode just fine.

rustdesk:
> Could you be more specific about more tests? I've already tested windows(fr input method) -> android(fr input method). I won't say it's useless because English input method users could use this mode just fine.

e.g. a Android device with FR system language connected with a French external keyboard, in this case can AnyDesk work as the same as the French external keyboard (not depend on soft French input method)?

mcfans:
> e.g. a Android device with FR system language connected with a French external keyboard, in this case can AnyDesk (1:1 map mode) work as the same as the French external keyboard (not depend on software French input method)?
Sorry I am not able to test it.I don’t have a French layout keyboard, even my android device is borrowed from my friend.

rustdesk:
You can use simulator.

mcfans:
I tested it. I am typing "qwe" the whole time.
Results：
Win(En)->Android(physical keyboard layout fr)
Typing in simulator directly: "aze"
Typing in AnyDesk(1:1): "qwe"
Typing in AnyDesk(translate): "aze"
Expected: "aze"

Win(Fr)->Android(physical keyboard layout fr)
Typing in simulator directly: "qwe"
Typing in AnyDesk(1:1): "qwe"
Typing in AnyDesk(translate): "aze"

I don't think we could trust simulator for such a test, but AnyDesk map mode never works like a french layout keyboard.

https://github.com/rustdesk/rustdesk/assets/15027395/2ef3bb61-5971-4c55-bdb4-099c1b74b6d3

rustdesk:
Good, let's disable map mode for x->android first.

fufesou:
@mcfans The keyboard mode is still "Map mode", if the previous keyboard mode is "Map mode".

Maybe we can check and set current keyboard mode to a supported one, if the previous keyboard mode is not supported, when the connection is established. (Handling the peer info on the flutter side.)

The check and changing may also be removed, because it is triggered when choosing the menu.

https://github.com/rustdesk/rustdesk/blob/69f643447d702dfbead012c45870df671e55729a/flutter/lib/desktop/widgets/remote_toolbar.dart#L1410

The "Translate mode" is better than the "Legacy mode".

peer_android_1

fufesou:
> The keyboard mode is still "Map mode", if the previous keyboard mode is "Map mode".

@rustdesk Or you can merge this PR first, and then fix it.

fufesou:
@mcfans Hi, can you please try the following test? Can they be implemented in RustDesk?
It seems there's a way to simulate the "scancode".

AnyDesk has a good experience of desktop -> mobile (simulator).

1. Mouse can click, change the cursor positions. (The **Mouse**, not the ***Keyboard**)
2. Keys of "Arrows" "Home/End" "PageUp/PageDown" work fine.
3. Shift + "Arrow" work fine.
4. Ctrl + A work fine.
5. Ctrl + C/V work fine. The reason for the garbled code is that there is a problem with my copy and paste on the mobile side.

https://github.com/rustdesk/rustdesk/assets/13586388/98629786-ad95-4f01-8a70-bd1040ccd1ad

mcfans:
Which API version is it? The cursor and arrow key also works when targeting API 33 in RustDesk.

fufesou:
I've tested RustDesk in the simulator.

The cursor works fine.
But I cannot input letters.

fufesou:
@mcfans Hi, I've tried Android 14 with API 34, Android 11 with API 30.

The cursor can be positioned at the location of the mouse click.

I can confirm that the keys can not be input on the remote side.

Moreover, the mouse cannot click after key input.

image

mcfans:
Thanks for the information, I’m investigating this.

rustdesk:
@algora-pbc why is not this labeled as `bounty-claim` automatically?

algora-pbc:
@rustdesk sorry about that. This was the first time a claim was submitted for a bounty pool (as opposed to a singular bounty) and we hit an edge case. Fixed it now!

mcfans:
@fufesou Hi!
I fixed the issue of not being able to input characters. It's a issue around findFocus method. Please take a look.
About the ctrl-a shortcut, It seems It's not an Android system shortcut, It's done by AnyDesk itself. I tried that shortcut directly in simulator, nothing happened. I don't know if we should do the same thing, implementing a shortcut collection that system doesn't have.
The cursor is a tricky problem, because we can't set text and selection at the same time, application might change text after we set text, if they did, the cursor position would be wrong. AnyDesk has the same problem, you could reproduce this bug by texting in RustDesk address field while using AnyDesk. The cursor would be in middle of the text not the end, you will need use arrow key to navigate to the end every time you input a letter. I think this is more annoying than you can't move the cursor with arrow key. So I choose to not update cursor location.
The mouse issue is another problem, sometimes it works. I don't know what's going on right now, but we could track it in a new issue?

fufesou:
> About the ctrl-a shortcut, It seems It's not an Android system shortcut, It's done by AnyDesk itself. I tried that shortcut directly in simulator, nothing happened. I don't know if we should do the same thing, implementing a shortcut collection that system doesn't have.

I think the `Ctrl+` shortcuts are supported by Android. The emulator does not handle the `Ctrl +` shortcut correctly.

I've tried to connect a physical mouse and keyboard to Android. The shortcut experience is very good.

> The cursor is a tricky problem, because we can't set text and selection at the same time, application might change text after we set text, if they did, the cursor position would be wrong. AnyDesk has the same problem, you could reproduce this bug by texting in RustDesk address field while using AnyDesk. 

I'll try, but the address field is a one line editor.

Maybe you can also try a notepad app?


> 2. Keys of "Arrows" "Home/End" "PageUp/PageDown" work fine.
> 3. Shift + "Arrow" work fine.

Not only the "Arrow" keys, but also the other keys work fine. So I think AnyDesk can handle these non-letter keys well.

mcfans:
> Not only the "Arrow" keys, but also the other keys work fine. So I think AnyDesk can handle these non-letter keys well.
Yeah, the main problem here is the cursor.

fufesou:
I've tested the last commit.

The key input is still not working on Android API 34.

mcfans:
It works for me. My emulator hardware profile is 

image

Which emulator profile are you using?

fufesou:
Here's mine

image

> The cursor is a tricky problem, because we can't set text and selection at the same time, application might change text after we set text, if they did, the cursor position would be wrong. AnyDesk has the same problem, you could reproduce this bug by texting in RustDesk address field while using AnyDesk.

I've tested, AnyDesk works fine for me. left/right move the cursor to left/right, up/down move the cursor to the begin/end.

https://github.com/rustdesk/rustdesk/assets/13586388/7d256d17-54e6-40f6-890c-c67945db0aa8

mcfans:
Sorry for the confusion, the bug occurs when continuously entering characters, especially after RustDesk inserted space between numbers, the cursor would be in the middle of the text, not the end, so the order of your input would be wrong. Can you reproduce that?

fufesou:
It is also ok to insert space.

Maybe you need a notepad app on Android for test.

https://github.com/rustdesk/rustdesk/assets/13586388/e2574554-61ad-4c8f-83f7-25c407047447

mcfans:
I recorded a video to demonstrate this issue. I am typing 1 continuously, and the cursor is moving around, I don't touch anything other than 1. I am using AnyDesk. 

https://github.com/rustdesk/rustdesk/assets/15027395/6899dcb2-ba96-4c81-b86a-f74ce5561954

mcfans:
@fufesou 
It also works on Pixel 2 API 34 arm64. I didn't managed to make it run on x86_64 emulator. It fails to load librustdesk.so. How did you do it? Did you compile RustDesk to x86?

https://github.com/rustdesk/rustdesk/assets/15027395/7230ed4e-3a17-49c3-a3ae-370fca22a610

fufesou:
> It fails to load librustdesk.so. How did you do it? Did you compile RustDesk to x86?

I also did not handle x86.

> I never changed the cursor in this video, It should be at the end of text all the time.

I've also reproduced it. I can see that the cursor is moving around.

fufesou:
> It fails to load librustdesk.so. How did you do it? Did you compile RustDesk to x86?

I have met this error. But I forgot how I solved it. Maybe you can run `flutter clean`, and then rebuild.

> I never changed the cursor in this video, It should be at the end of text all the time.

I've also reproduced it. I can see that the cursor is moving around.

mcfans:
> It fails to load librustdesk.so. How did you do it? Did you compile RustDesk to x86?

I am now able to run RustDesk on x86 emulator. I use `--split-per-abi` when building apk and then use arm64 version apk. But I failed to reproduce the issue, text input works perfectly for me on Android API 34 Pixel 2 x86_64 emulator.

> I've also reproduced it. I can see that the cursor is moving around.

Do you agree that we should not mess up with the cursor? Or should we do the same thing with AnyDesk?

fufesou:
The cursor should be important as we assume that the user on local side uses physical keyboard/mouse.

I'll test agian later.

mcfans:
> The cursor should be important as we assume that the user on local side uses physical keyboard/mouse.

Okay I will implement that for API level lower than 33, but it would have the same bug with AnyDesk.

fufesou:
> I've tested the last commit.
> 
> The key input is still not working on Android API 34.


Sorry what I tested is incorrect.
I installed the aarch64 installation package on the X86_64 emulator to test.

But my x86_64 always gets stuck at startup.

The aarch64 app runs on the emulator, and it seems normal.
I only have this problem when typing, the key inputs have no effect, and the mouse clicks will no longer have any effect after the key inputs.

mcfans:
Do you have a soft keyboard showing up?

fufesou:
Yes. The emulator is x86_64 while the apk is aarch64 in the following video


https://github.com/rustdesk/rustdesk/assets/13586388/0da5ba57-73d4-45f3-b0aa-5d90633c41da

mcfans:
Which note app are you using? Does it work for other app?

fufesou:
I've tried the ID edit in RustDesk.
The key input also does not work.

src/ui_session_interface.rs
```diff
@@ -216,6 +216,11 @@ impl<T: InvokeUiSession> Session<T> {
 
     pub fn get_keyboard_mode(&self) -> String {
         let mode = self.lc.read().unwrap().keyboard_mode.clone();
+        if self.peer_platform() == crate::PLATFORM_ANDROID {
```
fufesou: 👍 

This commit seems better than
https://github.com/rustdesk/rustdesk/pull/6168/

It's independent of the UI.

fufesou: But the logic here is the same as in flutter_ffi.rs. Maybe it's better to merge the code.
```rust
if session.peer_platform() == "Android" && mode == KeyboardMode::Map {
    return SyncReturn(false);
}
```

mcfans: The logic in flutter_ffi.rs is used to hide map mode in UI, ui_session_interface.rs is used to override keyboard mode saved in local storage. It’s two part of the same problem.

fufesou: > It’s two part of the same problem.

Yes, but if something changed, you have to change to two places, which is error prone.

They are doing almost the same filtering, maybe you can extract the filter code, or some func like `should_replace_mode_to(conds, mode) -> Option<mode>`

mcfans: I agree, that's better. Working on it.

mcfans:
I made the cursor works for API level lower than 33 just like AnyDesk.

src/ui_session_interface.rs
mcfans: Done.

fufesou:
@mcfans Hi, what architecture did you use for testing? It's strange that I can't run x86_64 on my emulator.

fufesou:
With the latest commit, I can't enter **"Backspace" "Enter"** "Left/Up/Right/Down".

mcfans:
> With the latest commit, I can't enter **"Backspace" "Enter"** "Left/Up/Right/Down".

Which API level are you using?

rustdesk:
https://github.com/rustdesk/rustdesk/commit/c2703d215b5451f9540a6088fb5fe8e8cd9917d1
flutter and bridge updated to new versions.
If you can not run on mac, please rm -rf macos/rustdesk.xcodeproj, pub get/upgrade,

rustdesk:
@mcfans please fix the conflicts. I will merge then.

src/server/connection.rs
```diff
@@ -1763,8 +1765,56 @@ impl Connection {
...
+
+                    let encode_result = me.write_to_bytes();
+
+                    if let Ok(data) = encode_result {
+                        let result = call_main_service_key_event(&data);
+                        if let Err(e) = result {
+                            log::debug!("call_main_service_key_event fail:{}", e);
+                        }
+                    } else {
+                        log::debug!("encode key event fail:{}", encode_result.err().unwrap());
```
rustdesk: Can we remove this `unwrap`? BTW, space between `:` and `{`.

mcfans: It's safe here, but I changed it to a `match` block anyway.

basilgello:
Thanks for so awaited feature! :)

rustdesk:
https://github.com/rustdesk/rustdesk/discussions/6417 big regression failure.

Magustructor:
I don't know if here's the right place for reporting it but I have a problem with using keyboard in Windows 10 (R. ver.1.2.4)-> Android 11 (R. ver. 1.2.4) scenario.
The keyboard generally works properly except Enter key.
The Enter moves the cursor to the next line instead of confirming/accepting text when used in any search field like google.
I tried both legacy and translation modes.

Am I doing something incorrectly or it's a bug?

rustdesk:
@mcfans wondering if you have interest in this issue, https://github.com/rustdesk/rustdesk/issues/7439, 900 usd, frankly, this is a simple task.

rustdesk:
In 1.2.4, current night build, https://github.com/rustdesk/rustdesk/releases/tag/nightly

Issue #4597, opened 2023-06-09
RustDeskIddDriver not installed automatically （failed to install driver sometimes)


davidebeatrici:
### Bug Description

Problem:

The remote setup consists in a Windows 11 virtual machine with an Nvidia Tesla M40 passed through it.

If I leave the SPICE interface with the corresponding virtual GPU enabled, the remote desktop experience is excellent, aside from the lack of 3D acceleration.

However, if I leave only the dedicated GPU the entire RustDesk session becomes extremely slow, to the point that the remote system seems frozen. It even takes a very long time to connect and sometimes goes into timeout or throws an error.

---

Solution:

Setting the SPICE display resolution to 640x480 (in order for the virtual display adapter to show up) and installing the corresponding driver in Device Manager fixes the issue.

### How to Reproduce

Install RustDesk when at least one display is present and with a resolution equal or higher to 800x600 (#4560).

### Expected Behavior

The driver should be installed automatically, so that Windows finds it when the virtual display adapter shows up.

### Operating system(s) on local side and remote side

Any -> Windows 11

### RustDesk Version(s) on local side and remote side

1.2.0 -> 1.2.0

### Screenshots

![](https://github.com/rustdesk/rustdesk/assets/5897523/8562a9ba-5878-48c0-9c70-a5350808ffa3)

rustdesk:
no idea what you are saying.

rustdesk:
what is `automatically`?

davidebeatrici:
The driver should be installed along with RustDesk. It can be done even if the device is not available yet.

rustdesk:
idddriver is installed with RustDesk, but if starting the virtual display depends on https://github.com/rustdesk/rustdesk/issues/4359#issuecomment-1582338893

davidebeatrici:
In my case it was not installed... I had to do it myself in Device Manager. The device was there but had the yellow exclamation mark next to it due to no driver present.

rustdesk:
we only install the cert, the driver will be installed when starting virtual display.

@fufesou  please correct me.

davidebeatrici:
I can confirm the cert was installed as the digital signature is recognized.

rustdesk:
We have no plan to install it automatically for the time being. But we will consider this together with https://github.com/rustdesk/rustdesk/issues/4359#issuecomment-1582338893 together in the future.

davidebeatrici:
Alright, no problem.

rustdesk:
[...]

PR #6156, opened 2023-10-24
fix, win, install cert

dignow:
1. Use `CertOpenStore()` instead of `CertOpenSystemStoreW()`.
2. Create a certificate context by `CertCreateCertificateContext()` and then install the cert.
3. Fix potential crash when uninstalling cert. `CertDeleteCertificateFromStore()` should not be called after cert context enumeration.

Registry operations are left as an insurance.

This PR may fix the exclamation issue. But I haven't reproduced it, so I can't confirm.

https://github.com/rustdesk/rustdesk/issues/4597#issuecomment-1616396696

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -1839,20 +1839,19 @@ pub fn uninstall_cert() -> ResultType<()> {
 
 mod cert {
     use hbb_common::{allow_err, bail, log, ResultType};
-    use std::{path::Path, str::from_utf8};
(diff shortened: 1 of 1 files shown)


rustdesk:
Confirm with guys who have problems

dignow:
They cannot test now. The virtual display is removed.
I'm trying to find a better way to detect the dummy display.

Maybe we can test it later, when the virtual display is added back.

PR #6185, opened 2023-10-26
fix close on taskbar

21pages:
old implement just hide it without closing the connection when closed on taskbar, for this pr when closed on the taskbar:
* Windows behaves as if clicked on the close button in the top right. 
* Linux simply close all tabs since it lacks the ability to top from taskbar. 
* MacOS can't close applications directly from the taskbar

tested remote/file transfer/port forward

https://github.com/rustdesk/rustdesk/assets/14891774/cf26f6d8-7492-4ba5-a4fc-bbb55e9c623d

https://github.com/rustdesk/rustdesk/assets/14891774/5663620a-b5f3-45d0-8da8-30d9409d58fe


rustdesk:
👍

PR #6193, opened 2023-10-27
fix, flutter, infinite loop

fufesou:
`final model = Provider.of<PeerTabModel>(context);` gets the model instance of `gFFI.peerTabModel`.

It also has a side effect of rebuild the widget.

The previous call of `gFFI.peerTabModel.setCurrentTabCachedPeers(peers);` will `notifyListeners()`.

Then infinite loop is build.

BTW, using `gFFI.peerTabModel` and `Provider.of<PeerTabModel>(context)` at the same time can be confusing. This place is prone to bugs.


fufesou:
This PR is incorrect. `final model = Provider.of<PeerTabModel>(context);` should not be replaced.
Maybe `gFFI.peerTabModel.setCurrentTabCachedPeers(peers);` need to be changed.

Issue #6198, opened 2023-08-22
[FEATURE] - Connect even if device are not on same server


Apashh:
Hey,

It will be interesting to connect to a public Rustdesk ID not hosting it on our servers while remaining connected to our hosted server.

Thanks for reply !

rustdesk:
https://github.com/rustdesk/rustdesk/issues/2118

rustdesk:
@sahilyeole Let's support `id@host?key=<value>` in the id input box.

rustdesk:
@sahilyeole Let's support `id@host?key=<value>` in the id input box.

rustdesk:
@sahilyeole add some tooltip there to tell users how to use

mycanaletto:
This is a good idea.

This will enable an isolated customer to install Rustesk without having to be given the key to connect to it the first time.

This raises the problem of the key. If the customer can copy the key, he can use our server for other devices without our knowledge.

We'd need to find a way of making it impossible to copy the key, or of making this possible only if the Rustdesk user is logged in. Otherwise, the key could spread and my server become public...

pedjas:
> We will address this issue in next Pro version.

This explains everything. We were just teased. FInal goal was to make commercial product an sell it to us.

Well, RustDesk is fine product, but while OpenSource. As commercial product it is far behind competition.

mycanaletto:
I bought a license that supports 500 devives. The group is not very small....

Perhaps it should be possible to define several keys and limit them to one group?

rustdesk:
> We'd need to find a way of making it impossible to copy the key, or of making this possible only if the Rustdesk user is logged in. Otherwise, the key could spread and my server become public...

This is pretty a good idea, we will think about this.

pdifeo:
I agree but only if the user can logout and the key  is still configured on
the client.

Il ven 27 ott 2023, 19:33 RustDesk ***@***.***> ha scritto:

pdifeo:
> I agree but only if the user can logout and the key is still configured on the client.

Another solution is protect configuration with an additional passwork, like TeamViewer with "Option Password"

<img width="369" alt="image" src="https://github.com/rustdesk/rustdesk/assets/8618744/60076adb-efc3-47ba-b510-5a145498f5b9">

rustdesk:
![image](https://github.com/rustdesk/rustdesk/assets/71636191/cd5e2959-7481-4309-9f1d-54b3027fab54)

[...]

PR #6206, opened 2023-10-28
Feat connect to peer on another server

beelchester:
**Still working**

https://github.com/rustdesk/rustdesk/issues/6198#issuecomment-1782696098
Currently, I'm able to connect a peer on the public server while the control side is connected to a private server.
Support for public -> private in progress.

Ex: 143242314@164.432.413: 21116?key=32432432Efdsf34 or 143242314@public

https://github.com/rustdesk/rustdesk/assets/73148455/82ee7e63-147a-4e98-aa18-2f993d20fc06

### DIFF
--- a/flutter/lib/desktop/pages/connection_page.dart
+++ b/flutter/lib/desktop/pages/connection_page.dart
@@ -147,6 +147,31 @@ class _ConnectionPageState extends State<ConnectionPage>
   /// Connects to the selected peer.
   void onConnect({bool isFileTransfer = false}) {
     var id = _idController.id;
+    if (id.contains('@')) {
+      var input = id;
+      id = id.split('@')[0];
+
+      // Ex: 143242314@164.432.413?key=32432432Efdsf34 or 143242314@public
+      RegExp regex = RegExp(r'(\d+)@([^:]+)(?::(\d+))?(?:\?key=(\d+))?');
+      Match? match = regex.firstMatch(input);
+
+      if (match != null) {
+        // String id = match.group(1)!;
+        String host = match.group(2)!;
+        String? port = match.group(3);
+        String? key = match.group(4);
+
+        if (host == "public"){
+          host = "rs-ny.rustdesk.com";
+          port = "21116";
+        }
+        var sValue = host + ":" + port!;
+        debugPrint('text_field_rendezvous_server: $sValue key: $key');
+        bind.mainSetOption(key: "text_field_rendezvous_server", value: sValue);
+      } else {
+        debugPrint('regex match failed');
+      }
+    }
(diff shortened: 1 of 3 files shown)


sahilyeole:
@rustdesk How can I know which public server the peer is connected to?
Currently, I have hardcoded it just for testing.
Or else I can loop and try every public server we have when user provides `@public`.

---
Also with this implementation, I'm getting the unencrypted connection. How can I encrypt it?

Screenshot_20231028_053044

rustdesk:
that is you need to figure out

rustdesk:
> public

You need to take care of this case https://rustdesk.com/docs/en/self-host/client-configuration/hardcode-settings/

rustdesk:
No need to take care of multiple public servers, just use the first one, in the future we will use only one.

sahilyeole:
> No need to take care of multiple public servers, just use the first one, in the future we will use only one.

Ok so I'm currently using `rs-ny.rustdesk.com:21116` server for `@public`. Is that enough?

sahilyeole:
I have added support to get the private server key from the text field but I'm getting this error on connecting:

Screenshot 2023-10-29 at 5 15 03 AM

as connection to private servers requires login credentials of the user allowed on that server.

So I can try this text field expression:
Ex: `143242314@164.432.413:21116?key="32432432Efdsf34"?username="admin"?password="admin1234"`
and authenticate the user for that server only for this particular connection.

Also, according to this, I think adding user credentials for the text field expression might be good. https://github.com/rustdesk/rustdesk/issues/6198#issuecomment-1782808127
>We'd need to find a way of making it impossible to copy the key, or of making this possible only if the Rustdesk user is logged in. Otherwise, the key could spread and my server become public...

Should we try this?

rustdesk:
> > No need to take care of multiple public servers, just use the first one, in the future we will use only one.
> 
> Ok so I'm currently using `rs-ny.rustdesk.com:21116` server for `@public`. Is that enough?

Can not be hard coded, it may change in the future.

rustdesk:
> Should we try this?

No

rustdesk:
> as connection to private servers requires login credentials of the user allowed on that server.

Expected, no need to consider for now.

rustdesk:
Also make sure `143242314@164.432.413: 21116?key=32432432Efdsf34` works for uri, e.g. `rustdesk://143242314@164.432.413: 21116?key=32432432Efdsf34`.

sahilyeole:
- Used RENDEZVOUS_SERVERS to get public server
- URL connection done on macos. Windows & linux yet to check
- Private -> public still working on the encryption part

rustdesk:
I will take over this issue.

PR #6257, opened 2023-11-01
Fixes for Android (ndk r26b)

basilgello:
- Refactor build_android_deps.sh
- Fix build on android-x86 and rename function

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -41,7 +41,7 @@ fn build_manifest() {
     }
 }
 
-fn install_oboe() {
+fn install_android_deps() {
     let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
     if target_os != "android" {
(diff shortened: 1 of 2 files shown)


basilgello:
These are the patches I added to fdroid build. There are also two other issues I'd like to have fixed:

* The versionCode in android gradle does not increment with time. I worked around by setting external CI project https://gitlab.com/basilgello/rustdesk-fdroid-versioner but properly it should be solved by adding the step to on-tag CI pipeline to generate rustdesk-version.txt inside build tree or in release.

* I added android-x86 and android-x64 builds into f-droid because emulators can use only them. Flutter, however, supports building x86 only with fpjit_release flutter engine which I build from sources. I can set separate CI task for flutter engine build if you are OK adding these architectures.

PR #6260, opened 2023-11-01
fix build

fufesou:


basilgello:
@fufesou build fails because https://github.com/Kingtous/rustdesk_thirdparty_lib lack ndk_compat from vcpkg package cpu-features that I added today to address crashes on android arm64 emulators. In flutter/build_android_deps.sh I added that dependency but the @kingtous repo does not use that script.

rustdesk:
https://github.com/rustdesk/rustdesk/commit/9096c29fef4f01bde555102873d75917a3f5a5ee

rustdesk:
> I'd better run flutter/build_android_deps.sh in CI rather than use third-party repo

Agree, I do not like third-party repo very much.

basilgello:
Oh, reverting my script fixes is not the outcome I expected! We need to merge it back after I revise dependencies from third-party repo.

basilgello:
@rustdesk I checked the thirdparty repo, vcpkg and upstreams of RD deps. I will update aom to 3.7.0, oboe to 1.8.0 and libvpx to 1.13.0. Also I would like to move all dependency builds into vcpkg custom registry and utilize Github Actions binary caching feature. Can you please create new empty repo https://github.com/rustdesk/rustdesk-vcpkg-repo so I can pull request changes there?

rustdesk:
Why do you need a new repo? can not do everyting on this repo?

basilgello:
I can try :) vcpkg documentation recommends moving custom deps into separate git registry: https://learn.microsoft.com/en-us/vcpkg/maintainers/registries but there is also local overlays available: https://learn.microsoft.com/en-us/vcpkg/consume/install-locally-modified-package?pivots=shell-bash

Let me try first with local overlays and if that fails, we'll go recommended way.

rustdesk:
Thanks

PR #6165, opened 2023-10-24
Feat single peer per row/list view

beelchester:
https://github.com/rustdesk/rustdesk/discussions/4731
New peer view option: 'List'
Updated other view type names to 'Big Tiles' & 'Small Tiles'

(old video)

https://github.com/rustdesk/rustdesk/assets/73148455/c9fd3aaf-9746-41fd-a014-a7f9cd7cd25c

(updated video)

https://github.com/rustdesk/rustdesk/assets/73148455/a9f223fd-33e6-43ec-87f6-1a2d49b513f8

In case of long alias name:
Screenshot 2023-10-24 at 11 27 23 PM

PR #6279, opened 2023-11-03
Fix: change view dropdown not working on macos clicking

beelchester:
Fixes: https://github.com/rustdesk/rustdesk/pull/6165#issuecomment-1791516777

Trackpad soft touch + click both working now.

### DIFF
--- a/flutter/lib/common/widgets/peer_tab_page.dart
+++ b/flutter/lib/common/widgets/peer_tab_page.dart
@@ -824,13 +824,14 @@ class _PeerViewDropdownState extends State<PeerViewDropdown> {
         setState(() {
           menuPos = RelativeRect.fromLTRB(x, y, x, y);
         });
+        showMenu(
+          context: context,
+          position: menuPos,
+          items: items,
+          elevation: 8,
+        );
       },
-      onTap: () => showMenu(
-        context: context,
-        position: menuPos,
-        items: items,
-        elevation: 8,
-      ),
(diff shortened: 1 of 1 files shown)


flutter/lib/common/widgets/peer_tab_page.dart
```diff
@@ -824,13 +824,14 @@ class _PeerViewDropdownState extends State<PeerViewDropdown> {
...
+          elevation: 8,
+        );
       },
-      onTap: () => showMenu(
-        context: context,
-        position: menuPos,
-        items: items,
-        elevation: 8,
-      ),
+      onTap: () => {}
```
rustdesk: This part of code looks stupid.

rustdesk: why setState here?

rustdesk: Why is different from `sort by`?

rustdesk: Why do not share the code with `sort by`?

`copy & paste` wrong code is rediculous.

beelchester: Ok, sorry. I used setState because the menu position was not updating (on `flutter run` linux. Sort by menu also sometimes give incorrect position on `flutter run` linux). But now I have removed it and made it work same as `Sort by`.

macos click is also working.

PR #6310, opened 2023-11-07
fix install after connection

21pages:
1. Fix install after connection.  If there is a multiwindow which sets prevent close, std::process::exit() will not work, ipc is occupied but can't be connected. Close all sub windows before install. https://github.com/rustdesk/rustdesk/blob/c69f4ba4a2fa28ae2fb60739db9f39ea3ad384dd/src/ui_interface.rs#L92 

https://github.com/rustdesk/rustdesk/assets/14891774/7baad843-d4a1-470f-95b2-090183f7934f

https://github.com/rustdesk/rustdesk/assets/14891774/d39f8992-7074-4be7-ae26-667d3eeeb053

2. Fix issue with missing logs when a deleted directory name contains another directory name and they are in the same directory (e.g. /a/b and /a/bc)


rustdesk:
👍

Issue #6298, opened 2023-11-06
Main Window Tabs missing


kb9gxk:
### Bug Description

Can't switch between settings and main window.  The tab bar is missing on Windows.

### How to Reproduce

Go to settings

### Expected Behavior

Having a tab bar to switch tabs

### Operating system(s) on local side and remote side

Windows 11

### RustDesk Version(s) on local side and remote side

1.2.4

### Screenshots

![image](https://github.com/rustdesk/rustdesk/assets/59369311/5d05e754-abe0-47d9-b167-b041ec4fa680)

rustdesk:
@fufesou same prolem related to flutter 3.13.

rustdesk:
@kb9gxk you see this every time?

kb9gxk:
90% of the time.  It happens at random times when the window is open.  In order to reset it, I have to use task manager to close it.  No issue on remote windows though.

fufesou:
Can you please upload a video to show how it is reproduced? Thx

I can hardly reproduce it.

rustdesk:
0% for me, you got the good luck. :(

PR #6312, opened 2023-11-07
fix, remote tab menu

fufesou:
https://github.com/rustdesk/rustdesk/issues/6298#issuecomment-1793923005


fragtion:
This fix is working for the missing tab menu @ #6298. Thank you. Hopefully @kb9gx's issue with wholly missing tab bar can also be fixed, although I have not seen that problem myself

PR #6325, opened 2023-11-07
Fixes for vcpkg-deps

basilgello:
We switched to unmodified oboe. If that gives weird
errors, we can add oboe.patch into res/vcpkg/oboe.


basilgello:
The second patch is an attempt to increase waiting time before the error is thrown from 12 to 128 seconds. In reality, reports show 30 seconds can be enough so lets check.

basilgello:
The last commit builds arm libvpx properly on arm64 machines.

Need merge and manual nightly build to see if all issues gone

basilgello:
And I added github action to publish version info file for F-Droid every time the tag is pushed. F-Droid will read it and update packages automatically.

basilgello:
I am moving oboe_wrapper to separate vcpkg port that will depend on oboe. This way build.rs can be simplified and all the heavy lifting will be done by vcpkg

rustdesk:
https://github.com/rustdesk/rustdesk/actions/runs/6809018374/job/18514500298 arm-sciter failed. :)

PR #6327, opened 2023-11-07
Update de.rs

Mr-Update:


grummbeer:
```
("Enable blocking user input", "Blockieren von Benutzereingaben zulassen"),
```

"Blockieren … zulassen" klingt uneindeutig?

https://github.com/rustdesk/rustdesk/pull/6303

permissions-allow-user-input


Analog zu den anderen müsste es heißen "Benutzereingaben aktivieren", was sich aber etwas seltsam liest. Es müsste eher lauten wie: "Benutzereingabe erlauben/zulassen" was im rechten Fenster wäre: "Benutzereingabe erlauben/zulassen: ON/OFF".

Das mit dem "zulassen" gib es schon bei "Enable file copy and paste". Wobei das aber klar aus der Reihe fällt.

```
("Enable file copy and paste", "Kopieren und Einfügen von Dateien zulassen"),
```

@21pages Does the left side "enable file transfer" do the same as the "Enable file copy and paste" on the right side? 

permissions-file-paste

21pages:
`Block user input` is used to block the input of controlled side, it's an action name. 
rustdesk_jYU25rXtfv
`copy and paste` is more accurate for remote connection, so I kept it, it also makes sense to replace it with file transfer for better consistency.

grummbeer:
### translations

> `copy and paste` is more accurate for remote connection, so I kept it, it also makes sense to replace it with `file transfer` for better consistency.

"copy and paste" sounds more like "clipboard" (already existing) while "file transfer" match really the transfer of files (which is what is meant here?).

So the action here is replaceing the translation key "Enable file copy and paste"

https://github.com/rustdesk/rustdesk/blob/7bafe142ca513d3247833f31e500911e81e389e6/flutter/lib/desktop/pages/server_page.dart#L631

with "Enable file transfer" and remove the key "Enable file copy and paste" from language files?

https://github.com/rustdesk/rustdesk/blob/7bafe142ca513d3247833f31e500911e81e389e6/src/lang/template.rs#L182

### block user input

> `Block user input` is used to block the input of the controlled side, it's an action name. 

sorry, but i do not understand how this work. This is for the controlled side to block the input from the controlling side? But why has the controlling side this "block user input" action in toolbar?

Linux > Windows

Windows (client) disabled "Enable blocking user input", but 

- controlling side can still do input
- the action "Block user input" is not visible in controlling toolbar (may intended)

permissions-allow-user-input-2

When controlled windows enable "Enable blocking user input" then at controlling side the action is visible …
permissions-allow-user-input-3

… but a error is thrown on click.

permissions-allow-user-input-4

grummbeer:
@Mr-Update what do you think of "Benutzereingaben aktivieren"? For consistency it is the best.

In general I think a permission is not something you can enable/disable, it can be granted or revoked.

21pages:
> This is for the controlled side to block the input from the controlling side

reverse, require controlled side to be installed

21pages:
> "copy and paste" sounds more like "clipboard" (already existing) while "file transfer" match really the transfer of files (which is what is meant here?).

They're different, maybe text and file but I'm not sure.
https://github.com/rustdesk/rustdesk/blob/7bafe142ca513d3247833f31e500911e81e389e6/flutter/lib/common/widgets/toolbar.dart#L439
https://github.com/rustdesk/rustdesk/blob/7bafe142ca513d3247833f31e500911e81e389e6/flutter/lib/common/widgets/toolbar.dart#L454

https://github.com/rustdesk/rustdesk/blob/7bafe142ca513d3247833f31e500911e81e389e6/src/server/connection.rs#L2487
https://github.com/rustdesk/rustdesk/blob/7bafe142ca513d3247833f31e500911e81e389e6/src/server/connection.rs#L2495

grummbeer:
@21pages in short "Enable file transfer" !== "Enable file copy and paste"?

21pages:
equals

grummbeer:
:-) OK, then the same wording should be used? If yes, the consequence is replaceing all "Enable file copy and paste" with "Enable file transfer", right?

21pages:
use enable file transfer to be consistent with permission settings may be better.

grummbeer:
So sorry, I apologize for the noise and the stolen time.

"Enable blocking user input" means "Allow the other party/controller to block _my_ input" (show/hide the toolbar action at controlling side)

@Mr-Update your translation is correct … so my suggestion is just using "aktivieren" instead of "zulassen".

> use enable file transfer to be consistent with permission settings may be better.

I tried, but i’am not longer sure of the purpose or intended function. Also found it under settings|display>others. Please don’t waste more time on this. I need to do more investigation.

Mr-Update:
@grummbeer I have made the change.

PR #6333, opened 2023-11-08
close all connections if cm exit unexpected, and allow retry

21pages:
close all connections if cm exit "unexpected", and allow retry.
"unexpected":  the return value of function start_ipc is not "expected", which always works in tests.

https://github.com/rustdesk/rustdesk/assets/14891774/1dcee426-1f1f-44f0-93d6-d4ba6941c91c


basilgello:
Any ideas why this fails on x86-android? https://gitlab.com/basilgello/fdroid-main-repo-data/-/jobs/5497763811#L6075

rustdesk:
Rerun

basilgello:
It was all easier: fdroid packaging had a step removing firebase and it was over-aggressive. Fixed and merged now - update its on its way to official fdroid repo :)

PR #6340, opened 2023-11-08
Fix. translation keys "Allow file copy and paste" > "Enable file copy and paste" 

grummbeer:
see https://github.com/rustdesk/rustdesk/pull/6303

> Permission unified into Enable...

"Allow file copy and paste" > "Enable file copy and paste"


rustdesk:
Thanks, :)

PR #6351, opened 2023-11-09
vcpkg-deps fixes

basilgello:


basilgello:
This should fix macos signing. And I am doing lib sciter build locally to see why rust toolchain does not like libvpx

basilgello:
Found the mess in sciter :) Should be fixed now

basilgello:
Well, it might be that scrap rlib is emitted as LLVM bitcode but fed into conventional `cc` linker. Last time there was a known architecture mismatch between libvpx and rest of build, linker complained pointing to `vpx_decoder.c.o` inside scrap rlib. This time, it complains on whole rlib. The same error message surfaced when LTO was used. Maybe using lld as linker helps.

rustdesk:
https://github.com/rustdesk/rustdesk/issues/3556#issuecomment-1804010109

Issue #6280, opened 2023-11-04
virtual display driver not getting installed properly


shamirahim111:
### Bug Description

i am on latest nightly versions, 4/11 on local side  (linux-wayland) and 31/10 on remote side (windows 10 ltsc)

i am trying to use virtual monitor but the drvier is not installed properly.

i tried to update driver with the dll files but that did not work.

### How to Reproduce

try to use virtual display driver. in device manager it says not installed properly

### Expected Behavior

it should get installed

### Operating system(s) on local side and remote side

linux>windows

### RustDesk Version(s) on local side and remote side

1.2.3>1.2.4

### Screenshots

![vdd1](https://github.com/rustdesk/rustdesk/assets/149858437/0e76a6df-74e5-4765-ac6a-ea5418d1bd9d)
![vdd2](https://github.com/rustdesk/rustdesk/assets/149858437/08c7bf6f-7414-4f78-a7d3-a559c826c847)
![vdd3](https://github.com/rustdesk/rustdesk/assets/149858437/0c6be455-a613-4054-ada5-47912d911cd0)

shamirahim111:
i am using aspia in the screenshots but rustdesk gets unresponsive if i click on virtual monitor, rustdesk freezes. sometimes it unsticks, sometimes i have to reconnect

rustdesk:
https://github.com/rustdesk/rustdesk/issues/5582#issuecomment-1785145975 if 1.2.4 still not work, please append to # 5582 issue.

rustdesk:
@fufesou let's still follow it up here.

rustdesk:
> i am using aspia in the screenshots but rustdesk gets unresponsive if i click on virtual monitor, rustdesk freezes. sometimes it unsticks, sometimes i have to reconnect

This case is similar to https://github.com/rustdesk/rustdesk/issues/4974.

rustdesk:
> @fufesou @rustdesk please check logs here https://cloudvyzor.com/logpad/?query&database=sandbox-e3065ec2639a2a5a9fff7ceb5a12e074

Thanks

fufesou:
@shamirahim111 Hi, thanks for your feedback and the logs.

I can only see the driver is not installed properly in the logs.

```
ERROR [src\virtual_display_manager.rs:134] Plug in monitor failed Plug in monitor failed RustDeskIdd: Idd device: GetDevicePath2 SetupDiEnumDeviceInterfaces failed, last error 0x103
```

[...]

PR #6446, opened 2023-11-17
Fix. Do not get displays on server start.

fufesou:
https://github.com/rustdesk/rustdesk/issues/6280#issuecomment-1815887300

Do not detect displays (or primary display) on server start.

RustDesk use the primary display as the default display.
The video service of the primary display should be added immediately after the first connection is ready.


shamirahim111:
@fufesou does this mean if user does not install the cert at first boot, then user wont see virtual display but if later user wants to install virtual display, is there a setting to install certificate and then install the driver?

Issue #6472, opened 2023-11-20
RustDeskIdd: Failed InstallUpdate in privacy mode 2


Creling:
### Bug Description

![image](https://github.com/rustdesk/rustdesk/assets/43109504/dd358de5-1893-4a2b-9c54-8e203410b6d5)

local device: windows11 + Rustdesk v1.2.4 2023-11-18 00:22
remote device: windows11 + Rustdesk v1.2.4 2023-11-18 00:22

### How to Reproduce

Just connect to the remote device and enable privacy mode 2

### Expected Behavior

N/A

### Operating system(s) on local side and remote side

Window 11 -> Windows 10

### RustDesk Version(s) on local side and remote side

Rustdesk v1.2.4 2023-11-18 00:22 -> Rustdesk v1.2.4 2023-11-18 00:22

### Screenshots

![image](https://github.com/rustdesk/rustdesk/assets/43109504/dd358de5-1893-4a2b-9c54-8e203410b6d5)

Creling:
Update: For anyone meet this issue like me, you should enable the "Install the Cert" when installing the Rustdesk.

fufesou:
@Creling Hi, do you solve this issue by installing the cert?

<img width="941" alt="1700451833968" src="https://github.com/rustdesk/rustdesk/assets/13586388/efb5e083-0424-4005-a652-d6de3083969d">

Creling:
> @Creling Hi, do you solve this issue by installing the cert?
> 
> <img alt="1700451833968" width="941" src="https://user-images.githubusercontent.com/13586388/284123958-efb5e083-0424-4005-a652-d6de3083969d.png">

yep, I have resolved it by installing the cert.

fufesou:
> Update: For anyone meet this issue like me, you should enable the "Install the Cert" when installing the Rustdesk.

I'll do it.

PR #6473, opened 2023-11-20
Feat/format idd error message

fufesou:
1. Format error message.  https://github.com/rustdesk/rustdesk/issues/6472#issuecomment-1818179586
2. Send a "message box" when failed to plug in/out a virtual display.
3. Replace `GetLastError()` with `std::io::Error::last_os_error()`  in rust source.

### SwDevice handle is held by another process

Format `HRESULT`

### No cert

Format `GetLastError`


rustdesk:
` fn get_error() -> String {` in windows.rs can be removed then.

PR #6487, opened 2023-11-21
Fix: wayland fractional scaling distortion

### LINKED ISSUE #6116: Wayland issues: Green lines on scaled screen + no input (SELinux, Fedora)
### Bug Description

This was tested on both 1.2.3 and October 19th nightly, on Fedora 38 GNOME with Intel graphics.

Xorg version works flawlessly.

On Wayland:

1. If I have a 125% screen scaling in Settings -> Screen, the connection via Rustdesk is completely distorted with the green lines you can see in the screenshot. I have tried different codecs and forcing software rendering, with same result.
2. If I set back the scaling to 100%, the image is fine. However, I cannot control the mouse from the remote device. I have tried doing the "sudo rustdesk --service" as README indicates, but the terminal blinks indefinitely without launching the server.

Thanks for the great product! I hope wayland support improves soon :)

### How to Reproduce

Just download rustdesk on Fedora GNOME wayland session and try to log in.

### Expected Behavior

Clear image and working mouse control.

Fedora 38

1.2.3 tested, 19-10 nightly also tested

### Screenshots

imagen

### Additional Context

_No response_

beelchester:
Fixes: #6116
Tested: arch, fedora, debian (gnome + kde). No issue with and without fractional scaling.

Screenshot 2023-11-21 at 6 55 42 PM

### DIFF
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -67,13 +67,23 @@ pub struct PipeWireCapturable {
 
 impl PipeWireCapturable {
     fn new(conn: Arc<SyncConnection>, fd: OwnedFd, stream: PwStreamInfo) -> Self {
+        // alternative to get screen resolution as stream.size is not always correct ex: on fractional scaling
+        // https://github.com/rustdesk/rustdesk/issues/6116#issuecomment-1817724244
+        let res = get_res(Self {
+            dbus_conn: conn.clone(),
+            fd: fd.clone(),
+            path: stream.path,
+            source_type: stream.source_type,
+            position: stream.position,
+            size: stream.size,
+        });
         Self {
             dbus_conn: conn,
             fd,
             path: stream.path,
             source_type: stream.source_type,
             position: stream.position,
-            size: stream.size,
+            size: res.unwrap_or(stream.size),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


manueldeprada:
nice job!!! thanks and looking forward to tomorrows nightly :))

sahilyeole:
@manueldeprada 
The mouse might not work properly because we are using xrandr to get max desktop resolution (addition of all the monitors) and xrandr gives wrong display resolution on gnome fractional scaling (same wrong value as in pipewire stream).

I'm trying to fix it.

rustdesk:
👍

manueldeprada:
> @manueldeprada The mouse might not work properly because we are using xrandr to get max desktop resolution (addition of all the monitors) and xrandr gives wrong display resolution on gnome fractional scaling (same wrong value as in pipewire stream).
> 
> I'm trying to fix it.

Thanks!! I see there is a mismatch, yes. The good part of only relying on the actual stream size is that now you can also stream windows or multi-monitors and all of this should just work

sahilyeole:
>you can also stream windows or multi-monitors and all of this should just work

Sorry, I could not get this part, can you elaborate?
Do you mean we can use the dbus data to get all monitors' resolution?

sahilyeole:
@manueldeprada https://github.com/rustdesk/rustdesk/pull/6502 will solve the cursor mismatch issue, please test it.

Also, **multi-monitor+gnome(other than ubuntu)+wayland+fractional scaling enabled** will still have cursor mismatch issue, that's because we only get the resolution of current streaming screen. Other than xrandr I could not find any other way to get max desktop resolution on wayland.

**multi-monitor+gnome(other than ubuntu)+wayland+fractional scaling enabled**, this can be a less common case, at this point user can just change scaling to 100% and cursor will work. We may look at this particular case in future if this issue still exists after official release of fractional scaling on gnome. Ubuntu gnome and kde have official release, not an issue there.

I think we have covered enough feasible support for fractional scaling for now.

PR #6504, opened 2023-11-23
feat: Support new tcp-mq rendezvous protocol

kaaass:
This PR provides client-side support for the new tcp-mq rendezvous protocol. The purpose of the new tcp-mq rendezvous protocol is to offer a rendezvous messaging communication method that doesn't require UDP, while maintaining support for the existing tcp-udp rendezvous protocol. Specifically, this PR implements client-side support for the following messages:

- TCP Messages
  - `RegisterPk` and `RegisterPkResponse`
    - Request-response type messages, previously sent by hbbs using UDP protocol. In tcp-mq protocol, these messages will be sent using a short TCP connection, i.e., a TCP connection is opened to the hbbs server for sending, and it is disconnected after receiving the expected response message.
    - `RegisterPk` message sent using TCP carries the semantics of a "handshake" for establishing a connection between the client and the server. The server's response message, `RegisterPkResponse`, now includes an extra `mq_token` field. The client will use the `mq_token` as an authentication credential to establish a connection with the mq server (Using port `21120/tcp`). This is expected to avoid abusing of mq server.

- MQ Messages
  - `RegisterPeer`, `RegisterPeerResponse`
    - Request-response type messages, previously sent by hbbs using UDP protocol. In tcp-mq protocol, these messages will be sent using mq, utilizing the topic `ping.[node_id]`. The direction of message sending remains the same as existing messages. That is, `RegisterPeer` is sent from the client to the server via topic `ping.[node_id]`, after which server sends `RegisterPeerResponse` to client using topic `push.[node_id]`.

  - `PunchHole`, `RequestRelay`, `FetchLocalAddr`
      - Previously pushed actively to the client by hbbs using the UDP protocol. In tcp-mq protocol, these messages will be sent using mq, utilizing the topic `push.[node_id]`.

  - `ConfigureUpdate`
     - This message is an optional response to the `RegisterPeer` message, previously sent by hbbs using UDP protocol after `RegisterPeerResponse`. In tcp-mq protocol, these messages will be sent using mq, utilizing the topic `push.[node_id]`.

## Limitations

1. The ability to establish a NATS MQ connection via a user-specified socks5 proxy has not yet been implemented.
2. The implementation of JWT token authentication on a per-peer basis is still pending. Currently, only a version based on token authentication has been implemented. Although this coarser-grained authentication method can prevent attackers from abusing the MQ for message passing (as long as the admin and regular user accounts are not compromised simultaneously), it still does not restrict peers from listening to messages from other peers. It is necessary to employ JWT tokens for fine-grained ACL (Access Control List) access control for different peers.

## Tests

- ✅ New Client → New Client, local network (Both clients using TCP + NATS, trying to establish remote control session)
- ✅ New Client → Old Client, hole punching (TCP + NATS client send remote control request to UDP client)
- ✅ Old Client → New Client, hole punching
- ✅ New Client → Old Client, relay
- ✅ Old Client → New Client, relay
- ✅ New Client connects to Old Server
- ✅ New Client change ID
- ✅ Old Client change ID
- ✅ When hbbs server restarts, new client should be able to reconnect
- ⏳ Online config update for new client

### DIFF
--- a/libs/hbb_common/src/socket_client.rs
+++ b/libs/hbb_common/src/socket_client.rs
@@ -154,7 +154,7 @@ pub fn ipv4_to_ipv6(addr: String, ipv4: bool) -> String {
     addr
 }
 
-async fn test_target(target: &str) -> ResultType<SocketAddr> {
+pub async fn test_target(target: &str) -> ResultType<SocketAddr> {
     if let Ok(Ok(s)) = super::timeout(1000, tokio::net::TcpStream::connect(target)).await {
         if let Ok(addr) = s.peer_addr() {
             return Ok(addr);
--- a/src/rendezvous_mediator.rs
+++ b/src/rendezvous_mediator.rs
@@ -6,6 +6,7 @@ use std::{
     },
     time::Instant,
 };
+use std::cmp::min;
 use async_nats::{ConnectOptions, Subscriber};
 
 use uuid::Uuid;
@@ -15,9 +16,9 @@ use hbb_common::{allow_err, anyhow::bail, config::{Config, CONNECT_TIMEOUT, READ
     self, select,
     time::{interval, Duration},
 }, udp::FramedSocket, AddrMangle, ResultType, IntoTargetAddr, anyhow};
-use hbb_common::anyhow::anyhow;
 use hbb_common::config::NetworkType;
 use hbb_common::socket_client::test_target;
+use hbb_common::futures::StreamExt;
 use crate::common::increase_port;
 
 use crate::server::{check_zombie, new as new_server, ServerPtr};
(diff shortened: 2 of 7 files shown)


src/rendezvous_mediator.rs
```diff
@@ -6,27 +6,20 @@ use std::{
...
-    tokio::{
-        self, select,
-        time::{interval, Duration},
-    },
-    udp::FramedSocket,
-    AddrMangle, ResultType,
-};
+use hbb_common::{allow_err, anyhow::bail, config::{Config, CONNECT_TIMEOUT, READ_TIMEOUT, REG_INTERVAL, RENDEZVOUS_PORT}, futures::future::join_all, log, protobuf::Message as _, rendezvous_proto::*, sleep, socket_client::{self, is_ipv4}, tokio::{
+    self, select,
+    time::{interval, Duration},
```
rustdesk: auto formatted?

src/rendezvous_mediator.rs
```diff
@@ -6,27 +6,20 @@ use std::{
...
+}, udp::FramedSocket, AddrMangle, ResultType, IntoTargetAddr, anyhow};
+use hbb_common::config::NetworkType;
+use hbb_common::socket_client::test_target;
+use hbb_common::futures::StreamExt;
```
rustdesk: Please follow our `use` style.

src/rendezvous_mediator.rs
```diff
@@ -92,7 +85,11 @@ impl RendezvousMediator {
                 for host in servers.clone() {
                     let server = server.clone();
                     futs.push(tokio::spawn(async move {
-                        if let Err(err) = Self::start(server, host).await {
+                        if let Err(err) = Self::start(server.clone(), host.clone()).await {
+                            log::error!("rendezvous mediator error: {err}");
+                        }
+                        // Fallback to legacy udp protocol
+                        if let Err(err) = Self::legacy_start(server, host).await {
```
rustdesk: Let's use udp_start, `legacy_start` make users think we are discarding udp.

libs/hbb_common/src/socket_client.rs
```diff
@@ -154,7 +154,7 @@ pub fn ipv4_to_ipv6(addr: String, ipv4: bool) -> String {
     addr
 }
 
-async fn test_target(target: &str) -> ResultType<SocketAddr> {
+pub async fn test_target(target: &str) -> ResultType<SocketAddr> {
```
rustdesk, line 157: socks5 proxy tested for nats connection?

src/rendezvous_mediator.rs
```diff
@@ -92,7 +85,11 @@ impl RendezvousMediator {
                 for host in servers.clone() {
                     let server = server.clone();
                     futs.push(tokio::spawn(async move {
-                        if let Err(err) = Self::start(server, host).await {
```
rustdesk, line 95: Let's use another loop. Let's define a new static variable in config.rs.

```
pub const NATS_SERVERS: &str = match option_env!("NATS_SERVER") {
    Some(server) if !server.is_empty() => [server],
    _ => ["nats://nats.rustdesk.com"],
};
```

 And should we connect to multiple nats servers the same time? We should loop to connect to only one avaiable nats server.

rustdesk (changes requested):

libs/hbb_common/src/socket_client.rs
kaaass: The ability to establish a NATS MQ connection via a user-specified socks5 proxy has not yet been implemented. This is mainly because [nats.rs](https://github.com/nats-io/nats.rs) doesn't expose the necessary API. Instead, it just using `TcpStream::connect` directly ([async-nats/src/connector.rs](https://github.com/nats-io/nats.rs/blob/c661d2147ccbbfc25a32fad8e5b9c1cb9ca077df/async-nats/src/connector.rs#L291C27-L291C27)). We might need to fork our own version to replace it with other `dyn AsyncWrite + AsyncRead + Send + Unpin`, e.g., getting from `FramedStream`.

src/rendezvous_mediator.rs
rustdesk: We have to fork then.

libs/hbb_common/src/socket_client.rs
rustdesk: We have to fork then.

src/rendezvous_mediator.rs
kaaass: OK. Let me check if I understand it right. This means we should consider the NATS server/cluster as a separate service, not just an extra port of a rendezvous server. Thus connecting to any of them would be enough.

rustdesk: Yes, it is a seperate service, just like https://azure.microsoft.com/en-us/products/iot-hub

kaaass: I got it. 

BTW, would it be acceptable if we send `NATS_SERVERS` in `RegisterPkResponse`? Since we could connect to MQ only after we get `mq_token`, IMO it is reasonable to also get `mq_servers` in the same way.

PR #6702, opened 2023-12-17
Estonian translation

Madis0:
Translated the UI to Estonian (et-EE) :)


rustdesk:
Thank you very much.
You also need to modify https://github.com/rustdesk/rustdesk/blob/master/src/lang.rs to make it effective.

Madis0:
Done.

PR #6675, opened 2023-12-13
Feat: Wayland flatpak input support | Remote desktop portal

beelchester:
## Description:
Utilise the [remote desktop portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.RemoteDesktop.html) interface (provided by xdg desktop portal) when uinput is unavailable for input capture on wayland.
## Advantages:
Wayland input is supported similar to uinput on:
1. Flatpak
2. While dev/debugging (cargo/flutter run)
## Supports:
It supports input from:
- Desktop
- Mobile
- Physical keyboard mobile

Events:
- Key inputs
- Mouse movements, clicks, and scroll (x & y axis)
## Tested:
Remote: ubuntu, arch (kde, gnome)
Local: mac, linux, ios, android
## Note:
During testing, I found an issue that occurs even without the changes of this PR (also occurs on 1.2.3):
- Mobile -> ubuntu wayland flatpak stream freezes but input still works (rdp input; enabled by this PR).
## Build for test/review:
[rustdesk-1.2.4-x86_64_wl_rdp.flatpak](https://github.com/sahilyeole/rustdesk/releases/download/nightly/rustdesk-1.2.4-x86_64_wl_rdp.flatpak)
## Test videos:

https://github.com/rustdesk/rustdesk/assets/73148455/643e0438-9bd7-48fc-8a73-1c1519103545

https://github.com/rustdesk/rustdesk/assets/73148455/31c45b29-b106-4afb-9d89-a50b550cb635

PR #6742, opened 2023-12-23
Fix xdg portal verison check for persist_mode

### LINKED ISSUE #6741: Persist_mode for xdg portal does not work
### Bug Description

The portal dialog on wayland shows upon connecting but it is not possible to make the remote desktop session persistent.

### How to Reproduce

Connect to a wayland host.

### Expected Behavior

Persistent session 

wayland -> wayland (Fedora 39)

nightly

### Screenshots

-

### Additional Context

_No response_

linuxrider:
fixes #6741 on Fedora 39.

- version check mixed up `ScreenCast` and `RemoteDesktop` which added `persist_mode` and `restore_token` on versions `4` and `2`, respectively.
- injection of `persist_mode` and `restore_token` happened in a too late step of setting up the session

### DIFF
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -563,7 +563,16 @@ fn on_create_session_response(
     move |r: OrgFreedesktopPortalRequestResponse, c, _| {
         let portal = get_portal(c);
         let mut args: PropMap = HashMap::new();
-
+        if let Ok(version) = remote_desktop_portal::version(&portal) {
+            if version >= 2 {
+                let restore_token = config::LocalConfig::get_option(RESTORE_TOKEN_CONF_KEY);
+                if !restore_token.is_empty() {
+                    args.insert(RESTORE_TOKEN.to_string(), Variant(Box::new(restore_token)));
+                }
+                // persist_mode may be configured by the user.
+                args.insert("persist_mode".to_string(), Variant(Box::new(2u32)));
+            }
+        }
         args.insert(
             "handle_token".to_string(),
             Variant(Box::new("u2".to_string())),
@@ -620,16 +629,6 @@ fn on_select_devices_response(
     move |_: OrgFreedesktopPortalRequestResponse, c, _| {
         let portal = get_portal(c);
         let mut args: PropMap = HashMap::new();
-        if let Ok(version) = remote_desktop_portal::version(&portal) {
-            if version >= 4 {
-                let restore_token = config::LocalConfig::get_option(RESTORE_TOKEN_CONF_KEY);
-                if !restore_token.is_empty() {
-                    args.insert(RESTORE_TOKEN.to_string(), Variant(Box::new(restore_token)));
-                }
-                // persist_mode may be configured by the user.
-                args.insert("persist_mode".to_string(), Variant(Box::new(2u32)));
-            }
-        }
         args.insert(
             "handle_token".to_string(),
             Variant(Box::new("u3".to_string())),
@@ -697,7 +696,7 @@ fn on_start_response(
     move |r: OrgFreedesktopPortalRequestResponse, c, _| {
         let portal = get_portal(c);
         if let Ok(version) = remote_desktop_portal::version(&portal) {
-            if version >= 4 {
+            if version >= 2 {
                 if let Some(restore_token) = r.results.get(RESTORE_TOKEN) {
                     if let Some(restore_token) = restore_token.as_str() {
                         config::LocalConfig::set_option(


rustdesk:
@fufesou

linuxrider:
Relevant API description: https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.RemoteDesktop.html#org-freedesktop-portal-remotedesktop-selectdevices

rustdesk:
@linuxrider have you tested this fixes your `persist` issue? we tested, it does not work.

linuxrider:
I have tested it for Fedora 39 Gnome 45 on wayland. With this the persistent checkbox is present in the source selection portal dialog. Furthermore after accepting with persistence option, closing rustdesk and reopening no further dialog is shown upon reconnecting and remote control works.

This is the only procedure that works for me. I have tried several other procedures. E.g. adding the `restore_token` in `on_create_session_response`, `on_select_devices_response` and `on_select_sources_response` or only the first two. But then it crashed and I got the message on the rustdesk client `X11 expected`.

Documentation of the portals is not that clear for me. I think this is how it is supposed to work:
The chain of action is `RemoteDesktop.CreateSession` -> `RemoteDesktop.SelectDevices` ("is called" in `on_create_session_response`, here `restore_token` needs to be provided ) -> `ScreenCast.SelectSources` (persistence managed by  `RemoteDesktop `) -> `RemoteDesktop.Start` (returns new token).

The `RemoteDesktop ` portal added `persist_mode` on version `2` and it is stated that persistence is managed by `RemoteDesktop` rather than `ScreenCast` portal https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html#org-freedesktop-portal-screencast-selectsources

rustdesk:
Thanks @linuxrider 

@fufesou

linuxrider:
It works also with the nightly flatpak. 
I am writing this in a remote session.
Really glad that it works now. Thank you.

fufesou:
There're two problems of this PR.

1. It does not fix the "restore_token" issue.
2. `persist_mode` is supported until [version 4]( https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html#:~:text=portal.RemoteDesktop%20interface.-,This%20option%20was%20added%20in%20version%204%20of%20this%20interface.,-persist_mode%20(u) for `org.freedesktop.portal.ScreenCast`. 

### 1.2.3


https://github.com/rustdesk/rustdesk/assets/13586388/94442120-354e-4002-8249-1b14c1c4512b

### master commit and this branch

https://github.com/rustdesk/rustdesk/assets/13586388/f313ad88-128a-4405-89e9-882ff451fa08

linuxrider:
So the version checking needs to be more fine grained and in more places to cover Fedora 39 and Ubuntu 22.04.
Your point 2: Does this mean that persistence handling is done in `ScreenCast.SelectSources` for portal versions < 4?

fufesou:
I don't know how portal handle it under version 4. I'm also testing.

Another thing is the sharing icon persists after the connection end.

In 1.2.3, the sharing icon disappears after the connection.

fufesou:
> So the version checking needs to be more fine grained and in more places to cover Fedora 39 and Ubuntu 22.04.

Does this PR work fine in Fedora 39?

I've tested Fedora 38, the same behavior to Ubuntu 22.04.

linuxrider:
@fufesou I have adapted the version checks and it still works for Fedora 39. 
Does this work in your tests, too? 
https://github.com/linuxrider/rustdesk/tree/persist_mode

linuxrider:
> I don't know how portal handle it under version 4. I'm also testing.
> 
> Another thing is the sharing icon persists after the connection end.
> 
> In 1.2.3, the sharing icon disappears after the connection.

The icon staying in the system tray after connection ends is also present in Fedora 39.

fufesou:
> The icon staying in the system tray after connection ends is also present in Fedora 39.

So your test may be invalid.

You need to run `sudo systemctl restart rustdesk` after the connection.

Or you can stop rustdesk service, do the following steps:

1. `sudo systemctl stop rustdesk`
2. `pkill rustdesk`
3. `rustdesk`
4. connect
5. stop rustdesk
6. `rustdesk`
7. connect

linuxrider:
I did the test without installation of the compiled executable directly and with the flatpak from github. Closing the flatpak removes rustdesk from memory. I also did a reboot and the session still persisted.
Here it the flatpak to flatpak Fedora 39 test.
[Bildschirmaufzeichnung vom 2023-12-24, 14-58-42.webm](https://github.com/rustdesk/rustdesk/assets/5882359/83a91ac7-399d-40b4-915b-a12ec149030a)

I could not get the rpm nor the deb to work properly in the vm.
Will try once more.

fufesou:
@linuxrider I've also tested with Fedora 39. This PR works fine.

It was my fault.

The versions of [`org.freedesktop.portal.RemoteDesktop`](https://github.com/rustdesk/rustdesk/blob/57acadd52a91150ed940df481d6eedb100bf877e/libs/scrap/src/wayland/remote_desktop_portal.rs#L308) and [`org.freedesktop.portal.ScreenCast`](https://github.com/rustdesk/rustdesk/blob/57acadd52a91150ed940df481d6eedb100bf877e/libs/scrap/src/wayland/screencast_portal.rs#L99) are different.

Your PR is good.

| Version |  org.freedesktop.portal.RemoteDesktop | org.freedesktop.portal.ScreenCast |
| --- | --- | --- |
| Ubuntu 22.04 | 1 | 4 |
| Fedora39 | 2 | 5 |

image

fufesou:
@sahilyeole FYI. There will be compatibility issues here. `persist_mode` is only supported when `org.freedesktop.portal.RemoteDesktop` version >= 2. The latest Fedora39 is the minimum version that can meet this requirement.

And there's one left strange thing. Even after the connection ends, the shared state persists.
It is introduced in https://github.com/rustdesk/rustdesk/pull/6675 . Maybe we can improve it.

image

sahilyeole:
Sure, I will work on it soon.

sahilyeole:
@linuxrider @fufesou 
Should we use `persist_mode` for [select_sources](https://arc.net/l/quote/tbdffnjx) as well?
Is it needed there too?

we can add a check there like `screencast_portal::version(&portal) >= 4`

linuxrider:
@sahilyeole For `org.freedesktop.portal.RemoteDesktop.version` >= 2 it is enough to set `persist_mode`  for`org.freedesktop.portal.RemoteDesktop.SelectDevices`. It does not work if it is also set in `SelectSources`. Persistence is managed in `org.freedesktop.portal.RemoteDesktop` (https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html).
For `org.freedesktop.portal.RemoteDesktop.version` < 2 it is a different story and I don't understand how it works exactly.

fufesou:
> we can add a check there like screencast_portal::version(&portal) >= 4

@sahilyeole I'm not sure if the restore token can be used for both `RemoteDesktop` and `ScreenCast`.

Maybe we can use "RemoteDesktop" and "uinput" together.

For `org.freedesktop.portal.RemoteDesktop.version` < 2 and `org.freedesktop.portal.RemoteDesktop.version` >= 4, use `uinput` as the input method.

Otherwise, use "RemoteDesktop" as the input method.

It's a little complicated.

What do you think?

sahilyeole:
Currently, we are only using rdp input only when uinput is not available, like on flatpak.
On normal wayland we are still using uinput.
One possible solution I can think of to maintain persistence same as before my PR for normal wayland:
Skip `org.freedesktop.portal.RemoteDesktop.SelectDevices` in the callback chain when uinput is already available. Use `persist_mode` for `select_sources`.
This way the function `request_remote_desktop` will work the same as it was before my PR only when uinput is already available (on normal wayland), also same persistence as before.

But the need for such implementation comes down to how much of a big problem the unavailability of `persist_mode` is because during my testing I didn't see any problem so I'm not sure.
Let me know if we should proceed with it.

fufesou:
> One possible solution I can think of to maintain persistence same as before my PR for normal wayland:
Skip `org.freedesktop.portal.RemoteDesktop.SelectDevices` in the callback chain when uinput is already available. Use persist_mode for select_sources.

Agree. I've also confirmed with @rustdesk

sahilyeole:
After this PR wayland arch linux (with version 2) as remote is not working. It asks for input permission but the connection is not made.
It might not be stable to use yet, not sure.
So should I comment the part added in this PR for flatpak? Because for normal wayland I'm turning back pipewire.rs code to how it was before #6675 anyway (with persist_mode in select_sources).
Also, version 2 is only available in rolling releases that too fail on arch, not tested fedora.

rustdesk:
> After this PR wayland arch linux (with version 2) as remote is not working

I will revert this PR.

rustdesk:
Reverted

rustdesk:
Reverted

RushingAlien:
Can this be added back? it's been over a year

PR #6765, opened 2023-12-29
use official async trait

21pages:
https://github.com/21pages/rustdesk/actions/runs/7354782147

developers need to update rust to 1.75 after merging


rustdesk:
@fufesou @sahilyeole upgrade rust please.

PR #6782, opened 2024-01-02
Better 2fa verification support

fufesou:
https://github.com/rustdesk/rustdesk/assets/13586388/d189ae51-4d8b-4984-bdf3-d1b5d3a1c846

Login to old/new server are both tested.

### DIFF
--- a/flutter/lib/common/hbbs/hbbs.dart
+++ b/flutter/lib/common/hbbs/hbbs.dart
@@ -11,9 +11,11 @@ class HttpType {
   static const kAuthReqTypeMobile = "mobile";
   static const kAuthReqTypeSMSCode = "sms_code";
   static const kAuthReqTypeEmailCode = "email_code";
+  static const kAuthReqTypeTfaCode = "tfa_code";
 
   static const kAuthResTypeToken = "access_token";
   static const kAuthResTypeEmailCheck = "email_check";
+  static const kAuthResTypeTfaCheck = "tfa_check";
 }
 
 enum UserStatus { kDisabled, kNormal, kUnverified }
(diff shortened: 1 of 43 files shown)


Mikesco3:
This would be such a welcome feature !!!

Mikesco3:
@fufesou would this add the 2FA prompt for logging in with the unattended password as well?
so for example when I install rustdesk on computer A and set a password under security, 
then on computer B, I go to login to computer A with the ID of computer A, plus the password I entered, will it prompt for the code from the authenticator app at that point as well?

rustdesk:
add sciter also please

fufesou:
> @fufesou would this add the 2FA prompt for logging in with the unattended password as well? so for example when I install rustdesk on computer A and set a password under security, then on computer B, I go to login to computer A with the ID of computer A, plus the password I entered, will it prompt for the code from the authenticator app at that point as well?

NO. Current implementation is only for login. 

For connections, perhaps it would be better to confirm via a pop-up window or message bar on the app.

rustdesk:
@Mikesco3 this is also in our prioirity.

Mikesco3:
> > @fufesou would this add the 2FA prompt for logging in with the unattended password as well? so for example when I install rustdesk on computer A and set a password under security, then on computer B, I go to login to computer A with the ID of computer A, plus the password I entered, will it prompt for the code from the authenticator app at that point as well?
> 
> No. Current implementation is only for login.
> 
> For connections, perhaps it would be better to confirm via a pop-up window or message bar on the app.

@fufesou thank you for answering.
I love rustdesk and I'm super grateful for benefitting of all of the effort by people like you.

What I would love to see at some point is the ability to enable a pop-up like what anydesk has, where it prompts for a totp code in the app when trying to login using the unattended password.

The scenario is useful for example if computer B is a laptop where I stored into the rustdesk app the id and login password of computer B, it would still prompt in the rustdesk app or computer B to put in the authenticator's rolling code before showing the screen of computer A.

It could be implemented by having added a check-box right below the setting up the passwords to enable it to generate the totp seed for the authenticator app, and disabling the box flushes the seed allowing again 1-click login with just the password.

So that way if computer B got compromised or is being used by someone else, that doesn't instantly give 1-click access to computer A.

I don't know much more work would that feature require, but it would be really appreciated.

All that being said, I really love this project and I'm super grateful for such an excellent open source alternative as Rustdesk.

Happy New Year

PR #6805, opened 2024-01-05
Update cn.rs

Integral-Tech:
Add Chinese translation for ```2FA code```.


zhangbo8418:
> Add Chinese translation for `2FA code`.

Issue #323, opened 2021-12-05
multiple concurrent users (multiple rdp sessions / terminals)


test2a:
would it be possible to get multiple clients to hit a windows machine multiple user logins at once? i mean i understand this to be an anydesk/teamviewer alternative than a rdp alternative but still, can users be selected and can multiple people hit a single server at once and not get the same account

fufesou:
For now rustdesk captures picture of the displays.
This idea can be implemented, if multiple accounts login on different displays.
But there seems no way to do this?

Rustdesk can also start rdp client.
On the other way, you can use rdp connection to do this on windows.
https://github.com/anhkgg/SuperRDP

test2a:
@fufesou i currently use rdpwrap to avoid buying cals or whatever but i don;t know what superrdp does. anyways, i haone "this" via anydesk but its janky. maybe you can "automate" it.

i have multiple terminals on the "server" that i need to remote into. instead of initiating the connection from the remote>server, i do the opposite, i start a connection server>remote then reverse the connection. 

i have only achieved this when the account on the server is on and someone can do this manually but i'm not sure if this is making much sense. 
oh, i have also had the idea of " different anydesk IDs per user" on a machine so a remote user can join their own account remotely with password so that is something you can look into. Anydesk does not allow that, they have a per machine id which makes this a bummer but maybe you can try.

fufesou:
@test2a  "SuperRDP is a new version rdpwarp base on the rdpwrap by stascorp." You can learn it very in few minutes.

As I know, SuperRDP may handle your request well.

Tapuzi:
Discussion was useful for me - thank you all.

rustdesk:
![890347658](https://user-images.githubusercontent.com/1255404/212705498-b0815476-52d5-410e-8960-db9fdd065c86.gif)

pata2004:
Thanks for the great work.
Can Rustdesk manage several virtual graphics desktops (different users) on a single remote server?
If yes, how can I set up that?
If not, do you plan to add this function?

rustdesk:
no plan yet

pata2004:
Hi,
I use the Rustdesk to connect to my Centos machine with 4 users/desktops. 
But I never know what user/desktops Rustdesk will go open.
Can I define what user/desktop to connect to?
Or, do I need to use Centos with only the root user?

test2a:
[...]

PR #6825, opened 2024-01-08
Feat: Windows connect to a specific user session

beelchester:
#2095
#323 related but no concurrent user sessions

Note: Both client and server-side needs to be updated. Also, installation is required.

---
Test:
RDPWrapper's RDPCheck program running on user `sahil_2` while the default (console) is the user `Sahil`.
I don't have a setup for terminal sessions so I haven't tested that. Feedback would be appreciated.

https://github.com/rustdesk/rustdesk/assets/73148455/ace53c41-7bb5-4ab2-84e1-8e0856609d7e

Tested mac, android -> windows too

---
Workflow:
image


sahilyeole:
File transfer is not working on multiple user sessions, fixing that.

libs/hbb_common/protos/message.proto
```diff
@@ -120,10 +120,16 @@ message PeerInfo {
...
+message MultipleUserSessions {
+  string user_session_ids = 1;
+  string user_names = 2;
+}
+
 message LoginResponse {
   oneof union {
     string error = 1;
     PeerInfo peer_info = 2;
+    MultipleUserSessions multiple_user_sessions = 3;
```
rustdesk: You should do like this.
```
message RdpUserSession {
  string user_session_id = 1;
  string user_name = 2;
}

repeated RdpUserSession rdp_user_sessions = 3;
```

src/client.rs
```diff
@@ -1562,6 +1562,7 @@ impl LoginConfigHandler {
                 self.adapter_luid,
             ));
         n += 1;
+        msg.user_session = Config::get_option("store_usid");
```
rustdesk: Config::get_option is used to store global option. store_usid is not global option, it is peer option.

src/server/connection.rs
```diff
@@ -14,6 +14,7 @@ use crate::platform::linux_desktop_manager;
 use crate::platform::WallPaperRemover;
 #[cfg(windows)]
 use crate::portable_service::client as portable_client;
+use crate::ui_interface::{get_local_option, set_local_option};
```
rustdesk: server/connection.rs does nothing with the UI, so you can not use any thing of ui_interface here.

src/flutter.rs
```diff
@@ -826,6 +826,16 @@ impl InvokeUiSession for FlutterHandler {
         )
     }
 
+    fn set_multiple_user_session(&self, u_sids: String, u_names: String) {
+        self.push_event(
+            "set_multiple_user_session",
+            vec![
+                ("u_sids", &u_sids),
+                ("u_names", &u_names),
+            ],
```
rustdesk: I guess sid and name are paired, you should not seperate them, and wast resource to double check sids.size == u_names.size. Please do as I suggested in the protobuf.

src/server/connection.rs
```diff
@@ -1482,8 +1485,49 @@ impl Connection {
         self.video_ack_required = lr.video_ack_required;
     }
 
+    #[cfg(all(feature = "flutter", target_os = "windows"))]
```
rustdesk: again, server/connection.rs does nothing with UI. Your logic is totally wrong here, and confused by remote and local side.

I know you want to tell Sciter (controlling side) does not support mutipe rdp. But here is controlled side, not controlling side.

So you have to add Sciter support. Otherwise, there will be problem if controlled side is the newest version with multiple rdp session support, and controlling side is Sciter which will hang there after input password.

src/server/input_service.rs
```diff
@@ -1571,6 +1571,13 @@ async fn send_sas() -> ResultType<()> {
     Ok(())
 }
 
+#[tokio::main(flavor = "current_thread")]
+pub async fn connect_to_user_session(usid:Option<u32>) -> ResultType<()> {
```
rustdesk: format. `userid: Option`. Pleaser auto format before submit

rustdesk: should put connect_to_user_session here. This is not part of input_service. Put in ipc.rs.

src/server/connection.rs
```diff
@@ -1482,8 +1485,49 @@ impl Connection {
...
+            .collect::<Vec<&str>>();
+        if usids_vec.len() <= 1 {
+            return;
+        }
+        let usid;
+        match lr.option.user_session.parse::<u32>() {
+            Ok(n) => usid = Some(n),
+            Err(..) => usid = None,
+        }
+        let is_usid_changed = get_local_option("is_usid_changed".to_string()) == "true";
```
rustdesk: We have our convention to use bool option, please follow.

src/server/connection.rs
```diff
@@ -1504,16 +1549,18 @@ impl Connection {
...
+                    crate::platform::run_as_user(vec!["--tray"], usid).ok();
                 }
             });
         }
     }
 
     async fn on_message(&mut self, msg: Message) -> bool {
         if let Some(message::Union::LoginRequest(lr)) = msg.union {
+            #[cfg(all(feature = "flutter", target_os = "windows"))]
+            self.handle_multiple_user_sessions(lr.clone()).await;
```
rustdesk: Where do you tell this process run as system? You have said this can only work if it is installed and run as system service.

src/server/connection.rs
```diff
@@ -1482,8 +1485,49 @@ impl Connection {
         self.video_ack_required = lr.video_ack_required;
     }
 
+    #[cfg(all(feature = "flutter", target_os = "windows"))]
+    async fn handle_multiple_user_sessions(&mut self, lr: LoginRequest){
+        if let Some(login_request::Union::FileTransfer(_)) = lr.union {
```
rustdesk: How about port forward session?

Again, I think we should follow current convention for this `if` like below.
```
         let (conn_type, auth_conn_type) = if self.file_transfer.is_some() {
             (1, AuthConnType::FileTransfer)
         } else if self.port_forward_socket.is_some() {
             (2, AuthConnType::PortForward)
         } else {
             (0, AuthConnType::Remote)
         };
```

rustdesk (changes requested):

rustdesk:
I will review again tomorrow, still a lot of problems

rustdesk:
How will you handle this?

https://github.com/rustdesk/rustdesk/blob/a2d08fa40dab796d20cf1d2940d77038454e52da/flutter/lib/desktop/pages/desktop_setting_page.dart#L803

Does AnyDesk have similar thing to turn on/off multiple rdp session sharing? Please confirm.

sahilyeole:
I missed multiple scenarios. Thanks for pointing it out.

---

> How will you handle this?
> 
> https://github.com/rustdesk/rustdesk/blob/a2d08fa40dab796d20cf1d2940d77038454e52da/flutter/lib/desktop/pages/desktop_setting_page.dart#L803

I can use `is_share_rdp()` from windows.rs to only have this feature when it is true.
> 
> Does AnyDesk have similar thing to turn on/off multiple rdp session sharing? Please confirm.

I could not find any such option in anydesk. Though this feature seem to be [only available for windows servers on anydesk](https://arc.net/l/quote/oloujbis) so I'm not completely sure.

> Where do you tell this process run as system? You have said this can only work if it is installed and run as system service.

```rust
let _ = connect_to_user_session(usid);
```
This part is responsible for telling the service to restart the server according to the given session id.
It does not work on non-installed, I can use `is_installed()` from windows.rs to ensure it is shown only when it is installed.

> So you have to add Sciter support. Otherwise, there will be problem if controlled side is the newest version with multiple rdp session support, and controlling side is Sciter which will hang there after input password.

Yes, got it. I'm working on adding this in sciter.

---
Rest I'm making the changes as suggested, let me know if more changes are needed to be done.

rustdesk:
Did you test the case use close the rdp user while you are connecting to?

rustdesk:
> 'Enable RDP session sharing

Let's keep this option, but change the default value from `false` to `true`

sahilyeole:
> Did you test the case the user closes the rdp user while you are connecting to?

It fails the connection (screen is not shared) and also there's a possibility that server might also not work if rdp user is closed and it was running on it. So I need to add a check in the service loop in windows.rs to ensure the current `session_id` it has is active or else change `session_id` to the value of `get_current_session()` like earlier.

shamirahim111:
@sahilyeole did you test from linux?  would x11 or wayland make a difference or those things are not relevant for this PR?

second, should the main user, like the admin be informed that there are x number of users connected to abc like they can do from task manager and active users tab? 

like there is an idea of a threat, like a bad guy installs RD/AD on a victims computer, creates a new user and sets up RDPWRAP for free/ease of use and now suddenly the bad guy can come in and out of the system and do anything without the user even knowing. like normally there is the RD window open in the taskbar or the info window but this way there wont be any info.  not sure if this needs to be added/implemented or can be ignored but just wanted to write this for future reference

sahilyeole:
Sciter:


https://github.com/rustdesk/rustdesk/assets/73148455/1ed18c3c-8fe8-4e0b-9062-0587c35e1466

rustdesk:
Please resolve previous conservations.

sahilyeole:
I have completed the changes required. Let me know if more changes are needed.

rustdesk:
@21pages help me review and test.

src/server/connection.rs
```diff
@@ -1491,8 +1494,55 @@ impl Connection {
         self.video_ack_required = lr.video_ack_required;
     }
 
+    #[cfg(target_os = "windows")]
+    async fn handle_multiple_user_sessions(&mut self, lr: LoginRequest) {
+        if self.file_transfer.is_some() {
```
21pages: For file transfer, Always pop `Multiple active user sessions found`

src/server/connection.rs
```diff
@@ -1491,8 +1494,55 @@ impl Connection {
...
+                    rdp.push(rdp_session);
+                }
+                res.rdp_user_sessions = rdp;
+                let mut msg_out = Message::new();
+                msg_out.set_login_response(res);
+                self.send(msg_out).await;
+            }
+            if usid != self.user_session_id && !is_usid_changed {
+                LocalConfig::set_option("is_usid_changed".to_string(), "Y".to_string());
+                self.send_close_reason_no_retry("Restarting...").await;
```
21pages: "Restarting..." is a retry message.

flutter/lib/common/widgets/dialog.dart
```diff
@@ -1861,3 +1861,104 @@ void enter2FaDialog(
...
+      dialogManager.showLoading(translate('Connecting...'),
+          onCancel: closeConnection);
+    }
+
+    return CustomAlertDialog(
+      title: null,
+      content: msgboxContent(type, title, text),
+      actions: [
+        SessionsDropdown(peerId, sessionId, sessions),
+        dialogButton('Connect', onPressed: onConnect, isOutline: true),
```
21pages: `isOutline: false` seems better

src/server/connection.rs
```diff
@@ -1491,8 +1494,55 @@ impl Connection {
...
+                return;
+            }
+            let usid;
+            match lr.option.user_session.parse::<u32>() {
+                Ok(n) => usid = Some(n),
+                Err(..) => usid = None,
+            }
+            let is_usid_changed = LocalConfig::get_option("is_usid_changed") == "Y";
+            if usid.is_none() {
+                LocalConfig::set_option("is_usid_changed".to_string(), "N".to_string());
```
21pages: I don't fully understand the logic yet,  here need a connection scope option or a global scope option?

src/ui_interface.rs
```diff
@@ -448,9 +448,9 @@ pub fn is_share_rdp() -> bool {
 }
 
 #[inline]
-pub fn set_share_rdp(_enable: bool) {
+pub fn set_share_rdp(_disable: bool) {
```
21pages: change var name to `_enable`

flutter/lib/common/widgets/dialog.dart
```diff
@@ -1861,3 +1861,104 @@ void enter2FaDialog(
...
+    String title,
+    String text,
+    OverlayDialogManager dialogManager,
+    SessionID sessionId,
+    String peerId,
+    String sessions) {
+  dialogManager.dismissAll();
+  dialogManager.show((setState, close, context) {
+    onConnect() {
+      bind.sessionReconnect(sessionId: sessionId, forceRelay: false);
```
21pages: just a idea, remove `selected_user_session_id` peer option and make it a optional parameter of sessionReconnect

21pages:
The review is invisible,  add comments here
1. For file transfer, Always pop `Multiple active user sessions found`
2. `self.send_close_reason_no_retry("Restarting...").await;`, "Restarting..." is a retry message.
3. `dialogButton('Connect', onPressed: onConnect, isOutline: true),` , isOutline: false seems better
4. `let is_usid_changed = LocalConfig::get_option("is_usid_changed") == "Y";` I don't fully understand the logic yet, here need a connection scope option or a global scope option?
5. `pub fn set_share_rdp(_disable: bool) {`, change parameter name to _enable
6. `bind.sessionReconnect(sessionId: sessionId, forceRelay: false);`, just a idea, remove `selected_user_session_id peer` option and make it a  sessionReconnect parameter

shamirahim111:
@sahilyeole the top says " #323  related but no concurrent user sessions"  what does that mean?

if 2 people try to log into a windows host using user 1 and user 2, will both get different "user accounts" open like rdp can do or will both share the same session?
what about multiple people?

21pages (changes requested):

src/server/connection.rs
```diff
@@ -384,6 +386,7 @@ impl Connection {
             file_remove_log_control: FileRemoveLogControl::new(id),
             #[cfg(feature = "gpucodec")]
             supported_encoding_flag: (false, None),
+            user_session_id: None,
```
21pages, line 389: https://learn.microsoft.com/en-us/windows/win32/devnotes/getting-the-session-id-of-the-current-process
You can try this if you didn't, and confirm if the user still needs to reconnect if they select the same session ID.

src/server/connection.rs
beelchester: I needed `is_usid_changed` to be stored on the server side to ensure it does not reconnect on loop.
Is there any better way we can handle this?

src/server/connection.rs
beelchester: Got it, Thanks.

sahilyeole:
> if 2 people try to log into a windows host using user 1 and user 2, will both get different "user accounts" open like rdp can do or will both share the same session?
what about multiple people?

Only one user session can be connected at one time for now.

src/server/connection.rs
21pages: If there are 2 connections, the LocalConfig options may be overridden, comparing session ids is one possible way to do it.

flutter/lib/common/widgets/dialog.dart
beelchester: Making it a parameter does not work because there is one more reconnection happening after the dialog's sessionReconnect on this line of connection.rs from the server:
 ```rust
 self.send_close_reason_no_retry("Restarting...").await;
 ```
I needed to add this to let the server process restart (with the specified user id). Otherwise, the connection was getting crashed.
After this change (of sessionReconnect's parameter), when there's another attempt to reconnect (after selecting, triggered by server), the usid will be None again, and the dialog will be shown in a loop.
Please let me know if I misunderstood.

src/server/connection.rs
beelchester: > comparing session ids is one possible way to do it.

Worked well, thanks.

src/server/connection.rs
beelchester: What can I use to make it a retry message?
We need this to be able to reconnect without user interaction automatically. This already seems fine.
Screenshot 2024-01-29 at 7 13 30 AM

21pages: `send_close_reason_no_retry`'s function name doen't match its real effect.You can try `on_close`.

flutter/lib/common/widgets/dialog.dart
21pages: Is this the correct order?
client -> server: LoginRequest
server->client: RdpUserSession
client->server: client choose session id, `bind.sessionReconnect`, loginRequest with session id
server->client: send `self.send_close_reason_no_retry("Restarting...").await;` to make client reconnect and then reset `--server`
client->server: reconnect

beelchester: Yes, it is correct.

21pages: Saving the chosen session id to LoginConfigHandler, it won't change during reconnection.

sahilyeole:
@21pages I have made the changes, please check.

src/platform/windows.cc
```diff
@@ -434,6 +434,25 @@ extern "C"
...
+    uint32_t get_current_process_session_id(PWSTR bufin, uint32_t nin)
+    {
+        uint32_t nout = 0;
+        PWSTR buf = NULL;
+        DWORD n = 0;
+        DWORD sessionId = 0;
+        HANDLE hProcess = GetCurrentProcess();
+        if (hProcess) {
+            ProcessIdToSessionId(GetCurrentProcessId(), &sessionId);
+            if (buf) {
```
21pages: buf is unused

src/ui_session_interface.rs
```diff
@@ -1023,6 +1023,13 @@ impl<T: InvokeUiSession> Session<T> {
         if true == force_relay {
             self.lc.write().unwrap().force_relay = true;
         }
+        if !user_session_id.is_empty() {
+            let mut config = self.load_config();
+            config
+                .options
+                .insert("selected_user_session_id".to_string(), user_session_id);
+            self.save_config(config);
+        }
```
21pages, line 1028: Works fine, the original meaning is not to store runtime variables in file , using the same method as force_relay.

21pages (changes requested):
Reconnection for same session and file transfer repeating session choosing dialog are fixed, and file tranfer support choose session now, great work.
 I didn't test port forward, windows server 2019 report 
`Bad packet length 1349676916.
ssh_dispatch_run_fatal: Connection to 127.0.0.1 port 1237: message authentication code incorrect` in nightly build, but works in a previous version, has nothing to do with this pr.

Did you test port forward?

src/server/connection.rs
```diff
@@ -1491,8 +1493,51 @@ impl Connection {
...
+                res.rdp_user_sessions = rdp;
+                let mut msg_out = Message::new();
+                msg_out.set_login_response(res);
+                self.send(msg_out).await;
+            }
+            if usid != self.user_session_id && usid != Some(current_process_usid) {
+                self.on_close("Reconnecting...", false).await;
+                std::thread::spawn(move || {
+                    let _ = ipc::connect_to_user_session(usid);
+                });
```
21pages, line 1532: Works fine, just a suggestion, `on_close` always exits the connection, it seems more reliable for `handle_multiple_user_sessions` to return false here and exit the connection without relaying on `--server` restart.

sahilyeole:
> Reconnection for same session and file transfer repeating session choosing dialog are fixed, and file tranfer support choose session now, great work.

Thanks :v:

> Did you test port forward?

No, I didn't. I do not have a setup for it.

rustdesk:
Port forward does not need to choose session.

src/ui/remote.rs
```diff
@@ -376,7 +386,7 @@ impl sciter::EventHandler for SciterSession {
                     let site = AssetPtr::adopt(ptr as *mut video_destination);
                     log::debug!("[video] start video");
                     *VIDEO.lock().unwrap() = Some(site);
-                    self.reconnect(false);
+                    self.reconnect(false, "".to_string());
```
21pages, line 389: I may have been pointed in the wrong direction, now the sciter version of the peer options code exists but is not used, it's hard to pass session id to reconnect funciton, the session selection dialog pops up repeatedly when connecting, may have to back to the peer options way both version or sciter version.

21pages (changes requested):

src/ui/remote.rs
beelchester: Sure, no problem.

beelchester: @21pages can you please test the new changes? The public server seems to be down.
I tested the changes yesterday, it was working. But I haven't tested thoroughly.

21pages: No problem

src/server/connection.rs
```diff
@@ -1491,8 +1493,55 @@ impl Connection {
         self.video_ack_required = lr.video_ack_required;
     }
 
+    #[cfg(target_os = "windows")]
+    async fn handle_multiple_user_sessions(&mut self, lr: LoginRequest) -> bool {
+        if self.port_forward_socket.is_some() {
+            return true;
```
21pages, line 1501: `self.port_forward_socket` is always none here, you should judge with `lr.union`

21pages (changes requested):

src/ui/remote.rs
beelchester: Sciter is working now

https://github.com/rustdesk/rustdesk/assets/73148455/3d200a66-74ab-4621-8472-06883a5fdc1c

flutter/lib/mobile/pages/remote_page.dart
```diff
@@ -78,6 +78,8 @@ class _RemotePageState extends State<RemotePage> {
   Future<void> dispose() async {
     // https://github.com/flutter/flutter/issues/64935
     super.dispose();
+    bind.sessionPeerOption(
+        sessionId: sessionId, name: 'selected_user_session_id', value: "");
```
21pages: seems these code can be removed.

21pages (changes requested):

21pages (approved):

21pages:
Another question is that the user can change the session ID without authenticating, but this also avoids double verification of the password.

sahilyeole:
Should I make the select session dialog appear after the password has been entered and verified?
But I think this way will prompt the user for a password 3 times.

21pages:
authenticating first

src/server/connection.rs
```diff
@@ -1817,6 +1811,15 @@ impl Connection {
...
         } else if self.authorized {
+            #[cfg(target_os = "windows")]
+            if crate::platform::is_installed()
+                && crate::platform::is_share_rdp()
+                && !*CONN_COUNT.lock().unwrap() > 1
+            {
+                if !self.handle_multiple_user_sessions(self.lr.clone()).await {
+                    return false;
+                }
+            }
```
21pages, line 1839: What about check only one time?

21pages (changes requested):

sahilyeole:
Should we add a check for the client version? I think it will cause a problem if client is using the older version and server is on the new version.

src/server/connection.rs
```diff
@@ -1815,10 +1817,12 @@ impl Connection {
             if crate::platform::is_installed()
                 && crate::platform::is_share_rdp()
                 && !*CONN_COUNT.lock().unwrap() > 1
+                && !self.checked_multiple_session
             {
                 if !self.handle_multiple_user_sessions(self.lr.clone()).await {
                     return false;
                 }
+                self.checked_multiple_session = true;
             }
```
21pages, line 1839:
```
if !self.checked_multiple_session
{
	self.checked_multiple_session = true;
	if crate::platform::is_installed()
		&& crate::platform::is_share_rdp()
		&& !*CONN_COUNT.lock().unwrap() > 1 {
		    if !self.handle_multiple_user_sessions(self.lr.clone()).await {
                          return false;
		}
	}
}
```
 This will be called on every message, optimization will be worth it, and it only needs to be checked when the connection authorization has just succeeded. Please do some optimization like code above if it works.

21pages (changes requested):

21pages:
> Should we add a check for the client version? I think it will cause a problem if client is using the older version and server is on the new version (with multiple user sessions).

What problem?

sahilyeole:
> > Should we add a check for the client version? I think it will cause a problem if client is using the older version and server is on the new version (with multiple user sessions).
> 
> What problem? Add version check is better.

It was reconnecting on a loop. I added the check, now it is fixed.

libs/hbb_common/protos/message.proto
```diff
@@ -124,11 +124,17 @@ message PeerInfo {
...
+  string user_session_id = 1;  
+  string user_name = 2;  
+}  
+
 message LoginResponse {
   oneof union {
     string error = 1;
     PeerInfo peer_info = 2;
   }
+  repeated RdpUserSession rdp_user_sessions = 3;
```
21pages: `rdp_user_sessions` can be moved out of `LoginResponse`

21pages (changes requested):

src/server/connection.rs
```diff
@@ -1491,8 +1495,55 @@ impl Connection {
...
+                    let mut rdp_session = RdpUserSession::new();
+                    rdp_session.user_session_id = u_sid.clone();
+                    rdp_session.user_name = u_name.clone();
+                    rdp.push(rdp_session);
+                }
+                res.rdp_user_sessions = rdp;
+                let mut msg_out = Message::new();
+                msg_out.set_login_response(res);
+                self.send(msg_out).await;
+                return false;
```
21pages: Why is false returned here?

beelchester: I thought the connection stops there so false would be correct. Should I make it true?

21pages: The server sends all sessions to the client here, and then disconnects itself. I think this may be the reason why the old version reconnects on a loop. You can try returning true, and see what happens.

libs/hbb_common/protos/message.proto
beelchester: Where should I place it? I tried placing it in `Message` it gives error `label not allowed`.
I was asked to use repeated here https://github.com/rustdesk/rustdesk/pull/6825#discussion_r1445021973
Also, if I move it out will this part work? As this is part of login response.
```rust
      res.rdp_user_sessions = rdp;
      let mut msg_out = Message::new();
      msg_out.set_login_response(res);
      self.send(msg_out).await;

Comment (21pages) [reply libs/hbb_common/protos/message.proto]: The login has ended, user is authorized already. This can be treated as a normal message. You can try putting `RdpUserSessions ` in `Misc`.  `repeated` is used to express arrays.  `label not allowed`: maybe oneof doesn't allow repeated

```
message RdpUserSessions {
 repeated RdpUserSession rdp_user_sessions = 1;
}
```

File: libs/hbb_common/protos/message.proto [hunk 27]
```diff
@@ -742,6 +741,7 @@ message Misc {
     TogglePrivacyMode toggle_privacy_mode = 33;
     SupportedEncoding supported_encoding = 34;
   }
+  repeated RdpUserSession rdp_user_sessions = 35;
```
21pages:
```
message RdpUserSessions {
 repeated RdpUserSession rdp_user_sessions = 1;
}
message Misc {
  oneof union {
   ...
    RdpUserSessions rdp_user_sessions = 35;
  }
}
```
You can try this.

21pages (changes requested):

libs/hbb_common/protos/message.proto
beelchester: If I do that I am not able to set rdp user sessions.
Screenshot 2024-02-02 at 1 24 13 PM

I think it is fine now with having it outside the union as repeated don't work inside it.

21pages: You can imitate `CaptureDisplays`

src/server/connection.rs
```diff
@@ -1491,8 +1495,58 @@ impl Connection {
...
+                self.user_session_id = usid.clone();
+                return true;
+            }
+            true
+        }
+    }
+
     #[cfg(not(any(target_os = "android", target_os = "ios")))]
     fn try_start_cm_ipc(&mut self) {
+        let usid = self.user_session_id;
```
21pages, line 1549: cm is always in the console session, related to auth order changes.

21pages, line 1541: cm is always in the console session, related to auth order changes.

sahilyeole: Got it. If there are any more other changes required let me know, I will finish it tonight.

src/server/connection.rs
```diff
@@ -1491,8 +1495,50 @@ impl Connection {
...
+                });
+                return false;
+            }
+            true
+        }
+    }
+
     #[cfg(not(any(target_os = "android", target_os = "ios")))]
     fn try_start_cm_ipc(&mut self) {
+        let usid = self.user_session_id;
```
21pages, line 1541: If there is a console session and an rdp session, 1.2.4 choose rdp session and then close the connection, the rustdesk server is still running on the rdp session, then an older version client connects, it will connect to the rdp session, but `self.user_session_id` is none and cm will be started on the console session. Just make cm/tray's session ID always the same as --server's session ID. `crate::platform::get_current_process_session_id()` can be used.

21pages (changes requested):

src/server/connection.rs
21pages: Switching between console and rdp for servel times, cm show in the wrong session.

src/server/connection.rs
```diff
@@ -1573,14 +1573,7 @@ impl Connection {
...
                 if !self.checked_multiple_session {
-                    let usid;
-                    match lr.option.user_session.parse::<u32>() {
-                        Ok(n) => usid = Some(n),
-                        Err(..) => usid = None,
-                    }
-                    if usid.is_some() {
-                        self.user_session_id = usid;
-                    }
+                    self.user_session_id = Some(crate::platform::get_current_process_session_id());
```
21pages: Latest version on both sides, switching between console and rdp session for several times, cm shows up in the wrong session.

21pages (changes requested):

src/server/connection.rs
```diff
@@ -1513,16 +1560,22 @@ impl Connection {
...
     }
 
     async fn on_message(&mut self, msg: Message) -> bool {
         if let Some(message::Union::LoginRequest(lr)) = msg.union {
+            #[cfg(target_os = "windows")]
+            {
+                if !self.checked_multiple_session {
+                    self.user_session_id = Some(crate::platform::get_current_process_session_id());
+                }
+            }
```
21pages: The problem is still existing. 

One way is to remove field `user_session_id ` and test `crate::platform::run_as_user(args.clone(), Some(crate::platform::get_current_process_session_id()));`

21pages (changes requested):

src/server/connection.rs
beelchester: It is still happening. But I think we can ignore it because in my testing it only occurred once in a while when rapidly reconnecting to different user sessions, it doesn't occur if there's a slight delay between connections. Maybe because server process id is not changed immediately when `run_as_user` is called.
I still did `crate::platform::run_as_user(args.clone(), Some(crate::platform::get_current_process_session_id()));` because it is cleaner.

21pages: The reason may be that cm does not exit so quickly. Please print some log and find the reason and a solution in order to avoid safety hazards.

beelchester: `get_current_process_session_id();` was giving incorrect id once in a while on repeatedly changing the sessions between connections. But since this problem was occurring only on changing sessions (i.e. on 1.2.4) I changed it back to how it was a few commits ago where I used the usid chosen by the user for `run_as_user`. And for older versions (i.e. when usid is none) I made it use `get_current_process_session_id();`.
I have tested both 1.2.4 and the older version, and now cm is working as expected.

src/server/connection.rs
```diff
@@ -3139,10 +3151,7 @@ async fn start_ipc(
                 #[cfg(target_os = "windows")]
                 {
                     log::debug!("Start cm");
-                    res = crate::platform::run_as_user(
-                        args.clone(),
-                        Some(crate::platform::get_current_process_session_id()),
-                    );
+                    res = crate::platform::run_as_user(args.clone(), user_session_id);
```
21pages, line 3154: https://github.com/rustdesk/rustdesk/pull/6825#discussion_r1476928541 Initially it was to solve this problem.

beelchester: https://github.com/rustdesk/rustdesk/pull/6825#discussion_r1479359255

This took care of it:
```rust
let mut session_id = get_current_process_session_id();
```

src/server/connection.rs
```diff
@@ -1761,6 +1821,22 @@ impl Connection {
                 }
             }
         } else if self.authorized {
+            #[cfg(target_os = "windows")]
+            if !self.checked_multiple_session {
+                self.checked_multiple_session = true;
+                if crate::platform::is_installed()
+                    && crate::platform::is_share_rdp()
+                    && !*CONN_COUNT.lock().unwrap() > 1
```
21pages, line 1829: when the second client connects, the session chosen dialog still pops up, it's not a big problem, but if the user choose a session different from the current one, the two clients will reconnect in a loop.

src/server/connection.rs
21pages: > get_current_process_session_id(); was giving incorrect id once in a while on repeatedly changing the sessions between connections

Please dig it out, because `get_current_process_session_id` is used in `run_as_user` now to replace `get_current_session(share_rdp())`

src/server/connection.rs
21pages: When the second client connects, the session chosen dialog still pops up, it's not a big problem, but if the user choose a session different from the current one, the two clients will reconnect in a loop.

And could you update the workflow on the top of the pr? Describe critical steps in more detail, including conditions, what messages to send, disconnection and reconnection.

21pages (commented):
I can't reproduce cm in the wrong session now.

src/server/connection.rs
beelchester: > get_current_process_session_id(); was giving incorrect id once in a while on repeatedly changing the sessions between connections

I think it's related to how windows handles processes, it might be due to some short-term cache.
https://www.geeksforgeeks.org/context-switch-in-operating-system/

It is not a problem after my last commit, since we are not using `get_current_process_session_id` in case of switching user sessions. I have tested thoroughly and got the correct session id every time.

src/server/connection.rs
beelchester: Why does CONN_COUNT not work, it has been used in other places too. Is there anything else I can use to ensure only one connection is allowed for multiple session dialog?

> And could you update the workflow on the top of the pr? Describe critical steps in more detail, including conditions, what messages to send, disconnection and reconnection.

How about this image?
Screenshot 2024-02-07 at 11 29 19 AM

21pages: `CONN_COUNT` is 2 in log.

rustdesk:
@21pages take over this, no need to waste time on communication.

sahilyeole:
@21pages I have fixed the multiple connections problem. Now the user sessions dialog doesn't appear when another connection exists on the server side.
I think it should be ready to merge now since I have completed all the required changes and there is no issue in the working of this feature. If there are more particular changes you wish to add you may push to this PR or do it later after merging.

rustdesk:
Thanks for your effort.

src/platform/windows.cc
```diff
@@ -433,6 +434,74 @@ extern "C"
...
+        std::wstring tmpStr;
+        for (size_t i = 0; i < sessionIds.size(); i++) {
+            if (i > 0) {
+                tmpStr += L",";
+            }
+            tmpStr += sessionIds[i];
+        }
+
+        if (buf && !tmpStr.empty() && tmpStr.size() < bufSize) {
+            memcpy(buf, tmpStr.c_str(), (tmpStr.size() + 1) * sizeof(wchar_t));
```
rustdesk, line 501: Change to wcscpy here.

src/platform/windows.cc
```diff
@@ -433,6 +434,74 @@ extern "C"
...
+                memcpy(bufin, buf, nout);
+                WTSFreeMemory(buf);
+            }
+        }
+        return nout;
+    }
+
+    void get_available_session_ids(PWSTR buf, uint32_t bufSize, BOOL include_rdp) {
+        std::vector<std::wstring> sessionIds;
+        PWTS_SESSION_INFOA pInfos;
```
rustdesk, line 467: set `pInfos = NULL` here so hat `WTSFreeMemory(pInfos);` more safer

src/platform/windows.rs
```diff
@@ -684,10 +712,10 @@ pub fn try_change_desktop() -> bool {
 }
 
 fn share_rdp() -> BOOL {
-    if get_reg("share_rdp") != "true" {
-        FALSE
-    } else {
+    if get_reg("share_rdp") != "false" {
         TRUE
+    } else {
+        FALSE
```
rustdesk, line 718: Make sure corresponding option in settings UI page also changed, and test please.

rustdesk:
@21pages I will merge first, later please make the code code for above code review.

PR #6856, opened 2024-01-12
Warn on MIT-SHM not working on Linux X11

wesley800:
This PR simply throws a warn into log, when rustdesk server runs on an Linux box with its X11 server having MIT-SHM set to disabled. Plus another minor fix on pointer usage.
Personally I'm very unsatisfied about the current way of passing MIT-SHM errors from `scrap` to the server. Need some hint on API designing flavors.

### DIFF
--- a/src/server/video_service.rs
+++ b/src/server/video_service.rs
@@ -193,17 +193,22 @@ fn create_capturer(
     match c {
         Some(c1) => return Ok(c1),
         None => {
-            log::debug!("Create capturer dxgi|gdi");
             #[cfg(windows)]
-            return crate::portable_service::client::create_capturer(
-                _current,
-                display,
-                _portable_service_running,
-            );
+            {
+                log::debug!("Create capturer dxgi|gdi");
+                return crate::portable_service::client::create_capturer(
+                    _current,
+                    display,
+                    _portable_service_running,
+                );
+            }
             #[cfg(not(windows))]
-            return Ok(Box::new(
-                Capturer::new(display).with_context(|| "Failed to create capturer")?,
-            ));
+            {
+                log::debug!("Create capturer from scrap");
+                return Ok(Box::new(
+                    Capturer::new(display).with_context(|| "Failed to create capturer")?,
+                ));
+            }
         }
     };
 }
--- a/src/server/video_service.rs
+++ b/src/server/video_service.rs
@@ -310,6 +310,7 @@ fn get_capturer(current: usize, portable_service_running: bool) -> ResultType<Ca
     }
     let display = displays.remove(current);
 
+    #[cfg(target_os = "linux")]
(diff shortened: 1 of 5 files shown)


rustdesk:
This fixes bug or adds feature?

wesley800:
> This fixes bug or adds feature?

I believe it's a feature, but only for debugging.

rustdesk:
Debug for what?

wesley800:
Currently when rustdesk runs server on SHM disabled environment, the client only sees black screen after connecting to the server, while keyboard and mouse works as normal. It is at least to me very confusing, when there's no relevant lines in rustdesk's log.
I tried to minimize the change. It also sounds reasonable to me to just throw a fatal error, cuz `scrap` requires SHM explicitly.

rustdesk:
It is better to show here.

image

https://github.com/rustdesk/rustdesk/blob/286181ca04c43455ccf183badcf418cc9edbc231/flutter/lib/desktop/pages/desktop_home_page.dart#L324

wesley800:
I agree to add a warn on GUI interface. What about headless setup? And I'm sorry that I can't write in flutter...

src/server/video_service.rs
```diff
@@ -304,6 +309,16 @@ fn get_capturer(current: usize, portable_service_running: bool) -> ResultType<Ca
         );
     }
     let display = displays.remove(current);
+
+    if let Display::X11(inner) = &display {
+        if let Err(err) = inner.get_shm_status() {
```
rustdesk, line 315: I do not think this can be compiled on non-Linux.

rustdesk (changes requested):

wesley800:
Could you explain a bit on why it's closed?

rustdesk:
You did not answer https://github.com/rustdesk/rustdesk/pull/6856#discussion_r1452037313. I thought I had given up.

wesley800:
I was expecting for ideas on how to pass the warns around, rather than a small `#[cfg]` modification. As I stated I have a strong feel that the PR needs refactor. Current commits are just for quick illustration and proof.

To be more specific, would you mind changing API of `scrap`? e.g. let `Capturer::new` returns `Result<(Capturer,WarningKind),ErrorKind>`, rather than `Result<Capturer,ErrorKind>`?

If API changing is not preferred, would you prefer the state (`shm_status` field) to be stored in `Server` instance, or checks on the fly each time `get_shm_status` is called?

rustdesk:
I prefer `#[cfg]`

wesley800:
BTW I'll try to add a tooltip to GUI, but it will take some time.

rustdesk:
👍

109-999:
👍

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #6939, opened 2024-01-21
video handler holds only one decoder of the current codec format

21pages:
1. For example: when receiving h264 video frames, only one h264 decoder is created, priority: vram > ram
2. For creation and decoding failed:

  * Remove real_supported_decodings, this will update real existing decoders, replace it with the `mark_unsupported` vector. After creating the decoder failure, marks the codec as unsupported and updates supported decoding to the controlled side
  *  Add `fail_counter` in the decoder. When decoding 10 consecutive frames failed, adding codec type to `mark_unsupported` vector
  *  The controlled end always ignores the unavailability of VP9

test:
* 1.2.4 control 1.1.9~1.2.4
* In this test video, the default codec  is AV1, let AV1 and H264 fail to be created, and let H265 fail to decode after 100 frames.

https://github.com/rustdesk/rustdesk/assets/14891774/4f033b03-33b0-4c19-b340-731b1842900d

### DIFF
--- a/libs/scrap/examples/benchmark.rs
+++ b/libs/scrap/examples/benchmark.rs
@@ -240,7 +240,10 @@ fn test_av1(
 #[cfg(feature = "hwcodec")]
 mod hw {
     use hwcodec::ffmpeg::CodecInfo;
-    use scrap::hwcodec::{HwDecoder, HwEncoder, HwEncoderConfig};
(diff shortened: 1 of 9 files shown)


rustdesk:
Add log::info! when one decoder created

PR #6980, opened 2024-01-24
Fix: Mac retina display video blur

### LINKED ISSUE #5246: Video encoding downsamples retina resolutions
### Bug Description

When connecting to a macOS device with a retina display, the video encoding uses the non-retina resolution.
When I connect from a Windows system with a high-density display to macOS system with a retina display, I end up getting a very blurry image (scaled down on the server from 2x to 1x and then scaled back up on the client).

### How to Reproduce

Install RustDesk on a macOS laptop with a retina display
Connect from a Windows machine with a 150% scaled display (or higher)
Notice that "Scale Original" results in a tiny display at the non-retina resolution
Notice that "Scale Adaptive" results in a blurry display

Also notice that there is no difference in RustDesk's apparent resolution when switching the macOS device from e.g. "1920x1080" to "1920x1080 low resolution"

### Expected Behavior

At least optionally, the macOS server should be encoding the full resolution display

Win11 -> macOS

1.2.1 -> 1.2.1

### Screenshots

Screenshot 2023-08-03 111354

### Additional Context

_No response_

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/rustdesk) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/rustdesk/rustdesk/issues/5246">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/rustdesk/rustdesk/issues/5246/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

beelchester:
Fixes: #5246 
Working: https://github.com/rustdesk/rustdesk/issues/5246#issuecomment-1907200031

`BackingScaleFactor` is 2.0 if it is a retina display. [Doc](https://developer.apple.com/documentation/appkit/nswindow/1419459-backingscalefactor)

# This merge has been reverted because causing more severe bug

### DIFF
--- a/libs/scrap/src/quartz/display.rs
+++ b/libs/scrap/src/quartz/display.rs
@@ -36,8 +36,9 @@ impl Display {
 
     pub fn width(self) -> usize {
         let w = unsafe { CGDisplayPixelsWide(self.0) };
-        if self.is_retina() {
-            w * 2
+        let s = self.scale();
+        if s > 1.0 {
+           ((w as f64) * s) as usize
         } else {
             w
         }
@@ -45,8 +46,9 @@ impl Display {
 
     pub fn height(self) -> usize {
         let h = unsafe { CGDisplayPixelsHigh(self.0) };
-        if self.is_retina() {
-            h * 2
+        let s = self.scale();
+        if s > 1.0 {
+           ((h as f64) * s) as usize
         } else {
             h
         }
@@ -68,8 +70,8 @@ impl Display {
         unsafe { CGDisplayIsOnline(self.0) != 0 }
     }
 
-    pub fn is_retina(self) -> bool {
-        unsafe { BackingScaleFactor() == 2.0 }
+    pub fn scale(self) -> f64 {
+        unsafe { BackingScaleFactor() } as _
     }
 
     pub fn bounds(self) -> CGRect {
--- a/libs/scrap/src/quartz/display.rs
+++ b/libs/scrap/src/quartz/display.rs
@@ -35,11 +35,21 @@ impl Display {
     }
 
     pub fn width(self) -> usize {
-        unsafe { CGDisplayPixelsWide(self.0) }
+        let w = unsafe { CGDisplayPixelsWide(self.0) };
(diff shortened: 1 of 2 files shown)


libs/scrap/src/quartz/display.rs
```diff
@@ -58,6 +68,10 @@ impl Display {
         unsafe { CGDisplayIsOnline(self.0) != 0 }
     }
 
+    pub fn is_retina(self) -> bool {
+        unsafe { BackingScaleFactor() == 2.0 }
```
rustdesk: It can not be 1.5, 2.5, 3?

rustdesk: Why not use > 1.0?

libs/scrap/src/quartz/display.rs
```diff
@@ -35,11 +35,21 @@ impl Display {
...
+        } else {
+            w
+        }
     }
 
     pub fn height(self) -> usize {
-        unsafe { CGDisplayPixelsHigh(self.0) }
+        let h = unsafe { CGDisplayPixelsHigh(self.0) };
+        if self.is_retina() {
+            h * 2
```
rustdesk: Why not use BackingScaleFactor directly?

rustdesk (changes requested):

zhangbo8418:
This will cause the mouse pointer to not hit the correct location. . .

rustdesk:
Revert for the time being

rustdesk:
@sahilyeole we will take care of this, because it is a little complex now.

rustdesk:
@21pages follow this up, let's translate the mouse event x/y in connection.rs when receiving mouse event.

21pages:
@zhangbo8418 Could you try https://github.com/21pages/rustdesk/releases/download/nightly/rustdesk-1.2.4-x86_64-retina.dmg ?

PR #7068, opened 2024-02-06
Refact. Disable idd driver

fufesou:


danielmarschall:
Why was this pull request merged, although the driver will be signed with a trusted certificate soon? (I can't find the post anymore. Someone offered to buy a EV certificate)

rustdesk:
Getting a EV will need 1 month or more.

danielmarschall:
> Getting a EV will need 1 month or more, you thought everything is simple, plese trust us we are not stupid. Stop push us please. We know what we do, and really hate to answer repeated questions and explain everyting.

I'm sorry, but I didn't found any post where you explained this. Actually, the posts that mention the EV certificate have been removed from the thread, because I know they were there, but are now deleted. This is very untransparent, and now you blame me because I asked a question.

Also, I did not find any information (or got a reply to my question) what the purpose of that IDD driver is. I installed the latest version that has the IDD driver removed, and I can still connect to servers without attached screens -  it works exactly like before. So why do we need a virtual display driver at all? Also, I cannot find any documentation about this...

fufesou:
@danielmarschall Thanks for your patient. 

We have to temporarily remove idd. Because it does cause trouble, and may cause more later.

>  I can still connect to servers without attached screens - it works exactly like before.

For connections: If you have a virtual display adapter, there is no need to create a virtual display. You can find the displays info here

image


The Idd driver is used on machines that don't even have a virtual display in this case.

> So why do we need a virtual display driver at all? Also, I cannot find any documentation about this...

A virtual display driver provides ability to another two features:

1. Privacy mode based on a virtual display.
2. Use multiple displays on machines that do not have so many physical monitors.

Sorry for causing confusion to you.

PR #7173, opened 2024-02-18
Refact. Replace all `tokio::time::interval()`

fufesou:
Replace all `tokio::time::interval()` with `ThrottledInterval` which checks `min_interval` when `tick()`.

### DIFF
--- a/src/client/io_loop.rs
+++ b/src/client/io_loop.rs
@@ -30,7 +30,7 @@ use hbb_common::{
     tokio::{
         self,
         sync::mpsc,
-        time::{self, Duration, Instant, Interval},
+        time::{self, Duration, Instant},
     },
     Stream,
 };
@@ -62,7 +62,7 @@ pub struct Remote<T: InvokeUiSession> {
     read_jobs: Vec<fs::TransferJob>,
     write_jobs: Vec<fs::TransferJob>,
     remove_jobs: HashMap<i32, RemoveJob>,
-    timer: Interval,
+    timer: crate::RustDeskInterval,
     last_update_jobs_status: (Instant, HashMap<i32, u64>),
     is_connected: bool,
     first_frame: bool,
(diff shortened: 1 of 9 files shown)


fufesou:
How about just use `tokio::time::sleep` instead of `tokio::time::Interval::tick` here.

Replace

```rust
pub struct ThrottledInterval {
    interval: Interval,
    last_tick: Instant,
    min_interval: Duration,
}

impl ThrottledInterval {
    pub fn new(i: Interval) -> ThrottledInterval {
        let period = i.period();
        ThrottledInterval {
            interval: i,
            last_tick: Instant::now() - period * 2,
            min_interval: Duration::from_secs_f64(period.as_secs_f64() * 0.9),
        }
    }

    pub async fn tick(&mut self) -> Instant {
        let instant = poll_fn(|cx| self.poll_tick(cx));
        instant.await
    }

    pub fn poll_tick(&mut self, cx: &mut std::task::Context<'_>) -> Poll<Instant> {
        match self.interval.poll_tick(cx) {
            Poll::Ready(instant) => {
                if self.last_tick.elapsed() >= self.min_interval {
                    self.last_tick = Instant::now();
                    Poll::Ready(instant)
                } else {
                    // This call is required since tokio 1.27
                    cx.waker().wake_by_ref();
                    Poll::Pending
                }
            }
            Poll::Pending => Poll::Pending,
        }
    }
}
```

to

```rust
pub struct ThrottledInterval {
    interval: f32,
}

impl ThrottledInterval {
    pub fn new(interval: f32) -> ThrottledInterval {
        ThrottledInterval { interval }
    }

    pub async fn tick(&mut self) -> Instant {
        hbb_common::sleep(interval).await;
        Instant::now()
    }
}
```

https://docs.rs/tokio/latest/tokio/time/struct.Interval.html#method.tick:~:text=This%20type%20allows%20you%20to%20wait%20on%20a%20sequence%20of%20instants%20with%20a%20certain%20duration%20between%20each%20instant.%20Unlike%20calling%20sleep%20in%20a%20loop%2C%20this%20lets%20you%20count%20the%20time%20spent%20between%20the%20calls%20to%20sleep%20as%20well.

PR #7184, opened 2024-02-19
refactor windows specific session, file transfer and waiting for image

21pages:
#7170
1. File transfer doesn't show directory until correct session id is ensured
2. Fix file transfer, caused by `pi.username = self.lc.read().unwrap().get_username(&pi);` in `handle_peer_info` override empty username and `get_active_username` doesn't return currect session username
* Fix home directory not change when session changed, or wrong home directory  (TODO: Not fix this through flutter, but sciter version still have problems, wrong home directory or empty remote directory)
* Fix show empty remote directory rather than error messagbox when current session is in login screen
3. Show `Connected, waiting for image` after user choose the same session id

File transfer delay show directory and correct home directory

https://github.com/rustdesk/rustdesk/assets/14891774/baacacc5-e9fa-49a3-a83b-02192bf4c063

File transfer console in login screen

https://github.com/rustdesk/rustdesk/assets/14891774/0fb1f398-bf54-4a63-bd5f-a056beba6aeb

Remote waiting for image

https://github.com/rustdesk/rustdesk/assets/14891774/3f96f494-9327-4c28-908e-49bc69b7857c


khoinp1012:
is this working on 1.2.4 ? I downloaded 1.2.4 nightly. But it does not work (installed on both server and client) ! Should I report it as bug because it is nightly version.

rustdesk:
Both side must install latest 1.2.4 (at least yesterday's nightly build).
if there is still issue, please show us your output of "Query User" on controlled machine. https://serverfault.com/questions/758930/how-can-i-view-active-remote-connections-rdp-to-a-windows-server

khoinp1012:
Ok thanks ! I restart the remote computer, and now it is working now ! Thank you !

PR #7229, opened 2024-02-22
shared address book

21pages:
https://github.com/rustdesk/rustdesk-server-pro/issues/49

With the current server version, the legacy address book will be used.
With the next server version, personal address book(named "My address book") and shared address books will be used.

# Details

### Migration
* The next version of the server will automatically migrate the address book, and the issue of address book loss will be fixed, and each address book will have its own recycle bin.
* After migration, the legacy address book will be named My address book, invalid IDs will be removed. This address book will never be shared.
 
| before migration| after migration |
| ------- | ------- |
|    | Only Accept exsiting ID, IP or ID with `@`, other ID will be removed, some tag color may change  |
| deae8fb2d31b9f6a570996c90ea6bae | 5618da5706accc5ef5426466a05a697 |

### Add ID/device

You can add the following devices to the shared address book:

* Device of the same team, eg: 1234567
* IP address, eg: 192.168.1.101
* ID of other servers, eg: 1234567@public, 1234567@1.1.1.1?key=abcd

When the added ID is not an IP address and does not contain '@', it will be checked whether the ID exists in the team. If not, the addition will fail.

### Shared address book

* Every person can create shared address books, and address books that haven't been shared are only visible to the administrator and oneself.

* How to share

https://github.com/rustdesk/rustdesk/assets/14891774/01a65510-e51e-4676-82dc-0bda78dc1511

* After sharing

https://github.com/rustdesk/rustdesk/assets/14891774/35b8a8c2-e5dd-4960-ae51-6b9976527656

* Permissions

  - Admin users always have full control permissions to all address books.
  - Address book owners always have full control permissions to his own address books.
  - Read-only:  Use address book
  - Read/Write: Change address devices and tags 
  - Full Control: Have the same permission as the owner, including re-share, rename address book, delete address book
  - Priority: User > Group > Everyone

### Password

* The shared password needs to be manually set in the menu.
* Devices with a shared password will display a key icon in the top left corner of the card.
* When connecting through the card, the corresponding shared password associated with that card will be used. However, when connecting through other means, the shared password will not be used.
* Password priority: There are currently locally saved passwords, passwords in the personal address book and passwords in the shared address book. When connecting through a shared address book card, the shared password is used and has the highest priority. If this shared password is wrong, the password will be re-entered when connecting. After a successful connection, the shared password will not be updated because this password can only be updated manually.
* When adding an ID from one address book to another, the password will not be added to.
* When connect to a device with shared password, the password will not save to recent peer config or personal address book.

https://github.com/rustdesk/rustdesk/assets/14891774/0b0eeba4-0d96-4857-9525-87be086d6a93

### Device information

* The device information of the ID in the team will be read directly from the server
* Other IDs rely on option `sync from recent`, and ID will not be automatically added from recent. 

### Audit

Add/delete address book, recycle bin operations and permission operations will be logged.


rustdesk:
todo:
- [ ] legacy ab
- [ ] https://github.com/rustdesk/rustdesk-server-pro/issues/186#issuecomment-1887221621

rustdesk:
Update: we are still working on this hard, may need 1 more week.

sesu-tech:
> Update: we are still working on this hard, may need 1 more week. Let's target releasing this by 15 Mar.

Thanks for the transparency! :)

Elilitha:
awesome update, 
is there a hard limit of 5 shared address books, or is that because I'm using a trial?

i would like to set up a shared address book for each company we manage

21pages:
@Elilitha It equals to max user number of your license.

Elilitha:
@21pages cool cool. 

Any plan to change that to unlimited in the future for the self-hosting plan?
We are a small business of 10 users looking after over 100 SMB's, 
if not, ill see what i can do with Tags

Regardless, shared addressbook is a huge bonus for us, we are likely going to move off TeamViewer once pretesting is completed

21pages:
https://github.com/rustdesk/rustdesk/discussions/7748#discussioncomment-9150692

21pages:
@Elilitha **Update**: No limit of the number of address books in the next version.

centropaghe:
is this for self hosted also?

PR #7269, opened 2024-02-27
enable retina scale factor

21pages:
https://github.com/rustdesk/rustdesk/pull/6980 https://github.com/rustdesk/rustdesk/issues/5246
* Enabled only when there are only one video service
* Scale mouse event, remote cursor position, remote menu display button, and adjust resolution menu

https://github.com/rustdesk/rustdesk/assets/14891774/1f20430b-0262-439b-b644-85243347dc14

multi window:

https://github.com/rustdesk/rustdesk/assets/14891774/af5b8c14-19bc-49dc-9628-5b614e1a2e56

Old version control mac retina screen
rustdesk_JrsBXWCcIA

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -283,6 +283,8 @@ impl Server {
                 s.on_subscribe(conn.clone());
             }
         }
+        #[cfg(target_os = "macos")]
+        self.set_enable_retina();
         self.connections.insert(conn.id(), conn);
         *CONN_COUNT.lock().unwrap() = self.connections.len();
     }
@@ -293,6 +295,8 @@ impl Server {
         }
         self.connections.remove(&conn.id());
         *CONN_COUNT.lock().unwrap() = self.connections.len();
+        #[cfg(target_os = "macos")]
+        self.set_enable_retina();
     }
 
     pub fn close_connections(&mut self) {
@@ -325,6 +329,8 @@ impl Server {
             } else {
                 s.on_unsubscribe(conn.id());
             }
+            #[cfg(target_os = "macos")]
+            self.set_enable_retina();
         }
     }
 
@@ -373,6 +379,17 @@ impl Server {
                 }
             }
         }
+    }
+
+    #[cfg(target_os = "macos")]
+    fn set_enable_retina(&self) {
+        let mut video_service_count = 0;
+        for (name, service) in self.services.iter() {
+            if Self::is_video_service_name(&name) && service.ok() {
+                video_service_count += 1;
+            }
+        }
+        *scrap::quartz::ENABLE_RETINA.lock().unwrap() = video_service_count < 2;
     }
 }
 
--- a/src/server.rs
+++ b/src/server.rs
@@ -284,7 +284,7 @@ impl Server {
             }
         }
         #[cfg(target_os = "macos")]
-        self.set_enable_retina();
+        self.update_enable_retina();
         self.connections.insert(conn.id(), conn);
         *CONN_COUNT.lock().unwrap() = self.connections.len();
     }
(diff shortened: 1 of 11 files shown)


src/server/connection.rs
```diff
@@ -651,14 +657,28 @@ impl Connection {
...
-                        Some(message::Union::PeerInfo(..)) => {
+                        Some(message::Union::PeerInfo(_pi)) => {
                             conn.refresh_video_display(None);
+                            #[cfg(target_os = "macos")]
+                            conn.retina.set_displays(&_pi.displays);
+                        }
+                        #[cfg(target_os = "macos")]
+                        Some(message::Union::CursorPosition(pos)) => {
+                            if let Some(msg) = conn.retina.on_cursor_pos(&pos, conn.display_idx) {
+                                sent = true;
```
rustdesk: why not rewrite msg here.
`msg = new_msg`

src/server.rs
```diff
@@ -374,6 +376,17 @@ impl Server {
...
+
+    #[cfg(target_os = "macos")]
+    fn set_enable_retina(&self) {
+        let mut video_service_count = 0;
+        for (name, service) in self.services.iter() {
+            if Self::is_video_service_name(&name) && service.ok() {
+                video_service_count += 1;
+            }
+        }
+        *scrap::quartz::ENABLE_RETINA.lock().unwrap() = video_service_count < 2;
```
rustdesk, line 392: So, the first one can be retina, then following subscription will be not-retina. If so, there is still issue for multiple window/display case, because the first display is retina, the other displays are not.

rustdesk (changes requested):

src/server.rs
rustdesk: Another worse case, the capture uses retina, but the subscriber's displays info shows non-retina.

21pages: `scrap::quartz::ENABLE_RETINA` is a global control and is used when querying the display size in display service in a loop , so "use retina as retina" or all non-retina. 
A display is retina when BackingScaleFactor > 1.0 and ENABLE_RETINA is true

rustdesk: You should not put here, you should put in connection.rs. because multiple connections with one subscription for each connection should still can use retina.

21pages: video_service_count is the number of video services with subscriber, or captured screen count. When multiple connections are connected to the same screen, they will share the same video service, and count is 1.

rustdesk: How about one subscribe to display 1, another one subscribe to display 2?

rustdesk: we can ignore this case.

21pages: video_service_count will be 2.

both display 1:
image

one connection display 1 and another disply 2

image

rustdesk: > How about one subscribe to display 1, another one subscribe to display 2?

For this, retina is fine. But we can ignore this case.

rustdesk:
I did not see you test multiple windows.

21pages:
> I did not see you test multiple windows.

Updated to the above, but found another problem. video_service_count can be 0 when there is a connection. It may be a sequence problem. I will find the reason.

src/server/connection.rs
```diff
@@ -629,7 +633,7 @@ impl Connection {
                 },
                 Some((instant, value)) = rx.recv() => {
                     let latency = instant.elapsed().as_millis() as i64;
-                    let msg: &Message = &value;
+                    let mut msg= value;
```
rustdesk: format

rustdesk (changes requested):

src/server/connection.rs
```diff
@@ -3488,6 +3509,51 @@ extern "C" fn connection_shutdown_hook() {
...
+            let mut pos = pos.clone();
+            pos.x = d.x + ((pos.x - d.x) as f64 * s) as i32;
+            pos.y = d.y + ((pos.y - d.y) as f64 * s) as i32;
+            let mut msg = Message::new();
+            msg.set_cursor_position(pos);
+            return Some(msg);
+        }
+        None
+    }
+}
```
rustdesk, line 3558: add #[inline] for above 3 functions

rustdesk (changes requested):

PR #7326, opened 2024-03-06
#7315 Fix lang detection for zh-TW

xlionjuan:
I use the sample code from [sys-locale](https://github.com/1Password/sys-locale).

````rust
use sys_locale::get_locale;

let locale = get_locale().unwrap_or_else(|| String::from("en-US"));

println!("The current locale is {}", locale);
````

It print this on my system:
```plaintext
The current locale is zh-TW
```


rustdesk:
Thank you very much.

PR #7429, opened 2024-03-19
Fix how 2FA codes are verified

p0358:
Currently the verification code simply verifies if the provided code is equal to current:
https://github.com/rustdesk/rustdesk/blob/b2290f5b4c3fd9233f6550b01c151401cdb3fed3/src/auth_2fa.rs#L90-L93

But the standard practice is to allow some skew (usually 1 before and after), to counteract desynchronized clocks or the user typing the code in the very last few seconds before clicking OK.

Instead of comparing to `generate_current()`, this dedicated method should be used:
https://docs.rs/totp-rs/latest/totp_rs/struct.TOTP.html#method.check_current
(accepts token code and returns boolean, which means simpler code too)

It accounts for skew, and skew defaults to 1 from what I see, meaning one below and one after:
https://github.com/rustdesk/rustdesk/blob/b2290f5b4c3fd9233f6550b01c151401cdb3fed3/src/auth_2fa.rs#L32

### DIFF
--- a/src/auth_2fa.rs
+++ b/src/auth_2fa.rs
@@ -89,16 +89,14 @@ pub fn generate2fa() -> String {
 
 pub fn verify2fa(code: String) -> bool {
     if let Some((info, totp)) = CURRENT_2FA.lock().unwrap().as_ref() {
-        if let Ok(cur) = totp.generate_current() {
-            let res = code == cur;
-            if res {
-                if let Ok(v) = info.into_string() {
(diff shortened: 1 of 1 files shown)


fufesou:
There's another place that can be improved.

https://github.com/rustdesk/rustdesk/blob/ecb70b43df51b194921e32440e07ba5ddf7620dc/src/server/connection.rs#L1753

PR #7619, opened 2024-04-05
Fix. Cert, remove strange cert folders .

fufesou:
RustDesk can remove test certs in the "Trusted Root Certification Authorities" on uninstall.

But 1.2.3-1 will still create unexpected cert folder, because the wrong parameter of `CertOpenSystemStoreW`

```rust
CertOpenSystemStoreW(0 as _, "ROOT\0".as_ptr() as _);
```

https://github.com/rustdesk/rustdesk/discussions/6444#discussioncomment-9017532

Thanks to @danielmarschall  for fingering out the cause and providing the code.

https://github.com/rustdesk/rustdesk/discussions/6444#discussioncomment-9010062
https://github.com/rustdesk/rustdesk/discussions/6444#discussioncomment-9014382

### DIFF
--- a/.github/workflows/flutter-build.yml
+++ b/.github/workflows/flutter-build.yml
@@ -8,7 +8,7 @@ on:
         default: true
       upload-tag:
         type: string
-        default: "1.2.3-1"
+        default: "1.2.3-2"
 
 env:
   CARGO_NDK_VERSION: "3.1.2"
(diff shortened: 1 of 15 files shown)


fufesou:
@danielmarschall Hi, you can compare the code in `windows_delete_test_cert.cc`.

I tried removing the fingerprint before deleting the directory.
Then delete only the directory itself and one-level subdirectories.

Maybe you have a better way of doing it.

danielmarschall:
> @danielmarschall Hi, you can compare the code in `windows_delete_test_cert.cc`.
> 
> I tried removing the fingerprint before deleting the directory. Then delete only the directory itself and one-level subdirectories.
> 
> Maybe you have a better way of doing it.

@fufesou Thank you for including the code!

I haven't compiled+executed the file, but looking at the code everything is fine.

Just two notes / requests:

(1) `RegDelnodeRecurseW()` with `bOneLevel=TRUE` has the problem that it will begin delete empty subdirs and then stop if a non-empty-subdir is found. The function then fails but has already deleted parts. This is not a problem in our case, because empty keys "Certificates", "CRLs", "CTLs" will be re-created if missing. However, you might want to add this note to the method header comment, because other people might use this method (like, copy paste this OpenSource code for their own projects, or someone from RustDesk might use this method in the future for something else).

(2) Maybe move the line `std::wcout << "Rogue Key Deleted!" << std::endl;` and the commented out line above into the if-statement, so that the message only appears if the deletion was successful. Or better this:

			//std::wcout << "Delete erroneous certificate store \"" << Complete << "\"" << std::endl; // TODO: Why does this break the console?
			std::wcout << "Delete erroneous certificate store... "; // no std::endl
			if (RegDelnodeW(RootKey, Complete, TRUE)) {
				std::wcout << "success" << std::endl;
				Index--; // Because index has moved due to the deletion
			} else {
				std::wcout << "failed" << std::endl;
			}

Issue #7667, opened 2024-04-09
No remote image, no cursor image and error message (Web Client - Mobile)


TBG-FR:
### Bug Description

Thanks to recent modifications to RustDesk Web Client, I managed to update it successfuly to latest version, and it works like a charm on Desktop 👏 However, I then tried on an Android tablet, and it looks broken 😞 

Upon connection : 
- **The whole screen is grey, we can't see remote screen**
- Mouse cursor image is broken (**however it moves, both on client and remote**)
- An error keeps spamming in logs, about a file not found (but this file is an UUID ?)

_I guess it's all related to the `Image`/`ImagePainter`, and the logs are related to each "refresh" of the image, but I can't say for sure, and I'm struggling (logs aren't that helpful on this error, but I still included the stacktrace below)_

### Web Client (Desktop) to Desktop app client => OK ✅ 
0. Open RustDesk Web Client on a Desktop computer (Windows), through your browser (Chrome), latest version
0. Connect to a remote client with Desktop app (Windows), enter the password
0. Everything is working well, you'll see the screen of the remote client, see the cursor, without errors in logs ✅ 

### Web Client (Mobile) to Desktop app client => KO 🛑 
0. Open RustDesk Web Client on a mobile device, like a tablet (Android), through your browser (Chrome), latest version
0. Connect to a remote client with Desktop app (Windows), enter the password
0. Not working at all, you'll see  a grey screen, with a broken image for the cursor, with errors spamming the logs 🛑 

### Expected Behavior

Remote Desktop should work well on mobile, as much as it works on desktop

(cc @fufesou I know you worked on the latest changes around Desktop part of the Web Client, can you provide any hints or guidance to fix the Mobile part ? I'm willing to do it, but I have hard time locating the issue... Even after reviewing both codes and your recent changes)

### Operating system(s) on local side and remote side

Web Client (Linux) -> Desktop (Windows 10)

### RustDesk Version(s) on local side and remote side

1.2.4 -> 1.2.3

### Screenshots

![Screenshot_20240409_181608_ChromeEdit](https://github.com/rustdesk/rustdesk/assets/18146038/b01c19f4-c488-45b5-bac6-be3711e2abd0)

### Additional Context

<h2>Browser console logs</h2>

```js
[...]

PR #7668, opened 2024-04-09
fix: do not call WakeLock on Web (mobile)

TBG-FR:
While doing tests and researchs for #7667 I stumbled upon a small issue

```js
Uncaught ReferenceError: "Wakelock is not defined"
```

I've done the same thing as in `flutter/lib/desktop/pages/remote_page.dart`, making `Wakelock` usage exclusive to non-web apps https://github.com/rustdesk/rustdesk/blob/master/flutter/lib/desktop/pages/remote_page.dart#L128


rustdesk:
But, I still prefer !isWeb

PR #7688, opened 2024-04-11
Feat/msi

fufesou:
1. Check if is self-installation.
2. Add firewall rule by custom action.


DmiRials:
Tell me, can I use additional installation parameters to specify, for example, my server or configuration file?
How can I do this?

fufesou:
Currently the only parameter is `INSTALLFOLDER`

```
msiexec /i package.msi /l*v install.log /qn INSTALLFOLDER="D:\Program Files\RustDesk\"
```

No additional parameters are supported for now. 

We may add some parameters in the future.

PR #7693, opened 2024-04-11
Fix. Msi. Terminate brokers.

fufesou:
1. Try terminate process "RuntimeBroker_rustdesk.exe".
2. Only do InstallValidate if is Install.
3. Create tray shortcut in the `StartupFolder`.
4. Move custom actions from `Package.wxs` to `RustDesk.wxs`.
5. Remove "RustDesk Tray.lnk" in Startmenu. Because it is rarely used.
6. Format `*.wsx` files.
7. Read the previous "stop-service" config.  Custom action in C++.
    a. Create and start services.
    b. Start tray.
    c. Create or delete tray link in the `StartupFolder`.

### DIFF
--- a/res/msi/CustomActions/Common.h
+++ b/res/msi/CustomActions/Common.h
@@ -1,0 +1,15 @@
+#pragma once
+
+#include <Windows.h>
+#include <string>
+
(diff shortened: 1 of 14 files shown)


rustdesk:
finish the todo

PR #7717, opened 2024-04-13
Feat: Follow remote cursor and window focus | Auto display switch

### LINKED ISSUE #7439: Auto Screen-switch / Mouse follow
### Discussed in https://github.com/rustdesk/rustdesk/discussions/7437

<div type='discussions-op-text'>

<sup>Originally posted by **Benno1308** March 20, 2024</sup>
### Short Summary
I would like to request a feature that allows the client to automatically follow the remote mouse cursor.

### Scenario
1. I'm connected to someone via RustDesk
2. The remote side has multiple monitors connected
3. The remote side is actively using his mouse across all monitors

### Expected Behaviour
Whenever the remote side moves the mouse cursor to a new monitor, 
I expect my RustDesk client to automatically follow the cursor to that monitor.

This can also be extended to be based on which window is currently focused, 
allowing to follow the remote work based on either where the mouse is, or which window is currently active.

### Bounty
I am willing to add a bounty of 750 USD via [polar.sh](https://polar.sh/rustdesk/rustdesk), or a similar service, if there is be a reasonable timeline to add this feature

### Reference
#2059 (Creating a seperate feature request as this one adds the request for follow window focus)

https://support.anydesk.com/knowledge/display
```
If the remote device has multiple monitors, enabling "Follow remote cursor" automatically switches the AnyDesk view to the correct monitor if the remote user moves the cursor to a different monitor.
If the remote device has multiple monitors, enabling "Follow remote window focus" automatically switches the AnyDesk view to the display where the foreground window is located.
```

</div>

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/rustdesk) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/rustdesk/rustdesk/issues/7439">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/rustdesk/rustdesk/issues/7439/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

beelchester:
### Mac:

https://github.com/rustdesk/rustdesk/assets/73148455/7f3b1e77-d954-4f3f-8764-c9d1c108ba00

### Linux:

https://github.com/rustdesk/rustdesk/assets/73148455/491a8ee5-9de9-4875-9f2d-7a1ce80ca49d

### Windows:

https://github.com/rustdesk/rustdesk/assets/73148455/6d01d75f-b1f2-40ed-bf25-9c50d5c8fcff


rustdesk:
@21pages please help review and test

src/client/io_loop.rs
```diff
@@ -1565,8 +1565,18 @@ impl<T: InvokeUiSession> Remote<T> {
                     }
                 }
                 Some(message::Union::PeerInfo(pi)) => {
-                    self.handler.set_displays(&pi.displays);
-                    self.handler.set_platform_additions(&pi.platform_additions);
+                    // For follow cursor, follow window focus
+                    if pi.displays.is_empty() && pi.platform_additions.is_empty() {
+                        let config = self.handler.load_config();
+                        if config.displays_as_individual_windows != "Y"
+                            && config.show_all_displays != "Y"
```
21pages: 1. When the previous connection ended with the option show_all_displays set to "Y", the new connection only displays one display, and the follow function doesn't work until you switch display manaully. Currently it is a runtime state not a config. the recommended approach is to use a session-level variable state instead of relying on the previous connection's settings. 

2. Opt:  Follow feature is used in the case of having only one display. When `displays_as_individual_windows` is enabled, there may be one display. If there are two displays, 'displays_as_individual_windows' can be deselected, so `displays_as_individual_windows ` can't represent display count. Because `displays_as_individual_windows` is only permitted when texture render is enabled, you can check the number of `map_display_sessions`  in flutter.rs to verify.

21pages (changes requested):

src/server/input_service.rs
```diff
@@ -352,6 +382,21 @@ fn run_cursor(sp: MouseCursorService, state: &mut StateCursor) -> ResultType<()>
     Ok(())
 }
 
+fn run_window_focus(sp: EmptyExtraFieldService, state: &mut StateWindowFocus) -> ResultType<()> {
+    let window_id = crate::get_focused_window_id();
+    if let Some(window_id) = window_id {
+        if state.is_changed(window_id) {
```
21pages: When `Follow remote cursor` is not enabled and `Follow remote window` is enabled, move focused window to another display,  it doesn't follow.

src/platform/windows.rs
```diff
@@ -66,6 +66,34 @@ use winreg::RegKey;
...
+        let hWnd = GetForegroundWindow();
+        let mut rect: RECT = mem::zeroed();
+        if GetWindowRect(hWnd, &mut rect as *mut RECT) == 0 {
+            return None;
+        }
+        displays.iter().position(|display| {
+            rect.left >= display.x
+                && rect.left <= display.x + display.width
+                && rect.top >= display.y
+                && rect.top <= display.y + display.height
```
21pages: AnyDesk uses the center of the focused window

src/server/connection.rs
```diff
@@ -3114,6 +3142,42 @@ impl Connection {
...
+                let mut msg_out = Message::new();
+                msg_out.set_peer_info(pi);
+                self.send(msg_out).await;
+            }
+        }
+    }
+
+    #[cfg(not(any(target_os = "android", target_os = "ios")))]
+    async fn handle_cursor_switch_display(&mut self, pos: CursorPosition) {
+        if let Ok(displays) = super::display_service::update_get_sync_displays().await {
```
21pages: Considering that it runs every 33 milliseconds, please test whether the cached display info in display_service.rs or the sent display information is sufficient for the execution.

21pages (changes requested):

src/platform/linux.rs
```diff
@@ -119,6 +119,41 @@ pub fn get_cursor_pos() -> Option<(i32, i32)> {
 
 pub fn reset_input_cache() {}
 
+pub fn get_focused_window_id() -> Option<i32> {
+    if let Ok(focused_window_id) = run_cmds("xdotool getactivewindow") {
```
21pages: xdotool is not installed on my ubuntu 22.04, it works afer installation.

21pages (changes requested):

src/platform/macos.rs
```diff
@@ -301,6 +306,43 @@ pub fn get_cursor_pos() -> Option<(i32, i32)> {
     */
 }
 
+pub fn get_focused_window_id() -> Option<i32> {
```
21pages: On my Mac it doesn't work, when I print out all the window ids and their layers I find that there are several windows with layer 0 and different window ids.

flutter/lib/common/widgets/setting_widgets.dart
```diff
@@ -212,6 +212,8 @@ List<(String, String)> otherDefaultSettings() {
     if ((isDesktop || isWebDesktop)) ('show_monitors_tip', kKeyShowMonitorsToolbar),
     if ((isDesktop || isWebDesktop)) ('Collapse toolbar', 'collapse_toolbar'),
     ('Show remote cursor', 'show_remote_cursor'),
+    ('Follow remote cursor', 'follow_remote_cursor'),
+    ('Follow remote window focus', 'follow_remote_window'),
```
21pages, line 216: Enabling these two default settings, but not "Show Remote Cursor", will not send the show remote cursor the first time you connect.

flutter/lib/common/widgets/toolbar.dart
```diff
@@ -399,6 +400,61 @@ Future<List<TToggleMenu>> toolbarDisplayToggle(
               }
             : null));
   }
+  // follow remote cursor
+  if (pi.platform != kPeerPlatformAndroid &&
```
21pages, line 408: 1. lower version check
2. multi display check
3. in view mode, these two options are sent, but they are unchecked.

src/server/connection.rs
```diff
@@ -3114,6 +3142,42 @@ impl Connection {
         self.inner.send(msg.into());
         self.supported_encoding_flag = (true, not_use);
     }
+
+    #[cfg(not(any(target_os = "android", target_os = "ios")))]
+    async fn handle_window_focus(&mut self) {
+        if let Ok(displays) = super::display_service::update_get_sync_displays().await {
+            let current_display = crate::get_focused_display(displays);
```
21pages: Instead of using the current_display in the peer info, it would be better to use a separate new message. Currently, the client handles it in `sync_peer_info` message, but the current_display doesn't actually change until the client sends a switch display command. So, it's not a true representation of peer information. What this message does is to send command to request the client to switch display, rather than sync the future display info.

src/client.rs
```diff
@@ -1456,6 +1467,22 @@ impl LoginConfigHandler {
                 BoolOption::No
             })
             .into();
+        } else if name == "follow-remote-cursor" {
```
21pages, line 1459: If it's not in sciter's menu, don't send it for sciter version, it would be better to add these two menu for sciter,

21pages (changes requested):

src/platform/linux.rs
beelchester: Hi, can you please try it out on your device now?
It should work now without xdotool.

21pages: Windows -> linux x11, follow menu doesn't show.
`IsMultiScreen` and `pi.isSupportMultiDisplay`  are used incorrectly.

21pages:
All platforms work now

Problems remaining:
https://github.com/rustdesk/rustdesk/pull/7717#discussion_r1564509789 1
https://github.com/rustdesk/rustdesk/pull/7717#discussion_r1566804048
https://github.com/rustdesk/rustdesk/pull/7717#discussion_r1564658989

Opt:
https://github.com/rustdesk/rustdesk/pull/7717#discussion_r1564509789 2
https://github.com/rustdesk/rustdesk/pull/7717#discussion_r1564662371
https://github.com/rustdesk/rustdesk/pull/7717#discussion_r1564662630
Ignore display changes at the edges of the screen, check AnyDesk

21pages:
For these optimizations, complete them or give your opinion.

sahilyeole:
Yes, I'm working on the sciter support.

>Ignore display changes at the edges of the screen, check whether AnyDesk has this behavior

AnyDesk doesn't have this behavior so I've skipped it.

> Opt: Follow feature is used in the case of having only one display. When displays_as_individual_windows is enabled, there may be one display. If there are two displays, 'displays_as_individual_windows' can be deselected, so displays_as_individual_windows  can't represent display count. Because displays_as_individual_windows is only permitted when texture render is enabled, you can check the number of map_display_sessions in flutter.rs to verify.

Since I've added a check for multi display for the follow options to show, do I still need to work on this?

src/client/io_loop.rs
```diff
@@ -1504,7 +1504,12 @@ impl<T: InvokeUiSession> Remote<T> {
                         log::info!("update supported encoding:{:?}", e);
                         self.handler.lc.write().unwrap().supported_encoding = e;
                     }
-
+                    Some(misc::Union::FollowCurrentDisplay(d_idx)) => {
+                        let config = self.handler.load_config();
+                        if config.displays_as_individual_windows != "Y" {
```
21pages: When the option is enabled, it may also be only one display used, which can use this feature. This is very likely to happen. Enable/disable it in the remote menu,  rather than click but not work.

21pages (changes requested):

src/server/connection.rs
```diff
@@ -3114,6 +3142,42 @@ impl Connection {
...
+                misc.set_follow_current_display(d_index as i32);
+                let mut msg_out = Message::new();
+                msg_out.set_misc(misc);
+                self.send(msg_out).await;
+            }
+        }
+    }
+
+    #[cfg(not(any(target_os = "android", target_os = "ios")))]
+    async fn handle_cursor_switch_display(&mut self, pos: CursorPosition) {
```
21pages: 1. When there are multiple displays, this is still running, but the current display is wrong, and messages will to be sent repeatly.
2. use cached `get_sync_displays`, both for `handle_window_focus`

src/platform/linux.rs
```diff
@@ -119,6 +127,40 @@ pub fn get_cursor_pos() -> Option<(i32, i32)> {
...
+                    &mut y as _,
+                    std::ptr::null_mut(),
+                ) != 0
+                {
+                    return;
+                }
+                res = displays
+                    .iter()
+                    .position(|d| x >= d.x && x < d.x + d.width && y >= d.y && y < d.y + d.height);
+            }
```
21pages, line 174: center, maybe `xdo_get_window_size`

flutter/lib/common/widgets/toolbar.dart
```diff
@@ -399,6 +400,63 @@ Future<List<TToggleMenu>> toolbarDisplayToggle(
...
+      !ffi.canvasModel.cursorEmbedded &&
+      !pi.isWayland &&
+      versionCmp(pi.version, "1.2.4") >= 0 &&
+      pi.displays.length > 1) {
```
21pages: 1. && pi.currentDisplay != kAllDisplayValue && self.session_handlers.read().unwrap().len() == 1
2. allow changing `show remote cursor` when follow menus are invisible, eg:  when pi.currentDisplay != kAllDisplayValue

src/flutter.rs
```diff
@@ -882,6 +882,18 @@ impl InvokeUiSession for FlutterHandler {
         );
     }
 
+    fn set_current_display(&self, disp_idx: i32) {
+        if self.session_handlers.read().unwrap().len() > 1 {
+            return;
+        }
+        self.peer_info.write().unwrap().current_display = disp_idx;
```
21pages: remove this line  unless it is a bugfix, there was no such line before when sending switch display

flutter/lib/models/model.dart
```diff
@@ -440,7 +442,7 @@ class FfiModel with ChangeNotifier {
     }
   }
 
-  updateCurDisplay(SessionID sessionId, {updateCursorPos = true}) {
+  updateCurDisplay(SessionID sessionId, {updateCursorPos = false}) {
```
21pages, line 445: Explain `updateCursorPos = false`

21pages (changes requested):

flutter/lib/common/widgets/toolbar.dart
```diff
@@ -390,6 +390,20 @@ Future<List<TToggleMenu>> toolbarCursor(
     if (pi.currentDisplay == kAllDisplayValue ||
         bind.sessionIsMultiUiSession(sessionId: sessionId)) {
       lockState.value = false;
+      final followCursorOption = 'follow-remote-cursor';
```
21pages: The "prevent" relies on user expanding the menu. The server side can know the display count of the current connection.

21pages (changes requested):

src/server/connection.rs
```diff
@@ -3105,6 +3143,40 @@ impl Connection {
...
+        let displays = super::display_service::get_sync_displays();
+        let d_index = displays.iter().position(|d| {
+            let scale = d.scale;
+            pos.x >= d.x
+                && pos.y >= d.y
+                && (pos.x - d.x) as f64 * scale < d.width as f64
+                && (pos.y - d.y) as f64 * scale < d.height as f64
+        });
+        if let Some(d_index) = d_index {
+            if self.display_idx != d_index {
```
21pages, line 3184: move the display sub count judgement here for efficiency

src/server/connection.rs
```diff
@@ -3105,6 +3143,40 @@ impl Connection {
         self.inner.send(msg.into());
         self.supported_encoding_flag = (true, not_use);
     }
+
+    #[cfg(not(any(target_os = "android", target_os = "ios")))]
+    async fn handle_window_focus(&mut self) {
+        let displays = super::display_service::get_sync_displays();
+        let current_display = crate::get_focused_display(displays);
+        if let Some(d_index) = current_display {
```
21pages: move the display sub count judgement here for efficiency

21pages (changes requested):

21pages:
Follow `NAME_POS`'s no_perms, subscribe

21pages (changes requested):
https://github.com/rustdesk/rustdesk/blob/d920953df1c781908bf9c1b6f5c8924ac30e9f9f/src/server/connection.rs#L2761

21pages:
I think it's ready to be merged.

21pages (approved):

rustdesk:
I will take a look of the code later

src/server/connection.rs
```diff
@@ -666,8 +677,14 @@ impl Connection {
                             #[cfg(target_os = "macos")]
                             conn.retina.set_displays(&_pi.displays);
                         }
-                        #[cfg(target_os = "macos")]
                         Some(message::Union::CursorPosition(pos)) => {
+                            #[cfg(not(any(target_os = "android", target_os = "ios")))]
```
rustdesk, line 675: Mac retina tested?

libs/hbb_common/src/config.rs
```diff
@@ -1258,6 +1264,19 @@ serde_field_bool!(
     default_show_remote_cursor,
     "ShowRemoteCursor::default_show_remote_cursor"
 );
+serde_field_bool!(
+    FollowRemoteCursor,
+    "follow_remote_cursor",
+    default_follow_remote_cursor,
+    "FollowRemoteCursor::default_follow_remote_cursor"
```
rustdesk, line 1271: Default is true or false here?

src/server/connection.rs
```diff
@@ -658,6 +661,14 @@ impl Connection {
...
                                     break;
                                 }
+                                Some(misc::Union::WindowFocus(_)) => {
+                                    #[cfg(not(any(target_os = "android", target_os = "ios")))]
+                                    {
+                                        if conn.follow_remote_window {
+                                            conn.handle_window_focus().await;
+                                        }
+                                    }
+                                }
```
rustdesk: Multiple problems here:
- [ ] since WindowFocus value is not used, why do you use uint?
- [ ] should `continue` here because this message is not needed to passed to peer
- [ ] I do not like your implementation here. Since it is a service, you should unsubscribe it if `get_subbed_displays_count` > 1 rather than creating additional protobuf message

src/server/connection.rs
```diff
@@ -3126,6 +3164,54 @@ impl Connection {
...
+            msg_out.set_misc(misc);
+            self.send(msg_out).await;
+        }
+    }
+
+    #[cfg(not(any(target_os = "android", target_os = "ios")))]
+    async fn handle_cursor_switch_display(&mut self, pos: CursorPosition) {
+        if let Some(s) = self.server.upgrade() {
+            let server = s.write().unwrap();
+            let subbed_disp_count = server.get_subbed_displays_count(self.inner.id());
```
rustdesk: No way to mark this status on Connection structure? Checking it every time is not good.

src/server/connection.rs
beelchester: Yes, tested.

libs/hbb_common/src/config.rs
beelchester: False

src/server/connection.rs
beelchester: New changes:
- Remove windowfocus protobuf message
- New bool field `multi_ui_session`. True when `get_subbed_displays_count > 1` (checked only when there's a change in display services)
- Unsub window focus if `multi_ui_session`

libs/hbb_common/src/config.rs
rustdesk: Could you please confirm the default value of "follow_remote_cursor" and "follow_focus_window" of AnyDesk? We need to follow AnyDesk.

beelchester: By default, both the features are off in AnyDesk.

hueldoeu:
hello, i am sorry but i don't understand follow-remote-window. what exactly does that mean? could someone please describe it easily to me?

PR #7736, opened 2024-04-15
Connection dialoge should be centered

Sen66:
The connection approval dialoge should be center as in other Remote-Apps such as Anydesk. On bigger monitors it is purely annoying and sometimes you do not even look on your top right corner.

### DIFF
--- a/flutter/lib/main.dart
+++ b/flutter/lib/main.dart
@@ -258,7 +258,7 @@ showCmWindow({bool isStartup = false}) async {
     ]);
     // ensure initial window size to be changed
     await windowManager.setSizeAlignment(
-        kConnectionManagerWindowSizeClosedChat, Alignment.topRight);
+        kConnectionManagerWindowSizeClosedChat, Alignment.center);
     _isCmReadyToShow = true;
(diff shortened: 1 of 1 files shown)


Benno1308:
Would be nice to see this getting merged,
I tend to miss the connection request from time to time too.

Maybe in a further PR add the option to force AlwaysOnTop

rustdesk:
I do not agree to make it default to center, please add an option in setting for this.

danielmarschall:
Indeed, AlwaysOnTop very high on my person wishlist, because so many people miss the incoming connection window because it vanishes behind the full-screen application if the user clicks

rustdesk:
> Indeed, AlwaysOnTop very high on my person wishlist, because so many people miss the incoming connection window because it vanishes behind the full-screen application if the user clicks

Yes, agree. @21pages follow this up.

rustdesk:
https://github.com/rustdesk/rustdesk/pull/7953

rustdesk:
> Apologize, I do not agree to make it default to center, please add an option in setting for this. AlwaysOnTop option is also welcome

Close since no reply.

Issue #5976, opened 2023-10-11
Connect Headless Linux Client report error: Unsupported display server type "", x11 or wayland expected


Bao0ne:
### Bug Description

Unsupported display server type "", x11 or wayland expected

Can connect successfully and run lightdm manually before connecting

my question is how does the headless client automatically create a session like xrdp

### How to Reproduce

running headless client with ubuntu:latest container .

get id and set passowrd by cli

### Expected Behavior

connection succeeded

### Operating system(s) on local side and remote side

osx -> ubuntu

### RustDesk Version(s) on local side and remote side

1.2.3

### Screenshots

![image](https://github.com/rustdesk/rustdesk/assets/37828835/8c0917b0-de11-45c5-a7f9-1cbd28d0ac0d)

### Additional Context

exec headless rustdesk client
```
root@u2-6758bb5fd-966q7:/# loginctl 
No sessions.
root@u2-6758bb5fd-966q7:/# /usr/lib/rustdesk/rustdesk --get-id
198527304
root@u2-6758bb5fd-966q7:/# ps -ef | grep rust
root     1040309  331871  0 Oct10 pts/5    00:05:45 /usr/lib/rustdesk/rustdesk --service
root     1713901 1040309  0 01:08 pts/5    00:00:04 /usr/lib/rustdesk/rustdesk --server
root     1719608 1718860  0 01:18 pts/12   00:00:00 grep --color=auto rust
```

when running lightdm manually
```
root@u2-6758bb5fd-966q7:/# systemctl start lightdm      
root@u2-6758bb5fd-966q7:/# loginctl 
SESSION UID USER    SEAT  TTY
     c8 112 lightdm seat0 

1 sessions listed.
```
Then connection succeeded

rustdesk:
no duplcate please.

PR #3902, opened 2023-04-03
Feat/linux virtual display (headless linux)

fufesou:
1. Linux virtual display support. https://github.com/rustdesk/rustdesk/issues/59
2. May fix https://github.com/rustdesk/rustdesk/issues/3592.
3. Refactor dialog. https://github.com/rustdesk/rustdesk/pull/3764#issuecomment-1482636535 @grummbeer 
4. Add "--cm-no-ui" for virtual display. There's some `segfault` bug with "--cm" with ui on headless env.

**Only tested on Ubuntu for now.**

Prerequirements:
```bash
# Only support gnome desktop. https://www.gnome.org
# xfce4 may be supported.

# for Ubuntu
$ sudo apt install ubuntu-desktop
$ sudo apt install xserver-xorg-video-dummy
$ sudo apt install lightdm

# for Fedora
$ sudo sudo yum groups install "GNOME"
$ sudo yum install xorg-x11-drv-dummy

# for archlinux
$ pacman -S xorg xorg-server xf86-video-dummy
$ pacman -S gnome

```

Steps
```
# install rustdesk
# get rustdesk id
$ sudo rustdesk --get-id
# set password
$ sudo rustdesk --password <password>

# connect
```

TODOs:
- [ ] Do not show "Connection Error" on logout.
- [ ] Failed to start Xorg on some PCs. `parse_vt_settings: Cannot open /dev/tty0 (Permission denied)`.

Issue #1131, opened 2022-08-01
Linux Headless Support


simeononsecurity:
One of the biggest things that TeamViewer has that no other alternative has is that TeamViewer can be installed on headless linux systems. Accessing the command line of a linux system remotely and securely is a every day task for system administrators. Please integrate this as a feature.

rustdesk:
How does it work? can you elaborate more?

simeononsecurity:
https://community.teamviewer.com/English/kb/articles/4352-install-teamviewer-on-linux-without-graphical-user-interface
https://community.teamviewer.com/English/kb/articles/6318-install-teamviewer-on-linux
https://community.teamviewer.com/English/kb/articles/4256-use-teamviewer-on-headless-systems

_Framebuffer console without Xorg
TeamViewer can be used to access machines without a graphical desktop environment (Xorg, Wayland). TeamViewer lets you connect to the foreground console (as indicated by fgconsole). In other words, the VT (virtual terminal) which is currently visible on the screen – or would be visible if a screen was attached._

mbennette68:
Yea, I noticed that the other day when I tried to remove into a headless machine. Error generated was no display found or something to that effect. 
I normally use [dwservice.net](https://www.dwservice.net/en/) which does seem to support headless remoting but sometimes audio playback does not work with older Linux distros.
I would say, adding Remote capabilities in to headless machines is a must.

Issue #634, opened 2022-05-27
add session manager to replace x2go and nomachine


congzhangzh:
Hi Rustdesk,

As title, x2go is very lagging, and nomachine is not free for multisession support。

if rustdesk support special mode to spawn more xephyr as standalone session, it will be very helpful and revolutionary of clouddesktop way of xwindows server.

with session manager that shoud be easy to do something like xrandr.

Tks,
Cong

rustdesk:
@fufesou

rustdesk:
Who are the target users for session manager? I can not understand the scenario.

congzhangzh:
> Who are the target users for session manager? I can not understand the scenario.

A team to access a super computer to develop or do some ui related work。

x2go is very slow, nomachine is very fast but limit one session for free version, btw nomachine is magic fast and powerful!

congzhangzh:
A session manager to start headless x server and gnone like things

rustdesk:
I used nomachine before, not saw magic.

congzhangzh:
> I used nomachine before, not saw magic.

seems your are right, I use rustdesk to connect to rustdesk in nomachine now, it's smooth:)

congzhangzh:
> > Who are the target users for session manager? I can not understand the scenario.
> 
> A team to access a super computer to develop or do some ui related work。
> 
> x2go is very slow, nomachine is very fast but limit one session for free version, btw nomachine is magic fast and powerful!

even x2go very laging, connect rustdesk in x2go very soooth, the only problem is that keyboard input seems be input twice each time when type from rustdesk, can you give a try @rustdesk

BTW, if rustdesk support run different instance for each x2go X session, the problem can be workaround, as connect to 21118+1/2/3/4 for session 1/2/3/4

rustdesk:
Sorry， I have no x2go server env to debug.

rustdesk:
Low priority.

congzhangzh:
Maybe we can split session manager to another  application and make rustdesk focus on remote control? Like add a one 2 one client/server mode with preshare key?

# basic idea, raw server client
server:

``` bash
rustdesk —just-server [xx.xx.xx.xx:]21118 —key-file ~/.rustdesk/prehare-key
```

client:

``` bash
rustdesk —connect-to 21118 —key-file prehare-key-file
```

# with session manager wrapper

``` bash
# client connect to server run desktop & rustdesk in xvfb
ssh xx@xx.xx <<EOF
[...]

PR #7811, opened 2024-04-23
fix: add support for headless in docker

congzhangzh:
# Try to make rustdesk work in the docker environment for testing, and even for Google Colab,  and computer datacenter, at least technically, two minor fix:

1. allow user to forced set desktop system by envirement variable like **RUSTDESK_FORCED_DISPLAY_SERVER**
2. back to default if loginctl can't be used(failed)

# Why need this fix:
When the system has no systemd context like run in docker, loginctl can't be used even 4 queries, you will get:

**System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down**

and **echo $?** get **1**

# Review our old code:

from the old code we can expect, that in docker, it should be x11 which is the default if nothing fit, but it's not, maybe because we can not use loginctl in docker?

```rust
//libs/hbb_common/src/platform/linux.rs
pub fn get_display_server() -> String {
    let mut session = get_values_of_seat0(&[0])[0].clone();
    if session.is_empty() {
        // loginctl has not given the expected output.  try something else.
        if let Ok(sid) = std::env::var("XDG_SESSION_ID") {
            // could also execute "cat /proc/self/sessionid"
            session = sid;
        }
        if session.is_empty() {
            session = run_cmds("cat /proc/self/sessionid").unwrap_or_default();
            if session == INVALID_SESSION {
                session = "".to_owned();
            }
        }
    }
    if session.is_empty() {
        "".to_owned()
    } else {
        get_display_server_of_session(&session)
    }
}

pub fn get_display_server_of_session(session: &str) -> String {
    let mut display_server = if let Ok(output) =
        run_loginctl(Some(vec!["show-session", "-p", "Type", session]))
    // Check session type of the session
    {
        let display_server = String::from_utf8_lossy(&output.stdout)
            .replace("Type=", "")
            .trim_end()
            .into();
        if display_server == "tty" {
            // If the type is tty...
            if let Ok(output) = run_loginctl(Some(vec!["show-session", "-p", "TTY", session]))
            // Get the tty number
            {
                let tty: String = String::from_utf8_lossy(&output.stdout)
                    .replace("TTY=", "")
                    .trim_end()
                    .into();
                if let Ok(xorg_results) = run_cmds(&format!("ps -e | grep \"{tty}.\\\\+Xorg\""))
                // And check if Xorg is running on that tty
                {
                    if xorg_results.trim_end() != "" {
                        // If it is, manually return "x11", otherwise return tty
                        return "x11".to_owned();
                    }
                }
            }
        }
        display_server
    } else {
        "".to_owned()
    };
    if display_server.is_empty() || display_server == "tty" {
        // loginctl has not given the expected output.  try something else.
        if let Ok(sestype) = std::env::var("XDG_SESSION_TYPE") {
            display_server = sestype;
        }
    }
    if display_server == "" {
        display_server = "x11".to_owned();
    }
    display_server.to_lowercase()
}
fn run_loginctl(args: Option<Vec<&str>>) -> std::io::Result<std::process::Output> {
    let mut cmd = std::process::Command::new("loginctl");
    if let Some(a) = args {
        return cmd.args(a).output();
    }
    cmd.output()
}
```
refs:
https://github.com/rustdesk/rustdesk/pull/3902
https://github.com/rustdesk/rustdesk/wiki/Headless-Linux-Support
https://github.com/rustdesk/rustdesk/issues/634
https://github.com/rustdesk/rustdesk/issues/192
https://github.com/rustdesk/rustdesk/issues/1131
https://github.com/rustdesk/rustdesk/issues/5976

@fufesou please have a review, I have very little rust and background of C++, hope I can do something

### DIFF
--- a/libs/hbb_common/src/platform/linux.rs
+++ b/libs/hbb_common/src/platform/linux.rs
@@ -49,6 +49,16 @@ pub fn is_x11_or_headless() -> bool {
 const INVALID_SESSION: &str = "4294967295";
 
 pub fn get_display_server() -> String {
+    // Check for forced display server environment variable first
+    if let Ok(forced_display) = std::env::var("RUSTDESK_FORCED_DISPLAY_SERVER") {
+        return forced_display;
+    }
+
+    // Check if `loginctl` can be called successfully
+    if run_loginctl(None).is_err() {
+        return DISPLAY_SERVER_X11.to_owned();
+    }
+
     let mut session = get_values_of_seat0(&[0])[0].clone();
     if session.is_empty() {
         // loginctl has not given the expected output.  try something else.


congzhangzh:
# to test rust in docker

# install
```bash
latest_rustdesk=$(curl -s https://api.github.com/repos/rustdesk/rustdesk/releases/latest | jq -r ".tag_name")
wget https://github.com/rustdesk/rustdesk/releases/download/${latest_rustdesk}/rustdesk-${latest_rustdesk}-x86_64.deb

dpkg -i rustdesk-*.deb -y
apt --fix-broken install
```
```
tee hi_rustdesk.sh <<EOF
export XDG_SESSION_ID=xxxxxx
export XDG_SESSION_TYPE=x11
export DISPLAY=:0
EOF
chmod +x hi_rustdesk.sh 
```
# by service mode
```
. ./hi_rustdesk.sh

/usr/lib/rustdesk/rustdesk --service
/usr/lib/rustdesk/rustdesk --get-id
/usr/lib/rustdesk/rustdesk --password youpassword
# connect now
```
# by standalone mode
```
. ./hi_rustdesk.sh

xvfb $DISPLAY &
/usr/lib/rustdesk/rustdesk
xwd -root -silent | gm convert xwd:- output.png
# get id and pwd from output.png
# connect now
```

congzhangzh:
BTW, it seems already works if I set to something, trick here:
```bash
export XDG_SESSION_ID=xxxxxx
```

rustdesk:
Good job

congzhangzh:
@rustdesk Hi, I add a guideline about how to make it work with colab

https://github.com/congzhangzh/daily-scripts-4-us/blob/main/Colab%26LM%20Studio%20By%20Rustdesk.ipynb

adrianog:
Does this run rustdesk in a docker container, on a headless remote host?

congzhangzh:
> Does this run rustdesk in a docker container, on a headless remote host?
> 
> Can you release a dockerfile?

The colab jupyterbook maybe help!

https://github.com/congzhangzh/daily-scripts-4-us/blob/main/Colab%26LM%20Studio%20By%20Rustdesk.ipynb

Tips: I use customized xdummy xserver & rustdesk command line to set password &get id

PR #7857, opened 2024-04-28
update hwcodec, fix screen jitter

21pages:
Fix jitter on ffmpeg vram decode and intel vram decode on my old pc, convert nv12 to bgra by shader rather than d3d11 video processor

https://github.com/rustdesk/rustdesk/assets/14891774/581db531-dee0-47bd-a4b8-f14c33aedfc1

https://github.com/rustdesk/rustdesk/assets/14891774/9e8d6959-be78-40f8-98ba-73d44310fc92


askdad:
my old pc has the same problem. gpu is intel hd grapgics 4000. rustdesk's version is 1.3.6

PR #7923, opened 2024-05-06
fix: translate, proxy

fufesou:
Add `default_proxy_tip` in `template.rs` and move it to the end.

https://github.com/rustdesk/rustdesk/assets/13586388/8e85f01a-a685-409a-8caf-7c14c84a5a65


rustdesk:
Do not put it as placehoder, since can not be display completely.
image

PR #4848, opened 2023-07-04
fix: bump winget releaser version & add dependabot

sitiom:
The Winget Releaser action failed (https://github.com/rustdesk/rustdesk/actions/runs/5446780380) for two reasons:
- The action was outdated (`v1`)
- The Windows builds weren't available when the release was made (https://github.com/rustdesk/rustdesk/releases/download/1.2.0/rustdesk-1.2.0-x86_64.exe and https://github.com/rustdesk/rustdesk/releases/download/1.2.0/rustdesk-1.2.0-x86-sciter.exe were added later)

This PR fixes the first issue by updating Winget Releaser to `v2` and adding Dependabot to ensure dependencies are up-to-date. For the second issue, you must **ensure that the Windows builds are available on release**.

For the meanwhile, I have manually created a PR for 1.2.0: https://github.com/microsoft/winget-pkgs/pull/111231

PR #7971, opened 2024-05-08
Revert "fix: bump winget releaser version & add dependabot"

rustdesk:
Reverts rustdesk/rustdesk#4848

depbot gives me big trouble, since some new dep will brreak us.

newer does not mean better, if you know xz event.


sitiom (commented):
https://github.com/rustdesk/rustdesk/pull/4848#issuecomment-2099823648

PR #7983, opened 2024-05-08
Tidy up configuration files UNIX permissions

HorlogeSkynet:
---

Hello :wave:

This patch enforces `0600` UNIX permission on configuration files, for non-Windows hosts.
As they usually contain remote machine credentials, their access should be restricted to their owner.

Thanks, bye :pray:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ RUN apt update -y && \
         libxcb-shape0-dev \
         libxcb-xfixes0-dev \
         libasound2-dev \
+        libpam0g-dev \
         libpulse-dev \
         make \
         cmake \
--- a/libs/hbb_common/src/config.rs
+++ b/libs/hbb_common/src/config.rs
@@ -491,7 +491,19 @@ pub fn load_path<T: serde::Serialize + serde::de::DeserializeOwned + Default + s
 
 #[inline]
 pub fn store_path<T: serde::Serialize>(path: PathBuf, cfg: T) -> crate::ResultType<()> {
-    Ok(confy::store_path(path, cfg)?)
+    #[cfg(not(windows))]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        Ok(confy::store_path_perms(
+            path,
+            cfg,
+            fs::Permissions::from_mode(0o600),
+        )?)
+    }
+    #[cfg(windows)]
+    {
+        Ok(confy::store_path(path, cfg)?)
+    }
 }
 
 impl Config {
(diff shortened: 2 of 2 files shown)


rustdesk:
then, testing video please.

HorlogeSkynet:
Hey @rustdesk, I'm really sorry but I can't manage to find "video" tests. Do you have a pointer or documentation that I could read to add such a test case ?

rustdesk:
Like this, https://github.com/rustdesk/rustdesk/pull/7988

Showing your PR taking effect, and does not break any function.

HorlogeSkynet:
So I eventually got what you mean by "testing video"...  
As I'm not setup to do so, and I fail to see how this would be somehow realistic, I've opted for two proper unit test cases for `hbb_common` (including one checking for file permissions mode, on non-Windows).

During building/testing, I've also encountered an issue related to Docker image and PAM, so I took the liberty to add the missing library to package dependencies list.

Bye :wave:

rustdesk:
Unit test is not enough for such case. Integration test is required.

HorlogeSkynet:
There isn't any integration tests from what I see...

I could make a screenshot of RD running and my terminal showing configuration files with `0600` permissions but I don't think it's really relevant as :

* There is not any "visible" change in the UI ;
* If peer configurations are `0644` or `0600` it's exactly the same thing from RD PoV (as the "user" dumping the configuration is the same one who loads it afterwards, unless I'm wrong here) ;
* I could be totally running another version of the codebase and I failed to see how this would be bullet-proof.

Thanks for your time, bye :pray:

rustdesk:
Let's give you an example.

If a ipc file created with 600 permission, does it mean only the creator can access it? then the other process running as the other user can not connect to the IPC?

HorlogeSkynet:
> If a ipc file created with 600 permission, does it mean only the creator can access it? then the other process running as the other user can not connect to the IPC?

Fully-agreeing here with your IPC example; Is `hbb_common::config::store_path` (thus relying on `confy`, dealing itself with configuration files) actually used for handling IPC (or something different than configuration files) that I'm not aware of ?

> You need to undertand test, "I think" is the worst world for testing.

:+1: that's why projects usually rely on testing workflows that maintainers trust, and not "videos" recorded on untrusted machines (?), with completely unknown codes/setups/configurations which result in non-reproducible scenarios.

Thanks, see you :wave:

rustdesk:
> that's why projects usually rely on testing workflows that maintainers trust

Such a flow is too hard for RustDesk, I hope you can understand.

HorlogeSkynet:
(rebased according to confy submodule "move" @rustdesk)

HorlogeSkynet:
Little up @rustdesk, thanks :pray:

HorlogeSkynet:
Little up @rustdesk, thanks bye :wave:

HorlogeSkynet:
Little up @rustdesk, thanks bye :pray:

rustdesk:
i would suggest a further enhancement.

move some ipc files to local directory, only _service ipc in tmp.

PR #7988, opened 2024-05-09
fix: wayland delete restore token

fufesou:
1. Add a button to clear wayland restore token.
2. If restore token is supported, do not hold the dbus session.
3. Support close session if restore is not supported. Then the user can select another screen without restarting the server process.

https://github.com/rustdesk/rustdesk/assets/13586388/bfbf7870-5521-4607-9de3-c2ab492623ae


rustdesk:
image

PR #8037, opened 2024-05-13
refact: android audio input, voice call

fufesou:
## Features

Support voice call for Android.

### Preview

https://github.com/rustdesk/rustdesk/assets/13586388/0a2a6047-9e33-46c8-969f-a7786a0ed5d0

https://github.com/rustdesk/rustdesk/assets/13586388/556eb001-44d7-412a-98f3-d95f0ee742db

### Desc

1. Extract audio process from `MainService.kt` to `AudioRecordHandle.kt`. 
The flollowing message is not translated and the message is simple. Because it rarely happens.
We may make it clearer in the future.
```kt
            flutterMethodChannel?.invokeMethod("msgbox", mapOf(
                "type" to "custom-nook-nocancel-hasclose-error",
                "title" to "Voice call",
                "text" to "Failed to stop voice call."))
```

## Known issue

1. Sometimes the sound played on Android is noisy, and the noise is still played after the connection is ended.
Simple Android as the control side, without voice call.
This issue in not introduced by this PR.
I haven't found how to reproduce it yet.

### DIFF
--- a/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/AudioRecordHandle.kt
+++ b/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/AudioRecordHandle.kt
@@ -1,0 +1,200 @@
+package com.carriez.flutter_hbb
+
+import ffi.FFI
+
+import android.Manifest
+import android.content.Context
+import android.media.*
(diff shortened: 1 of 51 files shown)


qwuip789:
Android voice call phone connected successfully, but no sound.

PR #8097, opened 2024-05-19
fix: custom client, whitelist, check default

fufesou:
whitelist option, `trim()` before checking if is default.


fufesou:
Closed because the web console will execute `trim()`. Even though it `trims()` the entire string instead of the `key` and `value`.

PR #8127, opened 2024-05-23
sort address book name dropdown

21pages:
https://github.com/rustdesk/rustdesk-server-pro/discussions/254#discussioncomment-9531340

Personal address book first, then `names.sort((a, b) => a.toLowerCase().compareTo(b.toLowerCase()));`

**Before**:

**After**:


Nawaki89:
nice! how long before I can use this future? is it released in your nightly build(using the pro server to generate custom clients, need to wait for an upgrade of the server?)
thanks for quick and awesome respone!

PR #8201, opened 2024-05-29
Add missing french language tips

aeltorio:


rustdesk:
Could you please avoid format change?

aeltorio:
I saw the problem, auto linting was turn on… now does it work? I manually copied back modification into the HEAD src/lang/fr.rs

rustdesk:
THAKS

PR #8261, opened 2024-06-05
fix: setMovable only on macos

fufesou:


rustdesk:
comments, and tell why

Issue #6192, opened 2023-10-25
Android Screen Record Stops after hours (service killed by system)


vavolkova:
Hi guys,

Not sure if that has been discussed before, but I didn't find anything under this topic. I was looking to set unattended access on few Android devices (Android 11). I have set Pro account with Linux server and it all worked properly until I noticed that after 2-3 hours the service stops (Screen Record automatically turns off) and I need to manually turn it on and continue using it.

Is there a solution for that?

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/rustdesk) so you can upvote and help fund this issue.
- We receive the funding once the issue is completed & confirmed by you.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/rustdesk/rustdesk/issues/6192">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/rustdesk/rustdesk/issues/6192/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/rustdesk/rustdesk/issues/6192/pledge.svg">
</picture>
</a>

rustdesk:
Frankly, we still have no good solution.

rustdesk:
https://support.anydesk.com/knowledge/anydesk-for-android-chromeos

![image](https://github.com/rustdesk/rustdesk/assets/71636191/dd3e4468-1a43-4b69-99c5-84247fe3db09)

rustdesk:
![image](https://github.com/rustdesk/rustdesk/assets/71636191/d5137876-eb94-4ecd-b18f-c317a15dec5e)

TheGrave:
I have the same issue on Xperia 5 and Xperia 1. It's utterly ridiculous - I need to keep TeamViewer Host installed so that I can re-enable the service. This also forces me to have no screen lock on these phones as TeamViewer Host cannot get to an unlock prompt with them.

vavolkova:
Hi @scshiv29-dev ,

I'm not sure how the bounty process works, but is there going to be a solution for this problem anytime soon?

Thanks!

scshiv29-dev:
Hey @vavolkova  the bounty process is designed to give incentives and get open source contributions to the project. The thing is I did the battery optimization setting on my phone and tried 2-3 other phones and rustdesk was working fine . I still think this is an android issue rather than the app code itself, like if the app is stopped from running in the background it will just terminate the session with the server and it would stop.

[...]

PR #8268, opened 2024-06-05
android floating window

21pages:
* The floating window shows when the service is on.
* It helps avoid services being killed by the system.

https://github.com/rustdesk/rustdesk/issues/6192

[floating-window.webm](https://github.com/rustdesk/rustdesk/assets/14891774/ef38476b-a3c3-467c-89f4-dc76def5698d)

https://rustdesk.com/docs/en/self-host/client-configuration/advanced-settings/
Custom client config of picture below:

`floating-window-svg=<?xml version="1.0" standalone="no"?><!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd"><svg t="1717559129252" class="icon" viewBox="0 0 1024 1024" version="1.1" xmlns="http://www.w3.org/2000/svg" p-id="4248" xmlns:xlink="http://www.w3.org/1999/xlink" width="32" height="32"><path d="M950.857143 512c0 242.285714-196.571429 438.857143-438.857143 438.857143S73.142857 754.285714 73.142857 512 269.714286 73.142857 512 73.142857s438.857143 196.571429 438.857143 438.857143z" fill="#1296db" p-id="4249"></path></svg>`
`floating-window-transparency=10`
`disable-floating-window=N`
`floating-window-size=240`

01da1e41bd06e290fee599dfac50fc8

# Other

* Becase `translate` has only one `new_string`, and `with_local_frame_returning_local` leak it for return, so remove `with_local_frame_returning_local`


xlionjuan:
How about implement this also when doing outgoing connections? Because outgoing connections will killed by system too if it isn't on foreground.

PR #8368, opened 2024-06-16
mac scale factor of each screen

21pages:
Each screen has an independent BackingScaleFactor

before change: scale factor is same with main screen, and may change when after click.

https://github.com/rustdesk/rustdesk/assets/14891774/2fbba55e-4568-4824-b9b1-7c632e99a87d

after change:  scale factor is same with setting

https://github.com/rustdesk/rustdesk/assets/14891774/0e597ef5-28c6-4257-bfe7-11d0cd891ebb

### DIFF
--- a/libs/scrap/src/quartz/display.rs
+++ b/libs/scrap/src/quartz/display.rs
@@ -71,7 +71,7 @@ impl Display {
     }
 
     pub fn scale(self) -> f64 {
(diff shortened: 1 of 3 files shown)


rustdesk:
reverted in https://github.com/rustdesk/rustdesk/commit/97aa739d69c67a9cd9d40da8ed7308224ffae875
because on my mac mini, display_index ==2

PR #8512, opened 2024-06-27
refact: mobile more actions, divider

fufesou:
https://github.com/rustdesk/rustdesk/assets/13586388/624cc514-60ac-42a0-9675-5a821dda13d2


it-boerner:
Is it possible, like with TeamViewer, to create a way for installed apps to be displayed and for them to be started or uninstalled?

PR #8582, opened 2024-07-03
Bump vcpkg baseline to 2024.06.15

basilgello:
- **res/vcpkg: drop old opus port overlay**
- **res/vcpkg: bump aom to 3.9.1**
- **res/vcpkg: bump libvpx to 1.14.1**
- **Bump vcpkg baseline to 2024.06.15**


rustdesk:
still failed, https://github.com/rustdesk/rustdesk/actions/runs/9773973076/job/26983236248

Interesting, it build arm on x64   $VCPKG_ROOT/vcpkg install --triplet arm-linux --x-install-root="$VCPKG_ROOT/installed"

and build arm64 on armv7

basilgello:
Likely the issue is that `-DCOMPILER_SUPPORTS_NEON=OFF` which is needed for opus 1.4.0 onwards was not ported by Cheney-W of vcpkg

basilgello:
What we see is a mix of minimum two issues:

1. https://github.com/rustdesk/rustdesk/blob/master/.github/workflows/flutter-build.yml#L1347 - vcpkg dependency step for sciter needs same case-esac as https://github.com/rustdesk/rustdesk/blob/master/.github/workflows/flutter-build.yml#L1046

2. The opus local port should be reinstated to 1.5.1 with one additional opus cmake option passed (and upstreamed!)

rustdesk:
Yes

PR #8583, opened 2024-07-03
Fix F-Droid version file workflow

basilgello:


basilgello:
@rustdesk this should return the latest final release regardless of how the workflow was called. The only way to test is merge and start workflow :)

rustdesk:
Why not https://gitlab.com/basilgello/rustdesk-fdroid-versioner detect it directly rather than creating a middle file here?

basilgello:
Because my goal is to keep everything here and not make unnecessary dependencies on myself and my private gitlab spaces

PR #8585, opened 2024-07-03
vcpkg: sort triplets and restore local opus port for 1.5.1

basilgello:
- **ci/flutter-build: Sort out triplets in job matrices**
- **Restore local port for opus 1.5.1**

### DIFF
--- a/.github/workflows/flutter-build.yml
+++ b/.github/workflows/flutter-build.yml
@@ -58,7 +58,7 @@ jobs:
         job:
           # - { target: i686-pc-windows-msvc        , os: windows-2022                  }
           # - { target: x86_64-pc-windows-gnu       , os: windows-2022                  }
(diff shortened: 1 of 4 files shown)


basilgello:
Lets try this one @rustdesk ^^

rustdesk:
Thank you very much, all work now.

basilgello:
I am still confused with `flutter/build_android.sh`, `flutter/build_ios.sh`, `flutter/ios_arm64.sh`, `flutter/ios_x64.sh`, `flutter/run.sh`. They duplicate steps in `.github/workflows/flutter-build.yml` but are used nowhere. Maybe it is better to consolidate them just like I did with `flutter/build_android_deps.sh` and call them in CI yaml so there is no need to fix same things here and there?

basilgello:
@rustdesk ^^ Thinking of re-organizing build steps and moving f-droid build step into this tree :)

PR #8589, opened 2024-07-03
Update hu.rs

FowlerJohn:
I modified the Hungarian translation.
A lot of strings were not translated.
And I fixed the translation of some strings.
John Fowler.


rustdesk:
Good job. Thanks

PR #8597, opened 2024-07-03
Add basque language

Thadah:


rustdesk:
Thank you very much. You need to modify https://github.com/rustdesk/rustdesk/blob/master/src/lang.rs also.

Thadah:
Ah, sorry, I didn't notice the file. I've updated the PR with lang.rs modified.

PR #8625, opened 2024-07-05
Move F-Droid build script to RustDesk source tree

basilgello:


.github/workflows/flutter-build.yml
```diff
@@ -10,6 +10,13 @@ on:
         type: string
         default: "nightly"
 
+# NOTE: Environment variables like `{ANDROID,}-FLUTTER_VERSION`, `RUST_VERSION`,
+# `NDK_VERSION` etc are used by `flutter/build_fdroid.sh` which in turn is used
+# by official F-Droid builder!
+#
+# Please make changes to F-Droid packaging whenever 'build-rustdesk-android'
+# is changed!
+
```
rustdesk, line 19: How about let flutter/build_fdroid.sh read envs from flutter-build.yml? No idea if it is easy in bash, but I know it is easy to do with Python.

basilgello:
However I tried enabling hwcodec in fdroid build and it seems @21pages is using prebuilt ffmpeg, amf and nvcodec libraries. I still think all native libs should be managed via vcpkg ports in one place. Are there any other prebuilt static libs (.a) other than 21pahes/deps and 21pages/externals ?

rustdesk:
> I still think all native libs should be managed via vcpkg ports in one place

Yes, we discussed, but I did not push him, he had his reason.

21pages:
@basilgello You can add vcpkg ffmpeg ports in ci follow https://github.com/21pages/deps, and after that, we can move hwcodec to https://github.com/rustdesk/rustdesk/tree/master/libs, and make better use of ffmpeg in audio and other aspects

basilgello:
> We are doing a product, I do care about code quality, but not pursue code beauty too much

Fair, but my "push" for central native library management is about quickest possible updates of libs for security fixes.

> Frankly, I do not like vcpkg and M$, If it is not you, I will never do such thing.

M$ is not my favorite either, but as a product, fixing security issues must be as automated as possible. And vcpkg was used un RustDesk codebase waaay before I started using it :) What I like about vcpkg is a proper portfile format, but we can pursue other options with maximum maintainability as goal.

rustdesk:
We may have some internship to do this in the future.

basilgello:
> We may have some internship to do this in the future.

See email :)

rustdesk:
I have seen, this is the reply, :(

basilgello:
> You can add vcpkg ffmpeg ports in ci follow https://github.com/21pages/deps, and after that, we can move hwcodec to https://github.com/rustdesk/rustdesk/tree/master/libs, and make better use of ffmpeg in audio and other aspects

My plan is to ship 1.2.6 to F-Droid with hwcodec off and then adapt vcpkg's ffmpeg port to values in deps repo (as well as libamf etc). Otherwise F-Droid build may stuck for days if not weeks, and that's not everyone wants.

basilgello:
@rustdesk This can be merged. Once you merge it, please run F-Droid version CI action again because I changed the numbering scheme. After that I will PR the F-Droid update.

Issue #8633, opened 2024-07-07
Settings > Security > Enable Remote Configuration Modification 


DCVolo:
### Bug Description

Settings > Security > Enable Remote Configuration Modification, default state is disabled.

But it can be modified remotely.

Litteraly my first 15s of use.

### How to Reproduce

Select any other window
ALT-TAB back to RustDeskp
Navigate through the menu using TAB and arrows
press enter to apply changes

### Expected Behavior

Window should be unresponsive from any type of input

### Operating system(s) on local side and remote side

Windows 11 -> Windows 10

### RustDesk Version(s) on local side and remote side

1.2.6

### Screenshots

[Screen Recording 2024-07-07 184923.zip](https://github.com/user-attachments/files/16120070/Screen.Recording.2024-07-07.184923.zip)

### Additional Context

file is MKV x265

First thing happening is me showing that having the option pre-selected i can enable/disable the option at will.
Then it describes the steps i provided.

rustdesk:
duplicate https://github.com/rustdesk/rustdesk/issues/2680#issuecomment-1975002181

rustdesk:
close since fix merged

PR #8640, opened 2024-07-08
main window add block mask, cm add keyboard block

21pages:
* block window body only so user can still click minimize button.
* cm doesn't show mask
* Remove focusable Offstage in tabbar_widget.dart

https://github.com/rustdesk/rustdesk/assets/14891774/52ca201e-fc4e-4f1c-94e8-14b3ffb69039


rustdesk:
https://github.com/rustdesk/rustdesk/issues/8633

PR #8659, opened 2024-07-09
Fix review points on F-Droid submission

basilgello:
- **Factor out F-Droid buildscript**
- **build_fdroid: Move sudo-deps back to f-droid spec**


basilgello:
@rustdesk need one more change to merge :( I thought all in one place is perfect but at least Debian deps are changed not so often :) And the actual buildscript remains under our supervision!

Issue #8688, opened 2024-07-12
wrong fullscreen behavior


twprh:
### Bug Description

if remote is fullscreen and stopping remote session - screen left blank

see video

### How to Reproduce

sett remote session to fullscreen and close session.

### Expected Behavior

no blank window in front

### Operating system(s) on local side and remote side

Windows -> Windows

### RustDesk Version(s) on local side and remote side

nightly: 2024-07.11

### Screenshots

[Video.webm](https://github.com/user-attachments/assets/8653fbca-60e0-4307-a4ee-a39b5df14cc8)

### Additional Context

if remote client is offline -> blank screen and only way out is to kill rustdesk

rustdesk:
@fufesou

rustdesk:
1.2.6 ok?

twprh:
yes 
nightly:1.2.6 2024-06-27 is ok

rustdesk:
Thanks for your report

rustdesk:
Close as fix merged, please try out tomorrow's nightly build.

Issue #8676, opened 2024-07-10
Windows 7 cannot close home page


zhangbo8418:
### Bug Description

Windows 7 cannot close home page

### How to Reproduce

Watch the video below

### Expected Behavior

Can be closed normally

### Operating system(s) on local side and remote side

Windows7

### RustDesk Version(s) on local side and remote side

1.2.7

### Screenshots

https://github.com/rustdesk/rustdesk/assets/46797568/75636fde-c69f-4e62-a264-a4b62f8ddb2e

21pages:
What version is your win7?

rustdesk:
We won't spend time on Win7, use sciter version please.

PR #8690, opened 2024-07-12
fix: unable to close on fullscreen

fufesou:
https://github.com/rustdesk/rustdesk/issues/8688

## Desc

The bug is introduced here.

image

Because `WindowActionPanel` handles window events.

https://github.com/rustdesk/rustdesk/blob/1d59a7fe5fe8acb8f13ffe02eedc291625c60eea/flutter/lib/desktop/widgets/tabbar_widget.dart#L469

https://github.com/rustdesk/rustdesk/blob/1d59a7fe5fe8acb8f13ffe02eedc291625c60eea/flutter/lib/desktop/widgets/tabbar_widget.dart#L622

With the changes above, `WindowActionPanel` is not built when `stateGlobal.showTabBar` is false.

It not a good idea to bind the window events with `WindowActionPanel`.

So I just move `MultiWindowListener` and `WindowListener` from  `WindowActionPanel` to `DesktopTab `.

```flutter
class _DesktopTabState extends State<DesktopTab>
    with MultiWindowListener, WindowListener {
}
```

### DIFF
--- a/flutter/lib/desktop/widgets/tabbar_widget.dart
+++ b/flutter/lib/desktop/widgets/tabbar_widget.dart
@@ -230,8 +230,7 @@ typedef LabelGetter = Rx<String> Function(String key);
 int _lastClickTime =
     DateTime.now().millisecondsSinceEpoch - bind.getDoubleClickTime() - 1000;
 
-// ignore: must_be_immutable
-class DesktopTab extends StatelessWidget {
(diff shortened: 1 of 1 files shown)


zhangbo8418:
Thank you very much, you also fixed the problem of Windows 7 homepage close button not working.https://github.com/rustdesk/rustdesk/issues/8676

PR #8693, opened 2024-07-12
refact: unwrap offstage for cursor paint

fufesou:
Remove `Offstage` wrap for cusor paint.

Cusor paint consums some calculations. Use Simple `Offstage()` may save some calculations.


zhangbo8418:
Thank you very much, you also fixed the problem of Windows 7 homepage close button not working.https://github.com/rustdesk/rustdesk/issues/8676

PR #8701, opened 2024-07-13
F-Droid buildscript: fix review points

basilgello:
- **build_fdroid: Move sudo-deps back to f-droid spec**
- **build_fdroid.sh: Fix review points by Licaon_Kter**


basilgello:
@rustdesk this needs to be merged so I can proceed with another iteration of F-Droid MR (see my update I tagged you about)

PR #8707, opened 2024-07-14
mobile view mode menu

21pages:
https://github.com/rustdesk/rustdesk/discussions/8689 

* Add view mode menu to mobile, disable the same menus as desktop
* Display forbidden cursor when not in view mode && no keyboard permission && not display remote cursor
* Hide keyboard/mouse bottom buttons when keyboard is disabled
* Fixe not listen keyboard permission change in some widgets
* Changing resolution and http proxy settings are missing for mobile devices, will be added if needed.

https://github.com/user-attachments/assets/2de11649-3b6e-487a-8e9d-fb79ff1b477c

### DIFF
--- a/flutter/lib/common/widgets/toolbar.dart
+++ b/flutter/lib/common/widgets/toolbar.dart
@@ -650,6 +650,18 @@ Future<List<TToggleMenu>> toolbarDisplayToggle(
     v.addAll(toolbarKeyboardToggles(ffi));
   }
 
+  // view mode (mobile only, desktop is in keyboard menu)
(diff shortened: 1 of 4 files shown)


Vasilis553:
I would like to see on mobile option to change resolution and add privacy screens(ex here at work we. Have an old printer and we connect from an old laptop wired on printer but for 4 people's to print we can't activate privacy mode because is exclusive but we use one privacy screen and cast to it but if computer restart we can't add privacy screen from phono we have to connect to a pc activate privacy screen and after use the phone

rustdesk:
- Changing resolution and http proxy settings are missing for mobile devices, will be added if needed.
Yes, continue to add these

PR #8724, opened 2024-07-16
install ffmpeg lib with vcpkg

21pages:
* Use vcpkg to install ffmpeg and build sdk from source, so no prebuild lib in hwcodec.
* Link ffmpeg in rustdesk directly, ffmpeg can be used as basic library.
* For windows developers, `VCPKG_DEFAULT_HOST_TRIPLET` env need to be set to `x64-windows-static` during installation.
* Most code in `res/vcpkg/ffmpeg` is from https://github.com/microsoft/vcpkg/tree/2024.06.15/ports/ffmpeg
* Tested windows/mac/linux/android, still work

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,9 +4,9 @@ env:
 #   MIN_SUPPORTED_RUST_VERSION: "1.46.0"
 #   CICD_INTERMEDIATES_DIR: "_cicd-intermediates"
   VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"
-  # vcpkg version: 2023.10.19
+  # vcpkg version: 2024.06.15
   # for multiarch gcc compatibility
-  VCPKG_COMMIT_ID: "8eb57355a4ffb410a2e94c07b4dca2dffbee8e50"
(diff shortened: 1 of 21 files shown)


rustdesk:
@basilgello can we enable hwcodec for fdroid?

basilgello:
Yes! @21pages does this imply libamf/libmfx are also installed from vcpkg now? I noticed rust-webm crate also builds libwebp from webm project. Maybe you can move it to vcpkg too?

basilgello:
BTW @rustdesk 1.2.6 was finally built and today it should become available to F-Droid public!

basilgello:
The vcpkg ffmpeg port depends on amd-amf, ffnvcodec and mfx-dispatch. Are you sure building amf from source in rust crate is still needed?

21pages:
Yes, vcpkg build the core code used by ffmpeg, there are some extra code used by sdk vram codec.

basilgello:
I think we can kang amd-amf port from vcpkg and add custom patches into it. Do you need my help with that?

21pages:
Can you tell why this is better?

Issue #8548, opened 2024-06-30
MacOS client session freezes when leaving RustDesk window


B4DschK4Pp:
### Bug Description

I just installed the RustDesk 1.2.6 client on MacOS Ventura and connected to a Windows PC running RustDesk 1.2.6 as well.
The initial connection is fine but when I switch to another window locally (or even just click on the local osx desktop with my mouse) and return to the RustDesk session window afterwards, the session seems to be frozen. But actually I can see on the remote side that input is still being forwarded (mouse and keyboard). The session window on the host side (osx) does not update accordingly.
This state remains until I restart the RustDesk client on MacOS and initiate a new connection.

### How to Reproduce

- Start RustDesk on osx and start a remote session to another client.
- Leave the session window/switch to another window on your MacOS client.
- Set focus on the active RustDesk session window again.
==> Video of the session is frozen.

### Expected Behavior

Reenter the session window and continue remote session with video working.

### Operating system(s) on local side and remote side

osx -> Windows 11

### RustDesk Version(s) on local side and remote side

1.2.6 -> 1.2.6

### Screenshots

https://github.com/rustdesk/rustdesk/assets/26527958/2884ea35-642d-4215-b8b1-8b5aa557252b

Left side: RustDesk on MacOS Ventura as the local side. 
Right side: remote VM being controlled.

### Additional Context

RustDesk 1.2.6 running on MacOS Ventura 13.6.7 on Intel Mac Book Pro 2020

fufesou:
@B4DschK4Pp Hi, thanks for your feedback.
Can you please also give a screenshot of the RustDesk build date on macOS side.

![image](https://github.com/rustdesk/rustdesk/assets/13586388/2d3cbb78-aa88-495c-8122-318c78972668)

gbyers:
Having the exact same issue 2020 Intel Mac Pro.

xavetar:
The last successful build version is 1.2.3-2. Subsequent versions freeze after exiting the window.

Tackyou:
I have the same issue, thought it's random. Good that @B4DschK4Pp found out it's caused by leaving the window. That should make it a little easier to debug. Have to stay on 1.2.3-2. for now. Hope this gets fixed.

gcharita:
It appears that this issue is not happening when the app is in Full Screen mode. Also, entering and then exiting Full Screen mode seams to also temporary fix the session freeze issue.

rustdesk:
[...]

PR #8738, opened 2024-07-17
fix: macos subwindow frozen, try another workaround

fufesou:


zhangbo8418:
It seems ok

rustdesk:
https://github.com/rustdesk/rustdesk/issues/8548

PR #8751, opened 2024-07-18
Revert vcpkg ffmpeg

21pages:
https://github.com/21pages/rustdesk/actions/runs/9990633075/job/27611806595


basilgello:
What's wrong with this @21pages ?

21pages:
Still have problems amd lots of time spent, https://github.com/rustdesk/rustdesk/actions/runs/9985958862/job/27598517672

basilgello:
Only CI issues with sciter or hwcodec not working elsewhere?

21pages:
Only ci currently.

basilgello:
Cool, I will look into it :)

21pages:
You can build from the commit before this pr. 

The issue with fcntl64 is likely due to a mismatch in the libc version between the Docker container and the host environment. What's strange is that the FFmpeg Linux version did not report this error. A potential solution could be to use vcpkg within the Docker container.

https://github.com/FFmpeg/FFmpeg/blob/7dabad079b783e921747de96597ea47cab244a11/libavutil/file_open.c#L88

The sciter linux armv7 version may also have this problem, hwcodec is not enabled so it is not exposed.

basilgello:
ubuntu 18.04 should definitely have fcntl64… but yes, moving vcpkg into docker is probably needed so different compilers dont come into mess

PR #8804, opened 2024-07-24
Remove existing /opt/artifacts/vcpkg in sciter builds

basilgello:

### DIFF
--- a/.github/workflows/flutter-build.yml
+++ b/.github/workflows/flutter-build.yml
@@ -1396,6 +1396,7 @@ jobs:
             # vcpkg
             export VCPKG_ROOT=/opt/artifacts/vcpkg
             pushd /opt/artifacts
+            rm -rf vcpkg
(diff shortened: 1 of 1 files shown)


rustdesk:
still not work

basilgello:
```
Downloading vcpkg tool sources 
Building vcpkg-tool...
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
CMake 3.15...3.24 or higher is required.  You are running version 3.10.2
```

Let me find cmake 3.15+ for bionic armhf

basilgello:
x86_64 linux sciter uses prebuilt CMake from Kitware but only aarch64 is built by them. Need to understand how do they build it. I know 3.16 can be built for 18.04 but we need 3.28 or newer

basilgello:
I got access to arm64 runner and solving the issue with sciter

Issue #1169, opened 2022-08-03
Extra Mouse Button and left/right wheel issues


Reikooters:
**Describe the bug you encountered:**

On a mouse with extra buttons:
- The scroll left and scroll right buttons are reversed.
- The back and forward mouse buttons are not supported at all.

**Scroll left/right:**
For example, on this page:
https://www.w3schools.com/howto/howto_css_menu_horizontal_scroll.asp

On the little example navigation area which has a horizontal scrollbar, in normal use (not using RustDesk) clicking the scroll right button (tiliting middle button to the right) will scroll to the right normally. However, when doing this on a remote machine using RustDesk, the behaviour of the scroll buttons are reversed - tilting the middle button to the right scrolls left, and tilting left scrolls right.

**Back and forward buttons:**
The back and forward buttons do not seem to be passed through to the remote machine at all (for example, these buttons are normally used to navigate back/forward in a web browser or through folders in Windows Explorer etc).

If it will help, according to what I have read online, the "back" button should be XBUTTON1 and "forward" should be XBUTTON2 [(source)](https://docs.microsoft.com/en-us/windows/apps/design/basics/navigation-history-and-backwards-navigation#handle-mouse-navigation-buttons).

https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes

For the right/left scroll, I'm unsure how they work whether it's as a button or maybe this - simply my guess.

https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-mousehwheel

On my mouse, it clicks like a button but maybe it's technically a wheel. In any case, as stated above, it is currently implemented but in reverse.

...

**What did you expect to happen instead?**
Reverse the current scroll left/right behaviour to fix the scrolling issue. Scrolling right on the mouse should scroll right, and left should scroll left.

Implement passthrough for back and forward mouse buttons.

...

**How did you install `RustDesk`?**

Build from source (yesterday)

---

**RustDesk version and environment**

Client: Windows 10, Rustdesk built from source
Remote: Windows 7, Rustdesk 1.1.9 downloaded from website.

rustdesk:
added in nigtly build, test please.

hatl:
Unfortunatel, this issue is still present in the latest version (1.2.3)
I'm using a Linux machine to control a Windows machine.

Scroll left/right are inverted.
Mouse4/5 (or back/forward) not working at all.

@rustdesk could you please check?

rustdesk:
[...]

PR #8806, opened 2024-07-24
Fix wheel drection revesed

zuiyu1998:
#1169 fix  wheel drection revesed.
Windows pc is test.


dignow:
Hi, this commit will reverse the direction vertically, not horizontally.

We also has this option for vertical direction.

f76e16e39ac20bbc66f3383c46ad7df

dignow:
Sorry, your commit reverses both horizontal and vertical values. Only horizontal direction seems reversed.

And it does not reverse horizontal values of trackpad.

PR #8855, opened 2024-07-27
Add built CMake to PATH in actual RD build step

basilgello:

### DIFF
--- a/.github/workflows/flutter-build.yml
+++ b/.github/workflows/flutter-build.yml
@@ -22,6 +22,7 @@ env:
   RUST_VERSION: "1.75" # sciter failed on m1 with 1.78 because of https://blog.rust-lang.org/2024/03/30/i128-layout-update.html
   CARGO_NDK_VERSION: "3.1.2"
   SCITER_ARMV7_CMAKE_VERSION: "3.29.7"
+  SCITER_NASM_DEBVERSION: "2.14-1"
   LLVM_VERSION: "15.0.6"
   FLUTTER_VERSION: "3.19.6"
   ANDROID_FLUTTER_VERSION: "3.13.9" # >= 3.16 is very slow on my android phone, but work well on most of others. We may switch to new flutter after changing to texture rendering (I believe it can solve my problem).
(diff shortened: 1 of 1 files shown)


basilgello:
@rustdesk ^^

.github/workflows/flutter-build.yml
```diff
@@ -1421,28 +1415,44 @@ jobs:
...
-            wget -O rust.tar.gz https://static.rust-lang.org/dist/rust-${{env.RUST_TOOLCHAIN_VERSION}}-${{ matrix.job.target }}.tar.gz
+            wget --output-document rust.tar.gz https://static.rust-lang.org/dist/rust-${{env.RUST_TOOLCHAIN_VERSION}}-${{ matrix.job.target }}.tar.gz
             tar -zxvf rust.tar.gz > /dev/null && rm rust.tar.gz
-            cd rust-${{env.RUST_TOOLCHAIN_VERSION}}-${{ matrix.job.target }} && ./install.sh
+            pushd rust-${{env.RUST_TOOLCHAIN_VERSION}}-${{ matrix.job.target }}
+            ./install.sh
+            popd
             rm -rf rust-${{env.RUST_TOOLCHAIN_VERSION}}-${{ matrix.job.target }}
-            # edit config
+            popd
```
rustdesk, line 1427: popd again?

basilgello: Yes, because the original line was 'cd rust-… && ./install.sh'and I splitted it to `pushd rust-… ` and `./install.sh` so one more popd is needed to go back to /workspace or /

basilgello: There is very simple method to check we land in the same dir we started: count that starting from `run:` statement, count of pushd == count of popd

rustdesk: it seems break x64

basilgello: It breaks both because suddenly /opt/artifacts is not generated outside… Should be easy to fix, but I assume reeulting debs are stored in workspace dir?

PR #8857, opened 2024-07-27
Try to detach /opt/artifacts from sciter

basilgello:

### DIFF
--- a/.github/workflows/flutter-build.yml
+++ b/.github/workflows/flutter-build.yml
@@ -1347,7 +1347,6 @@ jobs:
             ls -l "${PWD}"
           dockerRunArgs: |
             --volume "${PWD}:/workspace"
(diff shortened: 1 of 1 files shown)


basilgello:
@rustdesk

rustdesk:
don`t you have arm64 to test?

basilgello:
arm64 but 64-bit only :( And the onlty runner that can do armhf has no root access for me :(

rustdesk:
acutally our runner is on arm64 (64-bit) also.

basilgello:
Is it qemu? Because what I managed to find is old AMD Opteron (Seattle arch) that does armhf and aarch64. But no r00t so have to use proot.

rustdesk:
yes, arm32 runs on docker qemu of an arm64 host.

Issue #824, opened 2022-06-20
mouse bug on windows after locking screen


rustdesk:
Google translation: After locking the screen and then entering the desktop, you will encounter clicking the left button of the mouse, and the start menu will pop up once, but the mouse is not on the start button, you need to press the left button and drag it around for a few times to return to normal~~
> 最近才开始用这个，体验感比 RDP 差了点，但整体还是不错的，很满意，有个小 Bug 在这里说下：锁屏之后再进入桌面，会遇到点一下鼠标左键，就弹一次开始菜单，而鼠标是不在开始键上的，需要按着左键到处拖几下才会恢复正常~~

rustdesk:
https://github.com/rustdesk/rustdesk/releases/tag/nightly try it out on both sides

zuiyu1998:
i find that the bug  cannot be found in the latest version but the bug is esay found in the old version。 The possible reasons are that the old version will send CursorPosion on Connecting and alse send CursorPosion after locking screen.
we need check that the bug is existed on the latest version？

TheBongStack:
I am new to open source contribution and just learned rust. Can i get assigned to solving this issue?

zuiyu1998:
The bug is not find in the version.

PR #8889, opened 2024-07-30
Fix cursor lost after locking screen

zuiyu1998:
#824 
Fix cursor lost after locking screen。
I find updateCursorPosition will emit after locking screen.IsConnIn2Secs func will return false.CursorModel.gotMouseControl is false.
Simaple way is update _checkPeerControlProtected func.
Better way is not emit updateCursorPosition.

### DIFF
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -906,9 +906,7 @@ class InputModel {
     }
 
     if (!cursorModel.gotMouseControl) {
-      bool selfGetControl =
-          (x - lastMousePos.dx).abs() > kMouseControlDistance ||
-              (y - lastMousePos.dy).abs() > kMouseControlDistance;
+      bool selfGetControl = x != lastMousePos.dx || y != lastMousePos.dy;
       if (selfGetControl) {
         cursorModel.gotMouseControl = true;
       } else {


rustdesk:
Thanks a lot.

@fufesou please review

fufesou:
> I find `updateCursorPosition` will emit after locking screen. `IsConnIn2Secs` func will return `false`. `CursorModel.gotMouseControl` is false.
Simaple way is update `_checkPeerControlProtected` func.
Better way is not emit `updateCursorPosition`.

Thanks for pointing out the reason, though I cannot reproduce it.
`updateCursorPosition()` is not called when the cursor on the controlled side is not moved. I'll test more and read the code.

This PR may not be good.

1. It's not a good idea to remove the following code, because it will break some logic
```dart
      bool selfGetControl =
          (x - lastMousePos.dx).abs() > kMouseControlDistance ||
              (y - lastMousePos.dy).abs() > kMouseControlDistance;
```

2. `bool selfGetControl = x != lastMousePos.dx || y != lastMousePos.dy;` is also not a good idea.
Because it does not make any change for `updateCursorPosition()`, and directly comparing doubles is incorrect. Doubles and floats are usually compared like `(a - b).abs() < 0.00001`.

fufesou:
> I find `updateCursorPosition` will emit after locking screen.

Hi, what is the RustDesk version on the controlled side? Is your mouse moving or jittering?

The controlled side should not send the cursor movement message if the cursor is not moved.

https://github.com/rustdesk/rustdesk/blob/master/src/server/input_service.rs#L325

https://github.com/rustdesk/rustdesk/blob/master/src/server/input_service.rs#L341

zuiyu1998:
> I find updateCursorPosition will emit after locking screen.

这是我在log中看到的消息，我在家的rustdesk版本有可能是旧的。这个bug只会在勾选显示远程鼠标，且刚刚进入锁屏时触发。
```dart
  bool isConnIn2Secs() {
    if (_firstUpdateMouseTime == null) {
      _firstUpdateMouseTime = DateTime.now();
      return true;
    } else {
      return DateTime.now().difference(_firstUpdateMouseTime!).inSeconds < 2;
    }
  }
```
如果是第二次触发，这个函数会返false，并且下面这段代码不会将CursorModel.gotMouseControl置为true。因为此时的鼠标移动小于设定值，
```dart
  bool isConnIn2Secs() {
    if (_firstUpdateMouseTime == null) {
      _firstUpdateMouseTime = DateTime.now();
      return true;
    } else {
      return DateTime.now().difference(_firstUpdateMouseTime!).inSeconds < 2;
    }
  }

```

zuiyu1998:
我会将rustdesk版本更新在测试

zuiyu1998:
在使用最新版本后，updateCursorPosition并没有出现在短时间内多次触发的问题。这里的原因是旧版本会在连接成功后触发一次updateCursorPosition，在进入屏幕之后同样会触发一次updateCursorPosition，造成了短时间的多次触发，将CursorModel.gotMouseControl设置为false，进而导致远程鼠标失控。

PR #8912, opened 2024-07-31
fix: usage of `actions/checkout@v3`

hirehamir:
CI uses `actions/checkout@v3`, which uses a `deprecated` version of `Node.js`.

This pull request fixes that.


hirehamir (commented):
https://github.com/rustdesk/rustdesk/actions/runs/10180348568 has some additional context.

PR #8947, opened 2024-08-04
set max audio buffer to 150ms, clear audio buffer if full

21pages:
The device should have the capability to play a sufficient audio buffer during each period to meet the audio config, so the playback speed is not slow.

The audio delay is caused by network jitter. The controlled  side sends audio data every 10ms, but it often happens that multiple packets are sent together. During underrun periods, the controlling side plays extra silence data instead, resulting in the device playing more audio than the configured amount.

50 ms buffer length will cause frequent clear, 150 ms is ok in test. 

Clear cause short silence rather than noise.

70ab8e3b215a688a039275532c373bc


rustdesk:
https://github.com/rustdesk/rustdesk/issues/534

PR #8969, opened 2024-08-06
Set scale for rdp input.

### LINKED ISSUE #8524: Mouse Cursor Mismtach on scaled Wayland Screen
### Bug Description

Remote system is wayland with screen scaled to 150%.  Cursors are correct in the top left of the screen but diverge as the cursor moves towards the bottom right (remote cursor moves more towards the bottom right)

### How to Reproduce

Remote is a Manjaro system NVIDIA/Wayland (updated today, clean system, RustDesk just installed).
If the scaling is set to 100% the cursors track correctly.
I've tested from two local systems, the remove behaves the same with both, but both (all the systems involved) are Manjaro.

### Expected Behavior

Cursors should always align.

Manjaro->Manjaro

1.2.4->1.2.6

### Screenshots

Please let me know if these would help, but they'd just show cursors in two spots.

### Additional Context

_No response_

Shylock-Hg:
Fix #8524

### DIFF
--- a/res/DEBIAN/postinst
+++ b/res/DEBIAN/postinst
@@ -6,6 +6,7 @@ if [ "$1" = configure ]; then
 
 	INITSYS=$(ls -al /proc/1/exe | awk -F' ' '{print $NF}' | awk -F'/' '{print $NF}')
     ln -s /usr/lib/rustdesk/rustdesk /usr/bin/rustdesk
+    ln -s /usr/lib/rustdesk/get_scale /usr/bin/get_scale
 	
 	if [ "systemd" == "$INITSYS" ]; then
 
--- a/res/DEBIAN/postinst
+++ b/res/DEBIAN/postinst
@@ -6,7 +6,6 @@ if [ "$1" = configure ]; then
 
 	INITSYS=$(ls -al /proc/1/exe | awk -F' ' '{print $NF}' | awk -F'/' '{print $NF}')
     ln -s /usr/lib/rustdesk/rustdesk /usr/bin/rustdesk
-    ln -s /usr/lib/rustdesk/get_scale /usr/bin/get_scale
 	
 	if [ "systemd" == "$INITSYS" ]; then
 
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -9,36 +9,6 @@
       "host": false
     },
     {
-      "name": "glib",
-      "host": true,
-      "platform": "linux"
-    },
-    {
-      "name": "glib",
-      "host": false,
-      "platform": "linux"
-    },
-    {
-      "name": "pango",
-      "host": true,
-      "platform": "linux"
-    },
-    {
-      "name": "pango",
-      "host": false,
-      "platform": "linux"
-    },
-    {
-      "name": "graphene",
-      "host": true,
-      "platform": "linux"
-    },
-    {
-      "name": "graphene",
-      "host": false,
-      "platform": "linux"
-    },
-    {
       "name": "gtk",
       "host": true,
       "platform": "linux"
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -19,6 +19,16 @@
       "platform": "linux"
     },
     {
+      "name": "pango",
+      "host": true,
+      "platform": "linux"
+    },
+    {
+      "name": "pango",
+      "host": false,
+      "platform": "linux"
+    },
+    {
       "name": "cpu-features",
       "platform": "android"
     },
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -29,6 +29,16 @@
       "platform": "linux"
     },
     {
+      "name": "graphene",
+      "host": true,
+      "platform": "linux"
+    },
+    {
+      "name": "graphene",
+      "host": false,
+      "platform": "linux"
+    },
+    {
       "name": "cpu-features",
       "platform": "android"
     },
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -11,12 +11,14 @@
     {
       "name": "gtk",
       "host": true,
-      "platform": "linux"
+      "platform": "linux",
+      "version>=": "4.14.0"
     },
     {
       "name": "gtk",
       "host": false,
-      "platform": "linux"
+      "platform": "linux",
+      "version>=": "4.14.0"
     },
     {
       "name": "cpu-features",
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -9,6 +9,14 @@
       "host": false
     },
     {
+      "name": "glib",
+      "host": true
+    },
+    {
+      "name": "glib",
+      "host": false
+    },
+    {
       "name": "cpu-features",
       "platform": "android"
     },
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -39,6 +39,16 @@
       "platform": "linux"
     },
     {
+      "name": "gtk",
+      "host": true,
+      "platform": "linux"
+    },
+    {
+      "name": "gtk",
+      "host": false,
+      "platform": "linux"
+    },
+    {
       "name": "cpu-features",
       "platform": "android"
     },
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -10,11 +10,13 @@
     },
     {
       "name": "glib",
-      "host": true
+      "host": true,
+      "platform": "linux"
     },
     {
       "name": "glib",
-      "host": false
+      "host": false,
+      "platform": "linux"
     },
     {
       "name": "cpu-features",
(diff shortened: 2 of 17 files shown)


rustdesk:
Looks good, just need to add to the other packages, deb, yum, pkg, appimage, flatpak etc. and fix ci.

Shylock-Hg:
> Looks good, just need to
> 
> * [ ]  add to linux packages, deb, yum, pkg, appimage, flatpak etc.
> * [ ]  set correct path
> * [ ]  fix ci.
> * [ ]  more tests

What's the meaning of `set correct path`? And I don't see the test code, where to add it?

res/DEBIAN/postinst
```diff
@@ -6,6 +6,7 @@ if [ "$1" = configure ]; then
 
 	INITSYS=$(ls -al /proc/1/exe | awk -F' ' '{print $NF}' | awk -F'/' '{print $NF}')
     ln -s /usr/lib/rustdesk/rustdesk /usr/bin/rustdesk
+    ln -s /usr/lib/rustdesk/get_scale /usr/bin/get_scale
```
rustdesk: I do not think this is necessary, let us call this with abs path according to the path of current exe.

rustdesk:
> What's the meaning of `set correct path`? And I don't see the test code, where to add it?

https://github.com/rustdesk/rustdesk/pull/8969#discussion_r1705433875

res/DEBIAN/postinst
Shylock-Hg: ok

Shylock-Hg:
@rustdesk Could enable CI?

basilgello:
Re: addition of glib to vcpkg.json - is it really needed on mobile platforms (i.e, android) ?

Shylock-Hg:
> Re: addition of glib to vcpkg.json - is it really needed on mobile platforms (i.e, android) ?

Set it to Linux only now.

Shylock-Hg:
@rustdesk Could enable CI by default?

rustdesk:
No CI on your repo? no need to depends on our CI.

Shylock-Hg:
> No CI on your repo? no need to depends on our CI.

Ok, I will try it.

vcpkg.json
```diff
@@ -38,6 +38,16 @@
...
+    {
+      "name": "gtk",
+      "host": true,
+      "platform": "linux"
+    },
+    {
+      "name": "gtk",
+      "host": false,
+      "platform": "linux"
+    },
```
rustdesk, line 22: Why gtk in vcpkg? @basilgello you could help confirm.

basilgello: I looked through the PR and the approach is valid. However, bringing full GTK4 for just that purpose is MASSIVE waste of build resources.
I am looking if there is a some other standalone library for monitor things

basilgello: For reference, here it is all that needed for GdkMonitor:

https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/gdkmonitor.h
https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/gdkmonitor.c
https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/gdkmonitorprivate.h

https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/x11/gdkmonitor-x11.c
https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/x11/gdkmonitor-x11.h

https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/broadway/gdkmonitor-broadway.c
https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/broadway/gdkmonitor-broadway.h

https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/macos/gdkmacosmonitor-private.h
https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/macos/gdkmacosmonitor.h
https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/macos/gdkmacosmonitor.c

https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/wayland/gdkmonitor-wayland.c
https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/wayland/gdkmonitor-wayland.h

https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/win32/gdkmonitor-win32.c
https://chromium.googlesource.com/external/github.com/GNOME/gtk/+/refs/tags/upstream/4.15.4/gdk/win32/gdkmonitor-win32.h

That's for all supported platforms :)

rustdesk: Yes, it is such a small function, but introduce big complexity, seems not worthwhile.

basilgello: I'd isolate the minimal init code, make 1 c or cpp file and build it via vcpkg like oboe-wrapper.

Shylock-Hg: > I'd isolate the minimal init code, make 1 c or cpp file and build it via vcpkg like oboe-wrapper.

I can't get your point. Do you mean extract source code of GtkMonitor from gtk? But it's almost coupled with other components of gtk, I don't think it's possible.

basilgello: It is possible but I noticed that linux builds use "system" gtk headers. Even Flatpak and AppImage ship libsciter-gtk.so and gtk3.so. Can we use that lib instead of shipping one more instance of gtk4?

Shylock-Hg: This fractional scale is support only for gtk >= 4.14

rustdesk:
I am wondering how gtk4 get the scale? can not we find the root source?

vcpkg.json
rustdesk: > Can we use that lib instead of shipping one more instance of gtk4?

No, I do not wanna maintain more things, we are already busy enough.

rustdesk: > I'd isolate the minimal init code, make 1 c or cpp file and build it via vcpkg like oboe-wrapper.

You don't have to use vcpkg. We use vcpkg is for its cross-platform advantage. This is for Linux only. You do not need to waste time on vcpkg.

rustdesk: That's why somebody added `linux-pkg-config` feature.

https://github.com/rustdesk/rustdesk/blob/ae16b8975bdb66c7e74c600797fd98826f107eb7/libs/scrap/build.rs#L8

Shylock-Hg:
> I am wondering how gtk4 get the scale? can not we find the root source?

From here, https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/wayland/gdkdisplay-wayland.c?ref_type=heads#L2404

vcpkg.json
Shylock-Hg: > > I'd isolate the minimal init code, make 1 c or cpp file and build it via vcpkg like oboe-wrapper.
> 
> You don't have to use vcpkg. We use vcpkg is for its cross-platform advantage. This is for Linux only. You do not need to waste time on vcpkg.

I use vcpkg because older ubuntu in CI doesn't support the newer version of gtk.

basilgello:
One alternative is to extend https://crates.io/crates/display-info with wayland / broadway implementation

rustdesk:
Agree with @basilgello , there must be more smart way to get the scale, rather than compile giant gitk4.

basilgello:
@Shylock-Hg How do you feel with option above?

Shylock-Hg:
> @Shylock-Hg How do you feel with option above?
Screenshot_20240808_144352

`dispaly-info` has the same issue as `gtk3`, it get `2` when scale is `150%`,

Shylock-Hg:
If introducing entire `gtk4` by `vcpkg` is not acceptable, maybe there is only way: reference the logical of how `gtk4` get fractional scale to write a simple library.

basilgello:
Let's fix display_info then. The cratevis gonna be popular anyway :)

Shylock-Hg:
@rustdesk I have another way to do it: we could set this scale by command line argument by user. How do you think about it?

Shylock-Hg:
Follow computing to get scale as gtk4. Open new PR #9019

PR #8991, opened 2024-08-08
Android: Enable JNI mode on client too

basilgello:

### DIFF
--- a/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/FFIInitSingleton.kt
+++ b/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/FFIInitSingleton.kt
@@ -1,0 +1,22 @@
+package com.carriez.flutter_hbb
+
+import ffi.FFI
+
+import android.content.Context
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * Static singleton class to initialize NDK contexts in FFI for hardware codecs
+ */
+
+enum class FFIInitSingleton {
+    INSTANCE;
+
+    private val initCalled = AtomicBoolean(false)
+
+    fun init(ctx: Context) {
+        if (initCalled.compareAndSet(false, true)) {
+            FFI.init(ctx)
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


rustdesk:
So it will be called twice?

basilgello:
Service is AFAIK started in isolated process

rustdesk:
Yes, it can be configured to run in isolated process. But we run it in thread. Please do not suggest running it in process, which will introduce a lot of complexity and break the whole RustDesk.

basilgello:
OK :) Then a static singleton should do the trick between MainActivity and MainService in one thread. Actually, `av_jni_set_java_vm` will err out if the same pointer is about to be assigned but we should not rely on it and do things properly.

flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/FFIInitSingleton.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Static singleton class to initialize NDK contexts in FFI for hardware codecs
+ */
+
+enum class FFIInitSingleton {
+    INSTANCE;
+
+    fun init(ctx: Context) {
+        FFI.init(ctx)
+    }
+}
```
rustdesk, line 22: I do not think this is correct, though I have never learned Kotlin.

rustdesk: FFIInitSingleton.INSTANCE is singleton, it does mean init called once.

rustdesk: Chatgpt can help you.
```
import android.content.Context
import java.util.concurrent.atomic.AtomicBoolean

/**
 * Static singleton class to initialize NDK contexts in FFI for hardware codecs
 */

enum class FFIInitSingleton {
    INSTANCE;

    private val initCalled = AtomicBoolean(false)

    fun init(ctx: Context) {
        if (initCalled.compareAndSet(false, true)) {
            FFI.init(ctx)
        } else {
            // Optionally, you can log or handle the case where init has already been called
            println("Init has already been called.")
        }
    }
}
```

basilgello: Yup I noticed :) Thanks

PR #9012, opened 2024-08-09
trust this device to skip 2fa

21pages:
https://github.com/rustdesk/rustdesk/discussions/8513#discussioncomment-10234494

**Steps of trust one device:**

* The controlled side enables 2FA and the "Enable trusted devices" option.
* The control side connects to the controlled side, after the password is verified, if `enable_trusted_devices` proto is true,  there will be a checkbox under the 2FA input field. If the checkbox is checked and the 2FA is correct, the digest of the control side's HWID will be stored on the controlled side, and the checkbox option will be remembered in the `trust-this-device` peer option, which defaults to false.
* In the next connection, if the `trust-this-device` peer option is true, the HWID digest will be sent in the login request, and if the digest is found in the trusted devices list, the 2FA verification will be skipped

**Details**

* Trusted devices will be cleared when:
  * the permanent password is changed.
  * 2FA is disabled.
* The entire trusted devices list is encrypted, because we don't want the expiration time to be modified manually. The maximum size is 1MB, which is enough to store several thousand trusted devices.
* The expiration time is 90 days, and outdated items are removed when the trusted devices are read or set.
* The `enable_trusted_devices` proto is only set when the login error is `REQUIRE_2FA`, which can save network payload.
* In addition to HWID, the ID, platform, and name are also compared.
* When 2FA dialog is shown, the `trust-this-device peer` option is set to false. If the checkbox is selected, the `trust-this-device` peer option is set to true.

**Sciter**
* Sciter doesn't add `Enable trusted devices` option, only because `Manage trusted devices` dialog is not written, but sciter has grid example. I will add it to sciter and add bot to mobile and sciter.

**Test**
* Windows/Linux/Macos/Android

https://github.com/user-attachments/assets/30c93524-0d1d-4944-97d9-0836c99f2c4b

https://github.com/user-attachments/assets/917afd49-a233-4a8a-9824-7f46b4ac658c

https://github.com/user-attachments/assets/c891cd09-bff2-481e-bef4-640c71aba812


src/server/connection.rs
```diff
@@ -1628,6 +1628,20 @@ impl Connection {
         if let Some(o) = lr.option.as_ref() {
             self.options_in_login = Some(o.clone());
         }
+        if self.require_2fa.is_some() && !lr.hwid.is_empty() {
+            let devices = Config::get_trusted_devices();
+            let encrypted_hwid = TrustedDevice::encrypt_hwid(&lr.hwid);
+            if let Some(device) = devices.iter().find(|d| d.hwid == encrypted_hwid) {
+                if !device.outdate()
```
rustdesk, line 1645: Where do you remove outdated trust device?

flutter/lib/desktop/pages/desktop_setting_page.dart
```diff
@@ -783,8 +786,66 @@ class _SafetyState extends State<_Safety> with AutomaticKeepAliveClientMixin {
           onChangedBot(!hasBot.value);
         },
       ).marginOnly(left: _kCheckBoxLeftMargin + 30);
+
+      updateTrust() async {
+        hasTrust.value = option2bool(kOptionAllowTrustThisDevice,
```
rustdesk: The default value of trust this device on controlled side should be Y, so it should be `enable` prefix.

flutter/lib/desktop/pages/desktop_setting_page.dart
```diff
@@ -783,8 +786,66 @@ class _SafetyState extends State<_Safety> with AutomaticKeepAliveClientMixin {
...
+            bind.mainGetOptionSync(key: kOptionAllowTrustThisDevice));
+        setState(() {});
+      }
+
+      onChangeTrust(bool? checked) async {
```
rustdesk: Mobile?

libs/hbb_common/src/config.rs
```diff
@@ -210,6 +211,8 @@ pub struct Config2 {
     serial: i32,
     #[serde(default, deserialize_with = "deserialize_string")]
     unlock_pin: String,
+    #[serde(default, deserialize_with = "deserialize_string")]
+    trusted_device: String,
```
rustdesk: trusted_devices?

flutter/lib/common/widgets/dialog.dart
```diff
@@ -1921,9 +1981,28 @@ void enter2FaDialog(
...
+        ));
+
     return CustomAlertDialog(
         title: Text(translate('enter-2fa-title')),
-        content: codeField,
+        content: Column(
+          children: [
+            codeField,
+            if (option2bool(kOptionAllowTrustThisDevice,
+                bind.mainGetOptionSync(key: kOptionAllowTrustThisDevice)))
```
rustdesk: Is this peer option?

since you introduced controlled side `trust this device` option. So if controlled side turn off this, controlling side should not show this checkbox anymore, because it is useless.

src/server/connection.rs
21pages: In `set_trusted_devices` and `get_trusted_devices` in config.rs

rustdesk:
> I will add it to sciter and add bot to mobile and sciter.

No for Sciter. Next time, for anyting about Sciter, ask me before taking action. I need to judge if necessary.

rustdesk:
> > I will add it to sciter and add bot to mobile and sciter.
> 
> No for Sciter, at least for the time being. Next time, for anyting about Sciter, ask me before taking action. I need to judge if necessary to do Sciter.

I meant now further work on `Enable trusted devices` option and managed device dialog for Sciter, the other simple thing should add, e.g. the checkbox under 2fa inputbox.

src/server/connection.rs
```diff
@@ -1623,11 +1626,35 @@ impl Connection {
...
 
+    #[inline]
+    fn enable_trusted_devices() -> bool {
+        #[cfg(feature = "flutter")]
+        return config::option2bool(
+            config::keys::OPTION_ENABLE_TRUSTED_DEVICES,
+            &Config::get_option(config::keys::OPTION_ENABLE_TRUSTED_DEVICES),
+        );
+        #[cfg(not(feature = "flutter"))]
+        return false;
```
rustdesk, line 1637: change to `true`.

src/server/connection.rs
```diff
@@ -1841,6 +1865,15 @@ impl Connection {
...
                         }
+                        if !tfa.hwid.is_empty() && Self::enable_trusted_devices() {
+                            Config::add_trusted_device(TrustedDevice {
+                                hwid: tfa.hwid,
+                                time: hbb_common::get_time(),
+                                id: self.lr.my_id.clone(),
+                                name: self.lr.my_name.clone(),
+                                platform: self.lr.my_platform.clone(),
+                            });
+                        }
```
rustdesk, line 1876: Make sure not added again and again.

21pages: By removing the old one.
```
pub fn add_trusted_device(device: TrustedDevice) {
        let mut devices = Self::get_trusted_devices();
        devices.retain(|d| d.hwid != device.hwid);
        devices.push(device);
        Self::set_trusted_devices(devices);
    }
```

rustdesk: If no change, why still call `set_trusted_devices`? wasting io.

21pages: `time` may be updated. This can happen when `Enable trusted devices` is disabled, then being connected to,  so hwid will not be sent, then option enabled, then user send 2fa when in trusted devices. During one device is trusted and option is enabled, will not go to this line.

PR #9023, opened 2024-08-11
fix: win, virtual display

fufesou:
https://github.com/rustdesk/rustdesk/discussions/8044

### Fix "No physical displays, no need to use the privacy mode."

May fix https://github.com/rustdesk/rustdesk/discussions/8994#discussioncomment-10300985

#### Bug

https://github.com/user-attachments/assets/9d55eae0-2a6e-4973-873b-79070f49b6f7

#### Fixed

https://github.com/user-attachments/assets/72310e89-f12c-438b-9319-4fdec05cf2f7

### Default resolution 1920x1080

https://github.com/user-attachments/assets/42c39a49-2439-41d1-a739-0d6014de3586

### Restore on conn & disconn

https://github.com/user-attachments/assets/8f9c033d-0cc3-405d-a55d-f6b280c04d3c

The options is stored in the peer's configuration.

```toml
[options]
virtual-display = ''
```

or

```toml
[options]
virtual-display = ',0,0'
```

### Fix "No physical displays, no need to use the privacy mode."

The registry value is restored, but the settings in display settings is not refreshed if no virtual displays are plugged in/out.

We can manually trigger the refresh by the following call if no vitual displays are pluged in/out. 

image

### Default resolution 1920x1080

1. Move `mod reg_display_settings` from `src/privacy_mode/win_virtual_display.rs` to `src/platform/windows.rs`.
2. Spawn a new thread to check if the virtual display is plugged in for the first time.

image

image

### Restore on conn & disconn

1. Save option.

image

2. Remove virtual displays on disconnection.

image

3. Restore virtual displays.

image

**Note**:
1. Virtual displays are removed when the last connection is closed on the controlled side.  
If `A`,`B` -> `C`, `A` has one virtual display. Then `A` disconnect and connect, there will be two virtual displays on `C`.
Because `B` did not disconnect, `C` will not remove virtual displays. `A` will plug in one virtual display on conn.

We cannot control the display index to plug in with amyuni idd.

2. After all connections are closed, all virtual displays of amyuni idd will be deleted.

### DIFF
--- a/src/client/io_loop.rs
+++ b/src/client/io_loop.rs
@@ -937,6 +937,24 @@ impl<T: InvokeUiSession> Remote<T> {
         }
     }
 
+    async fn send_toggle_virtual_display_msg(&self, peer: &mut Stream) {
+        let lc = self.handler.lc.read().unwrap();
+        let displays = lc.get_option("virtual-display");
+        for d in displays.split(',') {
+            if let Ok(index) = d.parse::<i32>() {
+                let mut misc = Misc::new();
+                misc.set_toggle_virtual_display(ToggleVirtualDisplay {
+                    display: index,
+                    on: true,
+                    ..Default::default()
+                });
+                let mut msg_out = Message::new();
+                msg_out.set_misc(misc);
+                allow_err!(peer.send(&msg_out).await);
+            }
+        }
+    }
+
     async fn send_toggle_privacy_mode_msg(&self, peer: &mut Stream) {
         let lc = self.handler.lc.read().unwrap();
         if lc.version >= hbb_common::get_version_number("1.2.4")
@@ -1073,6 +1091,7 @@ impl<T: InvokeUiSession> Remote<T> {
                         self.handler.close_success();
                         self.handler.adapt_size();
                         self.send_opts_after_login(peer).await;
+                        self.send_toggle_virtual_display_msg(peer).await;
                         self.send_toggle_privacy_mode_msg(peer).await;
                     }
                     let incoming_format = CodecFormat::from(&vf);
--- a/src/flutter.rs
+++ b/src/flutter.rs
@@ -1780,6 +1780,54 @@ pub fn try_sync_peer_option(
     }
 }
 
+pub(super) fn session_update_virtual_display(session: &FlutterSession, index: i32, on: bool) {
+    let virtual_display_key = "virtual-display";
+    let displays = session.get_option(virtual_display_key.to_owned());
+    if !on {
+        if index == -1 {
+            if !displays.is_empty() {
+                session.set_option(virtual_display_key.to_owned(), "".to_owned());
+            }
+        } else {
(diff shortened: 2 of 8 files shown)


src/client/io_loop.rs
```diff
@@ -937,6 +937,24 @@ impl<T: InvokeUiSession> Remote<T> {
...
     }
 
+    async fn send_toggle_virtual_display_msg(&self, peer: &mut Stream) {
+        let lc = self.handler.lc.read().unwrap();
+        let displays = lc.get_option("virtual-display");
+        for d in displays.split(',') {
+            if let Ok(index) = d.parse::<i32>() {
+                let mut misc = Misc::new();
+                misc.set_toggle_virtual_display(ToggleVirtualDisplay {
+                    display: index,
```
rustdesk, line 947: No size (resolution)? Can we restore resolution after connect next time? Have you confirmed we have exactly the same behavior as ToDesk?

fufesou: > No size (resolution)? Can we restore resolution after connect next time?

Yes.
The resolutions in system `Display settings` (`HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\GraphicsDrivers\Connectivity`) are used.

ToDesk also use the system `Display settings`.

> Have you confirmed we have exactly the same behavior as ToDesk?

Not exactly.

1. ToDesk does not allow multiple control side, it requires external channel plug-in.
2. We can't control the virtual display index. So there is a drawback.

image

rustdesk: > The resolutions in system `Display settings` (`HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\GraphicsDrivers\Connectivity`) are used.

You have unplugged (all virtual displays of amyuni idd will be deleted), the resolution info (e.g. the user changes the virtual display resolution manually from our toolbar) still there?

fufesou: Yes.

The display settings are stored in registry `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\GraphicsDrivers\Connectivity`.
The displays settings will be applied for the same display combinations.

Eg. You have two displays `A`, `B`, you have ever set the display settings like:
1. `A`, `1920x1080`
2. `B`, `1440x900`
3. `A+B`, `2560x1440` and `1024x768`

Then, you have plugged out all displays
1. Plug in `A`, the resolution is `1920x1080`
2. Plug in `B`, the resolutions will be `2560x1440` and `1024x768`
3. Plug out `A`, the resolutions is `1440x900`



For `amyuni` idd, the resolutions are just changed without recording

https://github.com/rustdesk/rustdesk/blob/ce56be65077602d6fa4713ecb8d80d7ce6e71d42/src/server/connection.rs#L2711

Then, the controlled side will never try recovery the resolution set by self.

The plugged in display will always apply the system display settings. (The combination of the displays).

PR #9098, opened 2024-08-17
Vumstar patch 1

vumstar:


basilgello:
@rustdesk I wonder what these guys expect to achieve :) No bug bounty policy is clear

rustdesk:
No need to wonder, this world is full of such guys.

PR #9138, opened 2024-08-21
replace appindicator to ayatana-appindicator

21pages:
1. Replace  appindicator to ayatana-appindicator
* debian: tested on ubuntu 22.04
* debian flatpak: still use old appindicator, the control file of deb doesn't effect flatpak because it's not used, the tray doesn't show with nightly build.
* fedora: tested on fedora 38, the tray doesn't show with nightly build.
* suse: only test install because failed to install desktop
* arch: still use old one becase failed to build with `libayatana-appindicator`, althrough the package can be found on latest arch. https://github.com/21pages/rustdesk/actions/runs/10489496699/job/29054390628
2. Add to recommands, not implemented.
* debian: libayatana-appindicator3-1 is not added to `Recommands`, because libayatana-appindicator3-1 is not installed when it's in `Recommands` on ubuntu 22.04
* Rpm not well supported, https://unix.stackexchange.com/questions/73690/rpmbuild-is-there-a-way-to-specify-recommended-packages
3. Searched package, althrough not same as package in control file.
Arch:
9b20c850ae64af9a52cda45ff37b7b6
Suse:
3717f4935508eea79916f76cf6641c5
Fedora:
99989cc75bc6d9bc9833adcd2e954db

### DIFF
--- a/.github/workflows/flutter-build.yml
+++ b/.github/workflows/flutter-build.yml
@@ -884,7 +884,7 @@ jobs:
                git \
                g++ \
                g++-multilib \
-               libappindicator3-dev \
+               libayatana-appindicator3-dev \
                libasound2-dev \
                libc6-dev \
                libclang-10-dev \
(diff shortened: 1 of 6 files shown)


rustdesk:
Ubuntu 18 does not have libayatana-appindicator3-1

21pages:
libayatana-appindicator3-1 can't be added to Depends

PR #9192, opened 2024-08-27
fix: web, reset cursor on disconn, back to main page

fufesou:


flutter/lib/models/input_model.dart
```diff
@@ -784,7 +784,7 @@ class InputModel {
     if (!isInputSourceFlutter) {
       bind.sessionEnterOrLeave(sessionId: sessionId, enter: enter);
     }
-    if (enter) {
+    if (!isWeb && enter) {
```
rustdesk, line 787: Android / iOS?

fufesou: `enterOrLeave()`, no mobile.

Issue #9222, opened 2024-09-01
等待拷贝响应超时


Godor:
### Bug Description

等待拷贝响应超时

之前以为是版本不一样的问题，升级后依然存在这个问题

![screenshot-20240901-170552](https://github.com/user-attachments/assets/a7db3173-8bb9-4b81-b215-d07e3842663c)

### How to Reproduce

复制文本内容。

### Expected Behavior

复制文本内容出现 不能  复制到对方电脑。后来发现频繁有这个提示  

### Operating system(s) on local side and remote side

win10  - win10

### RustDesk Version(s) on local side and remote side

1.3.0-1.3.0

### Screenshots

![screenshot-20240901-170552](https://github.com/user-attachments/assets/906f33c4-e380-40e9-80b2-95e0f54972d3)

fufesou:
Does this issue occur every time or very often?

Strangely, this error message should only appears when copying files, while not when copying text.

fufesou:
Hi

1. Do you use the public server or self-host server?
2. Could you please confirm whether the issue occurs when copying files or text?
3. Could you please give a screenshot of the delay?

![image](https://github.com/user-attachments/assets/eef00848-7901-4a96-b24f-5b825bb8eeb0)

rustdesk:
Close since inactivity.

Godor:
utter: The Dart VM service is listening on http://127.0.0.1:58805/xurW0Bmm5_w=/
flutter: launch args: []
flutter: initializing FFI main
flutter: _appType:main,info1-id:Huolong,info2-name:Huolong,dir:G:\Users\鏂囨。
flutter: _globalFFI init
flutter: _globalFFI init end
flutter: registerEventHandler native_ui native_ui
flutter: reset ab model
[2024-09-04T14:24:29Z INFO&nbsp; scrap::common::hwcodec] set hwcodec config
[...]

PR #9243, opened 2024-09-03
fix: win, file clipboard

fufesou:
Related issue https://github.com/rustdesk/rustdesk/issues/9222
But this commit probably didn't fix it.

1. Return the result of `wait_response_event()` in `cliprdr_send_format_list()`

```C
static UINT cliprdr_send_data_request(UINT32 connID, wfClipboard *clipboard, UINT32 formatId)
```

The last line does not return any value, which causes **undefined behavior**.

image

2. Add recv flags to avoid waiting a long time if failed to `SetEvent()`.

image

image

3. Return `ERR_CODE_SEND_MSG` if failed to send clipboard message. Then the thread will not wait the response message.

## The possible reasons for https://github.com/rustdesk/rustdesk/issues/9222

1. Failed to send clipboard message. (Local side)
2. Failed to receive clipboard message. (Remote side)
3. Failed to send response. (Remote side)
4. Failed to receive response. (Local side)
5. Failed to notify the waiting thread of the received response.
6. Dead lock.

I did not find dead lock.

This commit improves the logic of steps 1 and 5.

## Tests

Copy&Paste files is tested.

### DIFF
--- a/libs/clipboard/src/lib.rs
+++ b/libs/clipboard/src/lib.rs
@@ -5,7 +5,7 @@ use std::{
 };
 
 #[cfg(any(target_os = "windows", feature = "unix-file-copy-paste",))]
-use hbb_common::{allow_err, log};
+use hbb_common::{allow_err, bail};
 use hbb_common::{
     lazy_static,
     tokio::sync::{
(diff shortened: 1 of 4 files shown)


fufesou:
I will keep looking for possible causes and fixes.

PR #9387, opened 2024-09-18
fix: option OPTION_ONE_WAY_FILE_TRANSFER

fufesou:
Fix, copy&paste files one direction. https://rustdesk.com/docs/en/self-host/client-configuration/advanced-settings/#one-way-clipboard-redirection

1. Rename `is_stopping_allowed_from_peer()` to `is_beginning_message()` in `ClipboardFile`.
2. Check `_clip.is_beginning_message()` and then option `OPTION_ONE_WAY_FILE_TRANSFER`. Because some messages must be sent back to the control side when "coping&pasting".

### DIFF
--- a/libs/clipboard/src/lib.rs
+++ b/libs/clipboard/src/lib.rs
@@ -132,7 +132,7 @@ impl ClipboardFile {
         )
     }
 
-    pub fn is_stopping_allowed_from_peer(&self) -> bool {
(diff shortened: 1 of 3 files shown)


fufesou:
> THe other 3 options tested?

Yes.

`OPTION_ONE_WAY_FILE_TRANSFER` is still not good in file transfer connection. I'll submit another PR.

PR #9428, opened 2024-09-23
Fix F-Droid build and bump Android NDK

basilgello:
Second attempt to fix F-Droid build after @21pages showed that the flutter-build CI fix was not propagated to `build_fdroid.sh`.

Alsobump Android NDK to r27b from r27.


basilgello:
@rustdesk :)

PR #9540, opened 2024-10-01
Update build.rs

akshitbansal2005:
Analysis of the Code
Configuration Macros:

The use of #[cfg(...)] is appropriate for conditional compilation based on the target operating system or feature flags. Error Handling:

There is a lack of error handling in several places. For example, std::env::var calls could fail if the environment variable is not set. This can lead to panics, which is undesirable in production code. Code Duplication:

The install_android_deps function has repetitive code for setting target_arch. This could be streamlined. Unused Imports:

There are some unused imports, such as use std::io::Write; inside build_manifest. They should be scoped or imported only if necessary. Formatting and Readability:

The use of format! in println! statements can be simplified. In most cases, the format string can be passed directly to println!. Potential Logical Errors:

In install_android_deps, if the CARGO_CFG_TARGET_OS is not android, the function returns immediately. However, it could be clearer if it included a log statement indicating this early exit. Code Organization:

The function main could benefit from better organization or helper functions to improve readability.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,31 +1,37 @@
 #[cfg(windows)]
 fn build_windows() {
-    let file = "src/platform/windows.cc";
-    let file2  = "src/platform/windows_delete_test_cert.cc";
-    cc::Build::new().file(file).file(file2).compile("windows");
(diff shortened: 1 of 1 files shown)


rustdesk:
Thanks but sorry.

Issue #8760, opened 2024-07-19
CapsLock fail (11 -> Ubuntu 22, lang?)


alanmcanonical:
### Bug Description

Client side:
OS: Windows 10 x64
version: rustdesk-1.2.6-x86_64.exe
Server side:
OS: Ubuntu 22.04
version: 1.2.6 (apt install)

The CapsLocks works fine outside the control window but only outputs lower case char inside the control window

### How to Reproduce

Use the portable rustdesk on Windows 11 x64 and use apt to install rustdesk on Ubuntu. The Ubuntu serves as server side.
Turn on the CapsLock and type whatever character inside the **Firefox/librewolf** of remote side (which I think could be the root cause)

### Expected Behavior

Type in uppercase char

### Operating system(s) on local side and remote side

Windows 11 -> Ubuntu 22.04 amd64 

### RustDesk Version(s) on local side and remote side

1.2.6 -> 1.2.6

### Screenshots

N/A

fufesou:
Hi, thanks for your feedback.

Do you use the "Legacy mode"? Please use the "Map mode" or "Translate mode".    https://github.com/rustdesk/rustdesk/wiki/FAQ#keyboard-translation-modes

![image](https://github.com/user-attachments/assets/a708a3e4-355c-49c1-9275-aa24c24552c8)

rustdesk:
similar, https://github.com/rustdesk/rustdesk/issues/8599, just as a ref

rustdesk:
tracked by https://github.com/rustdesk/rustdesk/issues/4367

rustdesk:
https://github.com/rustdesk/rustdesk/discussions/9060#discussioncomment-10910327

PR #9641, opened 2024-10-14
fix: wayland CapsLock

fufesou:
https://github.com/rustdesk/rustdesk/issues/8760

Fix `CapsLock` when the remote side is Wayland.

Flatpak and Appimage, which use rdp input(xdg desktop portal) is still not fixed. https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.RemoteDesktop.html

### uinput

https://github.com/user-attachments/assets/72865cd3-11fe-4506-b9cd-a0d900ed9e89

### rdp

https://github.com/user-attachments/assets/b864c90a-785f-463e-8eec-873d6ae03025

## Desc

RustDesk sends keys with additional `CapsLock` state to the remote side.

The remote side will 
1. Click `CapsLock`.
2. Simulate key.
3. Click `CapsLock` to revert the state.

But for uinput and rdp input, there should be a delay to ensure the `CapsLock` state take effect.

### Uinput

The remote side will click `CapsLock` and wait the state to be changed when it finds current state is not the same to the key event from the local side.

But it will not change the `CapsLock` state back.
Because there must be a delay to ensure the lock state is applied on Wayland uinput, which may affect the user experience.

**Note**: So there's a side effect that the `CapsLock` state may be changed after the remote control

### Rdp input

Rdp input just add a delay to wait the `CapsLock` state be changed. So the letters may not be correct.

Because there's no api to query the `CapsLock` state.
`/usr/input/eventX` also requires the root permission.

## Tests

The following test are `Windows -> `, "Map mode" and "Translate mode".

- [x] Linux
  - [x] X11
  - [x] Wayland
    - [x] Installed version
    - [x] Appimage and Flatpak.
- [x] Windows
- [x] MacOS

**Note**: Appimage and Flatpak are not good as descripted above.

### DIFF
--- a/src/server/input_service.rs
+++ b/src/server/input_service.rs
@@ -175,6 +175,22 @@ impl LockModesHandler {
         }
     }
 
+    #[cfg(target_os = "linux")]
+    fn sleep_to_ensure_locked(v: bool, k: enigo::Key, en: &mut Enigo) {
+        if wayland_use_uinput() {
+            // Sleep at most 500ms to ensure the lock state is applied.
+            for _ in 0..50 {
+                std::thread::sleep(std::time::Duration::from_millis(10));
+                if en.get_key_state(k) == v {
+                    break;
+                }
+            }
+        } else if wayland_use_rdp_input() {
+            // We can't call `en.get_key_state(k)` because there's no api for this.
+            std::thread::sleep(std::time::Duration::from_millis(50));
+        }
+    }
+
     #[cfg(any(target_os = "windows", target_os = "linux"))]
     fn new(key_event: &KeyEvent, is_numpad_key: bool) -> Self {
         let mut en = ENIGO.lock().unwrap();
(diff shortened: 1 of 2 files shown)


src/server/input_service.rs
```diff
@@ -1609,6 +1638,7 @@ pub fn handle_key_(evt: &KeyEvent) {
     #[cfg(not(any(target_os = "android", target_os = "ios")))]
     let mut _lock_mode_handler = None;
     #[cfg(not(any(target_os = "android", target_os = "ios")))]
+    if evt.down {
```
rustdesk: 100% sure?

fufesou: Yes, I've tested again. Windows, Linux and MacOS as the controlled side, "Map mode" and "Translate mode".

fufesou:
Shall we just simulate `CapsLock` click and ignore the `Capslock` state in the key events if the controlled side is Appimage and Flatpak?

PR #9643, opened 2024-10-14
combine upload files/folder button

21pages:
1. Combine upload files/folder button
2. Compress web cache

https://github.com/user-attachments/assets/25e5ea25-0574-4f76-8f4d-535c415058d1


rustdesk:
do not forget.change left/right icon to down/up icon.

Issue #8786, opened 2024-07-23
when I type in Korean the first character automatically displays the number 1, how to fix it?


rustdesk:
### Discussed in https://github.com/rustdesk/rustdesk/discussions/8601

<div type='discussions-op-text'>

<sup>Originally posted by **B-YS** July  4, 2024</sup>
I try connect from iphone to imac and window, then get the same error

https://github.com/rustdesk/rustdesk/assets/66666649/4fc91209-e0b7-4936-b614-98a985fabc89

</div>

rustdesk:
related to this, https://github.com/rustdesk/rustdesk/pull/8471

khieu-dv:
This error reappears on the latest version 1.3.5.

rustdesk:
The fix has been reverted, because it caused more problem in other languages. So far, we have no plan to fix it yet.

PR #9644, opened 2024-10-14
fix: TextFormField onChanged event triggered multiple times when Korean input

shleyZ:
issue: [(8786)](https://github.com/rustdesk/rustdesk/issues/8786#issuecomment-2401286686)

The cause of this bug has been identified. When entering and clicking on Korean content, the onChanged event of the TextFormField is triggered multiple times.

For example: 
When inputting ㅎ → ㅏ → 하, the onChanged event is triggered 4 times: ㅎ → ' ' → ㅏ → ' ' → 하. 
When inputting ㄱ → ㅏ → ㅇ → 강, the event is triggered 6 times: ㄱ → ' ' → ㅏ → ' ' → ㅇ → ' ' → 강.

This issue is caused by the positional changes of Korean characters with each new symbol. These multiple unnecessary onChanged triggers cause our _handleIOSSoftKeyboardInput function to execute incorrectly, which leads to the previous bug.

We only need to give a little more time to allow the keyboard input engine to finish constructing the Korean characters before triggering the onChanged event. Of course, this is somewhat of a hacky fix.

### DIFF
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -524,6 +524,8 @@ class _RemotePageState extends State<RemotePage> {
                       keyboardType: TextInputType.multiline,
                       onChanged: (newValue) async {
                         if (isIOS) {
+                          // fix: TextFormField onChanged event triggered multiple times when Korean input
+                          // https://github.com/rustdesk/rustdesk/pull/9644
                           await Future.delayed(const Duration(milliseconds: 10));
 
                           if (newValue != _textController.text) return;
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -242,15 +242,6 @@ class _RemotePageState extends State<RemotePage> {
     }
   }
 
-  // handle mobile virtual keyboard
-  void handleSoftKeyboardInput(String newValue) {
-    if (isIOS) {
-      _handleIOSSoftKeyboardInput(newValue);
-    } else {
-      _handleNonIOSSoftKeyboardInput(newValue);
-    }
-  }
-
   void inputChar(String char) {
     if (char == '\n') {
       char = 'VK_RETURN';
@@ -531,12 +522,15 @@ class _RemotePageState extends State<RemotePage> {
                       controller: _textController,
                       // trick way to make backspace work always
                       keyboardType: TextInputType.multiline,
-                      // onChanged: handleSoftKeyboardInput,
                       onChanged: (newValue) async {
-                        await Future.delayed(const Duration(milliseconds: 10));
+                        if (isIOS) {
+                          await Future.delayed(const Duration(milliseconds: 10));
 
-                        if (newValue != _textController.text) return;
-                        return handleSoftKeyboardInput(_textController.text);
+                          if (newValue != _textController.text) return;
+                          return _handleIOSSoftKeyboardInput(_textController.text);
+                        } else {
+                          return _handleNonIOSSoftKeyboardInput(newValue);
+                        }
                       },
                     ),
             ),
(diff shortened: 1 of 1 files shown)


rustdesk:
@fufesou review and test

fufesou:
@shleyZ Hi, which keyboard layout did you test?

image

## Current code

Current master branch, Korean input is ok for me. My test keyboard layout seems to be not good.

https://github.com/user-attachments/assets/b4aa67c0-9aef-4cc0-8a19-c05c6df5ed34


## This PR

But your commit breaks some composition input.

https://github.com/user-attachments/assets/88ff73fd-255a-4c8e-b06b-ecd3233011b2

fufesou:
There may be some problems with my input settings. I'll test more.

https://github.com/user-attachments/assets/771742a6-c0ec-4303-8f5c-1561a0533542


> breaks some composition input

It's my mistake.

I found that after I switched to Korean input method and then switched to Pinyin, composition became impossible.

shleyZ:
@fufesou This is my test video for your reference: [RPReplay_Final1728789493.mov](https://github.com/shleyZ/rustdesk/blob/master/RPReplay_Final1728789493.mov) (The video file is a bit large, you need to download it manually)

Also, please confirm whether you can input Korean normally after removed this logic. Or try to type the Korean word "하" in other input boxes.

fufesou:
> Also, please confirm whether you can input Korean normally after removed this logic.

No.

I have tested the system message app and safari, the letters cannot be composed.

rustdesk:
Please let the OP of the issue test it.

rustdesk:
Let me merge it first. You can test with testflight.

fufesou:
> You can test with testflight.

Ok

flutter/lib/mobile/pages/remote_page.dart
```diff
@@ -531,7 +531,13 @@ class _RemotePageState extends State<RemotePage> {
                       controller: _textController,
                       // trick way to make backspace work always
                       keyboardType: TextInputType.multiline,
-                      onChanged: handleSoftKeyboardInput,
+                      // onChanged: handleSoftKeyboardInput,
+                      onChanged: (newValue) async {
+                        await Future.delayed(const Duration(milliseconds: 10));
```
rustdesk: since only ios has this problem, please enable this for ios only. No need to introduce extra code change for working env.

rustdesk (changes requested):

flutter/lib/mobile/pages/remote_page.dart
rustdesk: Also add comment here, who can remember why modify this after several months?

rustdesk (changes requested):

flutter/lib/mobile/pages/remote_page.dart
shleyZ: Fixed and append some comments.

flutter/lib/mobile/pages/remote_page.dart
```diff
@@ -531,7 +522,18 @@ class _RemotePageState extends State<RemotePage> {
...
+                          // fix: TextFormField onChanged event triggered multiple times when Korean input
+                          // https://github.com/rustdesk/rustdesk/pull/9644
+                          await Future.delayed(const Duration(milliseconds: 10));
+
+                          if (newValue != _textController.text) return;
+                          return _handleIOSSoftKeyboardInput(_textController.text);
+                        } else {
+                          return _handleNonIOSSoftKeyboardInput(newValue);
+                        }
+                      },
```
rustdesk: I think modifying in handleSoftKeyboardInput is better.

rustdesk (changes requested):

flutter/lib/mobile/pages/remote_page.dart
rustdesk: https://github.com/rustdesk/rustdesk/actions/runs/11364668257/job/31611171027 You did not test after change?

shleyZ: @rustdesk Sorry, there is indeed no test, the position of async is wrong, can you revert this merge, I will resubmit the PR.

PR #9653, opened 2024-10-14
fix: wayland, rdp, simualte lock keys

fufesou:
https://github.com/rustdesk/rustdesk/issues/8760

https://github.com/rustdesk/rustdesk/pull/9641

1. The controlling side: send `CapsLock` and `NumLock` to the controlled side if it is Linux.
2. The controlled side:
    (1) `Wayland` rdp input: handle the `CapsLock` and `NumLock` key events and ignore the lock states within the key events.
    (2) X11 or `Wayland` uinput: handle the lock states within the key events and ignore the `CapsLock` and `NumLock` key events.

**Note**: There's also a side effect that the led may be changed after the remote control. The same to https://github.com/rustdesk/rustdesk/pull/9641

## Tests

Windows -> Linux, "Map mode" and "Translate mode"

- [x] X11
- [x] Wayland rdp input
- [x] Wayland uinput

## TODO

1. Remember the lock state before connecting? What if there're multiple controlling end.

### DIFF
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -249,21 +249,35 @@ fn get_keyboard_mode() -> String {
     "legacy".to_string()
 }
 
+#[inline]
+pub fn should_send_lock_key(peer: &str) -> bool {
+    // We need to send `CapsLock` and `NumLock` keys to the controlled side if the peer is Linux.
+    // Because Linux(Wayland, rdp input) may not be able to get the lock key state,
+    // so it cannot handle the lock key state within the key events.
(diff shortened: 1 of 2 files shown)


rustdesk:
This PR is pending, because it is not really better than old. Both have big problem.

hhypnos:
What about implementing a VirtualLockState system that manages lock states independently of physical keyboard LED indicators?

hhypnos:
@fufesou 
"1.Remember the lock state before connecting? What if there're multiple controlling end."
about this: 
image

in each tab to be button of capslock and numlock and with virtual key management it will stay for each session

Issue #9301, opened 2024-09-09
 Incorrect Image Orientation when Remoting into Microsoft Surface or Windows Tablet


deadboy18:
### Bug Description

When using RustDesk version 1.3.0 on Windows 10 to connect to a Microsoft Surface Pro 4 or other Windows tablets with gyroscopic capabilities, the image orientation on the local side adapts to the tablet's orientation rather than maintaining a consistent orientation relative to the remote desktop view.

### How to Reproduce

Start RustDesk version 1.3.0 on Windows 10.
Connect to a Microsoft Surface Pro 4 or any Windows tablet with gyro.
Observe the image orientation on the local screen.
Rotate the tablet and notice how the image orientation on the local screen changes accordingly.

### Expected Behavior

The image orientation on the local screen should remain consistent regardless of the tablet's orientation. Similar behavior to AnyDesk where the image auto-rotates to match the correct orientation for the local viewer.

### Operating system(s) on local side and remote side

Windows 10 22H2 on both end local and remote side

### RustDesk Version(s) on local side and remote side

1.3.0 on both side 

### Screenshots

![image](https://github.com/user-attachments/assets/4248d6f5-1c03-412c-aa30-3ea16135de89)
![image](https://github.com/user-attachments/assets/f357ab51-0425-45c4-9931-ee04af5406f7)
Below is Anydesk
![image](https://github.com/user-attachments/assets/e6a79ae5-2c3b-43e6-a214-d9aaeb3b3c60)

### Additional Context

 This issue has been tested and observed consistently on Microsoft Surface Pro 4 devices. The issue impacts user experience when remote desktop orientation does not match local expectations, unlike behavior observed with other remote desktop solutions such as AnyDesk.
 
 Discord Message Link: https://discord.com/channels/804630702657110016/804630702657110018/1281075454777753664

fufesou:
@deadboy18 Hi, thanks for your feedback.

Only this orientation has this issue?

<img width="162" alt="d73d03f6eb86e20cefb18f9cc2a07b5" src="https://github.com/user-attachments/assets/a48890f0-cd2b-4c2e-855b-a3a7cdaf2b51">

How about the other two "Portait" orientation?

fufesou:
Could you also test if this build works fine?

PR #9696, opened 2024-10-19
fix: workaround physical display rotation

fufesou:
https://github.com/rustdesk/rustdesk/discussions/8322

https://github.com/rustdesk/rustdesk/issues/9301

## Preview

https://github.com/user-attachments/assets/430b2c7a-2597-4718-8f73-9417e836b574

## Desc

The only change is returning error if rotation is not 0 for `vram` encoder. 

image

The capture will fallback `PixelBuffer` to and  the encoder will fallback to HwRamEncoding.

https://github.com/rustdesk/rustdesk/blob/675ffe038164c802233428736af013a1e0e1e844/libs/scrap/src/dxgi/mod.rs#L260

This is a simple workaround.

We cannot just rotate the image in GPU memory. The only way to do it is rotate in encoding. But it need more changes besides `shader`.

### DIFF
--- a/libs/scrap/src/common/vram.rs
+++ b/libs/scrap/src/common/vram.rs
@@ -101,7 +101,12 @@ impl EncoderApi for VRamEncoder {
         frame: EncodeInput,
         ms: i64,
     ) -> ResultType<hbb_common::message_proto::VideoFrame> {
-        let texture = frame.texture()?;
+        let (texture, rotation) = frame.texture()?;
+        if rotation != 0 {
+            // to-do: support rotation
+            // Both the encoder and display(w,h) information need to be changed.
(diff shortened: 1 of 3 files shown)


libs/scrap/src/common/vram.rs
```diff
@@ -101,7 +101,12 @@ impl EncoderApi for VRamEncoder {
         frame: EncodeInput,
         ms: i64,
     ) -> ResultType<hbb_common::message_proto::VideoFrame> {
-        let texture = frame.texture()?;
+        let (texture, rotation) = frame.texture()?;
+        if rotation != 0 {
+            // to-do: support rotation
```
rustdesk, line 106: Put in priority list, but lowest priority.

PR #9859, opened 2024-11-07
Update build.py

thomas-mc-work:
Make the clean up command silent and sane (status code is always `0`).


rustdesk:
It return 1 if no permission.

thomas-mc-work:
What's the point of `|| true` though? It'll let the script continue until it requires to write that file which will result in an error, too. Just later. Isn't the point to fail as early as possible?

Issue #9770, opened 2024-10-29
Audio cracked when streaming from Android to Windows


Ihfu:
### Bug Description

Getting broken audio (no latency though) when using Windows pc to control Android phone. Having tried with 2 pc running Windows 10 (19041 / 19045), and 2 phones running Android (10 / 14). This will happen at Windows .exe program with version higher than 1.2.7, the 1.2.7 and below runs well so seems it's nothing about the Android side. On Android I always run version 1.3.1.

### How to Reproduce

Start a connection from Windows, tried using ID numbers in different networks and using direct IP in a LAN, both get a flawless video stream with broken audio stream.

### Expected Behavior

Audio should not be cracked (like it used to be).

### Operating system(s) on local side and remote side

Android 14 -> Windows 10

### RustDesk Version(s) on local side and remote side

1.3.1 -> 1.3.1

### Screenshots

No need screenshots it's a sound bug

rustdesk:
There is no code change about audio since 1.2.7

21pages:
Thanks for your feedback, https://github.com/rustdesk/rustdesk/pull/8947 caused this, it will be optimized soon.

rustdesk:
Please test the nightly build (windows controlling side), https://github.com/rustdesk/rustdesk/releases/nightly

Ihfu:
> Please test the nightly build (windows controlling side), https://github.com/rustdesk/rustdesk/releases/nightly

Tried the nightly build 1.3.3, it's fixed. Thanks for the work.

zhangbo8418:
> > Please test the nightly build (windows controlling side), https://github.com/rustdesk/rustdesk/releases/nightly
> 
> Tried the nightly build 1.3.3, it's fixed. Thanks for the work.


Didn’t you find that the audio and video were out of sync after a while?

rustdesk:
> Didn’t you find that the audio and video were out of sync after a while?

Your `out of sync after a while` is same to this https://github.com/rustdesk/rustdesk/issues/534?

PR #9893, opened 2024-11-12
fix: windows, improve audio buffer (#9770)

zylthinking:
add some logic to make the sound better in windows

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -118,7 +118,7 @@ pub const SCRAP_X11_REQUIRED: &str = "x11 expected";
 pub const SCRAP_X11_REF_URL: &str = "https://rustdesk.com/docs/en/manual/linux/#x11-required";
 
 #[cfg(not(any(target_os = "android", target_os = "linux")))]
-pub const AUDIO_BUFFER_MS: usize = 150;
+pub const AUDIO_BUFFER_MS: usize = 3000;
 
 #[cfg(feature = "flutter")]
 #[cfg(not(any(target_os = "android", target_os = "ios")))]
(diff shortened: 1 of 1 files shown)


rustdesk:
@21pages test and review

rustdesk:
@zylthinking please add some comment to code to explain some machnism?

src/client.rs
```diff
@@ -904,36 +904,102 @@ pub struct AudioHandler {
...
+            if tms == 0 {
+                tms = dt;
+                return;
+            } else if dt < tms + 12000 {
+                return;
+            }
+            tms = dt;
+        }
+
+        let mut i = (having * 10) / self.1;
```
21pages: Do you want to count whether the buffer is always in a high-capacity state originally? If it is, maybe should put the counter before the 12s period check.

And how about another solution: the underrun bytes length are the real accumelated bytes length, skip it when it reaches a specific threshold.

src/client.rs
```diff
@@ -1117,18 +1183,36 @@ impl AudioHandler {
...
-                let mut lock = audio_buffer.lock().unwrap();
+
                 let mut n = data.len();
-                if lock.occupied_len() < n {
-                    n = lock.occupied_len();
+                let mut lock = audio_buffer.lock().unwrap();
+                let mut having = lock.occupied_len();
+                if having < n {
+                    let tms = info.timestamp();
+                    let how_long = tms.playback.duration_since(&tms.callback).unwrap();
```
21pages: remove possible unwrap

src/client.rs
zylthinking: yes, I mdee this change jsut before the final commit to make the code more beautiful, without considering it carefully

zylthinking: > And how about another solution: the underrun bytes length are the real accumelated bytes length, skip it when it reaches a specific threshold.

It maybe be good in therory, but considering the jitter in audio streams, this can lead to more data insufficiency from the device pulling view and more packet discard at audio pushing site.

src/client.rs
zylthinking: Don't understand.
The unwrap is impossible to lead panic, and the container function does not allow use ? operator.
Uses unwrap seems to be the most clear style

rustdesk: Change to something like `unwrap_or(0)`

PR #9940, opened 2024-11-16
quick fix win ci, replace patch with sed

21pages:
https://github.com/21pages/rustdesk/actions/runs/11871167953/job/33083445009


21pages:
Tried using sed in all jobs before, failed in some jobs.

rustdesk:
Not use /flutter_3.24.4_dropdown_menu_enableFilter.diff any more, why not remove it?

21pages:
Now only windows uses sed, others uses diff file, will try to unify it.

PR #9957, opened 2024-11-18
Change RPM compression to zstd

xlionjuan:
## Summry

Almost all RPM ecosystems are support `rpm` package file that using `zstd` compression and changing default compression method to `zstd`. Some very old enterprise distributions may affected, check *Affect platforms* section.

## Benefit

* Reduce package size, in latest nightly for example, from 25.4 MB to 20.9 MB
* Reduce bandwidth cost

## Requirement

[RPM >= 4.14.0](https://rpm.org/wiki/Releases/4.14.0)

## Affect platforms

* All CentOS non Stream (All EOL)
* Red Hat Enterprise Linux <= 7
* SUSE Linux Enterprise Server 12 SP5  (Maybe, can't verify because they may backport)

## Verify the compression method of the RPM package

Run:

```
bash-5.2# rpm -q --queryformat '%{PAYLOADCOMPRESSOR}\n' -p rustdesk-1.3.3-0.x86_64.rpm 
zstd
```
If not specified while building package, it will uses `gzip`.

## References

[Fedora Proposal: 
Changes/Switch RPMs to zstd compression
](https://fedoraproject.org/wiki/Changes/Switch_RPMs_to_zstd_compression)
[pagure.io/fesco/issue/2144: F31 System-Wide Change: Switch RPMs to zstd compression](https://pagure.io/fesco/issue/2144)

https://endoflife.date/centos (ALL EOL)
https://endoflife.date/rhel

## Note

If merged, refer this PR in release notes, in case someone really need `gzip` compression.


rustdesk:
Sorry, I do not want to change for now.

Issue #9096, opened 2024-08-16
File transfer won't do anything on empty folders


jh961202:
### Bug Description

When using File Transfer, if the transfer target contains one or more *empty* folder, then the empty ones *WON'T* be transferred for no reason, even though the status said that transfer has completed.
If a folder contains one or more files, then the folder could be transferred without problems.

### How to Reproduce

1. Make up some empty folders on one or both computers.
2. Use File Transfer to try to send empty folders to each other.
3. See the error, which is the folder won't be transferred.

### Expected Behavior

The folder should be transferred and appeared in the receiving computer without problems.

### Operating system(s) on local side and remote side

Windows 11 -> Windows 11

### RustDesk Version(s) on local side and remote side

1.2.3-1 ->1.2.3-1

### Screenshots

https://github.com/user-attachments/assets/8bca5c00-3805-4ba9-b767-edd43296be86

rustdesk:
We do not transer empty foler yet.

rustdesk:
@21pages

rustdesk:
Of course

Lumiphare:
It looks like this issue has already been fixed?

PR #9993, opened 2024-11-21
Create empty dir on send files in local

zuiyu1998:


rustdesk:
Could ypu please remove pubspec.lock change?

rustdesk:
@21pages review and test please

rustdesk:
https://github.com/rustdesk/rustdesk/issues/9096

zuiyu1998:
> Could ypu please remove pubspec.lock change?

ok

flutter/lib/models/file_model.dart
```diff
@@ -496,6 +497,31 @@ class FileController {
...
+      }
+    });
+
+    final dirs = paths.map((path) {
+      return PathUtil.getOtherSidePath(directory.value.path, path,
+          options.value.isWindows, toPath, isWindows);
+    });
+
+    for (var dir in dirs) {
+      createDir(dir);
```
21pages: `createDir` creates an empty directory on the operating side, call a `createOppositeDir` function then it'll send the emtpy dir. Since this operation only works from local to remote, check the `isLocal` variable.

21pages (changes requested):

rustdesk:
This PR only do one side? not both sides?

zuiyu1998:
this is local to remote.

rustdesk:
Could you finish both side? It is weird to merge a PR which only finish half. :(

zuiyu1998:
yes

flutter/lib/models/file_model.dart
zuiyu1998: ok

flutter/lib/models/file_model.dart
```diff
@@ -496,6 +501,35 @@ class FileController {
...
+    final List<Entry> entrys = items.items.toList();
+
+    await Future.forEach(entrys, (Entry item) async {
+      final List<String> paths = [];
+
+      final emptyDirs =
+          await fileFetcher.readEmptyDirs(item.path, isLocal, showHidden);
+
+      if (emptyDirs.isEmpty) {
+        paths.add(item.path);
```
21pages: 1. If `emptyDirs` is empty, dothing, otherwise there will be a empty directory whose name is the same as the current item path. After changing this, tests between win&win, win&linux, linux&linux, win&andrioid are good.
2. Add version check for remote to local:
```
if (versionCmp(rootState.target!.ffiModel.pi.version, '1.3.3') < 0) {
      return;
    }
```
3. Remove changes in `pubspec.lock` and code format in `message.proto`.
4. There will be creating folder logs in connection manager, but it's ok.

21pages (changes requested):

flutter/lib/models/file_model.dart
zuiyu1998: the test is include sessionReadRemoteEmptyDirsRecursiveSync?because，It have error in my pc.

21pages: Yes, test with sending a single file.

21pages: > It have error in my pc

What's the error?

zuiyu1998: registerReadEmptyDirsTask throw "Failed to registerReadEmptyDirsTask, already have same read job" in romote to local.

zuiyu1998: throw “Failed to read dir, timeout”

21pages: > throw "Failed to registerReadEmptyDirsTask, already have same read job"

Can you reproduce this? For open directory tasks, this happens often because there are 2 tasks executing at the same time.

> throw “Failed to read dir, timeout”

This happens when the controlled side is not built with this pr, or when network is bad or directory is very large in theory . Change the log to "Failed to read empty dir" to distinguish it from the original one.

zuiyu1998:
please test remote to local on sendfiles。

flutter/lib/models/file_model.dart
zuiyu1998: > This happens when the controlled side is not built with this pr.
可能是这个原因，我会更改报错的原因

21pages:
```
if (emptyDirs.isEmpty) {
        paths.add(item.path);
}
```
Transfer empty directory works without this. So it should be
```
if (emptyDirs.isEmpty) {
       return;
}
```

flutter/lib/models/file_model.dart
```diff
@@ -521,7 +521,11 @@ class FileController {
           await fileFetcher.readEmptyDirs(item.path, isLocal, showHidden);
 
       if (emptyDirs.isEmpty) {
-        paths.add(item.path);
+        if (item.size != 0) {
+          return;
+        } else {
+          paths.add(item.path);
+        }
```
21pages: empty directory return itself, if there is a child empty directory, the parent is not empty.

https://github.com/user-attachments/assets/701bf23b-e880-4580-92a5-d5134219b107

21pages (changes requested):

flutter/lib/models/file_model.dart
zuiyu1998: ok

21pages (approved):

flutter/lib/models/file_model.dart
21pages: > throw "Failed to registerReadEmptyDirsTask, already have same read job"

Can you reproduce this?

libs/hbb_common/protos/message.proto
```diff
@@ -31,8 +31,7 @@ message VideoFrame {
     EncodedVideoFrames h265s = 11;
     EncodedVideoFrames vp8s = 12;
     EncodedVideoFrames av1s = 13;
-  }
-  int32 display = 14;
+  } int32 display = 14;
```
21pages: remove code format in ` message.proto` please

21pages (changes requested):

libs/hbb_common/protos/message.proto
zuiyu1998: ok

21pages: https://github.com/rustdesk/rustdesk/pull/9993/files#diff-26d052b111f4884d9e4345bcc01bab2b566825cadbeb217c4d52ac66748427cc

zuiyu1998: fix

21pages (approved):

PR #10105, opened 2024-11-29
Fix F-Droid build and bump Android NDK to r27c

basilgello:
- **Fix fdroid build**
- **Bump Android NDK to r27c**

### DIFF
--- a/flutter/build_fdroid.sh
+++ b/flutter/build_fdroid.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -x
-
 #
 # Script to build F-Droid release of RustDesk
 #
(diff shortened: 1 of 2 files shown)


flutter/build_fdroid.sh
```diff
@@ -207,14 +232,16 @@ prebuild)
 
 	cargo install \
 		cargo-ndk \
-		--version "${CARGO_NDK_VERSION}"
+		--version "${CARGO_NDK_VERSION}" \
+		--locked
 
 	# Install rust bridge generator
 
 	cargo install cargo-expand
```
linsui, line 240: Lol, you forgot to add `--locked` to this line.

basilgello: You are right, will PR the change after I am back on laptop.

PR #10157, opened 2024-12-03
Update hu.rs

JohnFowler58:
Inserting a new line as line 60 -> "("Key:", "Kulcs:")". In six translations, corrections were made due to correctness of the language. Translation of two lines at the end.

Thanks!


JohnFowler58:
Thank you very much!

PR #9581, opened 2024-10-07
Flatpak manifest updates from the Flathub's PR

besdar:
Hello @rustdesk!

As you know, I recently [passed a code review in the Flathub repository](https://github.com/flathub/flathub/pull/5670). During the review, Flathub's reviewers suggested a few changes to the manifest file and asked that you also add a metainfo file to your repository. I have checked locally that everything seems to work with updated manifes, so I think it would be a good thing to modify your manifest file as well, to ensure consistency between your flatpak configuration and the one from the Flathub. If you decide to merge these changes, the only difference between the configurations would be the `"--talk-name=org.freedesktop.Flatpak"` finish argument, which was removed in Fluthub's version due to security issues.

Please feel free to implement only the changes that you find suitable for your Flatpak build. Thank you for considering this suggestion.

PR #10273, opened 2024-12-13
Flatpak manifest updates

besdar:
Hello @rustdesk,

This is a follow-up pull request to [PR #9581](https://github.com/rustdesk/rustdesk/pull/9581). After investigating the error in the "Flutter Nightly Build" workflow, I discovered that the issue was caused by the line including flatpak's metainfo in the build. To resolve this, I have removed the problematic line, and the GitHub workflow is now functioning correctly. You can verify this by checking the workflow run at https://github.com/besdar/rustdesk/actions/runs/12300041131, where I have disabled all other jobs except those for flatpak and removed the build for the aarch64 architecture (which initially used a custom external environment). You can also check the final build here: https://github.com/besdar/rustdesk/releases/download/nightly/rustdesk-1.3.5-x86_64.flatpak

As a result, the manifest is now aligned with the one in the Flathub repository, except for the removed line and one permission.

In addition to fixing the error, I have also optimized the job's build to make it more efficient and similar to the one recommended by Flathub in their [submission guidelines](https://docs.flathub.org/docs/for-app-authors/submission/#build-and-install). If you feel that these changes are excessive, I can revert the job to its previous state.

I apologize for the delay in submitting this follow-up request. It took some time to research the issue and try different approaches to resolve it. Thank you for considering this updated pull request.


rustdesk:
Doe the generated flatpak work?

besdar:
> Doe the generated flatpak file work?

@rustdesk of course it works. Why would I create a pull request for something that doesn't work? But please note that I've only checked basic scenarios like remote control and only for Fedora 41, so you'll probably need to recheck this build with more advanced test scenarios.

rustdesk:
BTW, we have removed pulseaudio. Is there something similar to this, but pipewire?

image

besdar:
> have removed pulseaudio. Is there something similar to this, but pipewire?

@rustdesk I don't have extensive knowledge on flatpak configuration, but based on my current understanding, it seems that this permission can be replaced with `--filesystem=xdg-run/pipewire-0`, as suggested in [this GitHub issue comment](https://github.com/flatpak/flatpak/issues/5130#issuecomment-1272348665). Alternatively, it's also possible to completely remove this permission, as described in [this pull request](https://github.com/flathub-infra/website/pull/3731) (i.e. this permission is included in the default ones).

PR #10288, opened 2024-12-15
Flathub badge in the README

besdar:
Hello @rustdesk,

Your Flathub package is now verified and the builds are consistent. I think it would be beneficial to add a Flathub badge to your README file to reflect this. You can find the badge source at [Flathub Badges](https://flathub.org/badges).

This addition would provide a clear visual indicator of your package's status and make it easier for users to verify its authenticity.

I have also checked the [source of the F-Droid badge](https://gitlab.com/fdroid/artwork/tree/master/badge) and updated it too.


xlionjuan:
Consider SVG? https://flathub.org/api/badge?svg&locale=en
Because image one is only 240x80, which is very blurry.

PR #10326, opened 2024-12-20
Revert "Revert "revert linux use cpal ""

21pages:
Fix https://github.com/rustdesk/rustdesk/discussions/10197


rustdesk:
though pulseaudio is not ready on new Linux, but libpulse is still availalbe, and pipewire provide pipewire-pulse bridge, the simple pulse crate still works (cpal on linux sucks).

PR #10293, opened 2024-12-16
refact: linux, chcon, bin_t

fufesou:
Fix SELinux limits on Fedora.

https://github.com/rustdesk/rustdesk/issues/6116
https://github.com/rustdesk/rustdesk/issues/4267
https://github.com/rustdesk/rustdesk/issues/7107
https://github.com/rustdesk/rustdesk/issues/5458
https://github.com/rustdesk/rustdesk/issues/5098
https://github.com/rustdesk/rustdesk/issues/7164
https://github.com/rustdesk/rustdesk/issues/6301

## Desc

Change the security context of `/usr/lib/rustdesk/rustdesk` from `lib_t` to `bin_t`.

Then the process security context will be `system_u:system_r:unconfined_service_t:s0`.

Although the service spec uses `/usr/bin/rustdesk` and the security context of the soft symlink is `bin_t`. The process actually uses the security context of `/usr/lib/rustdesk/rustdesk`.

All distros can apply the following changes. But we only do it for `rpm` packages for now.
Because the other distros dose not support `chcon` well (or maybe SELinux) for now.

```bash
# Change the security context of /usr/lib/rustdesk/rustdesk from `lib_t` to `bin_t`.
if command -v getenforce >/dev/null 2>&1; then
  if [ "$(getenforce)" == "Enforcing" ]; then
    file_security_context=$(ls -lZ /usr/lib/rustdesk/rustdesk 2>/dev/null | awk -F':' '{print $3}')
    if [ "${file_security_context}" == "lib_t" ]; then
      chcon -t bin_t /usr/lib/rustdesk/rustdesk || true
    fi
  fi
fi
```

The above code can also be applied to Ubuntu, openSUSE, Archlinux. But it takes no effect.
1. SELinux is not enabled by default, even command `getenforce` is not included.
2. `ls -lZ /usr/lib/rustdesk/rustdesk` results no security context
  `-rwxr-xr-x 1 root root ? 19184 Sep 20 14:10 /usr/lib/rustdesk//rustdesk`
3. `chcon -t bin_t /usr/lib/rustdesk/rustdesk` results errors.
  `chcon: can't apply partial context to unlabeled file '/usr/lib/rustdesk/rustdesk'`
  `chcon: failed to get security context of '/usr/lib/rustdesk/rustdesk': Operation not supported`

This documentation may no longer be needed https://rustdesk.com/docs/en/client/linux/selinux/.

## Tests

[...]

Issue #7164, opened 2024-02-17
Massive SELinux errors on Fedora-39


tuxmaster5000:
### Bug Description

After installing it the the system will show one SELinux error after the next. :(

### How to Reproduce

Simple only install it.

### Expected Behavior

Working app without massive SELinux alert messages.

### Operating system(s) on local side and remote side

Fedora-39

### RustDesk Version(s) on local side and remote side

1.2.3

### Screenshots

No

### Additional Context

SELinux alert log:
cat /var/log/audit/audit.log|grep rust > /tmp/alert.txt
[alert.txt](https://github.com/rustdesk/rustdesk/files/14319503/alert.txt)

rustdesk:
https://github.com/rustdesk/rustdesk/issues/4267

Issue #7107, opened 2024-02-12
unable to control my linux wayland machine from a windows machine


back-lacking:
### Bug Description

what the tin says
when i connect to my linux machine from my windows machine, i am unable to control my linux desktop. on the linux side, i have given the windows machine permission to use mouse and keyboard. when i try to go control from windows machine, nothing happens, no keyboard shortcuts go through, typing in any letters into my terminal doesnt work, clicking on any open windows doesnt work. 
Also tried installing from chocolatey (1.2.2 installer) and scoop (1.2.3 portable)
Windows client info
- Win10 22H2, Build 19045.3930
- Rustdesk 1.2.3, installer, installed using winget
Linux client info
- Fedora 38
- Gnome 44.8, Wayland
- Rustdesk 1.2.3 (RPM package from github releases)

### How to Reproduce

1. install up rustdesk on linux and windows clients
2. connect to linux (with wayland) from windows client
3. on linux machine, accept connection, and tick on "allow keyboard and mouse"
4. try to do things to linux machine from windows machine 

### Expected Behavior

be able to use my keyboard input on linux machine from windows machine. 

### Operating system(s) on local side and remote side

Windows 10 -> Fedora 38

### RustDesk Version(s) on local side and remote side

1.2.3 (from winget) -> 1.2.3 (RPM package)

### Screenshots

https://github.com/rustdesk/rustdesk/assets/86173810/5a538d58-55bd-4e37-bc08-833287e5367a

rustdesk:
tracked by https://github.com/rustdesk/rustdesk/issues/3565

back-lacking:
> Try out nightly build, and make sure selinux is set correctly.
> 
> https://github.com/rustdesk/rustdesk/releases/tag/nightly https://rustdesk.com/docs/en/client/linux/selinux/

so, first on 1.2.3, i did the selinux guide on the docs. 
im now swapped over to the nightly build, just went to the nightly build section in releases, grabbed the rpm, and then installed it. 
the issue is still persistent. 
![image](https://github.com/rustdesk/rustdesk/assets/86173810/e1acd315-addb-42b5-b634-398e5a62049d)
this is the rust desk im currently using, do i need to go reapply the rule since i installed the se linux rules and then installed a new version of rustdesk?

rustdesk:
No idea, our fedora 38 works well.

back-lacking:
> No idea, our fedora 38 works well.
> 
[...]

Issue #6301, opened 2023-11-06
Rustdesk SELinux Log spam / sudo access


BentHaase:
### Bug Description

Rustdesk spams systems logs (about every 1s) with denied access to `sudo` due to SELinux.

```bash
~ # journalctl --boot 0 --follow
[...]
Nov 06 06:15:29 fedora setroubleshoot[23731]: SELinux is preventing rustdesk from execute access on the file /usr/bin/sudo. For complete SELinux messages run: sealert -l aad61d44-49b7-415b-95d3-aaa7dfa04e9f
Nov 06 06:15:29 fedora setroubleshoot[23731]: SELinux is preventing rustdesk from execute access on the file /usr/bin/sudo.
                                              
                                              *****  Plugin catchall (100. confidence) suggests   **************************
                                              
                                              If you believe that rustdesk should be allowed execute access on the sudo file by default.
                                              Then you should report this as a bug.
                                              You can generate a local policy module to allow this access.
                                              Do
                                              allow this access for now by executing:
                                              # ausearch -c 'rustdesk' --raw | audit2allow -M my-rustdesk
                                              # semodule -X 300 -i my-rustdesk.pp

```

### How to Reproduce

Install Rustdesk on Fedora 38 (soon 39), run Rustdesk.

Show logs using: `journalctl --boot 0 --follow`

### Expected Behavior

Rustdesk should not constantly try to gain `sudo` access when doing nothing (no connection in/outbound) unless user requested it to do so. Giving Rustdesk (client) total system access as [shown here](https://rustdesk.com/docs/en/client/linux/selinux/) via custom policy is not the correct solution. 

### Operating system(s) on local side and remote side

Fedora 38

### RustDesk Version(s) on local side and remote side

1.2.2

### Screenshots

See log above.

### Additional Context

Why exactly is Rustdesk trying to gain access to `sudo` for when doing nothing (Can someone point me to the code snippet)?

PR #10327, opened 2024-12-20
refact: linux, move rustdesk into /usr/share

fufesou:
https://github.com/rustdesk/rustdesk/pull/10293#issuecomment-2548609489

Fix SELinux limits on Fedora.

https://github.com/rustdesk/rustdesk/issues/6116
https://github.com/rustdesk/rustdesk/issues/4267
https://github.com/rustdesk/rustdesk/issues/7107
https://github.com/rustdesk/rustdesk/issues/5458
https://github.com/rustdesk/rustdesk/issues/5098
https://github.com/rustdesk/rustdesk/issues/7164
https://github.com/rustdesk/rustdesk/issues/6301

## Desc

Change the install path from `/usr/lib/rustdesk` to `/usr/share/rustdesk`.

```
$ ls -lZ /usr/share/rustdesk/rustdesk
-rwxr-xr-x. 1 root root system_u:object_r:usr_t:s0 30608 Dec 20 21:49 /usr/share/rustdesk/rustdesk

$ ps -eZ | grep rustdesk
system_u:system_r:unconfined_service_t:s0 964 ?  00:00:04 rustdesk
```

## Tests

- [x] Flutter build. Install and upgrade on Ubuntu, Archlinux, OpenSUSE, Fedora.
- [x] Sciter build. Install and upgrade on Ubuntu.
- [x] AppImage on Ubuntu.
- [ ] Flatpak, flutter and sciter. The flutter version is tested. Sciter version is not tested, it's a little hard to build this package. But the only file need to be changed is `src/ui.rs`.
- [ ] Rpm package directly downgrade is broken. Since the older spec file uses `ln -s /usr/lib/rustdesk/rustdesk /usr/bin/rustdesk`. This command always fails. Previous versions worked because the location was consistent. No fix idea, the only way is to uninstall before installing the older version.

### DIFF
--- a/.github/workflows/flutter-build.yml
+++ b/.github/workflows/flutter-build.yml
@@ -719,7 +719,7 @@ jobs:
         shell: bash
         run: |
           cd "$(dirname "$(which flutter)")"
-          # https://github.com/flutter/flutter/issues/1.3.53
+          # https://github.com/flutter/flutter/issues/133533
           sed -i -e 's/_setFramesEnabledState(false);/\/\/_setFramesEnabledState(false);/g' ../packages/flutter/lib/src/scheduler/binding.dart
           grep -n '_setFramesEnabledState(false);' ../packages/flutter/lib/src/scheduler/binding.dart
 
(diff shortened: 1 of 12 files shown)


xlionjuan:
Solved, thanks @travier 's advise!

PR #10332, opened 2024-12-21
refactor: replace &PathBuf with &Path to enhance generality

Integral-Tech:
- `&Path` provides better generality than `&PathBuf` in function parameters.


rustdesk:
You should also remove unused "use pathbuf".

Integral-Tech:
@rustdesk Done :)

Issue #10005, opened 2024-11-21
Online status indicator doesn't work for remotes added via `id@server/r`


sepsol:
### Bug Description

I have a remote client that I usually connect to. The remote client is using a relay server and an ID server on its end. When I want to connect to that specific remote via its pure `<id>`, I have to change the networking settings on my client to add that relay and ID server globally. When I do that, everything works just fine and the online status indicator for that remote updates correctly.

Since the adding and deleting of relay and ID servers on my client for connecting to that remote client was becoming cumbersome, I researched this repository and found out that [an alternative method has been added in v1.3.0](https://github.com/rustdesk/rustdesk/issues/6198#issuecomment-1794694273) that allows connecting to remotes behind an ID and a relay server without having to change these settings globally.

Therefore, I gave it a shot and am now able to connect to my remote client via `<id>@<server>/r` without having to set the relay and ID servers globally on my client. Although this works great, the issue I'm experiencing with this method is that it no longer updates the small online status indicator for that remote correctly and it always stays orange, even though I'm able to connect to it just fine and if I use the old method it is able to detect when the remote is online and shows it as green.

### How to Reproduce

1. Install RustDesk on 2 machines (I tested on 2 Windows machines).
2. Open up the settings on machine 1 and go to networking and assign it to a custom (public) ID and relay server.
3. Now try connecting to machine 1 from machine 2 via `<id>@<server>/r` without changing the global networking settings on machine 2.
4. You should be able to connect to machine 1 but the online status indicator won't work and it'll stay orange.

### Expected Behavior

The online status indicator should turn green when the remote behind the ID and relay server is ready to receive connections.

### Operating system(s) on local (controlling) side and remote (controlled) side

Windows 11 -> Windows 11

### RustDesk Version(s) on local (controlling) side and remote (controlled) side

1.3.0 -> 1.3.0

### Screenshots

![image](https://github.com/user-attachments/assets/df2d3012-f879-437f-9df4-54990b919908)

![image](https://github.com/user-attachments/assets/bd7f761c-5576-44f9-8272-9809b81d6017)

rustdesk:
No plan for this.

rustdesk:
PR is welcome

Issue #498, opened 2022-05-10
Does Rustdesk support modifying the remote desktop resolution?


lllAlex2333:
I'm using  rustdesk, but the resolution issue is bothering me. The window is too small now, and when I choose to stretch, the content becomes blurry.

rustdesk:
https://github.com/rustdesk/rustdesk/issues/1053

rustdesk:
We are screen capturing, not creating new session as what mstsc does. But in some special case (e.g. no physical display), we can create a virtual display with some resolution, but still impossible as flexible as mstsc. If you insist it, please our RDP sharing feature https://github.com/rustdesk/rustdesk/issues/914#issuecomment-1173839858.

jwgn:
Ok so we are unable to fully control clients which have a higher screen resolution. Workaround: ask the client to lower the resolution below ours, then restart the service. Works, yea. Pretty messy solution, though.

Products like TeamViewer are able to scale the resolution to fit the operator's screen. Why is this not possible here?

rustdesk:
goto  toolbar, there is shrink/scale button.

PR #10408, opened 2024-12-31
fix online status indicator

Lumiphare:
#10005 

The original online status query simply checked whether the rendezvous_server contained the given ID. After the modification, the server address is obtained from the original set of IDs, and then a socket connection is established with that server in order to query the online status.

By using a HashMap, the same server address only needs one socket connection to send a batch of IDs at once. In the HashMap’s key, “rendezvous_server” is used to store cases where there is no “@” in the ID (i.e., the original situation where IDs only have an ID). “public” is used when the string after “@” is “public”. In other scenarios, the string after “@” is treated as a domain name or IP address (which can be IPv4 or IPv6, with optional ports; if a port is provided with an IPv6 address, it should be enclosed in brackets; no brackets are needed if there is no port). The value in the HashMap contains all IDs that need to be queried on the server corresponding to that key.

Originally, config.rs’s get_rendezvous_server function simply checked for the presence of “:” to determine if there was a port, ignoring IPv6. Therefore, IPv6 detection has been added by counting the number of “:” to handle IPv6 addresses.

I’ve tested the scenario between two Windows 10 machines myself, covering ID, ID@public, ID@IPv4 , ID@IPv4 :port, ID@IPv6, and ID@[IPv6]:port. The “online indicator” lights up as expected in all cases, but turning it off responds a bit slowly

PixPin_2025-01-01_06-54-39

### DIFF
--- a/libs/hbb_common/src/config.rs
+++ b/libs/hbb_common/src/config.rs
@@ -736,10 +736,8 @@ impl Config {
                 .next()
                 .unwrap_or_default();
         }
-        if !rendezvous_server.contains(':') {
-            rendezvous_server = format!("{rendezvous_server}:{RENDEZVOUS_PORT}");
-        }
-        rendezvous_server
+
+        crate::socket_client::check_port(rendezvous_server, RENDEZVOUS_PORT)
     }
 
     pub fn get_rendezvous_servers() -> Vec<String> {
--- a/src/client.rs
+++ b/src/client.rs
@@ -3607,7 +3607,11 @@ pub mod peer_online {
             let tmp = hbb_common::socket_client::check_port(server_addr.clone(), hbb_common::config::RENDEZVOUS_PORT);
             if let Some(pos) = tmp.rfind(':') {
                 host = tmp[.. pos].to_string();
-                port = tmp[pos + 1 ..].parse().unwrap_or(0);
+                port = if pos + 1 < tmp.len() {
+                    tmp[pos + 1..].parse().unwrap_or(0)
+                } else {
+                    0
+                };
             } else {
                 help_query_error(ids).await;
                 bail!("Invalid server address: {server_addr}");
--- a/src/client.rs
+++ b/src/client.rs
@@ -3485,16 +3485,15 @@ async fn hc_connection_(
 }
 
 pub mod peer_online {
+    use std::sync::{Arc, atomic::{AtomicBool, Ordering}};
+
     use hbb_common::{
-        anyhow::bail,
-        config::{Config, CONNECT_TIMEOUT, READ_TIMEOUT},
-        log,
-        rendezvous_proto::*,
-        sleep,
-        socket_client::connect_tcp,
-        tcp::FramedStream,
-        ResultType,
+        anyhow::bail, config::{Config, CONNECT_TIMEOUT, READ_TIMEOUT}, log, rendezvous_proto::*, sleep, socket_client::connect_tcp, tcp::FramedStream, ResultType, tokio
     };
+
+    use std::collections::{
+        HashMap, HashSet,
+    }; 
 
     pub async fn query_online_states<F: FnOnce(Vec<String>, Vec<String>)>(ids: Vec<String>, f: F) {
         let test = false;
@@ -3505,7 +3504,7 @@ pub mod peer_online {
             f(onlines, offlines)
         } else {
             let query_timeout = std::time::Duration::from_millis(3_000);
-            match query_online_states_(&ids, query_timeout).await {
+            match query_online_states_with_server_addr(&ids, query_timeout).await {
                 Ok((onlines, offlines)) => {
                     f(onlines, offlines);
                 }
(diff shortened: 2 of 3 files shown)


libs/hbb_common/src/config.rs
```diff
@@ -736,10 +736,17 @@ impl Config {
...
                 .unwrap_or_default();
         }
-        if !rendezvous_server.contains(':') {
-            rendezvous_server = format!("{rendezvous_server}:{RENDEZVOUS_PORT}");
+        match rendezvous_server.matches(':').count() {
+            0 => format!("{rendezvous_server}:{RENDEZVOUS_PORT}"),
+            1 => rendezvous_server,
+            _ => {
+                if rendezvous_server.starts_with('[') {
+                    format!("{rendezvous_server}:{RENDEZVOUS_PORT}")
```
rustdesk: Starting with `[` means it has ending port, why append again?

rustdesk (changes requested):

libs/hbb_common/src/config.rs
Lumiphare: Sorry, that was my oversight. I've replaced it with the already implemented function as an alternative.

src/client.rs
```diff
@@ -3516,82 +3516,121 @@ pub mod peer_online {
...
-        let tmp: Vec<&str> = rendezvous_server.split(":").collect();
-        if tmp.len() != 2 {
-            bail!("Invalid server address: {}", rendezvous_server);
-        }
-        let port: u16 = tmp[1].parse()?;
-        if port == 0 {
+    async fn create_online_stream(server_addr: String) -> ResultType<FramedStream> {
+        let (rendezvous_server, _servers, _contained) = match server_addr.as_str() {
+            "public" => {
+                (format!("{}:{}",hbb_common::config::RENDEZVOUS_SERVERS[0],hbb_common::config::RENDEZVOUS_PORT), vec![], false)
```
rustdesk: format?

src/client.rs
```diff
@@ -3516,82 +3516,121 @@ pub mod peer_online {
...
+                crate::get_rendezvous_server(READ_TIMEOUT).await
+            },
+            _ => {
+                (hbb_common::socket_client::check_port(server_addr, hbb_common::config::RENDEZVOUS_PORT), vec![], false)
+            }
+        };
+        let port;
+        let online_server;
+        if let Some(pos) = rendezvous_server.rfind(':') {
+            port = rendezvous_server[pos + 1 ..].parse().unwrap_or(0);
```
rustdesk: How about if rendezvous_server ends with ':'? What will happen to [pos + 1 ..]?

src/client.rs
```diff
@@ -3516,82 +3516,121 @@ pub mod peer_online {
...
+            online_server = format!("{}:{}", rendezvous_server[.. pos].to_owned(), port - 1);    
+        } else {
             bail!("Invalid server address: {}", rendezvous_server);
-        }
-        let online_server = format!("{}:{}", tmp[0], port - 1);
+        }    
         connect_tcp(online_server, CONNECT_TIMEOUT).await
     }
 
     async fn query_online_states_(
```
rustdesk: I would suggest you create another function 
```
query_online_states_with_ip_port(...) 
```
to wrap old query_online_states_, so that we have minimum code change.

rustdesk (changes requested):

Lumiphare:
I found that this issue isn’t that easy to solve properly. The previous implementation was too rough, so I spent a bit longer rewriting it. There may still be some issues.


I moved the part I implemented myself into a new function. Previously it was called `ip_port,` which was not comprehensive because it didn’t include `public` or domain names, so I changed it to `server_addr.` Most of the logic has been moved to the new `query_online_states_with_server_addr` function. I changed the function to a concurrent + polling mode. `query_online_states_with_server_addr` primarily uses the current record `ONLINE_IDS` to return user status, while user status updates are handled by each polling task. Therefore, based on the incoming  `ids`, the function also creates polling tasks. It uses `RUNNING_POLLING` to keep track of each async task, whose value is an `AtomicBool` used to signal the task to stop. Currently, the polling interval is set to 5 seconds. `FIRST_CALL` is mainly for the first load so that a portion of the user status can be displayed quickly. Since the `CONNECT_TIMEOUT` is set to 18,000 ms, whether a user goes from online to offline or offline to online, there might be situations where an offline state needs a long query time. Hence, user status changes won’t be reflected immediately.


Besides the changes I made in the new function, in order to implement concurrency, I modified the original ``#[tokio::main(flavor = "current_thread")]`` in `start_flutter_async_runner_` to **#[tokio::main(flavor = "multi_thread")]** and also updated `check_port` in `config.rs`.


For the issue at `pos + 1,` previously I considered that each `server_addr` call would run through the `check_port` function, which automatically appends a port, so it’s unlikely to end with `:`. But to be safe, I still added a check.


There may still be some parts of the code that need further changes. I’ll go ahead and submit this PR so we can revise it afterward.

Lumiphare:
I've now switched to a persistent connection mode. Each time, I first attempt to retrieve a `FramedStream` from `ONLINE_SOCKETS`; if it isn't found, a new socket is created. The original code was unable to establish a persistent connection because on the server side, after calling `next_timeout()`, the socket was immediately dropped—no loop was used, so each socket could only be used once.  I've changed it to a loop-based approach. This is the PR I submitted on the server side ([Solve the issue of reusing the client's socket for multiple send operations. #498](https://github.com/rustdesk/rustdesk-server/pull/498))

@rustdesk Are there any other parts that need to be modified?

PR #10541, opened 2025-01-19
fix(debian): makes postinst/prerm scripts idempotent

HorlogeSkynet:


rustdesk:
We do not agree with -f

HorlogeSkynet:
> We do not agree with `-f`

You want the symlink to be created during install, `-f` prevents `ln` from failing if target already exists.
You want your symlink to be removed during uninstall, `-f` prevents `rm` from failing if target is already missing.

These are usual ways of getting "idempotent" and bullet-proof packaging scripts.

rustdesk:
We do not want to change anything if there is no bug.

PR #10614, opened 2025-01-28
misused by bad guys

rustdesk:


tutsde:
was it a security hole ?
what does this changes do ?

PR #10677, opened 2025-02-05
refact: update path to git url about hbb_common

zuiyu1998:
hbb_common is not in local.update path to git url about hbb_common in Cargo.toml.


rustdesk:
We have not decided yet.

zuiyu1998:
lib/hbb_common is empty in master.this is ok?

zuiyu1998:
use git submodule update --init

PR #10796, opened 2025-02-15
Italian language update

bovirus:
@rustdesk 

Please check and merge. Thanks.

Note: it seems that there is a serious issue regarding use RD in pc with double video board (ex. Intel + Nvidia) because in this context the main windows of RD is always light gray and you cannot setup nothing. I didn't find any open ticket about it (I tested on several computer with smilar hardware with two video boards).


rustdesk:
> main windows of RD is always light gray and you cannot setup nothing

We can do nothing except waiting for google flutter update, you'd better run some simple flutter app to see if can report to flutter.

bovirus:
@rustdesk 

Then can you confirm that you get report about this?
The main problem is a new installation with this issue is to setup a default password for remote access (to avoid manual confirmation).
Is there a way to setup via command line or other way a default password for remote access?
Thanks.

PR #10813, opened 2025-02-17
NL lang further improvements

ANB5Dev:
enhanced a few more unclear texts, corrected one change made in a27fa43 (translated literally: Windows service -> Background service)


ANB5Dev:
note that it needs merging with #10812 - since (AFAIK) there's no guide to where these strings are used, I only translated them as soon as I encounter them in their natural habitat

PR #10941, opened 2025-02-27
Sardinian translation

asereze:
Sardinian ("Sardu", ISO 639-1 code: "sc") translation.


rustdesk:
Thank you very much, you also need to modify https://github.com/rustdesk/rustdesk/blob/master/src/lang.rs

asereze:
Done

PR #11029, opened 2025-03-06
Update reference to flutter web UI

manquer:
- the src js web UI has moved to directory v1 in flutter/web/js as v2 is under development . The link in README.md is returning 404 currently and needs update


annpeter:
when will web v2 release?

PR #11068, opened 2025-03-09
Create ta.rs

naveendev97:
Tamil தமிழ் language


rustdesk:
Thank you, but you need also to modify https://github.com/rustdesk/rustdesk/blob/master/src/lang.rs

PR #11201, opened 2025-03-20
Win input privacy mode refactor

tidely:
The motivation behind this PR is to make more of the implicit logic explicit. In this case it means instead of treating 0 as a special case for the current state of the thread id. We move it to an `Option`. Additionally the windows api expects thread id's to never be zero. So we wrap it in `NonZero`.

Overall this makes no functional changes, but more explicitly defines the intent of the code.

The core change is:

```rust
static ref CUR_HOOK_THREAD_ID: Mutex<DWORD> = Mutex::new(0);
```

```rust
static ref CUR_HOOK_THREAD_ID: Mutex<Option<NonZero<DWORD>>> = Mutex::new(None);
```

### DIFF
--- a/src/privacy_mode/win_input.rs
+++ b/src/privacy_mode/win_input.rs
@@ -29,7 +29,7 @@ lazy_static::lazy_static! {
 
 /// Hook keyboard and mouse to the current thread
 fn do_hook(tx: Sender<String>) -> ResultType<(HHOOK, HHOOK)> {
-    let invalid_ret = (0 as HHOOK, 0 as HHOOK);
+    let invalid_ret: (HHOOK, HHOOK) = (0, 0);
 
     let mut cur_hook_thread_id = CUR_HOOK_THREAD_ID.lock().unwrap();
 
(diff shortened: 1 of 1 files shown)


src/privacy_mode/win_input.rs
```diff
@@ -22,14 +23,18 @@ const GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS: u32 = 4;
 const WM_USER_EXIT_HOOK: u32 = WM_USER + 1;
 
 lazy_static::lazy_static! {
-    static ref CUR_HOOK_THREAD_ID: Mutex<DWORD> = Mutex::new(0);
+    /// thread id of the currently attached to thread (non-zero)
+    static ref CUR_HOOK_THREAD_ID: Mutex<Option<NonZero<DWORD>>> = Mutex::new(None);
 }
 
+/// Hook keyboard and mouse to the current thread
 fn do_hook(tx: Sender<String>) -> ResultType<(HHOOK, HHOOK)> {
```
tidely, line 31: We validate the HHOOK pointers to not be null, so we could make this return `NonNull<HHOOK>`. For example we currently recheck that the keyboard hook isn't null for a second time in `fn hook()` even though we have already validated it.

PR #11519, opened 2025-04-21
fix: add missed translation

fufesou:


rustdesk:
put in https://github.com/rustdesk/rustdesk/pull/11476

Issue #1046, opened 2022-07-25
Languages: Translation platform


tiagogaspar8:
Hi,

Can I ask for the creation of a translation platform? Something like Weblate? It would be great to have users without git knowledge to contribute.

Thanks.

itamarjp:
the portuguese brazilian translation is very poor, I think M$ AI or google translator can translate better

rustdesk:
> poor

It is a manual translation. :)

tiagogaspar8:
Like, I have git knowledge, and I'd like to add Portuguese for Portugal, but adding a platform for it would be better.

rustdesk:
Yes

tiagogaspar8:
Were there any plans? Is there any way we can proceed with this?

rustdesk:
I do not know weblate, but you just see it supports json. I can give you template.json, and you give me pt.json.

tiagogaspar8:
Weblate would be deployed (probably) to rust.com server and would allow users to participate in the translation process easily and clearly with a web interface.
I can investigate how one would deploy and integrate weblate to rustdesk but I need some confirmation that you guys are on board with me to actually do it. 😄

rustdesk:
no problem.

rustdesk:
csv format please. I just write a csv <-> rs translator.

rustdesk:
[ptbr.csv](https://github.com/rustdesk/rustdesk/files/9183327/ptbr.csv)

rustdesk:
It is better to integrate to the github repo.

WofWca:
Weblate is very well integrated with git, translations are automatically committed to a remote repo and all you have to do is merge them. I'd say it's a must-have.
And you don't actually need to self-host a Weblate server, they provide hosting for FOSS software https://hosted.weblate.org/hosting/

bigretromike:
I did update for Polish translation not so long ago. Yes doing this via text editor is not the best, but to be honest Its better to have more features like address book thank easy to use 1-2-3-times-use platform for translation.

Coool:
@yurtpage, what is the current progress? What still needs to be implemented?

Coool:
@yurtpage, convert them to CSV as @rustdesk suggested in comment[^convert_comment] and then import them in your project. I can help you if you give me some more privileges on your project. 

[^comment1]: [Convert](https://github.com/rustdesk/rustdesk/issues/1046#issuecomment-1194360218) to CSV

[...]

PR #11521, opened 2025-04-21
Docs: migrate to ISO lang codes

yurtpage:
We can add the `/docs/*.md` files to be translatable on Weblate, but their language code is in upper case and Weblate can't recognize them. So I renamed them e.g. `/docs/README-PTBR.md` to `/docs/pt-BR/README.md`.
Also this makes navigation and maintain easier.

See https://github.com/rustdesk/rustdesk/issues/1046#issuecomment-2817132785

As a downside old links may be broken.


rustdesk:
Thank you very much, but I do not like this structure. very sorry.

Coool:
@rustdesk, @yurtpage looks like somebody is not happy with neighbor, who makes good suggestion how to live their life. Get over your childish perception of right and wrong - I don't like it. Swallow your internal ego and make it better for all of us. @yurtpage made great suggestion. Just merge it.

PR #11532, opened 2025-04-22
fix build

fufesou:
1. Fix build msi.
2. Add some comments.
3. Remove unused items.

`Package.ru_ru.wxl` is not actually used. Because the build outputs are `en-us/Package.msi` and `ru-ru/Package.msi`. We only publish the `en-us/Package.msi` in the CI.


rustdesk:
reverted the PR

PR #11562, opened 2025-04-24
feat: removed install requirement for setting password

rk-esselink:


rk-esselink:
Sorry, this PR was to the wrong repo

PR #11591, opened 2025-04-28
feat: take screenshot

fufesou:
## Preview

https://github.com/user-attachments/assets/ea4d3bca-7660-4fe0-b34a-5ef0783c34dc

https://github.com/user-attachments/assets/f0e93fd0-a478-4411-9bcd-0c0944a9e5bb

https://github.com/user-attachments/assets/2fc96762-d269-445c-8121-a8b9ef063965

## Note

1. Taking screenshot is very slow, because encoding png is slow.
2. Windows, Linux, MacOS are tested. "Save as...", "Copy to clipboard", "Cancel".
3. Default png filename in sciter version does not contain the timestamp, because `Date` in sciter is different to js. I did not find a way to get the timestamp.
4. The sciter version does not disable the action "Take screenshot" for now. Because we can't cancel a timer in sciter, `self.timer()` always return `undefined` in sciter.


rustdesk:
@21pages review please, especially about vram set not use etc

21pages (approved):

PR #11708, opened 2025-05-11
refact: mouse scroll, remote tabs

fufesou:
https://github.com/rustdesk/rustdesk/discussions/11705

## Preview

https://github.com/user-attachments/assets/906dd021-6029-4fff-b2df-8e54b516a164

https://github.com/user-attachments/assets/fc3d1ffd-d841-41cc-b54c-f7bbea4783db

With `2.5`x rate applied, when peer ID is 10 digits, one mouse scroll distance is slightly larger than one tab.

## Note

A better way to switch between the tabs is using the dropdown list.


chmichael:
2.5x is still slow! Please increase it more or let us customize it. In Firefox/Chrome for example it's faster

PR #11718, opened 2025-05-11
Update hu.rs

JohnFowler58:
Translation of the new character strings into Hungarian.


JohnFowler58 (commented):
Translation of the new character strings into Hungarian.
I replaced bad characters with good characters.

PR #11727, opened 2025-05-11
fix: typos in src/ and subdirectories

luzpaz:
Found via codespell

### DIFF
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -902,7 +902,7 @@ fn _map_keyboard_mode(_peer: &str, event: &Event, mut key_event: KeyEvent) -> Op
     let keycode = match _peer {
         OS_LOWER_WINDOWS => {
             // https://github.com/rustdesk/rustdesk/issues/1371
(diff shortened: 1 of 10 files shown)


src/keyboard.rs
```diff
@@ -902,7 +902,7 @@ fn _map_keyboard_mode(_peer: &str, event: &Event, mut key_event: KeyEvent) -> Op
     let keycode = match _peer {
         OS_LOWER_WINDOWS => {
             // https://github.com/rustdesk/rustdesk/issues/1371
-            // Filter scancodes that are greater than 255 and the hight word is not 0xE0.
+            // Filter scancodes that are greater than 255 and the height word is not 0xE0.
```
luzpaz, line 905: This looks weird, please advise

rustdesk: both hight and height seem wrong.

rustdesk: It seems `high`.

luzpaz: `and the high word is not...` ?

PR #11914, opened 2025-05-29
Fix GUI closing immediately when enabling service (#7010)

GlassOnTin:
## Summary
This PR fixes the issue where the RustDesk GUI closes immediately (~500ms) when attempting to enable the service on Windows.

## Problem
When enabling the service, the IPC connection fails with "Access is denied" errors, causing the GUI to quit prematurely before the service has time to start properly.

## Solution
Modified the IPC connection logic in `ui_interface.rs` to:
- Add retry attempts (up to 10) before giving up on the IPC connection
- Detect when the service is being installed or started using `installing_service()` flag
- Prevent the GUI from quitting during service initialization
- Implement exponential backoff for connection retries

## Testing
To test this fix:
1. Start RustDesk without the service enabled
2. Attempt to enable the service from the settings
3. The GUI should remain open while the service starts, instead of closing immediately

Fixes #7010

## Related Discussion
https://github.com/rustdesk/rustdesk/discussions/7010

### DIFF
--- a/src/ui_interface.rs
+++ b/src/ui_interface.rs
@@ -1157,9 +1157,14 @@ async fn check_connect_status_(reconnect: bool, rx: mpsc::UnboundedReceiver<ipc:
     #[cfg(target_os = "windows")]
     let mut enable_file_transfer = "".to_owned();
     let is_cm = crate::common::is_cm();
+    let mut connection_attempts = 0;
+    let max_attempts = 10; // Give service time to start
 
     loop {
-        if let Ok(mut c) = ipc::connect(1000, "").await {
+        // Increase timeout and retry attempts when service might be starting
+        let timeout = if connection_attempts < 3 { 1000 } else { 2000 };
+        if let Ok(mut c) = ipc::connect(timeout, "").await {
+            connection_attempts = 0; // Reset on successful connection
             let mut timer = crate::rustdesk_interval(time::interval(time::Duration::from_secs(1)));
             loop {
                 tokio::select! {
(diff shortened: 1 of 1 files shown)


rustdesk:
Could you upload two videos? one is without this fix (window close), one is with your fix (window not close).

GlassOnTin:
@rustdesk You're absolutely right - I misunderstood the issue. The fix I proposed only affects the CM (connection manager) popup window, not the main RustDesk window that users are reporting closes immediately.

I believe the issue is that when enabling the service:
     1. The main window's IPC connection fails with 'Access is denied'
     2. Since `reconnect=true` for the main window, it should keep retrying
     3. However, the main window might be closing due to a different mechanism

We'll need to investigate why the main window actually closes when the IPC connection fails. The current fix won't help the reported issue since it only prevents the CM popup from closing, not the main window.

rustdesk:
We know the underlying reason as you said. But We can not find a good machnism to prevent it 100%, becaues we can not reproduce it. Do you really reproduce it easily?

GlassOnTin:
@rustdesk After deeper investigation...

 In `server.rs` line 551-556, when the IPC server fails to start with "Access is denied", the process exits with `std::process::exit(-1)`.

The problem occurs because:
     1. The GUI process has the IPC pipe open at \\.\\pipe\\RustDesk\\query
     2. When enabling the service, the service process tries to create the same IPC pipe
     3. It fails with "Access is denied" because the pipe is already in use
     4. The service process exits, but this might also affect the GUI

     The fix should probably be:
     - When installing/starting the service, the GUI should close its IPC server first
     - Or the service should use a different pipe name during startup
     - Or handle the "pipe already in use" error more gracefully

My current PR doesn't address the root cause. I'll close it and investigate a proper solution

Neustradamus:
@rustdesk: Replaced by:
- https://github.com/rustdesk/rustdesk/pull/11915

PR #11915, opened 2025-05-29
Fix GUI closing when enabling service on Windows (#7010)

GlassOnTin:
Fixes #7010 — on Windows, enabling the service closes the GUI within about half a second and it never comes back.

The sequence behind it: the GUI process holds the IPC server on `\\.\pipe\RustDesk\query`; the newly started service tries to create its own server on the same pipe and gets "Access is denied"; it then calls `stop_main_window_process()`, which sends `Close` to the GUI. On Linux and macOS the GUI restarts at that point; on Windows it just exits.

Three changes:

- `server.rs`: retry the IPC server creation with exponential backoff (up to 10 attempts) instead of giving up on the first denial, and check a `RUSTDESK_SERVICE_INSTALLATION` environment variable so the GUI isn't killed while a service installation is in progress.
- `ipc.rs`: on receiving `Close`, restart the GUI with `--no-server` after a 2-second delay — the behaviour Linux and macOS already have.
- More generous retry counts and wait times so the IPC handover has room to complete.

To verify: start the GUI on Windows and enable the service from Settings. Before this change the GUI closes permanently; after it, it closes briefly and restarts.

Two other shapes considered and rejected: having the GUI close its IPC server before the service starts (more moving parts to sequence), and separate pipe names for GUI and service (breaks compatibility with existing installs). This version leaves the pipe protocol untouched.

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -420,10 +420,10 @@ async fn handle(data: Data, stream: &mut Connection) {
                 if is_server() {
                     let _ = privacy_mode::turn_off_privacy(0, Some(PrivacyModeState::OffByPeer));
                 }
-                #[cfg(any(target_os = "macos", target_os = "linux"))]
                 if crate::is_main() {
                     // below part is for main windows can be reopen during rustdesk installation and installing service from UI
                     // this make new ipc server (domain socket) can be created.
+                    #[cfg(not(windows))]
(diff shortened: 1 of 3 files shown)


Neustradamus:
@rustdesk: It replaces:
- https://github.com/rustdesk/rustdesk/pull/11914

What do you think?

rustdesk:
@21pages let's polish this with codex.

PR #12114, opened 2025-06-18
fix: win, privacy mode 2

fufesou:
https://github.com/rustdesk/rustdesk/discussions/10540

1. Show all message boxes about privacy mode after 3 seconds delay. `if (text =='no_need_privacy_mode_no_physical_displays_tip' || text == 'Enter privacy mode')` can't cover all cases, the message box does not show sometimes.
2. Fix `ChangeDisplaySettingsEx()` failure with -1. The virtual display must be changed first on `Windows 24H2`, no idea why.
3. Add some delay waiting for the virtual display to be plugged in.

## issues

1. This line may fail sometimes https://github.com/rustdesk/rustdesk/blob/87545791813287feaceb4364aa5aa7564dac912c/src/privacy_mode/win_virtual_display.rs#L437 , causing the resolution to be `1024x768`. Maybe we could add a retry here.
2. The privacy mode may fail if the display settings are some like

95be8ff9136500a22d58a8f7b5238cc

3. If there're some virtual diplays created by the other virtual display drivers, the virtual display may be hidden after exiting the privacy mode 2.

## todos

1. Issue 3 above in another PR.
2. Replace the virtual display implementation. So Issue 1 will be fixed completely.

### DIFF
--- a/flutter/lib/models/model.dart
+++ b/flutter/lib/models/model.dart
@@ -991,17 +991,12 @@ class FfiModel with ChangeNotifier {
       String link,
       bool hasRetry,
       OverlayDialogManager dialogManager) {
-    if (text == 'no_need_privacy_mode_no_physical_displays_tip' ||
-        text == 'Enter privacy mode') {
-      // There are display changes on the remote side,
-      // which will cause some messages to refresh the canvas and dismiss dialogs.
-      // So we add a delay here to ensure the dialog is displayed.
-      Future.delayed(Duration(milliseconds: 3000), () {
-        showMsgBox(sessionId, type, title, text, link, hasRetry, dialogManager);
-      });
-    } else {
+    // There are display changes on the remote side,
+    // which will cause some messages to refresh the canvas and dismiss dialogs.
+    // So we add a delay here to ensure the dialog is displayed.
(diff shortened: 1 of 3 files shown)


fufesou:
This PR is closed because the display settings cannot be restored after exiting Privacy Mode 2 if there are virtual displays created by other virtual display drivers on `Windows 11 24H2`.

`ChangeDisplaySettingsEx()` will fail with `-1` on restoring the displays  https://github.com/rustdesk/rustdesk/blob/87545791813287feaceb4364aa5aa7564dac912c/src/privacy_mode/win_virtual_display.rs#L364

If we restore with the following order, the physical displays can't be restored in my tests.

```rust
        Self::restore_displays(&self.virtual_displays);
        Self::restore_displays(&self.displays);
```

Although the display settings in the registry is already restored, and disconnecting or reconnecting RustDesk can restore the settings.  https://github.com/rustdesk/rustdesk/blob/87545791813287feaceb4364aa5aa7564dac912c/src/privacy_mode/win_virtual_display.rs#L469

Not sure if there are other cases that could cause the display settings changes to fail.

Still very dangerous if the physical display cannot be restored right after exiting Privacy Mode 2.

PR #12218, opened 2025-07-02
Update hu.rs

JohnFowler58:
Missing strings in language files.

The image below shows which strings are missing.
rust1-4
or
https://i.ibb.co/b576g6H0/40e7b986bdbd.png


JohnFowler58 (commented):
Translation of missing strings in the language file.
The image below shows which strings are missing.
rust1-4

PR #12226, opened 2025-07-03
improve sas

rustdesk:


c-michail:
Does this work on Windows Server ?

hel652:
> Does this work on Windows Server ?
Sure, it can working on Windows Server. 
I think its a very good remote tools for us.

PR #12249, opened 2025-07-07
ios screen sharing

rustdesk:


safwenbarhoumi:
This merge need check because it make basic issues

florian-obradovic:
Is this still under development? I would love to remote assist our iOS users to help find settings on their iPhone.
Unfortunately we still rely on TeamViewer for this.

PR #12449, opened 2025-07-28
refact: suppress warns on macos

fufesou:
1. Suppress warnings on macOS.
2. Replace `whoami::hostname()` with `whoami_hostname()`.
3. Remove unused code `let is_keyboard_enabled = self.peer_keyboard_enabled();` in 
```rust
#[cfg(feature = "unix-file-copy-paste")]
async fn handle_file_clip(&mut self, clip: clipboard::ClipboardFile) {
}
```

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -8,7 +8,7 @@ use std::{
 
 use serde_json::{json, Map, Value};
 
-#[cfg(not(any(target_os = "android", target_os = "ios")))]
+#[cfg(not(target_os = "ios"))]
 use hbb_common::whoami;
 use hbb_common::{
     allow_err,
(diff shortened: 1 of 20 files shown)


src/common.rs
```diff
@@ -776,12 +778,23 @@ pub fn username() -> String {
...
 }
 
+// Exactly the implementation of "whoami::hostname()".
+// This wrapper is to suppress warnings.
+#[inline(always)]
+#[cfg(not(target_os = "ios"))]
+pub fn whoami_hostname() -> String {
+    let mut hostname =
+        whoami::fallible::hostname().unwrap_or_else(|_| WHOAMI_DEFAULT_HOSTNAME.to_lowercase());
+    hostname.make_ascii_lowercase();
```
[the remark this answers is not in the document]
fufesou: Just for keeping the same implemtation of the original `whoami::hostmane()`.

src/common.rs
```diff
@@ -50,6 +50,8 @@ pub enum GrabState {
 
 pub type NotifyMessageBox = fn(String, String, String, String) -> dyn Future<Output = ()>;
 
+pub const WHOAMI_DEFAULT_HOSTNAME: &str = "LocalHost";
```
rustdesk: Have you ever seen "localhost" writen as "LocalHost" on internet?

rustdesk (changes requested):

src/common.rs
fufesou: No. I'll change it.

I kept the same logic and names to https://github.com/ardaku/whoami/blob/8170882f0962107c7b897a5688f77c97e9e3388e/src/api.rs#L16

PR #12465, opened 2025-07-29
Update bg.rs

nevaran:
Fixed and finished translation using LLM and used the **en.rs** base for template


nevaran (commented):
using template.rs instead of en.rs

nevaran (commented):
wrong file

rustdesk:
Sorry， we prefer not using LLM.

nevaran:
I have looked through the translations and they seem to be overall better...

PR #12480, opened 2025-07-30
Update ko.rs

VenusGirl:
Update Korean


VenusGirl (commented):
Have you increased the space for displaying letters?

VenusGirl (commented):
Update Korean...

PR #12510, opened 2025-08-01
Update README-KR.md

VenusGirl:
Add missing countries and links...


VenusGirl (commented):
Update translation and add missing country list link

PR #12838, opened 2025-09-05
Set WindowClass & Update Persian

xmha97:


flutter/windows/runner/win32_window.cpp
```diff
@@ -10,7 +10,7 @@
 
 namespace {
 
-constexpr const wchar_t kWindowClassName[] = L"FLUTTER_RUNNER_WIN32_WINDOW";
+constexpr const wchar_t kWindowClassName[] = L"RustDeskWindow";
```
rustdesk, line 13: @fufesou  any impact?

rustdesk: @xmha97 is this used for? any real use case?

rustdesk (changes requested):

PR #12903, opened 2025-09-12
fix websocket reconnect

21pages:
# Description

Fix the issue where the WebSocket cannot automatically reconnect when it disconnects (for example, when switching the controlled user).

daa0a06c3ab765344039cdec15b7db03

# Test

* Errors like “ID not exist” will not trigger a reconnect.

* Handshake errors in a secure connection will not trigger a reconnect.

* When the controlled side actively closes the connection, it will not reconnect.

* Windows / Linux / Mac platforms.

https://github.com/user-attachments/assets/f3d872c5-44fa-4465-a33a-02b0773c8e0f

https://github.com/user-attachments/assets/9de690d0-73cb-4046-9377-7148132d4c94

https://github.com/user-attachments/assets/80f860f4-fa0f-4cc2-9e0b-a12f97cc1897


rustdesk:
web?

PR #12962, opened 2025-09-18
Centre the main window (Windows)

logiclrd:
When the Windows app is starting, it places the window with a specific location and size. That location is currently hardcoded to be the pixel offset (10, 10). Most parts of the system, including all subsidiary windows, automatically centre themselves. This PR updates the top-level window to be centered as well.

This PR is built on top of #12933. If that PR gets merged first, I will rebase this one as needed. If this PR gets merged instead of that PR, the changes from that PR will be included via this PR. As of creation, the diff on this PR shows the combined effect of both PRs, but this PR properly only includes this commit:

https://github.com/rustdesk/rustdesk/commit/04cc078afe64387a94253a213332f57e56e5bf7d

### DIFF
--- a/flutter/windows/runner/main.cpp
+++ b/flutter/windows/runner/main.cpp
@@ -135,13 +135,11 @@ int APIENTRY wWinMain(_In_ HINSTANCE instance, _In_opt_ HINSTANCE prev,
   Win32Desktop::GetWorkArea(workarea_origin, workarea_size);
 
   // Compute window bounds for default main window position: (10, 10) x(800, 600)
-  Win32Window::Point relative_origin(10, 10);
-
-  Win32Window::Point origin(workarea_origin.x + relative_origin.x, workarea_origin.y + relative_origin.y);
+  Win32Window::Point origin = workarea_origin;
   Win32Window::Size size(800u, 600u);
 
   // Fit the window to the monitor's work area.
-  Win32Desktop::FitToWorkArea(origin, size);
+  Win32Desktop::CentreInWorkArea(origin, size);
 
   std::wstring window_title;
   if (is_cm_page) {
--- a/flutter/windows/runner/win32_desktop.cpp
+++ b/flutter/windows/runner/win32_desktop.cpp
@@ -21,10 +21,7 @@ namespace Win32Desktop
       hMonitor = MonitorFromWindow(NULL, MONITOR_DEFAULTTOPRIMARY);
 
     RECT workAreaRect;
-    workAreaRect.left = 0;
-    workAreaRect.top = 0;
-    workAreaRect.right = 1280;
-    workAreaRect.bottom = 1024 - 40; // default Windows 10 task bar height
+    bool haveWorkAreaRect = false;
 
     if (hMonitor != NULL)
     {
@@ -35,7 +32,18 @@ namespace Win32Desktop
       if (GetMonitorInfoW(hMonitor, &monitorInfo))
       {
         workAreaRect = monitorInfo.rcWork;
+        haveWorkAreaRect = true;
       }
+    }
+
+    if (!haveWorkAreaRect)
+    {
+      // I don't think this is possible, but just in case, some
+      // reasonably sane fallbacks.
+      workAreaRect.left = 0;
+      workAreaRect.top = 0;
+      workAreaRect.right = 1280;
+      workAreaRect.bottom = 1024 - 40; // default Windows 10 task bar height
     }
 
     origin.x = workAreaRect.left;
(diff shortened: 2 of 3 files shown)


fufesou:
I think https://github.com/rustdesk/rustdesk/pull/12933 is necessary, as it does correct the previously incorrect calculation.

But why isn't Dart used in this PR? This is a cross-platform issue, so wouldn't Dart be more appropriate for this modification?
We also use `WindowManager.center()` for the other windows, "Remote windows", "Filter transfer window" and "Installation window"...
It's probably better to modify it within the window manager.

In fact, if it's centered, then an `800 x 600` window should display good on a typical monitor, although the taskbar offset is not excluded.

logiclrd:
Rebased. There is now a utility function that encapsulates the logic for centering a window on the monitor that contains it/is nearest.

logiclrd:
I'm not super familiar with Dart. I did make an upstream ~PR to correct~ issue to highlight the bug in `window_manager`'s `center` function:

https://github.com/leanflutter/window_manager/issues/564

I don't know if it's going to get any traction, because the codebase at this point seems pretty dormant. The last commit merged into the main line was over 2 months ago. Most issues never get commented on at all, it seems. I do see one of your PRs merged last year, though. :-) Crossing fingers.

Can everything that's done here be done through the Flutter abstractions??

logiclrd:
Okay, with further testing I've ascertained that there's a case of mistaken identity here.

The existing code is *not* using `window_manager`'s `center`. It is using `desktop_multi_window`'s `center`. I have managed an isolated test case of `windowManager.center` and observed it working perfectly -- even in the out-of-date local copy that the RustDesk codebase is using.

Also, confounding my testing was that, at first, I didn't realize that the window wasn't being centered at all because it now had previous saved bounds.

In any case, the code in the `desktop_multi_window` plugin *is* using the wrong rectangle to centre:

https://github.com/MixinNetwork/flutter-plugins/blob/985d81661d00715a37ba1767ab2d22ff64641d43/packages/desktop_multi_window/windows/base_flutter_window.cc#L26

(should be `rcWork`)

*This* is what I observed, once, up at the start of testing, on the initial run where there were no pre-existing saved bounds. :-)

So, I'll make a PR for *that* change upstream.

Meanwhile, I am looking into the structure of the Dart code that sits on top of the platform-specific code. At first glance, it actually looks like it already tries to center, and actually also already tries to load/save the window position. But, every time I restart RustDesk, the window goes straight to (10, 10). Hmm. Either I'm looking at the wrong place, or something isn't working properly.

fufesou:
> Meanwhile, I am looking into the structure of the Dart code that sits on top of the platform-specific code. At first glance, it actually looks like it already tries to center, and actually also already tries to load/save the window position. But, every time I restart RustDesk, the window goes straight to (10, 10). Hmm. Either I'm looking at the wrong place, or something isn't working properly.

The window will always be at `(10,10)` if you have never moved it.

logiclrd:
> The window will always be at `(10,10)` if you have never moved it.

It doesn't matter what I do to it or where I leave it. It always goes back to (10, 10) on the next launch.

fufesou:
> It doesn't matter what I do to it or where I leave it. It always goes back to (10, 10) on the next launch.

It's a strange issue. Position saving and restoring has been working well for a long time.

Could you please open a command line, and run `rustdesk.exe | more` ?

You can see some outputs like `flutter: Saving frame ...` after moving the window.

You should also see `flutter: restore lpos ...` on the next run.

Please also paste the outputs here.

logiclrd:
Well that's extremely weird. I located my installation and ran `rustdesk` from the command-line, piping the output into `more` as you suggested, and on the first run, I observed this:

```
flutter: no window position saved, ignoring position restoration
```

Then I moved the window a few times, and observed lines like this:

```
flutter: Saving frame: null: 1199.0/900.0, offset:97.0/69.0, isMaximized:false, isFullscreen:false
flutter: Saving frame: null: 1199.0/900.0, offset:583.0/60.0, isMaximized:false, isFullscreen:false
flutter: Saving frame: null: 1199.0/900.0, offset:816.0/1024.0, isMaximized:false, isFullscreen:false
flutter: Saving frame: null: 1199.0/900.0, offset:307.0/348.0, isMaximized:false, isFullscreen:false
```

Then I closed it and restarted it, and saw:

```
flutter: restore lpos: 1199.0/900.0, offset:307.0/348.0, isMaximized: false, isFullscreen: false
```

And now, the window position is saving and reloading -- even if I re-open it through the shortcut from the installation. Heckin'?? Running it this way fixed something, seemingly permanently.

logiclrd:
I have rebased this branch, which was previously based on #12933 and thus duplicated its changes, onto `master`, into which those base changes are now merged, leaving only the window-centering commit. 🙂

logiclrd:
I have also removed the comment about saving/restoring the position, since the Dart code does that and my comment was the result of not knowing that because for some reason it wasn't working so I didn't know that functionality existed.

logiclrd:
I'm on another machine now. I just ran RustDesk, it placed itself at (10, 10). I dragged the window, exited, re-opened it, back to (10, 10). Did it once more, back to (10, 10). So I opened up OBS Studio, started recording my desktop, moved RustDesk, exited and reopened, and now it's saving the position here too.

I should have saved the `RustDesk_local.toml` file before I started, that was an oversight on my part. This is a frustrating issue 🙂

logiclrd:
Huh. I'm not entirely certain, but I did some more testing and it looks like it might be down to a delay between the window being moved and the new position being saved. I just ran a test where I ran RustDesk, moved the window, and then immediately closed the window with Alt-F4 to minimize the delay. Back to (10, 10). Did it again. Back to (10, 10). Then I ran it, moved the window, waited 3 seconds, and closed it. Now that position is remembered.

logiclrd:
Sigh 🙂 

https://github.com/rustdesk/rustdesk/blob/b11a8dfe54d5de72113da30ce27f381a69f4ebbb/flutter/lib/desktop/widgets/tabbar_widget.dart#L295

Literally, every time I moved the window out from under the taskbar to get access to the titlebar, it was in order to close the window and I closed it before one second elapsed and the debouncer callback hadn't yet triggered. That seems to explain everything I've observed.

Maybe it would make sense to include a `_saveFrame` call at the time the main window is closed as well. Could either just be a redundant save in most cases, or could track whether it's dirty or not to avoid a superfluous save, either way.

logiclrd:
> Maybe it would make sense to include a `_saveFrame` call at the time the main window is closed as well. Could either just be a redundant save in most cases, or could track whether it's dirty or not to avoid a superfluous save, either way.

#12974

logiclrd:
> #12974 

#12987 instead.

rustdesk:
https://github.com/rustdesk/rustdesk/pull/13003

PR #12979, opened 2025-09-20
Fix super call in onWindowResized handler

logiclrd:
Highlighted by Copilot in automated review on an unrelated PR, this override looks wrong in that it calls a different method in the superclass. Maybe this is intentional, in which case this PR can be summarily closed. :-) (Though if it is intentional, it should probably be explained with a comment.)


fufesou:
This was written by mistake, not intentional.

This PR is good.

PR #12987, opened 2025-09-21
Centralize debounce of save window pos and save window pos on close

logiclrd:
This PR reworks the saving of window positions in the following ways:

- A new debouncer is added in `common.dart` for calls to `saveWindowPosition`. This implementation includes a `flush` optional parameter that saves immediately through the debouncer.
    - I didn't want to make a decision about whether to promote `LastWindowPos` to full by value comparison by implementing `operator ==` and `hashCode`, so I added a plain old method `equals` instead.
- The debouncer specific `_saveFrame` in `tabbar_widget.dart` is removed.
- The `onWindowClose` handler in `tabbar_widget.dart` is updated to call a flushing `saveWindowPosition` to immediately write out the position, in case it has changed recently and the debouncer hasn't fired yet.

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -1796,14 +1796,17 @@ Future<void> saveWindowPosition(WindowType type, {int? windowId, bool? flush}) a
 
   final WindowKey key = (type: type, windowId: windowId);
 
+  final bool haveNewWindowPosition = (_lastWindowPosition == null) || !pos.equals(_lastWindowPosition!);
+  final bool isPreviousNewWindowPositionPending = _saveWindowDebounce.isRunning;
 
-  if ((_lastWindowPosition == null) || !pos.equals(_lastWindowPosition!) || _saveWindowDebounce.isRunning) {
+  if (haveNewWindowPosition || isPreviousNewWindowPositionPending) {
     _lastWindowPosition = pos;
 
     if (flush ?? false) {
+      // If a previous update is pending, replace it.
       _saveWindowDebounce.cancel();
       await _saveWindowPositionActual(key);
-    } else if (!_saveWindowDebounce.isRunning) {
+    } else if (haveNewWindowPosition) {
       _saveWindowDebounce.call(() => _saveWindowPositionActual(key));
     }
   }
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -18,6 +18,7 @@ import 'package:flutter_hbb/utils/multi_window_manager.dart';
 import 'package:flutter_hbb/utils/platform_channel.dart';
 import 'package:flutter_svg/flutter_svg.dart';
 import 'package:get/get.dart';
+import 'package:get/get_rx/src/rx_workers/utils/debouncer.dart';
 import 'package:provider/provider.dart';
 import 'package:uni_links/uni_links.dart';
 import 'package:url_launcher/url_launcher.dart';
@@ -1667,6 +1668,16 @@ class LastWindowPosition {
   LastWindowPosition(this.width, this.height, this.offsetWidth,
       this.offsetHeight, this.isMaximized, this.isFullscreen);
 
+  bool equals(LastWindowPosition other) {
+    return (
+      (width == other.width) &&
+      (height == other.height) &&
+      (offsetWidth == other.offsetWidth) &&
+      (offsetHeight == other.offsetHeight) &&
+      (isMaximized == other.isMaximized) &&
+      (isFullscreen == other.isFullscreen));
+  }
+
   Map<String, dynamic> toJson() {
     return <String, dynamic>{
       "width": width,
@@ -1706,9 +1717,14 @@ String get windowFramePrefix =>
         ? "incoming_"
         : (bind.isOutgoingOnly() ? "outgoing_" : ""));
 
+typedef WindowKey = ({WindowType type, int? windowId});
+
+LastWindowPosition? _lastWindowPosition = null;
+final Debouncer _saveWindowDebounce = Debouncer(delay: Duration(seconds: 1));
+
 /// Save window position and size on exit
 /// Note that windowId must be provided if it's subwindow
-Future<void> saveWindowPosition(WindowType type, {int? windowId}) async {
+Future<void> saveWindowPosition(WindowType type, {int? windowId, bool? flush}) async {
   if (type != WindowType.Main && windowId == null) {
     debugPrint(
         "Error: windowId cannot be null when saving positions for sub window");
@@ -1777,16 +1793,37 @@ Future<void> saveWindowPosition(WindowType type, {int? windowId}) async {
 
   final pos = LastWindowPosition(
       sz.width, sz.height, position.dx, position.dy, isMaximized, isFullscreen);
-  debugPrint(
-      "Saving frame: $windowId: ${pos.width}/${pos.height}, offset:${pos.offsetWidth}/${pos.offsetHeight}, isMaximized:${pos.isMaximized}, isFullscreen:${pos.isFullscreen}");
 
-  await bind.setLocalFlutterOption(
-      k: windowFramePrefix + type.name, v: pos.toString());
+  final WindowKey key = (type: type, windowId: windowId);
 
-  if ((type == WindowType.RemoteDesktop || type == WindowType.ViewCamera) &&
-      windowId != null) {
-    await _saveSessionWindowPosition(
-        type, windowId, isMaximized, isFullscreen, pos);
+
+  if ((_lastWindowPosition == null) || !pos.equals(_lastWindowPosition!) || _saveWindowDebounce.isRunning) {
+    _lastWindowPosition = pos;
+
+    if (flush ?? false) {
+      _saveWindowDebounce.cancel();
+      await _saveWindowPositionActual(key);
+    } else {
+      _saveWindowDebounce.call(() => _saveWindowPositionActual(key));
+    }
+  }
+}
+
+Future<void> _saveWindowPositionActual(WindowKey key) async {
+  LastWindowPosition? pos = _lastWindowPosition;
+
+  if (pos != null) {
+    debugPrint(
+        "Saving frame: ${key.windowId}: ${pos.width}/${pos.height}, offset:${pos.offsetWidth}/${pos.offsetHeight}, isMaximized:${pos.isMaximized}, isFullscreen:${pos.isFullscreen}");
+
+    await bind.setLocalFlutterOption(
+        k: windowFramePrefix + key.type.name, v: pos.toString());
+
+    if ((key.type == WindowType.RemoteDesktop || key.type == WindowType.ViewCamera) &&
+        key.windowId != null) {
+      await _saveSessionWindowPosition(
+          key.type, key.windowId!, pos.isMaximized ?? false, pos.isFullscreen ?? false, pos);
+    }
   }
 }
 
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -1803,7 +1803,7 @@ Future<void> saveWindowPosition(WindowType type, {int? windowId, bool? flush}) a
     if (flush ?? false) {
       _saveWindowDebounce.cancel();
       await _saveWindowPositionActual(key);
-    } else {
+    } else if (!_saveWindowDebounce.isRunning) {
       _saveWindowDebounce.call(() => _saveWindowPositionActual(key));
     }
   }
--- a/flutter/lib/desktop/widgets/tabbar_widget.dart
+++ b/flutter/lib/desktop/widgets/tabbar_widget.dart
@@ -292,7 +292,6 @@ class DesktopTab extends StatefulWidget {
 // ignore: must_be_immutable
 class _DesktopTabState extends State<DesktopTab>
     with MultiWindowListener, WindowListener {
-  final _saveFrameDebounce = Debouncer(delay: Duration(seconds: 1));
   Timer? _macOSCheckRestoreTimer;
   int _macOSCheckRestoreCounter = 0;
 
@@ -370,7 +369,7 @@ class _DesktopTabState extends State<DesktopTab>
 
   void _setMaximized(bool maximize) {
     stateGlobal.setMaximized(maximize);
-    _saveFrameDebounce.call(_saveFrame);
+    _saveFrame();
     setState(() {});
   }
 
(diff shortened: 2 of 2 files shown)


logiclrd:
I sorted out my local debug environment, by the way. Turns out that if you just install the latest Flutter version, it contains at least one breaking change compared to what the RustDesk code expects to be run against (Google says it is between Flutter versions 3.26 and 3.27, which is a flagrant violation of semantic versioning 🙂). The `ThemeData` class `dialogTheme` and `tabBarTheme` properties (and possibly others) have changed in type. This newer version is also incompatible with `extended_text` version 14.0.0, to which that dependency is locked in `pubspec.yaml`. The following local diff finally allowed me to `flutter run` from code:

```
diff --git a/flutter/lib/common.dart b/flutter/lib/common.dart
index ee509f3..523e066 100644
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -372,7 +372,7 @@ class MyTheme {
     appBarTheme: AppBarTheme(
       shadowColor: Colors.transparent,
     ),
-    dialogTheme: DialogTheme(
+    dialogTheme: DialogThemeData(
       elevation: 15,
       shape: RoundedRectangleBorder(
         borderRadius: BorderRadius.circular(18.0),
@@ -403,7 +403,7 @@ class MyTheme {
     cardColor: grayBg,
     hintColor: Color(0xFFAAAAAA),
     visualDensity: VisualDensity.adaptivePlatformDensity,
-    tabBarTheme: const TabBarTheme(
+    tabBarTheme: const TabBarThemeData(
       labelColor: Colors.black87,
     ),
     tooltipTheme: tooltipTheme(),
@@ -470,7 +470,7 @@ class MyTheme {
     appBarTheme: AppBarTheme(
       shadowColor: Colors.transparent,
     ),
-    dialogTheme: DialogTheme(
+    dialogTheme: DialogThemeData(
       elevation: 15,
       shape: RoundedRectangleBorder(
         borderRadius: BorderRadius.circular(18.0),
@@ -504,7 +504,7 @@ class MyTheme {
     ),
     cardColor: Color(0xFF24252B),
     visualDensity: VisualDensity.adaptivePlatformDensity,
-    tabBarTheme: const TabBarTheme(
+    tabBarTheme: const TabBarThemeData(
       labelColor: Colors.white70,
     ),
     tooltipTheme: tooltipTheme(),
diff --git a/flutter/pubspec.yaml b/flutter/pubspec.yaml
index 9172703..016eb8e 100644
--- a/flutter/pubspec.yaml
+++ b/flutter/pubspec.yaml
@@ -105,7 +105,7 @@ dependencies:
   pull_down_button: ^0.9.3
   device_info_plus: ^9.1.0
   qr_flutter: ^4.1.0
-  extended_text: 14.0.0
+  extended_text: ^15.0.0
   xterm: 4.0.0
   sqflite: 2.2.0
   google_fonts: ^6.2.1
```

I installed Flutter via the Visual Studio Code extension, per the recommendation on the Flutter home page, and ended up with:

```
C:\code\rustdesk\flutter>flutter --version
Flutter 3.35.4 • channel stable • https://github.com/flutter/flutter.git
Framework • revision d693b4b9db (5 days ago) • 2025-09-16 14:27:41 +0000
Engine • hash feee8ee8fb8b975dd9990f86d3bda11e6e75faf3 (revision c298091351) (6 days ago) • 2025-09-15 14:04:24.000Z
Tools • Dart 3.9.2 • DevTools 2.48.0

C:\code\rustdesk\flutter>
```

Anyway, it certainly unlocks things to be able to build and debug locally 🙂

logiclrd:
Removed the Linux-specific save call. Tested working on Ubuntu.

logiclrd:
Rebased (#12979).

flutter/lib/common.dart
```diff
@@ -18,6 +18,7 @@ import 'package:flutter_hbb/utils/multi_window_manager.dart';
 import 'package:flutter_hbb/utils/platform_channel.dart';
 import 'package:flutter_svg/flutter_svg.dart';
 import 'package:get/get.dart';
+import 'package:get/get_rx/src/rx_workers/utils/debouncer.dart';
```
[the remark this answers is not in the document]
fufesou: This import is required.

flutter/lib/common.dart
```diff
@@ -1706,9 +1717,14 @@ String get windowFramePrefix =>
         ? "incoming_"
         : (bind.isOutgoingOnly() ? "outgoing_" : ""));
 
+typedef WindowKey = ({WindowType type, int? windowId});
+
+final Map<WindowKey, LastWindowPosition> _saveWindowData = Map<WindowKey, LastWindowPosition>();
```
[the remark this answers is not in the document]
fufesou: IDE (the flutter plugin) also prompts this suggestion.

fufesou:
@logiclrd To me, this PR is well designed.

But maybe `_saveWindowData` is not needed.
Because each window runs in a separate Dart isolate, they maintain independent global variable instances.

You can open different windows, and print `_saveWindowData.length`, it's always `1`.

You can use a variable instead of the map.

logiclrd:
Ahh, cool beans :-) That makes things a little simpler.

I also found a bug in the logic. If there's an outstanding call within the debouncer, then `flush` doesn't work, because the state data has already been updated (so it thinks it doesn't have any work to do) but the write hasn't yet happened. If this happens during shutdown, then the process exits before the write takes place. I fixed that in this update too.

fufesou:
>  If ... then `flush` doesn't work

You added a condition, which should make flushing more harder. 
I don't understand why this check is needed.

2f0f02bc-d2ac-4070-9063-7f8864d7a2c2

flutter/lib/common.dart
```diff
@@ -1706,9 +1717,14 @@ String get windowFramePrefix =>
         ? "incoming_"
         : (bind.isOutgoingOnly() ? "outgoing_" : ""));
 
+typedef WindowKey = ({WindowType type, int? windowId});
+
+LastWindowPosition? _lastWindowPosition = null;
```
fufesou, line 1722: No need to set `null` here. `LastWindowPosition? _lastWindowPosition;`

image

logiclrd:
> > If ... then `flush` doesn't work
> 
> You added a condition, which should make flushing more harder.

It actually makes flushing easier, because it's a third alternative that is "OR"ed into the condition expression. Specifically, if a previous call was made without `flush` and it latched the new bounds into `_lastWindowPosition`, and now there's a debouncer `.call` in flight, and then the user elects to close the application _right now_, then the `onWindowClose` handler runs and it makes another call to `saveWindowPosition` -- but now, the latched `_lastWindowPosition` matches the current physical window bounds. If that's the only check done, then it will not enter that branch and won't see that `flush` is true. Nothing gets saved, and then the process exits before the debounce timer goes off. With this change, it'll enter that branch if it sees a change in the bounds _or_ if there's a debounce in flight.

I just realized another layer, though. I think it's strictly hypothetical, but if the window bounds changed once and then something were to regularly call `saveWindowPosition` without the physical bounds changing, it'd continually reset the debounce timer and never actually save anything. So, one more check needed: if there's already a debounce in flight, no need to start a new one.

logiclrd:
Hmm, the other side of that hypothetical is that if the window bounds _don't stop changing_, then now it'll persist a snapshot of them every debounce interval. Is that a good or bad thing? If it's a good thing (you continue dragging the window for a really long time and it saves the position once a second while you're doing it), then it's good to go. If not, then the condition can be altered so that it _does_ reschedule the debounce if the position is actually changing.

logiclrd:
Okay I just tested and it doesn't get realtime updates on window drags anyway, so it's a non-issue, at least on Windows. And if I'm not mistaken, the Linux and OS X wrappers don't subscribe to bounds change notifications at all.

fufesou:
> It actually makes flushing easier, because it's a third alternative that is "OR"ed into the condition expression. 

I don't think so.

```dart
if ((_lastWindowPosition == null) || !pos.equals(_lastWindowPosition!))
```

equals

```dart
if ((_lastWindowPosition == null) || !pos.equals(_lastWindowPosition!) || true)
```

while the following code do one more check.

```dart
if ((_lastWindowPosition == null) || !pos.equals(_lastWindowPosition!) || _saveWindowDebounce.isRunning)
```

logiclrd:
> > It actually makes flushing easier, because it's a third alternative that is "OR"ed into the condition expression.
> 
> I don't think so.
> 
> ```dart
> if ((_lastWindowPosition == null) || !pos.equals(_lastWindowPosition!))
> ```
> 
> equals
> 
> ```dart
> if ((_lastWindowPosition == null) || !pos.equals(_lastWindowPosition!) || true)
> ```
> 
> while the following code do one more check.
> 
> ```dart
> if ((_lastWindowPosition == null) || !pos.equals(_lastWindowPosition!) || _saveWindowDebounce.isRunning)
> ```

I believe you're mistaken. If it were `|| true`, then it would _always enter the `if`, regardless of any other conditions_. The default is `|| false` (or `&& true`, for the equivalent no-op on AND expressions).

fufesou:
> Okay I just tested and it doesn't get realtime updates on window drags anyway, so it's a non-issue, at least on Windows.

Maybe you only tested the main window.

The remote (sub) window calls "saveWindowPosition()" frequently while being dragged, saving the position once a second.

logiclrd:
> > Okay I just tested and it doesn't get realtime updates on window drags anyway, so it's a non-issue, at least on Windows.
> 
> Maybe you only tested the main window.
> 
> The remote (sub) window calls "saveWindowPosition()" frequently while being dragged, saving the position once a second.

That is indeed the case. I will switch the condition, in that case.

logiclrd:
That seems to do the trick. :-)

logiclrd:
Thinking about things, I've realized that part of the reason I assumed that this method was multi-tenant is the fact that `saveWindowPosition` takes parameters identifying the window. But in fact, these will always be the same for a given isolate. They shouldn't be parameters to `saveWindowPosition`, but rather persistent state set once up front. That's an architectural issue outside of the scope of this PR.

(I just scouted it out, and perhaps a more logical design would have `saveWindowPosition` be a method in the `WindowController` class -- a common implementation provided by the abstract base, since there's nothing platform-specific about it. Then, `main.dart` could initialize a reference to a `WindowController` for the window it is driving -- as soon as it knows a value for `kWindowId`, it can be used with `WindowController.fromWindowId`, right? And then all of the calls to `WindowController.fromWindowId(kWindowId!)` peppered through the code would just use that global instead, and the call to save the window frame would simply be `kWindowController!.saveFrame()`, regardless of whether the window is the main window or is identified by an ID. I don't know the architecture well enough to know if there are caveats this would run into.)

Also, in pondering this, I realized my original solution wouldn't actually have worked 100% of the time were the windows not isolated, because the debouncer isn't multi-tenant. If the user moved one window and then quickly moved another window, then the first window's update would get scheduled through the debouncer, and could then be *replaced* by the second window's update before firing. Just because the `key` is different doesn't mean the debouncer sees them as different. 😛 Non-issue because of the use of isolates, though.

flutter/lib/common.dart
```diff
@@ -1777,16 +1793,40 @@ Future<void> saveWindowPosition(WindowType type, {int? windowId}) async {
...
+  final bool isPreviousNewWindowPositionPending = _saveWindowDebounce.isRunning;
+
+  if (haveNewWindowPosition || isPreviousNewWindowPositionPending) {
+    _lastWindowPosition = pos;
+
+    if (flush ?? false) {
+      // If a previous update is pending, replace it.
+      _saveWindowDebounce.cancel();
+      await _saveWindowPositionActual(key);
+    } else if (haveNewWindowPosition) {
```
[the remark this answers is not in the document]
logiclrd: This seems to be a false positive. It thinks that the inner check can just be removed entirely, but it's missing the fact that the outer `if` could be entered with `haveNewWindowPosition == false` when `isPreviousNewWindowPositionPending == true`. Its suggestion will alter the behaviour and, in fact, reintroduce the bug that the last commit fixes. 😛

flutter/lib/common.dart
```diff
@@ -1777,16 +1793,40 @@ Future<void> saveWindowPosition(WindowType type, {int? windowId}) async {
...
+  final WindowKey key = (type: type, windowId: windowId);
 
-  if ((type == WindowType.RemoteDesktop || type == WindowType.ViewCamera) &&
-      windowId != null) {
-    await _saveSessionWindowPosition(
-        type, windowId, isMaximized, isFullscreen, pos);
+  final bool haveNewWindowPosition = (_lastWindowPosition == null) || !pos.equals(_lastWindowPosition!);
+  final bool isPreviousNewWindowPositionPending = _saveWindowDebounce.isRunning;
+
+  if (haveNewWindowPosition || isPreviousNewWindowPositionPending) {
```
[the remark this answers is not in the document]
logiclrd: The conditions `haveNewWindowPosition || isPreviousNewWindowPositionPending` and `haveNewWindowPosition` are not equivalent. 🙂

flutter/lib/common.dart
fufesou: @logiclrd Please also address this suggestion.

fufesou (approved):
Only one suggestion, `LastWindowPosition? _lastWindowPosition = null;` can be `LastWindowPosition? _lastWindowPosition;`. But it is not a problem.

flutter/lib/common.dart
logiclrd: Wait, you wrote these comments 2, 3 days ago? I didn't receive any notification of them, despite your direct tag. I only just saw them now, after just seeing the notification of the approval. Weird. I did not ignore this on purpose!

Regarding this particular suggestion, my personal feeling on this is that this kind of suggestion (and there are plenty like them in diverse languages) is taking prescriptivism a little bit too far. I look at this and my thought is, "Does it actually improve the code to make the suggested change?" Sure, the default value of the field is `null`, but explicitly showing that it is `null` at initialization time, at least to me, can sometimes make it easier to reason about the code. Just because you *can* omit the initializer doesn't mean it's *better* that way.

I looked up the warning specifically, and it gives the following example:

```
  Item? bestItem = null;

  for (final item in cart) {
    if (bestItem == null || item.price < bestItem.price) {
      bestItem = item;
    }
  }
```

In this case, yes, it is definitely unnecessary; it isn't really instructive of anything. The amount of time where `bestItem` could be `null`, if the list isn't empty, is extremely short, and the explicit comparison with `null` reminds the reader just 3 lines away from the declaration that `bestItem` could be `null` under *some* circumstances.

But in the code here, `_lastWindowPosition` is a state variable. Other code that isn't close to the declaration needs to contend with the fact that a valid state for `_lastWindowPosition` is that it be `null`. It's not the case here and now, but it could conceivably also return to that state at some point. If you strictly follow `avoid_init_to_null` then now you have two different expressions of setting the state to `null`. I think it's an antipattern, personally, and that `avoid_init_to_null` *sometimes* could make code worse.

Anyway, rant over, that's just one programmer's thought process on the matter. Let me know if you'd still like me to make a commit to remove it.

fufesou: Ok. That's a fair point.

I had initially set it to null but changed it after following the Flutter plugin's instructions.

In some languages, if a value isn't explicitly assigned, it defaults to a random or undefined state, which can be problematic.

However, here, it is more of a personal preference than a problem.

logiclrd: Indeed :-) As long as it doesn't cause compile-time warnings or other problems.

PR #12989, opened 2025-09-22
add ab.py, audits.py

21pages:
Script for the address book and audit APIs
* address book: view/add/update/delete for ab/peer/tag/rule
* audits: view conn/file/alarm/console audits

# Preview

### ab
ab

### tag
tag

### peer
peer

### rule
rule

## audits

audit1

audit2

audit3

### DIFF
--- a/res/ab.py
+++ b/res/ab.py
@@ -1,0 +1,771 @@
+#!/usr/bin/env python3
+
+import requests
+import argparse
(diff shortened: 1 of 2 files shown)


cbka:
will the endpoint also provide a list of all shared ABs with their GUIDs ?

PR #13045, opened 2025-09-25
feat(ui): custom scale mode with inline controls and live apply

deblasis:
## Summary

- Adds a third view style: Scale custom.
- Inline controls under Display → Ratio when Custom is active:
  - “Scale ” label with current percent (e.g., 125%)
  - - and + buttons for 1% steps
  - Slider (5%–1000%, 1% step)
- Changes apply immediately without dismissing the menu.
- Persists the custom percentage per peer and auto-applies on reconnect when Custom was last used.

## Motivation
- Provide precise zoom control beyond original/adaptive.
- Improve UX with instant feedback and fewer clicks.
- Honor user preference per remote peer.

## Behavior
- While Custom is active:
  - Pressing -/+ updates scale by 1% and re-renders.
  - Slider updates scale in real time.
- If Custom wasn’t active, selecting it switches mode and shows the inline controls.
- The percentage is persisted per peer and restored next session if Custom was last used.

## Implementation notes
- Flutter UI:
  - Desktop remote toolbar: inline Custom controls shown only when Custom is selected.
  - No modal auto-open on Custom selection.
  - Percentage is displayed as a label with a % suffix.
- State:
  - Live apply via existing `sessionSetFlutterOption(kCustomScalePercentKey)` and `canvasModel.updateViewStyle()`.
  - Persistence per peer via existing session/peer option storage.
- No protocol or Rust core changes; FFI path unchanged.

## Screenshot / Videos

- V0
image

- V1
image

https://github.com/user-attachments/assets/c8d01a19-524b-4b0c-93b2-7dcf88cf3697

- V2

Two different approaches have been evaluated

### Uneven (preferred choice, reasoning https://github.com/rustdesk/rustdesk/pull/13045#issuecomment-3342148958)

https://github.com/user-attachments/assets/85233ff2-e027-4f2f-9302-4e0eab48abbb

### Dynamic

https://github.com/user-attachments/assets/9141302d-ddf3-4c0f-91fa-0b49b1563c05

## Testing
- Manual:
  - Switch between Original/Adaptive/Custom and verify scale changes.
  - Adjust via +/- and slider; ensure immediate canvas update.
  - Reconnect to same peer with Custom last used; verify auto-apply of saved percent.
  - Clamp behavior at 5% and 1000%.
- Platforms covered: macOS desktop (others share same Flutter UI path).

## Risk/Regression
- Low. Changes are scoped to Flutter UI and canvas scale application.
- Scale computation uses existing code paths; Custom-percent handling reuses session options.

## Coding assistance disclosure

I am a novice in Rust and Flutter even if I have 20+ years of programming experience. 
I used Cursor with GPT-5 to help me understand how to implement this and tested it locally at every iteration almost like I learnt PHP back in the day... 

I use RustDesk almost everyday ❤️  and I found this useful for myself. I thought I'd share 😄 

Sign-off
Signed-off-by: Alessandro De Blasis <alex@deblasis.net>


rustdesk:
Good job.

A video please

rustdesk:
Honestly, this menu design is ugly.

image

If you can ask cursor to make this menu better look and design. :(

rustdesk:
> I used Cursor with GPT-5 to help me understand how to implement this and tested it locally at every iteration almost like I learnt PHP back in the day...

You are great.

deblasis:
> Honestly, this menu design is ugly.
> 
> image
> 
> If you can ask cursor to make this menu better look and design. :(

Guilty as charged! 

I would appreciate some pointers. I was inspired by the "virtual display" submenu. I didn't want to go overboard.

This current version is essentially my MVP but super happy to polish it to be production ready.  

Once the design is better looking I will also post a video in the PR description. In the meantime I'll clean-up the code.

fufesou:
1. The scroll options are missed when using "Scale custom".

image

2. Maybe entering a number would be useful.

image

rustdesk:
> Honestly, this menu design is ugly.
> 
> image
> 
> If you can ask cursor to make this menu better look and design. :(

Virtual display is in second submenu, that's fine. You can also try to put the scale controls in to second submenu.

deblasis:
I am playing with something like this:

image

I have updated the PR description, adding also a video of the feature. I hope you can appreciate the way I **dove** into the codebase :)

rustdesk:
Will this make original menu much wider?



> I am playing with something like this:
> 
> image
> I have updated the PR description, adding also a video of the feature. I hope you can appreciate the way I **dove** into the codebase :)

deblasis:
> Will this make original menu much wider?
> 
> > I am playing with something like this:
> > image
> > I have updated the PR description, adding also a video of the feature. I hope you can appreciate the way I **dove** into the codebase :)

The slider will only fill the available width and as far as I understand, the width is like that to make space for the "Use all my displays for the remote session" menu item. In this screenshot I selected another Scale setting and the width is consistent.

image

Also confirmed from my latest released version without my change, the width seems to be the same.

rustdesk:
How about just put an editbox on the right side of `Scale custom`? Clicking on - + seems not friendly, need to click many times.

Or put an edit icon on the right side, when user clicks on the icon, pop out an edit dialog. menu item text changes to `Scale custom (50%)`.

deblasis:
> How about just put an editbox on the right side of `Scale custom`? Clicking on - + seems not friendly, need to click many times.
> 
> Or put an edit icon on the right side, when user clicks on the icon, pop out an edit dialog. menu item text changes to `Scale custom (50%)`.

Let me first describe how I use this feature: 
If I want to look at a particular section/window of the remote screen, I use the slider to quickly match more or less the height of the remote window with my RustDesk window, then I use the +/- to fine tune and get almost perfectly to the scale I want.

Having an input field would be just guessing numbers until it works and you have to type, close the menu and do it again many times. A slider gives you immediate visual feedback as you move it around. 
Also, nesting it under a submenu or dialog would add even more friction (clicks). In fact, initially, I had developed a small dialog but it was painful to go back and open it again, having to guess numbers.

So, this is my use case. Not sure if you have one I am missing that would justify the use of a textbox with a scale %. How do you know beforehand what custom scale you need?

That being said, think about this and let me know so that I can make the required changes. It's your product ultimately and I am here to contribute.
 
Personally, I am pretty happy with the way it is currently. It solves my specific problem. I can always keep running a custom version for my use case and every now and then integrate pulling changes from here.

rustdesk:
Agree with you. I accept your design. Good job.

image

image

rustdesk:
Let's wait for @fufesou's code review and test.

flutter/lib/models/model.dart
```diff
@@ -1699,6 +1699,8 @@ class ViewStyle {
         final s2 = height / displayHeight;
         s = s1 < s2 ? s1 : s2;
       }
+    } else if (style == kRemoteViewStyleCustom) {
+      // Read custom scale percent from flutter option; fallback to 100%
```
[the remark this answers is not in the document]
deblasis: I thought the same but then I asked GPT-5 and it had a convincing argument along these lines:

ViewStyle.scale is synchronous and has no session context. It can’t fetch the per-session custom percent. That’s why it’s a no-op for Custom. Actual scaling is applied in `CanvasModel.updateViewStyle` after reading `kCustomScalePercentKey`, which is why it works at runtime.

fufesou:
Hi, this PR is excellent.

The regions of 5-100 and 100-1000 is a bit odd. I'm not sure if a uniform slider would be appropriate.

Of course, this is just my slight quirk, perhaps it's normal.

image


https://github.com/user-attachments/assets/b09b234c-ac0d-42f5-91ec-f3ffac2e234a

flutter/lib/desktop/widgets/remote_toolbar.dart
```diff
@@ -1146,7 +1207,7 @@ class _DisplayMenuState extends State<_DisplayMenu> {
         widget.ffi.canvasModel.updateScrollStyle();
       }
 
-      final enabled = widget.ffi.canvasModel.imageOverflow.value;
+      final enabled = _scrollEnabled;
       return Column(children: [
```
fufesou: Maybe you can remove `_scrollEnabled` and use `Obx()` here.

`widget.ffi.canvasModel.addListener(_onCanvasModelChanged);` looks strange.

```dart
      return Obx(() => Column(children: [
            RdoMenuButton<String>(
              child: Text(translate('ScrollAuto')),
              value: kRemoteScrollStyleAuto,
              groupValue: groupValue,
              onChanged: widget.ffi.canvasModel.imageOverflow.value
                  ? (value) => onChange(value)
                  : null,
              ffi: widget.ffi,
            ),
            RdoMenuButton<String>(
              child: Text(translate('Scrollbar')),
              value: kRemoteScrollStyleBar,
              groupValue: groupValue,
              onChanged: widget.ffi.canvasModel.imageOverflow.value
                  ? (value) => onChange(value)
                  : null,
              ffi: widget.ffi,
            ),
            Divider(),
          ]));
```

deblasis:
> Hi, this PR is excellent.

I am glad you like it. I had fun and I learnt a few things along the way which is always great.

> The regions of 5-100 and 100-1000 is a bit odd. I'm not sure if a uniform slider would be appropriate.

I agree with you, I used wide ranges because what could happen is that you could have a wide range of resolution ratios between remote and client. Like a massive remote screen on a tiny device and viceversa.

I don't have a solution to this at the moment, something "smarter" that maybe adjusts the slider range depending on remote screen and client would be great I suppose but I need to play with it a bit more to understand what would work across the board, using different devices combos (big->small, small->big, same->same) in terms of screen sizes and also areas of the screens that I want to zoom in/out. 

🤔 

if you have any ideas about what would be an improvement please tell me. 
Maybe also have a chat about it internally as well or, not sure if it's viable, try to get some easly feedback from users that vocalized the need for a zoom feature like here (I saw the reference) https://github.com/rustdesk/rustdesk/issues/13011#issuecomment-3339621448

flutter/lib/desktop/widgets/remote_toolbar.dart
```diff
@@ -1245,6 +1306,231 @@ class _DisplayMenuState extends State<_DisplayMenu> {
...
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final colorScheme = Theme.of(context).colorScheme;
+    const smallBtnConstraints = BoxConstraints(minWidth: 28, minHeight: 28);
+
+    final sliderControl = Semantics(
+      label: translate('Custom scale slider'),
```
fufesou, line 1448: I can't see this label and maybe it is not required.

flutter/lib/desktop/widgets/remote_toolbar.dart
```diff
@@ -1245,6 +1306,231 @@ class _DisplayMenuState extends State<_DisplayMenu> {
...
+        ),
+      ),
+    );
+
+    return Column(children: [
+      Padding(
+        padding: const EdgeInsets.symmetric(horizontal: 12.0),
+        child: Row(children: [
+          Tooltip(
+            message: translate('Decrease'),
```
fufesou, line 1494: Maybe this tooltip is not required.

flutter/lib/desktop/widgets/remote_toolbar.dart
```diff
@@ -1245,6 +1306,231 @@ class _DisplayMenuState extends State<_DisplayMenu> {
...
+              iconSize: 16,
+              padding: EdgeInsets.all(1),
+              constraints: smallBtnConstraints,
+              icon: const Icon(Icons.remove),
+              onPressed: () => _nudge(-1),
+            ),
+          ),
+          Expanded(child: sliderControl),
+          Tooltip(
+            message: translate('Increase'),
```
fufesou, line 1505: Maybe this tooltip is not required.

flutter/lib/common/widgets/toolbar.dart
```diff
@@ -363,6 +363,11 @@ Future<List<TRadioMenu<String>>> toolbarViewStyle(
         child: Text(translate('Scale adaptive')),
         value: kRemoteViewStyleAdaptive,
         groupValue: groupValue,
+        onChanged: onChanged),
+    TRadioMenu<String>(
+        child: Text(translate('Scale custom')),
```
fufesou, line 368: For translation `translate('Scale custom')`, you can add `Scale custom` at the end of `src/lang/template.rs` and run `python res/lang.py` to create placeholders in other translation files.

fufesou:
I'm not sure if the mobile version also need this feature, but even if it do. It's ok do it in another PR.

fufesou:
> if you have any ideas about what would be an improvement please tell me.

How about this design? It might be a bit complicated though.

https://github.com/user-attachments/assets/71680a84-f535-4e5e-864e-93f0e6bf6726

deblasis:
> > if you have any ideas about what would be an improvement please tell me.
> 
> How about this design? It might be a bit complicated though.
> 
>  custom.mp4

🤔 Maybe we could display two buttons. one is "zoom out" and one is "zoom in" 

Zoom out sets the range of the slider between 0-100% and zoom in from 100% to 5000%

Or maybe, even better if we want to make it even more frictionless:
if you drag to a ratio that is below 100%, the range updates automatically to be 0-101% (you are in zoom out mode, inferred by the value)
the moment the value goes again > 100%, the range updates to be 100% to 5000% (you are in zoom in mode, inferred by the value)

If flutter allows this kind of interactivity I think it would feel pretty good to use, at least the way I imagine it.

fufesou:
> if you drag to a ratio that is below 100%, the range updates automatically to be 0-101% (you are in zoom out mode, inferred by the value)
the moment the value goes again > 100%, the range updates to be 100% to 5000% (you are in zoom in mode, inferred by the value)

Probably won't be very easy to implement.

How about this slider? GPT can generate this full demo.


https://github.com/user-attachments/assets/29e6a2e3-8ec1-4943-8254-0dadba8879d2

deblasis:
image


# Uneven

https://github.com/user-attachments/assets/85233ff2-e027-4f2f-9302-4e0eab48abbb

# Dynamic

https://github.com/user-attachments/assets/9141302d-ddf3-4c0f-91fa-0b49b1563c05

deblasis:
Which one of the two implementations should we go for?

fufesou:
I'm not sure which one is better, but I tend to prefer the **uneven** one.

rustdesk:
@deblasis  which one do you prefer? You are the real user of this feature.

deblasis:
> I'm not sure which one is better, but I tend to prefer the **uneven** one.

TL;DR: I agree. I pushed the corresponding changes for review.

> @deblasis which one do you prefer? You are the real user of this feature.

I played a bit with both. 

**Uneven** is more intuitive, I guess most users would find it more familiar.

**Dynamic** feels more like "magic", I have a slight preference for this one because it's explicit about zooming in or out and gives finer control BUT this promise is broken by the fact that the UX in the switching modes needs some refinement that's probably not worth the hassle. 
If the value is around 100% and the user drags even a tiny bit in the opposite direction they intend to go (it happens a lot, especially while using touchpads and tapping), the slider switches "mode" and the user is caught off guard. There should be some logic around detecting these cases but I am not sure if it's worth the time/effort. Because of this issue, I prefer the **uneven** version which is definitely an improvement compared to the V1 in the PR description.

I have pushed also the branch with the changes to enable the **dynamic** mode here https://github.com/deblasis/rustdesk/tree/feat/scale-custom-dynamic-ranges just in case you want to play with it or if in the future you decide to opt for this approach but for now I feel like it's best to go for **uneven**.

rustdesk:
Then, let's use the uneven one, since you both prefer it.

rustdesk:
I think you can have enough time to improve the uneven logic with the help of ai.

flutter/lib/models/model.dart
```diff
@@ -1824,12 +1833,25 @@ class CanvasModel with ChangeNotifier {
...
     _devicePixelRatio = ui.window.devicePixelRatio;
     if (kIgnoreDpi && style == kRemoteViewStyleOriginal) {
       _scale = 1.0 / _devicePixelRatio;
     }
     _resetCanvasOffset(displayWidth, displayHeight);
-    _imageOverflow.value = _x < 0 || y < 0;
+    final overflow = _x < 0 || y < 0;
+    if (_imageOverflow.value != overflow) {
+      _imageOverflow.value = overflow;
+    }
```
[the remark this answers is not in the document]
deblasis: I think we should keep the guard (defensive programming).

From my research, it avoids redundant Rx writes + listener notifications and prevents unnecessary UI/state churn.

flutter/lib/desktop/widgets/remote_toolbar.dart
```diff
@@ -1024,6 +1031,8 @@ class _DisplayMenu extends StatefulWidget {
 }
 
 class _DisplayMenuState extends State<_DisplayMenu> {
+  final RxInt _customPercent = 100.obs;
+  bool _scrollEnabled = false;
```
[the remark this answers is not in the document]
deblasis: I think we should keep this.

Reasons: 
- It’s a local snapshot used only to enable/disable menu items; you update state only on actual toggles, not on every canvas model change.
- It avoids wiring Obx inside submenu items, which can behave poorly while menus are open; a local setState on boolean flip is predictable and light.
- It’s easy to extend (e.g., combine with other gating logic without touching the observable).
- It decouples FutureBuilder/menu composition from GetX reactivity while still reacting to changes via a narrow listener that ignores no-ops (see the next != _scrollEnabled check above).

I fidgeted a lot with these things and learnt about Obx. I was having freezes in the remote connection window before settling to this. I mentioned this in the video as well I think.

flutter/lib/desktop/widgets/remote_toolbar.dart
```diff
@@ -1245,6 +1306,290 @@ class _DisplayMenuState extends State<_DisplayMenu> {
...
+    if (p > 1.0) return 1.0;
+    return p;
+  }
+
+  @override
+  void initState() {
+    super.initState();
+    _value = 100;
+    _debouncerScale = Debouncer<int>(
+       Duration(milliseconds: 300),
```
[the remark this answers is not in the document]
deblasis: I agree, I created a named constant and I also tidied up the other constants

flutter/lib/desktop/widgets/remote_toolbar.dart
fufesou: https://github.com/rustdesk/rustdesk/pull/13045#discussion_r2384435059

> I was having freezes in the remote connection window before settling to this.

Do you mean the above code causes freezes?
I didn't experience such freezes in my testing.

Please add comments above `_scrollEnabled` if it is really needed.

fufesou:
There's one other usability issue: when I select "Scale custom," I expect to see the current scale value immediately and start adjusting it. However, the menu closes right away, so I have to click again to reopen it.

https://github.com/user-attachments/assets/befda322-63d2-4aaf-98cc-96923ef206ae

deblasis:
> There's one other usability issue: when I select "Scale custom," I expect to see the current scale value immediately and start adjusting it. However, the menu closes right away, so I have to click again to reopen it.

Fair point. I made a few changes. Try again now.

fufesou (changes requested):

flutter/lib/desktop/widgets/remote_toolbar.dart
deblasis: > Do you mean the above code causes freezes?
> I didn't experience such freezes in my testing.

No, a version of the code that I had that was using Obx I think it was causing freezes. I'll try again and let's see how that goes. 

> Please add comments above _scrollEnabled if it is really needed.

OK

flutter/lib/desktop/widgets/remote_toolbar.dart
deblasis: It's for accessibility, not a visual element, the text would be used by screenreaders like VoiceOver.

flutter/lib/desktop/widgets/remote_toolbar.dart
deblasis: Similarly, I believe these tooltops would be used by screenreaders. I might be wrong.

flutter/lib/desktop/widgets/remote_toolbar.dart
deblasis: Similarly, I believe these tooltops would be used by screenreaders. I might be wrong.

flutter/lib/desktop/widgets/remote_toolbar.dart
deblasis: You definitely know more than me so I will execute. 

I refactored the code in 49af244f1f51b868f4c11c6f3b9d8cc8a3e0775a

To me, that was an optimization, perhaps a bit unclear in intent but certainly the code is simpler with Obx. If it doesn't cause issues, it should be OK I guess.

I tested a bit and I haven't experienced any freezes now.

flutter/lib/models/model.dart
```diff
@@ -1824,12 +1833,25 @@ class CanvasModel with ChangeNotifier {
...
+
     _devicePixelRatio = ui.window.devicePixelRatio;
     if (kIgnoreDpi && style == kRemoteViewStyleOriginal) {
       _scale = 1.0 / _devicePixelRatio;
     }
     _resetCanvasOffset(displayWidth, displayHeight);
-    _imageOverflow.value = _x < 0 || y < 0;
+    final overflow = _x < 0 || y < 0;
+    if (_imageOverflow.value != overflow) {
```
fufesou, line 1853: There is actually no need to do this check.

https://github.com/jonataslaw/getx/blob/dae1f77021149615ed756bc9b26ee4c961f460d9/lib/get_rx/src/rx_types/rx_core/rx_impl.dart#L101


It's Ok though, there is no problem with this check.

flutter/lib/utils/scale.dart
```diff
@@ -0,0 +1,36 @@
+import 'package:flutter_hbb/consts.dart';
+import 'package:flutter_hbb/models/platform_model.dart';
+
+import 'package:flutter_hbb/generated_bridge.dart'
```
fufesou:
```dart
import 'package:flutter_hbb/generated_bridge.dart'
    if (dart.library.html) 'package:flutter_hbb/web/bridge.dart';
```

This line is not required.

flutter/lib/models/model.dart
```diff
@@ -1824,12 +1833,25 @@ class CanvasModel with ChangeNotifier {
     _lastViewStyle = viewStyle;
     _scale = viewStyle.scale;
 
+    // Apply custom scale percent when in Custom mode
+    if (style == kRemoteViewStyleCustom) {
+      try {
+        _scale = await getSessionCustomScale(sessionId);
+      } catch (e, stack) {
+        print('Error in getSessionCustomScale: \$e\n\$stack');
```
fufesou: I think it should be `print('Error in getSessionCustomScale: $e\n$stack');`

flutter/lib/utils/scale.dart
```diff
@@ -0,0 +1,36 @@
...
+    if (dart.library.html) 'package:flutter_hbb/web/bridge.dart';
+import 'package:uuid/uuid.dart';
+
+/// Clamp custom scale percent to supported bounds.
+/// Keep this in sync with the slider's minimum in the desktop toolbar UI.
+int clampCustomScalePercent(int percent) {
```
fufesou, line 10: Why not use `percent.clamp(kScaleCustomMinPercent, kScaleCustomMaxPercent)`, please add comments if you have to use this implement.

fufesou (changes requested):

flutter/lib/models/model.dart
```diff
@@ -1824,12 +1833,25 @@ class CanvasModel with ChangeNotifier {
     _lastViewStyle = viewStyle;
     _scale = viewStyle.scale;
 
+    // Apply custom scale percent when in Custom mode
+    if (style == kRemoteViewStyleCustom) {
+      try {
+        _scale = await getSessionCustomScale(sessionId);
+      } catch (e, stack) {
+        print('Error in getSessionCustomScale: $e\n\$stack');
```
fufesou: You should also remove `\` before `$stack`.

fufesou (changes requested):

flutter/lib/models/model.dart
deblasis: sorry

fufesou (approved):

flutter/lib/desktop/widgets/remote_toolbar.dart
```diff
@@ -1245,6 +1309,290 @@ class _DisplayMenuState extends State<_DisplayMenu> {
...
+            radius: 4,
+            // Display the mapped percent for the current normalized value
+            displayValueForNormalized: (t) => _mapPosToPercent(t),
+          ),
+        ),
+        child: Slider(
+          value: _pos,
+          min: 0.0,
+          max: 1.0,
+          divisions: (_maxPercent - _minPercent).round(),
```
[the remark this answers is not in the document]
deblasis: Sorry Copilot but no...
Capping divisions at 100 would reduce precision from about 1% increments to about 10% increments, preventing users from setting common precise scale values like 125% or 150% that are essential for good UX IMHO

flutter/lib/desktop/widgets/remote_toolbar.dart
```diff
@@ -1245,6 +1309,290 @@ class _DisplayMenuState extends State<_DisplayMenu> {
...
```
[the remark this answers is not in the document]
deblasis: I have added a comment to motivate/clarify why the code is like this and might sound overkill. Let's see if it helps.

PR #13070, opened 2025-09-29
feat: Add IPv6 prefix-based rate limiting on login failures

mbacarella:
## Summary

Harden RustDesk against IPv6 originating brute force password attacks.

**Problem**: RustDesk has IP-based blacklisting which works well against IPv4, but not on IPv6. On IPv6, addresses are trivial to generate; a single client can select up to 2^64 unique addresses within one connection, making simple IP-based blacklisting ineffective.

**Solution**: Also track login failures by IPv6 prefixes (/64, /56, /48), and apply escalating thresholds (60, 80, and 100 attempts, respectively).

## Changes

- Add `get_ipv6_prefixes()` method to calculate network prefixes from IPv6 addresses
- Implement `check_failure_netblock()` for prefix-specific rate limiting
- Refactor `update_failure()` to handle both individual IPs and network prefixes  
- Add `ExceedIPv6PrefixAttempts` variant to `AlarmAuditType` enum
- Enhanced login failure tracking with network-aware logic

## Tests done

- [x] Test IPv6 addresses with different prefixes
- [x] Verify rate limiting triggers at correct thresholds
- [x] Test IPv4 addresses remain unaffected

## Security Impact

Frustrates IPv6 brute-force password attacks as a vector.

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -50,8 +50,10 @@ use serde_json::{json, value::Value};
 #[cfg(not(any(target_os = "android", target_os = "ios")))]
 use std::sync::atomic::Ordering;
 use std::{
+    net::Ipv6Addr,
     num::NonZeroI64,
     path::PathBuf,
+    str::FromStr,
     sync::{atomic::AtomicI64, mpsc as std_mpsc},
 };
 #[cfg(not(any(target_os = "android", target_os = "ios")))]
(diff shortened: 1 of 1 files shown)


mbacarella:
@fufesou implemented copilot's suggestions in a subsequent push

src/server/connection.rs
```diff
@@ -3173,35 +3175,134 @@ impl Connection {
...
+        let as_u128 = u128::from_be_bytes(ip.octets());
+
+        let p64 = Ipv6Addr::from(mask_u128(as_u128, 64).to_be_bytes()).to_string() + "/64";
+        let p56 = Ipv6Addr::from(mask_u128(as_u128, 56).to_be_bytes()).to_string() + "/56";
+        let p48 = Ipv6Addr::from(mask_u128(as_u128, 48).to_be_bytes()).to_string() + "/48";
+
+        Some((p64, p56, p48))
+    }
+
     fn update_failure(&self, (mut failure, time): ((i32, i32, i32), i32), remove: bool, i: usize) {
```
fufesou: `(mut failure, time)`, `mut` can be removed now.

src/server/connection.rs
```diff
@@ -3173,35 +3175,134 @@ impl Connection {
...
+            None
+        }
     }
 
     async fn check_failure(&mut self, i: usize) -> (((i32, i32, i32), i32), bool) {
+        let time = (get_time() / 60_000) as i32;
+
+        // IPv6 addresses are cheap to make so we check prefix/netblock as well
+        if let Some((p64, p56, p48)) = self.get_ipv6_prefixes() {
+            if let Some(res) = self.check_failure_ipv6_prefix(i, time, &p64, 64, 60).await {
```
fufesou, line 3287: > a single client can select up to 2^64 unique addresses within one connection

Is a threshold of 60 too lenient?
Sorry, I'm not sure, just a question.

fufesou (changes requested):
The code looks good to me. Though I have not tested yet.

I will test it and give feedback.

src/server/connection.rs
mbacarella: Yeah it's tricky. We have a lot of head-room since an attacker needs 2 billion-ish tries (2^36 * 6) to brute force 6-digit OTP space. So 60 could just as well be 1000. 

But I at least wanted some number >30, the limit for a single IP ban. I'm thinking of unknown unknowns. Maybe this is being used where a /64 is actually a LAN (perhaps some dinky W-ISP) where you wouldn't want others on the LAN to be able to cause you to get locked out.

fufesou (approved):
This PR is good.

**/64 blocking**: Tested.

**/56** & **/48 blocking**: Not fully tested, but the login failure counters for these prefixes are correctly.

southerndoc:
Glad to see this feature.  I think a /48 block would be great.  I have a static /48 if you need testers.

PR #13081, opened 2025-10-02
fix: Update Swedish translations that were missing

loako:
Translated most of the items that did not have a translation


loako (commented):
Translated most items that did not have a translation.

PR #13128, opened 2025-10-09
fix: CI, android build oom

fufesou:
https://github.com/rustdesk/rustdesk/actions/runs/18361667429/job/52311126057


rustdesk:
use https://github.com/rustdesk/rustdesk/commit/4e953291ede2edff251a59152690a41b2b38da2c because custom build server do not have big memory.

PR #13208, opened 2025-10-18
fix: video service, wait timeout, single frame fetched notifier

fufesou:
## Desc

Fix the issue where multiple video services share a single frame fetched notifier, causing 3-second wait ( the images are paused ).

https://github.com/rustdesk/rustdesk/blob/f2dc8e21a8dff9fc74ea55725d38f8561fb1fdf7/src/server/video_service.rs#L804

**Problem Scenario:**

```
A -> C
B -> C
```

`C` has two displays, and both `A` and `B` are viewing these displays.

Both video services send frames and wait for `A` and `B` to fetch them.

However, the frame-fetched notifications are randomly dispatched to video services:

- Video service **0** receives: conn1(A), conn1(A)
- Video service **1** receives: conn2(B), conn2(B)

This results in:
1. `fetched_conn_ids` = `[conn1]` in video service **0**
2. `fetched_conn_ids` = `[conn2]` in video service **1**

The condition `if fetched_conn_ids.len() >= frame_controller.send_conn_ids.len()` will never be true, because each video service only receives notifications for one connection, even though both services are waiting for two connections each.

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -776,7 +776,9 @@ impl Connection {
                 }
                 Some((instant, value)) = rx_video.recv() => {
                     if !conn.video_ack_required {
-                        video_service::notify_video_frame_fetched(id, Some(instant.into()));
+                        if let Some(message::Union::VideoFrame(vf)) = &value.union {
+                            video_service::notify_video_frame_fetched(vf.display as usize, id, Some(instant.into()));
+                        }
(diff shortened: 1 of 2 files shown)


rustdesk:
fix https://github.com/rustdesk/rustdesk/issues/13193

21pages (approved):

PR #13212, opened 2025-10-18
Create fi.rs

tomzza:
Please add Finnish language rustdesk.


rustdesk:
Thank you very much. You also need to modify `src/lang.rs` file, so that it takes effect.

tomzza:
I made pull requests add finnish lang.rs file.

PR #13242, opened 2025-10-21
Fix relay routing for devices after WebSocket is disabled

alonginwind:
Fix the bug where, after enabling WebSocket and then disabling it, devices that did not previously select 'force relay' still route through the relay.


rustdesk:
Yes, this can fix your problem, but it is not a good fix.

alonginwind:
OK, then I'll close this PR.

PR #13247, opened 2025-10-21
Edge scrolling

logiclrd:
This PR adds a new scroll style: Edge scrolling.

https://github.com/user-attachments/assets/eac377e9-1848-48e8-b596-9a6a29bd315d

Edge scrolling mode behaves like Scrollbar mode, with the added function that when the mouse cursor approaches the edge of the viewport, the view automatically scrolls in that direction if possible. This allows for convenient scrolling around the view while maintaining stability in the view position, as compared to the Auto scrolling mode in which every mouse cursor position in the view corresponds to a unique scroll offset.

Closes: #3053

### "Bump Mouse"

For edge scrolling to have a natural "feel" to it, the mouse cursor must "bump" against the edge. That is, when scrolling occurs as a result of edge scrolling, the mouse cursor should stay anchored to the point in the underlying view that it is overtop of before the scroll. In order to achieve this, it is necessary to force the position of the mouse cursor to change, called "warping" the mouse cursor in some toolkits.

This functionality is not a part of the base feature set exposed by Flutter, so it must be implemented by means of a Platform Channel. I noticed that the code already has a Platform Channel, but only for the MacOS platform. This PR co-opts that platform, so that there aren't _two_ platform channels when running on MacOS. The name of the channel within its namespace is changed from `macos` to simply `host`.

I did my best to implement the functionality for all three platform layers. I am currently developing on a Linux host, and was able to test and verify that the Gdk-based implementation of `bumpMouse` works, at least on my system running KDE. I have access to a Windows host and should be able to verify that implementation later, but it is as yet untested.

_I do not have access to an OS X development environment and have no way to test the `macos` implementation._ I wouldn't be surprised if it doesn't even build. I have no clue what I'm doing with Swift code :slightly_smiling_face: 

The Platform Channel can only be accessed from the root isolate, so this PR also adds a message type to the existing window event channel system so that the isolates hosting remote desktop windows can direct their `bumpMouse` actions properly.

### `ScrollStyle`

I noticed a number of places independently implementing conversion of `ScrollStyle` values, in one place to/from JSON and in another to/from an arbitrary string form for method channel encoding. Before proceeding, I centralized the formatting logic.

### Configuration

I have at this point not investigated the possibility of making the width of the edge that triggers scrolling configurable. It can't be too small, because if the mouse is moved quickly enough, it can skip it entirely between events and escape the window. I'm not convinced that this is a bug. In my debugging, the value of 120 that is presently hardcoded worked pretty well, for me on my trackpad at least.

### Automated Testing

There are some things conducive to automated testing, and I haven't written any of those yet at this point. With a bit of abstraction, the edge scrolling algorithm itself wouldn't be a bad thing to have tests for.


logiclrd:
Outstanding issue: If the cursor strays into the client area while the window frame is being dragged in a resize operation, then the mouse movement gets processed as an edge scroll action and `bumpMouse` moves the cursor, which causes a feedback effect of window resize effects that rapidly squishes the window down to minimum size. To prevent this, I need edge scrolling, or at the very least `bumpMouse` actions, to be suppressed during resize drag operations. I'm not sure if Flutter itself exposes this information.

logiclrd:
One possible approach: hook `didChangeMetrics` and disable edge scrolling briefly (200ms or something). Pros: Implemented entirely within the abstraction. Cons: Possibly imprecise? Edge cases?

logiclrd:
Okay that fixes the resize glitch. :slightly_smiling_face:

rustdesk:
https://github.com/rustdesk/rustdesk/pull/13247/files#diff-5454e76e9f6df3858c389abe87ce8a73d054902a802d0c1ed324be29b595a399 has grammer error. building failed.

logiclrd:
> https://github.com/rustdesk/rustdesk/pull/13247/files#diff-5454e76e9f6df3858c389abe87ce8a73d054902a802d0c1ed324be29b595a399 has grammer error. building failed.

Shucks. I don't know the language and I don't have any local feedback on it. I got feedback from the first round of automated builds and pushed up some fixes. Let's see if I can't figure out further fixes from the latest build. Ultimately, though, I'm going to need the help of someone who has an actual OS X dev setup for this. I have no way to build or test it myself. 🙁

rustdesk:
ChatGPT? :(

fufesou:
This patch fixes the build error on macOS, but it introduces unexpected behavior: the mouse cursor jumps to the edge of the screen when it moves to the edge of the remote window.

I need to test more on Linux.

```
--- a/flutter/macos/Runner/MainFlutterWindow.swift
+++ b/flutter/macos/Runner/MainFlutterWindow.swift
@@ -112,12 +112,14 @@ class MainFlutterWindow: NSWindow {
                     let dx = (arg["dx"] as? Int) ?? 0;
                     let dy = (arg["dy"] as? Int) ?? 0;
 
-                    if let mouseLoc = self.rustDeskViewController?.mouseLocation {
-                        mouseLoc.y = NSHeight(NSScreen.screens()![0].frame) - mouseLoc.y;
+                    if var mouseLoc = self.rustDeskViewController?.mouseLocation {
+                        if let screenFrame = NSScreen.screens.first?.frame {
+                            mouseLoc.y = NSHeight(screenFrame) - mouseLoc.y
 
-                        let newLoc = CGPoint(x: mouseLoc.x + dx, y: mouseLoc.y + dy);
+                            let newLoc = CGPoint(x: mouseLoc.x + CGFloat(dx), y: mouseLoc.
y + CGFloat(dy))
 
-                        CGDisplayMoveCursorToPoint(0, newLoc);
+                            CGDisplayMoveCursorToPoint(CGMainDisplayID(), newLoc)
+                        }
                     }
 
                     result(nil)
```

flutter/windows/runner/flutter_window.cpp
```diff
@@ -1,13 +1,24 @@
...
 #include <texture_rgba_renderer/texture_rgba_renderer_plugin_c_api.h>
 #include <flutter_gpu_texture_renderer/flutter_gpu_texture_renderer_plugin_c_api.h>
 
 #include "flutter/generated_plugin_registrant.h"
 
+#include <flutter/event_channel.h>
+#include <flutter/event_sink.h>
+#include <flutter/event_stream_handler_functions.h>
+#include <flutter/method_channel.h>
+#include <flutter-standard_method_codec.h>
```
fufesou: Should be `#include <flutter/standard_method_codec.h>`

flutter/windows/runner/flutter_window.cpp
```diff
@@ -29,6 +40,48 @@ bool FlutterWindow::OnCreate() {
...
+    flutter_controller_->engine()->messenger(),
+    "org.rustdesk.rustdesk/host",
+    &flutter::StandardMethodCodec::GetInstance());
+
+  channel.SetMethodCallHandler(
+    [](const flutter::MethodCall<>& call, std::unique_ptr<flutter::MethodResult<>> result) {
+      if (call.method_name() == "bumpMouse") {
+        auto arguments = call.arguments();
+
+        int dx = 0, dy = 0
```
fufesou: `;` is missing at the end of the line.

fufesou (changes requested):

flutter/windows/runner/flutter_window.cpp
logiclrd: Whoops, yep. Late last night I started poking on my Windows machine, and it's behaving oddly right now. I have Visual Studio 2026 installed (version 18), and that's listed as one of the recognized generators for CMake, but when I `flutter run`, it just immediately bombs out telling me that it can't find Visual Studio 16 2019. But `flutter doctor` finds it correctly and the only error it lists is that I don't have the Android SDK installed. Gotta be some small thing I'm missing somewhere...

src/lang/template.rs
```diff
@@ -122,6 +122,7 @@ pub static ref T: std::collections::HashMap<&'static str, &'static str> =
         ("Stretch", ""),
         ("Scrollbar", ""),
         ("ScrollAuto", ""),
+        ("ScrollEdge", ""),
```
fufesou: Please append the new translation to this file.

And you can run `python res/lang.py` to generate the placeholders for the other language files.

fufesou (changes requested):

fufesou:
Hi, the scrolling speed behaves differently between Linux and Windows.
On Linux, it's too fast for precise adjustments but beneficial for selecting entire lines.
On Windows, the slower speed is ideal for fine-tuning, though less efficient for large selections.

I'm not sure if the following suggestion is appropriate.

### three-stage speed

https://github.com/user-attachments/assets/ae26a00a-b997-4695-963c-d34919513b29

To improve the user experience, is it ok to implement a variable scrolling speed that slows down near the trigger boundary, resumes normal speed at the boundary, and accelerates after crossing it—similar to the behavior in the reference video?

image

Maybe, using a debouncer to determine current cursor area is possible. You can move the cursor the to near inner edge of the area once it stops moving.

Maybe this is a bit complicated and not very practical. Current implementation is good enough.

fufesou:
This function on macOS does not work.

https://github.com/user-attachments/assets/52817413-829d-4bc4-ac8d-b56016aa28b8

logiclrd:
> On Linux, it's too fast for precise adjustments but beneficial for selecting entire lines.

It sounds like `bumpMouse` isn't working in your configuration. The scrolling speed should be limited to _actual movement of the mouse_, because once the mouse enters the edge scrolling region, it gets warped out of it immediately. What you're describing sounds like the mouse fails to warp out of the edge scrolling region, so each subsequent mouse event repeats the scrolling. In my original video in the PR description, if you watch closely you'll see that each time the mouse bumps into the edge, it gets pushed back -- in essence, the mouse is scrolled _with_ the underlying view.

logiclrd:
> This function on macOS does not work.

It looks like the snippet I found for retrieving the current mouse position isn't working.

I found a discussion elsewhere on the Internet that said that you can just poke into `self.view.window.mouseLocationOutsideOfEventStream` in the context of the view controller. I copy/pasted their code.

The core of the issue comes down to the fact that the mouse position is coming from one place and the functionality for warping the mouse to a new position is in a different place, and these subsystems aren't working with the same coordinate space. But, I don't know the system, and I don't have the means to experiment.

In short: The code expects that `flutterViewController.mouseLocation` will return the absolute position of the cursor within the coordinate system that encapsulates all screens. It then has to contend, from the things I read, with the fact that Core Graphics for some reason uses an inverted coordinate system with the Y axis pointing upward and the origin in the lower-left of the primary monitor, so it converts the coordinate into that space before applying the delta. Maybe that conversion isn't actually needed??

The function `CGDisplayMoveCursorToPoint` is clearly documented to take a point in the absolute coordinate space of the displays, anchored at the primary screen's origin. So, logically, it must be that the code that establishes the current cursor position based on `mouseLocationOutsideOfEventStream` isn't getting the expected value.

<hr />

And I've just read the documentation for `mouseLocationOutsideOfEventStream` and it is explicitly documented as being relative to the receiving window's coordinates. There is a static property on `NSEvent` called `mouseLocation` documented as returning the position in screen coordinates. So, let's give that a go. :slightly_smiling_face: 

Google's AI result suggests also that the Y coordinate conversion isn't needed. I'll push up a commit that uses `NSEvent.mouseLocation` with no conversion. Crossing fingers. :crossed_fingers:

logiclrd:
Great news 🙂 I figured out by local environment problems (turns out it's just that Flutter itself doesn't support Visual Studio 2026 yet, but I managed to work around that), and the Windows code builds and works properly 🙂

logiclrd:
@fufesou Please give the latest code a try when it's convenient and let me know if it behaves better on OS X. 🙂

fufesou:
https://github.com/user-attachments/assets/4b702fc7-41f2-4add-aa27-cc7fa78f5da1

1. Cursor still jumps.
2. Resizing window is broken.

logiclrd:
Thanks 🙂 It looks like maybe the coordinate systems _are_ mismatched. Also, the suppression of edge scrolling during resizing depends on it getting continuous window resized events during the resize operation. It's one possibility that that isn't happening on OS X, maybe...

fufesou:
https://github.com/user-attachments/assets/3a5fe4c1-2032-4c70-ab8d-10684b30b12e

The cursor does not jump out of the window now.

There are 3 issues:
1. It's hard to scroll.
2. The cursor may jump when I move it into the window. The first time I tried to resize the window in the video.
3. Resizing the window is still problematic.

logiclrd:
> 1. It's hard to scroll.

That's weird. The way it's written, you should basically accomplish exactly the mouse movement you intended, with respect to the remote window, it's just that for some of the movement, the underlying pixels moved instead of the mouse cursor. It's not working that way on OS X? I wonder if OS X has a brief suppression of mouse movement after a call to `CGDisplayMoveCursorToPoint`...

Okay I just looked this up and it's _exactly_ what's happening. There's a function that can be used to control it: `CGSetLocalEventsSuppressionInterval`. Apparently, though, it's deprecated:

https://stackoverflow.com/questions/10196603/using-cgeventsourcesetlocaleventssuppressioninterval-instead-of-the-deprecated

The answers there contain two work-arounds:

1. Apparently, the function was deprecated in favour of a more specialized version, `CGEventSourceSetLocalEventsSuppressionInterval`, and it's possible to use this _but_ you need an `EventSource`. One can be transiently created for the purpose of calling the function.

2. But, also, there's a function `CGAssociateMouseAndMouseCursorPosition` that allows you to specify that mouse events should deliver x/y deltas _but not actually affect the mouse cursor position_. You call it with `false` to get this disconnected behaviour and `true` to reconnect it. The accepted answer states that if you call this with `true`, even if it is already `true`, any ongoing suppression interval is cancelled.

So let's give that a go :slightly_smiling_face: 

> 2. The cursor may jump when I move it into the window. The first time I tried to resize the window in the video.

Hmm, this is definitely expected with the design; the pixel at the very edge of the border counts as scrollable, and it sees it as quite a considerable scroll, which then warps the cursor in so that it's not over the edge any more. And I can see how if you accidentally hit that while trying to resize the window, it'd be a pain. I'll add a small buffer region around the edge that doesn't activate edge scrolling.

> 3. Resizing the window is still problematic.

I'm going to annotate the `window resized` event. There _should_ be a flurry of them in real-time during resize activity, but the behaviour your describing sounds like on this platform, maybe it's only sending one `window resized` event at the end. (Maybe there's a separate `window resizing` event that needs to be handled as well?)

logiclrd:
@fufesou Alright, assuming I haven't made any silly mistakes, this next build should address points 1 and 2 and add debug output to help figure out what's going on with point 3.

It occurs to me that when the resize speed is low, the mitigation for point 2 will actually incidentally fix the resizing behaviour as well. So unless you're resizing the window quickly, you might actually see normal resize behaviour. But what's of interest is whether it's receiving continuous `onWindowResize` events during the resize action or not.

Thanks so much for trying out these changes on OS X and helping me get them polished up :smile:

logiclrd:
Huh :stuck_out_tongue: The documentation for `CGAssociateMouseAndMouseCursorPosition` says to pass `true` in, but apparently that's the wrong data type.

fufesou:
Points 1 and 2 are fixed 👍 

Point 3, it is only a bit problematic now.
Resizing just jump sometimes, not often.

In the video, the most recent print are shown at the bottom.

https://github.com/user-attachments/assets/182d327b-4d99-4e81-9245-4ce13bbb1789

logiclrd:
Mm, okay, so you're not seeing `RECEIVED WINDOW RESIZE EVENT` over and over in quick succession _during the drag_:

image

Once that's figured out I think it'll work the same on all platforms. :slightly_smiling_face:

fufesou:
> so you're not seeing `RECEIVED WINDOW RESIZE EVENT` over and over in quick succession during the drag

Yes.

`RECEIVED WINDOW RESIZE EVENT` is printed only when I stop resizing.

logiclrd:
Okay I think I may have found the solution. Apparently, `onWindowResized` is explicitly documented to behave the way we're seeing on OS X.

> [onWindowResized](https://pub.dev/documentation/window_manager/latest/window_manager/WindowListener/onWindowResized.html)() → void
> Emitted once when the window has finished being resized.

On the other platforms, the behaviour is actually contrary to the documentation.

But, there is another event method, `onWindowResize` (no `d` at the end):

> [onWindowResize](https://pub.dev/documentation/window_manager/latest/window_manager/WindowListener/onWindowResize.html)() → void
> Emitted after the window has been resized.

If I'm understanding correctly, all that's needed here is to use `onWindowResize` instead of `onWindowResized`.

logiclrd:
Hmm, another thing: the Android and iOS platforms, they don't seem to have an API that could implement `bumpMouse`. People aren't used to using a mouse for input on these devices, but you can in fact do so; plugging a mouse into an Android or iOS device (even a cell phone) gives you a working mouse cursor. Also, I think tap events might trigger edge scrolling functionality as well -- but without a working `bumpMouse`, it won't behave well. Should perhaps hide the `ScrollEdge` option on these platforms??

EDIT: I just installed and ran the RustDesk Android app and found that the scrolling mode is _already_ hidden there. If it's that way on iOS too, then this is a non-issue :slightly_smiling_face:

logiclrd:
I actually can't figure out _how_ the option is suppressed when running the Android app. :thinking:

I notice that the Android app is version 1.4.1, which precedes the introduction of the "Custom" fixed-scaling view style, but the remote toolbar in that tag has the widgets for scrollauto vs. scrollbar, and the code _seems_ to be set up so that the scroll style is only hidden when the view style is set to "Adaptive". But in practice, the scroll style seems to always be hidden.

The reason I've been trying to locate it is in order to have confidence that the edge scrolling functionality won't cause a bad experience on Android/iOS, and right now that's based on an assumption.

fufesou:
Hi, resizing works perfectly on macOS now.

But `_saveFrame();` may still need to be called in `onWindowResized()`, not in `onWindowResize()`.

And please remove some unnecessary prints, as too much print can cause confusion.

logiclrd:
> Hi, resizing works perfectly on macOS now.

Yey! 🎉 

> But `_saveFrame();` may still need to be called in `onWindowResized()`, not in `onWindowResize()`.

No worries about `_saveFrame` -- it's already expecting the chatty calls because that's what Linux and Windows do for `onWindowResized` even though they're not supposed to. The switch from `onWindowResized` to `onWindowResize` only actually had an effect on OS X. The implementation of `_saveFrame` debounces the calls, and there's a mechanism to force an immediate save on window close already in place.

> And please remove some unnecessary prints, as too much print can cause confusion.

Oh, yes, they were only ever meant to be temporary. 🙂

logiclrd:
Apparently there _is_ a way to get the mouse position within the Core Graphics API. It doesn't seem to be documented, but a) `CGEventGetLocation` returns the coordinates of a mouse event, and b) `CGEventCreate(NULL)`, documented as returning "A new event to be filled in", actually returns an event _that has the current mouse cursor position already filled in_. Done this way, I believe there's no mismatch in the coordinate system.

fufesou:
This PR works perfectly on Windows and macOS now.

1. 
> > On Linux, it's too fast for precise adjustments but beneficial for selecting entire lines.
> 
> It sounds like `bumpMouse` isn't working in your configuration. The scrolling speed should be limited to _actual movement of the mouse_, because once the mouse enters the edge scrolling region, it gets warped out of it immediately. What you're describing sounds like the mouse fails to warp out of the edge scrolling region, so each subsequent mouse event repeats the scrolling. In my original video in the PR description, if you watch closely you'll see that each time the mouse bumps into the edge, it gets pushed back -- in essence, the mouse is scrolled _with_ the underlying view.

Yes.
**X11** works fine, **Wayland** has this issue.

Maybe https://github.com/wxWidgets/wxWidgets/pull/25721 helps, but it's a little complicated.

2. I can also see the print on Linux sometimes.
```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Null check operator used on a null value
#0      RemotePage.ffi (package:flutter_hbb/desktop/pages/remote_page.dart:64:35)
#1      _DesktopTabState.onWindowResize (package:flutter_hbb/desktop/widgets/tabbar_widget.dart:429:46)
#2      DesktopMultiWindow._windowMethodCallHandler (package:desktop_multi_window/desktop_multi_window.dart:129:27)
#3      DesktopMultiWindow.setMethodHandler.<anonymous closure> (package:desktop_multi_window/desktop_multi_window.dart:78:9)
#4      MethodChannel._handleAsMethodCall (package:flutter/src/services/platform_channel.dart:571:55)
#5      MethodChannel.setMethodCallHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:564:34)
#6      _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:618:35)
#7      _invoke2 (dart:ui/hooks.dart:344:13)
#8      _ChannelCallbackRecord.invoke (dart:ui/channel_buffers.dart:45:5)
#9      _Channel.push (dart:ui/channel_buffers.dart:135:31)
#10     ChannelBuffers.push (dart:ui/channel_buffers.dart:343:17)
#11     PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:750:22)
#12     _dispatchPlatformMessage (dart:ui/hooks.dart:257:31)
```


3. There are also too many prints like
```
flutter: [Main] call bump_mouse with args {dx: 55, dy: 0} from window 1
flutter: [Window Main] bump mouse cursor by (55, 0)
flutter: [Main] call bump_mouse with args {dx: 63, dy: 0} from window 1
flutter: [Window Main] bump mouse cursor by (63, 0)
```

Perhaps we can filter these outputs by the event name, as we may not need to print these outputs anymore.

logiclrd:
Ugh. Wayland. 😛 I have had some issues in the past with their philosophies. Looks like this bumps into them again.

It seems like Wayland's initial stance has been that, fundamentally, the ability to warp the mouse pointer is both a security risk and just automatically bad UI/UX, and Wayland deliberately did not support it. Instead, they provided a couple of protocols called `lock_pointer` and `confine_pointer`, where `lock_pointer` is in essence just a special case of `confine_pointer`. `confine_pointer` allows the application to specify a bounding rectangle and the pointer is not allowed to leave that rectangle. `lock_pointer` is basically just that with a rectangle of size (1, 1). It is deliberately left up to the compositor implementation to decide whether the cursor is forced into the region immediately or only after the next mouse event. Crucially, though, with `lock_pointer` it is possible to set a "hint" location to which the cursor should be warped when _unlocking_ the pointer

From the issue to which you linked, it seems like they did finally relent and accede to a `pointer_warp_v1` protocol. But, this was done only recently (version 1.45 of Wayland Protocols came out in June), and so far only a single compositor has actually implemented it.

So it seems to me that:

1. We need to try to use `pointer_warp_v1`.
2. If we can't, we need to try to use `lock_pointer`.
3. If neither of those works, we need to push the failure back up to the UI and forcibly switch the scrolling style away from `scrolledge`, because we don't have the necessary support for it to work properly.

logiclrd:
I have no idea where to even start writing code to interact directly with Wayland protocols. Every reference I've found thus far assumes that I can e.g. just `#include "wayland-pointer-constraints-unstable-v1.h"` and start calling functions, but I don't have that file or any other Wayland-related header and don't know how to get it. I'll try some other approaches tomorrow. 🙂

logiclrd:
Okay, I've found about about `wayland-scanner` and how libwayland doesn't distribute headers but rather XML definitions of Wayland protocols, and then your build incorporates a step translating the XML definitions into headers and glue code as needed. How does that interact with the automated RustDesk builds? Do they have access to `wayland-scanner`? I guess maybe new associated dependencies need to be added to the build agent initialization?? I saw reference to a Rust version of the scanner, `wayland-rs`, but it doesn't seem to be involved presently in Wayland-specific code in RustDesk.

fufesou:
> So it seems to me that:
> 
> 1. We need to try to use `pointer_warp_v1`.
> 2. If we can't, we need to try to use `lock_pointer`.
> 3. If neither of those works, we need to push the failure back up to the UI and forcibly switch the scrolling style away from `scrolledge`, because we don't have the necessary support for it to work properly.

Maybe we only need to use `pointer_warp_v1`.

Because it works on Wayland, but it's too fast for precise adjustments.


> How does that interact with the automated RustDesk builds? Do they have access to `wayland-scanner`?

I have no idea too.

Maybe there is no need to call wayland APIs in rust, you only need to change the `cc` file.

You may need to modify the CI task on Linux. https://github.com/rustdesk/rustdesk/blob/master/.github/workflows/flutter-build.yml 

Run `wayland-scanner` with the XML file, and then compile the project.

But there may be some version errors.
RustDesk still uses `ubuntu 18.04` to run CI https://github.com/rustdesk/rustdesk/blob/e66d2facd461a57ad24d77c50cf1515c4eee025a/.github/workflows/flutter-build.yml#L1289

logiclrd:
> Maybe we only need to use `pointer_warp_v1`.
> 
> Because it works on Wayland, but it's too fast for precise adjustments.

Mm, except that it's not working as designed.

The design is that whatever mouse movement you do is captured virtually, so that if the cursor bumps into the edge scrolling region, then the underlying canvas _including the position of the mouse cursor_ is scrolled accordingly. This can only work if Bump Mouse is functional.

I'm working on code that detects that Bump Mouse wasn't able to do its thing, and if so, I'm thinking maybe instead of just disabling edge scrolling entirely, it can fall back on a mode where the scroll velocity is a constant, rather than it being based on how far the mouse has moved into the edge scrolling region.

But there is literally only one compositor right now that supports `pointer_warp_v1`.

Which also means that to test this, it will be necessary to configure a Linux system using Wayland compositor "Jay".

> Maybe there is no need to call wayland APIs in rust, you only need to change the `cc` file.

Yep, that's my plan. This will be embedded in the `linux` platform-specific code underlying the Flutter app. Basically, where the current `bump_mouse` is implemented but only works on X11, that implementation will be broadened to support Wayland.

But, done properly, it's still necessary to integrate `wayland_scanner` into the build process. With most APIs, you just locate their `.h` files and `#include` them, but Wayland _doesn't have `.h` files_. At least, if I'm understanding correctly, it has XML protocol definitions and a code generator, and you're expected to autogenerate transient glue code within the build. Fun! 😁 

> But there may be some version errors.
> RustDesk still uses `ubuntu 18.04` to run CI

More fun 😁 Well, one thing at a time. First step: See it working on my local machine.

logiclrd:
I was doing some reading, and `lock_pointer` might or might not actually work for this. The documentation explicitly says that the call _doesn't actually obtain the lock_, just declares that you _want_ the lock, and the compositor decides when and whether to actually do a lock at some indeterminate point in the future, and the app doesn't even get notified as to whether it happened or not. :-/

I wonder if perhaps it would make sense to just give up on Bump Mouse with Wayland, and fall back on the constant scroll speed idea instead...

fufesou:
> I wonder if perhaps it would make sense to just give up on Bump Mouse with Wayland, and fall back on the constant scroll speed idea instead...

Agree.

If it is still not feasible, perhaps we can just hide this option on Wayland and add a comment explaining why.

logiclrd:
Okay, I've reached a point where the build process is reliably running `wayland-scanner` to generate the header files and glue code, and now I'm bumping into a more fundamental problem: It looks like Gtk/GDK 3 simply doesn't expose the necessary Wayland functionality. Gtk 4 does, but both of the Wayland protocols need, for instance, a pointer to the `wl_surface` that provides the context for the operation, and there's a method `gdk_wayland_surface_get_wl_surface` that provides this, but as far as I can tell, that method doesn't exist in the GDK 3 header files, nor for that matter does the entire concept of a `GdkSurface *` in the first place.

So, if I've got the right end of the stick, implementing the Wayland interaction will/would require an upgrade to Gtk 4. I'm going to assume that that's solidly out of scope at this point :grin: 

As a result, I think all the Wayland code I've been working on basically has nowhere to go right now, and the edge scrolling functionality will simply _always_ need to fall back on the constant-speed no-bump strategy when running under Wayland.

Is RustDesk likely to migrate to Gtk 4 at some point in the future? If so, maybe there's some way that the Wayland code using the pointer protocols could be stashed somewhere so it isn't lost. It's kind of ridiculously longer than any of the other platforms. :-P Windows is like 2 lines, OS X and Linux are like 15 lines, Wayland requires you to extract a bunch of pointers, to listen for seat capability messages and to listen for registry updates to find your interfaces before you can even start writing the code, and as a result, `bump_mouse_wayland.cc` is currently 259 lines long. :grin: Would be a pity to lose that if it _could_ have future value.

logiclrd:
I guess it's ultimately up to Flutter. I found this:

https://github.com/flutter/flutter/issues/94804

Looks like they see the value in it and are tinkering with the idea but are pretty far from even a proof of concept. So that's that for now, it would seem.

logiclrd:
The Wayland work-in-progress: https://github.com/logiclrd/rustdesk/pull/1

logiclrd:
To explain a little: I had made changes to support X11 and Wayland, and as part of those changes I reworked the Linux platform back-end, refactoring bump_mouse out into multiple files to allow the code to be kept separated. After determining that the Wayland implementation is a no-go today, I split the Git commit apart, keeping the end result but separating the Wayland-specific stuff into an isolated commit. I also split up other unrelated changes into their own commits.

Everything but the Wayland commit is now on this branch, and the https://github.com/logiclrd/rustdesk/pull/1 PR in my fork shows the Wayland changes on top of these changes.

logiclrd:
Should I rebase to resolve the `ca.rs` and `ja.rs` merge conflicts?

fufesou:
> Is RustDesk likely to migrate to Gtk 4 at some point in the future?
> Would be a pity to lose that if it could have future value.
> I guess it's ultimately up to Flutter.

Yes. It's up to Flutter.

> Should I rebase to resolve the `ca.rs` and `ja.rs` merge conflicts?

Yes, please.

flutter/pubspec.yaml
```diff
@@ -109,6 +109,7 @@ dependencies:
   xterm: 4.0.0
   sqflite: 2.2.0
   google_fonts: ^6.2.1
+  vector_math: ^2.2.0
```
fufesou: Build error on my machine (flutter 3.24.5, the version of github CI), maybe you should downgrade to `vector_math: 2.1.4`.

```
Note: vector_math is pinned to version 2.1.4 by flutter_localizations from the flutter SDK.
See https://dart.dev/go/sdk-version-pinning for details.


Because flutter_hbb depends on flutter_localizations from sdk which depends on
  vector_math 2.1.4, vector_math 2.1.4 is required.
So, because flutter_hbb depends on vector_math ^2.2.0, version solving failed.
```

fufesou (changes requested):

flutter/lib/models/model.dart
```diff
@@ -2052,6 +2121,119 @@ class CanvasModel with ChangeNotifier {
...
+
+        // Don't keep hammering BumpMouse if it's not working.
+        _bumpMouseIsWorking = false;
+
+        // Keep scrolling as long as the user is overtop of an edge. Each time the
+        // timer fires sets a new timer.
+        setEdgeScrollTimer();
+
+        // When autorepeating, scale down the extent of the scroll, otherwise it'll be way too fast.
+        encroachment *= 0.2;
```
fufesou: My testing shows that `0.2` is still fast.

`0.05` seems acceptable to me. What do you think?

fufesou (changes requested):

flutter/lib/models/model.dart
logiclrd: Mm, interesting. On my end, 0.2 felt exactly right. Ultimately, it will depend on how precise the timer is. The code currently requests a 10ms interval. Maybe on my system, it is in practice getting a longer delay?? Weird that it would be different.

I'm pretty sure 0.05 would be glacially slow on my end... Let me do some testing. Maybe an algorithm that automatically tunes based on feedback from the system high-precision timer is needed.

flutter/pubspec.yaml
logiclrd: Ahh, I am indeed running a newer Flutter SDK and have a handful of local updates to accomodate that.

flutter/lib/models/model.dart
logiclrd: I just tested this and found that, indeed, 0.05 feels really slow:

https://github.com/user-attachments/assets/0dd45f77-4577-495f-ad27-3b1ac973307d

Compare with 0.2:

https://github.com/user-attachments/assets/8b38d1ee-4f00-422a-b7de-f5f377dc68c7

But in both videos, in the final bit, I tested what happens when the mouse is actually moving vs. just the repeat timer, and the rate is considerably faster when mouse events are being received. The rate of mouse events is higher than the rate of timer ticks with a 10ms interval. So, I'm going to see if I can't come up with some math that adapts the rate to match so that there's no visual speed change with the mouse moving. The speed with the mouse moving is indeed much faster, so with this in place, a scale like 0.05 might be better.

logiclrd: Or, even better, make it so that mouse events don't trigger movement in the fallback. It's _entirely_ driven by the timer. I've just stumbled across the `Ticker` class as well, which ought to drive the timer with better regularity.

logiclrd: Much smoother :-)

https://github.com/user-attachments/assets/f752530e-a8f0-4aed-9a61-e94abe50121e

fufesou:
The speed on Wayland is acceptable to me now.

Please also address Point 2 and Point 3 in this  comment https://github.com/rustdesk/rustdesk/pull/13247#issuecomment-3447925276

flutter/lib/models/model.dart
```diff
@@ -1789,6 +1833,54 @@ class ViewStyle {
...
+
+      _owner.performEdgeScroll(delta);
+
+      _lastTotalElapsed = totalElapsed;
+    }
+  }
+
+  start() {
+    if (!_ticker.isActive) {
+      debugPrint("RESTARTING TICKER");
```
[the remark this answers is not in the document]
logiclrd: Oops

flutter/lib/models/model.dart
```diff
@@ -2052,6 +2170,126 @@ class CanvasModel with ChangeNotifier {
...
+
+    var scrollPixel = Vector2(
+      _horizontal.position.pixels,
+      _vertical.position.pixels);
+
+    final max = Vector2(
+      _horizontal.position.maxScrollExtent,
+      _vertical.position.maxScrollExtent);
+
+    encroachment.clamp(-scrollPixel, max - scrollPixel);
```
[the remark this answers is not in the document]
logiclrd: You'd expect that to be the case, but:

image

flutter/lib/models/model.dart
```diff
@@ -2052,6 +2170,126 @@ class CanvasModel with ChangeNotifier {
...
+      _horizontal.position.maxScrollExtent,
+      _vertical.position.maxScrollExtent);
+
+    var scrollPixel = Vector2(
+      _horizontal.position.pixels,
+      _vertical.position.pixels);
+
+    scrollPixel += delta;
+
+    scrollPixel.clamp(Vector2.zero(), max);
```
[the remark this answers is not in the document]
logiclrd: (incorrect in the same way)

fufesou:
Hi, there may be another cursor jump issue.

https://github.com/user-attachments/assets/e2ca7946-0d43-480a-9caf-262766495341

1. The cursor jumps when I move out the cursor from the left edge and move in from the right side.
2. The scrollbar (image) jumps.


Maybe, edge scrolling should only trigger when the cursor moves from inside the window to outside, not when it enters from outside?

logiclrd:
> The speed on Wayland is acceptable to me now.
> 
> Please also address Point 2 and Point 3 in this comment [#13247 (comment)](https://github.com/rustdesk/rustdesk/pull/13247#issuecomment-3447925276)

Regarding the debug output, it's specific to MethodChannels. There are two involved, and they each log every call going across the boundary. In the case of `PlatformChannel`, it would be trivial to make `bump_mouse` not log because each stub does its own log output. In the case of the root isolate channel in `DesktopHomePage`'s state, it's a common logging call for all calls across the channel:

```
    rustDeskWinManager.setMethodHandler((call, fromWindowId) async {
      debugPrint(
          "[Main] call ${call.method} with args ${call.arguments} from window $fromWindowId");
      ...
```

Filtering this would add one layer of complexity to code that's only there in a debugging context. I guess the condition could itself be wrapped in `if (kDebugMode)`...

In both cases, this output only appears for debug builds. Should I still suppress it?

fufesou:
> In both cases, this output only appears for debug builds. Should I still suppress it?

Yes.

There are too many prints here.
The ones for this call are no longer needed now that the logic has been debugged.

logiclrd:
Oh, interesting, I was wrong.

> The [debugPrint](https://api.flutter.dev/flutter/rendering/debugPrint.html) function logs to console even in [release mode](https://docs.flutter.dev/testing/build-modes#release). As per convention, calls to [debugPrint](https://api.flutter.dev/flutter/rendering/debugPrint.html) should be within a debug mode check or an assert:

Do we want to keep the debug output in the inter-isolate channel defined in `desktop_home_page.dart` for other calls even in release builds, or would it be appropriate for _all_ of that output to be debug-mode-only?

fufesou:
> The [debugPrint](https://api.flutter.dev/flutter/rendering/debugPrint.html) function logs to console even in [release mode](https://docs.flutter.dev/testing/build-modes#release). As per convention, calls to [debugPrint](https://api.flutter.dev/flutter/rendering/debugPrint.html) should be within a debug mode check or an assert:
> 

Yes, it's interesting.

> Do we want to keep the debug output in the inter-isolate channel defined in `desktop_home_page.dart` for other calls even in release builds, or would it be appropriate for _all_ of that output to be debug-mode-only?

No need to change the other prints.

The bump cursor logs should be filtered due to their excessive frequency, which leads to confusion.


These prints are not visible to the users unless they run RustDesk via the command line.

logiclrd:
> > Do we want to keep the debug output in the inter-isolate channel defined in `desktop_home_page.dart` for other calls even in release builds, or would it be appropriate for _all_ of that output to be debug-mode-only?
> 
> No need to change the other prints.

There's only one call for all methods, though, so it's not a matter of removing the one we don't want but of actively checking whether each `debugPrint` call should be done. If that code were placed inside `if (kDebugMode)` then that additional filtering logic wouldn't need to run in a release situation at all -- a situation where the standard output stream is never expected to be visible anyway.

logiclrd:
This is what I have in mind:

image

But if you'd rather the `debugPrint` be kept for release builds, just not for `kWindowBumpMouse` calls, then I'll do that. 🙂

flutter/lib/models/model.dart
```diff
@@ -2052,6 +2169,126 @@ class CanvasModel with ChangeNotifier {
...
+        dyOffset = y - edgeThickness;
+      } else if ((y >= size.height - edgeThickness)) {
+        dyOffset = y - (size.height - edgeThickness);
+      }
+    }
+
+    var encroachment = Vector2(dxOffset, dyOffset);
+
+    var (scrollPixel, max) = getScrollInfo();
+
```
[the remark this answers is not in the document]
logiclrd: Copilot is very confused about mutator methods, it seems. 🙂

flutter/lib/models/model.dart
```diff
@@ -2052,6 +2169,126 @@ class CanvasModel with ChangeNotifier {
...
+      }
+    }
+  }
+
+  performEdgeScroll(Vector2 delta) {
+    var (scrollPixel, max) = getScrollInfo();
+
+    scrollPixel += delta;
+
+    scrollPixel.clamp(Vector2.zero(), max);
```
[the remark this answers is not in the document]
logiclrd: There is no wrong method call. In the preceding comment, the AI hallucinated a method `clampVector` parallel to `clampScalar`. The `clampScalar` method is real, but the one that takes `Vector2` parameters is just called `clamp`. Note that this line is calling _the same method_ as the one it thinks is wrong on line 2243. 😛

fufesou:
> There's only one call for all methods, though, so it's not a matter of removing the one we don't want but of actively checking whether each `debugPrint` call should be done. If that code were placed inside `if (kDebugMode)` then that additional filtering logic wouldn't need to run in a release situation at all -- a situation where the standard output stream is never expected to be visible anyway.

Yes, you're absolutely right.

I also understand that `debugPrint` is typically used for debug-mode prints.

However, I wasn't clear that the current `debugPrint` could also be helpful for debugging the release build.
When instructing users to start the application via the command line and share the printed output, the `debugPrint` statements in the current `desktop_home_page.dart` aren't excessive and are placed in fairly important locations.

Therefore, I think it's best to just leave them as they are.

Perhaps we should consider a **separate PR** to wrap these prints with `kDebugMode`, so we can filter out the ones that aren't useful in release builds.

flutter/lib/models/model.dart
```diff
@@ -2052,6 +2169,126 @@ class CanvasModel with ChangeNotifier {
...
+
+    var scrollPixelPercent = scrollPixel.clone();
+
+    scrollPixelPercent.divide(max);
```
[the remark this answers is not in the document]
logiclrd: This is actually true, _but_ `double` division by 0 is not an error; it simply returns `double.infinity`. I must admit that I hadn't thought of this, and a part of the reason is that a NaN value in `_scrollX` or `_scrollY` doesn't seem to have any deleterious effect. The code runs and does the right thing when only one axis is scrolling despite NaN values. Still, it's smelly to have NaN values there. They need to be handled on a per-axis basis. I'll pushing the handling downstream, because that keeps local complexity lower while also making the logic more general.

flutter/windows/runner/flutter_window.cpp
```diff
@@ -29,6 +40,48 @@ bool FlutterWindow::OnCreate() {
...
+          }
+        } else if (std::holds_alternative<flutter::EncodableList>(*arguments)) {
+          auto argsList = std::get<flutter::EncodableList>(*arguments);
+
+          if (std::holds_alternative<int>(argsList[0])) {
+            dx = std::get<int>(argsList[0]);
+          }
+          if (std::holds_alternative<int>(argsList[1])) {
+            dy = std::get<int>(argsList[1]);
+          }
```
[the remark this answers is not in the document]
logiclrd: I mean, in practice it won't ever happen because the only caller definitely always passes two arguments. But, the absence of the check does stand out.

flutter/linux/my_application.cc
```diff
@@ -131,6 +149,60 @@ MyApplication* my_application_new() {
...
+
+    if (dyValue) {
+      dy = fl_value_get_int(dyValue);
+    }
+
+    bool result = bump_mouse(dx, dy);
+
+    FlValue *result_value = fl_value_new_bool(result);
+
+    GError *error;
```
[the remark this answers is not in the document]
logiclrd: Interesting. Flutter does not document that `error` should be `NULL` on entry. But, the call sequence eventually drills down to GLib's `g_set_error` which does indeed state:

> if `err` is non-`NULL`, then `*err` must be `NULL`

logiclrd:
> Therefore, I think it's best to just leave them as they are.

Righto, I'll leave off the `if (kDebugMode)` condition in that case.

fufesou:
This function is working excellently across all platforms - Windows, X11, Wayland, and macOS. Let's go through a few rounds of Copilot review to make sure everything is solid.

flutter/lib/models/model.dart
```diff
@@ -2052,6 +2169,131 @@ class CanvasModel with ChangeNotifier {
...
+    scrollPixelPercent.scale(100.0);
```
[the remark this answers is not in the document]
logiclrd: That's okay. NaN's are coalesced to 0.0 in `setScrollPercent`.

flutter/lib/models/model.dart
```diff
@@ -2052,6 +2169,131 @@ class CanvasModel with ChangeNotifier {
...
+        _edgeScrollFallbackState.start();
+      }
+    }
+  }
+
+  performEdgeScroll(Vector2 delta) {
+    var (scrollPixel, max) = getScrollInfo();
+
+    scrollPixel += delta;
+
```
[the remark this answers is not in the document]
logiclrd: Very confused AI. The code doesn't appear to be expecting `scrollPixel.clamp` to return a value at all.
```
    scrollPixel.clamp(Vector2.zero(), max);
```

flutter/lib/models/model.dart
```diff
@@ -1789,6 +1832,59 @@ class ViewStyle {
...
+      var delta = _encroachment * (SpeedFactor * thisTickElapsed.inMilliseconds / FrameTime);
+
+      _owner.performEdgeScroll(delta);
+
+      _lastTotalElapsed = totalElapsed;
+    }
+  }
+
+  start() {
```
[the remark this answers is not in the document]
logiclrd: I just looked it up, and if you leave off any return type annotation, the method implicitly returns `dynamic`. With no `return` statements, the returned value is `null`.

Before these changes, `CanvasModel` already has 15 methods implicitly returning `dynamic`. It has two that explicitly specify `void`.

It seems out of scope to fix up the class in this PR, but perhaps it would make sense to have a work item in the backlog to identify all methods that _should_ have `void` return type and fix them.

I'll add `void` annotations to all methods added by this PR, though, for sure.

flutter/linux/my_application.cc
```diff
@@ -131,6 +149,61 @@ MyApplication* my_application_new() {
...
+      default: break;
+    }
+
+    int dx = 0, dy = 0;
+
+    if (dxValue) {
+      dx = fl_value_get_int(dxValue);
+    }
+
+    if (dyValue) {
```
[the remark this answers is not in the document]
logiclrd: Ugh, it's true. The underlying implementation definitely checks that it's an `int`, but the documentation has this annoying line in it:

> Calling this with an [FlValue](https://api.flutter.dev/linux-embedder/fl__value_8h.html#a248fb199aa2cc2b92f692067e2d15cd9) that is not of type [FL_VALUE_TYPE_INT](https://api.flutter.dev/linux-embedder/fl__value_8h.html#ae514203eea95fec354c1648957f77a4da3201dc28f5f6b3237481e0fe9a326bab) is a programming error.

Sigh. 🙂

flutter/lib/models/model.dart
```diff
@@ -1810,6 +1906,15 @@ class CanvasModel with ChangeNotifier {
...
   ScrollStyle _scrollStyle = ScrollStyle.scrollauto;
+  // tracks whether edge scroll should be active, prevents spurious
+  // scrolling when the cursor enters the view from outside
+  EdgeScrollState _edgeScrollState = EdgeScrollState.inactive;
+  // fallback strategy for when Bump Mouse isn't available
+  late EdgeScrollFallbackState _edgeScrollFallbackState;
+  // to avoid hammering a non-functional Bump Mouse
+  bool _bumpMouseIsWorking = true;
+  // briefly set to true when the canvas detects a resize
+  bool _suppressEdgeScroll = false;
```
fufesou: `_suppressEdgeScroll` is never set, is the comment above wrong?

Image


The `Flutter` extension in VS Code can help the Dart diagnostics. https://dart.dev/tools/diagnostics

flutter/lib/models/model.dart
```diff
@@ -1789,6 +1832,59 @@ class ViewStyle {
...
+  }
+
+  emitTick(Duration totalElapsed) {
+    if (_nextEventIsFirst) {
+      _lastTotalElapsed = totalElapsed;
+      _nextEventIsFirst = false;
+    } else {
+      final thisTickElapsed = totalElapsed - _lastTotalElapsed;
+
+      const double FrameTime = 1000.0 / 60.0;
```
fufesou: Suggestion from the `Flutter` extension in vscode.

```
The constant name 'FrameTime' isn't a lowerCamelCase identifier.
Try changing the name to follow the lowerCamelCase style.
```


The following names may be better.

```dart
      const double kFrameTime = 1000.0 / 60.0;
      const double kSpeedFactor = 0.1;
```

fufesou (changes requested):

flutter/macos/Runner/MainFlutterWindow.swift
```diff
@@ -99,6 +99,51 @@ class MainFlutterWindow: NSWindow {
                         result(granted)
                     })
                     break
+                case "bumpMouse":
+                    let arg = call.arguments as! [String: Any];
+
+                    let dx = (arg["dx"] as? Int) ?? 0;
+                    let dy = (arg["dy"] as? Int) ?? 0;
```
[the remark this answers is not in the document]
logiclrd: I find it unclear whether the list argument format handling is actually needed. I can't remember now where I found it, but I defniitely found example code that checked for both map and list argument styles. That's why the Linux and Windows versions are written that way. But, the `macos` variant, as highlighted here, has only used the map style the whole time and has, in practice, worked. But, I suppose I'd better update it to match for consistency.

flutter/lib/models/model.dart
```diff
@@ -2052,6 +2169,131 @@ class CanvasModel with ChangeNotifier {
...
+        return;
+      }
+    }
+
+    var dxOffset = 0.0;
+    var dyOffset = 0.0;
+
+    if (x < edgeThickness) {
+      dxOffset = x - edgeThickness;
+    } else if ((x >= size.width - edgeThickness)) {
```
fufesou: Should it be?

```dart
} else if (x >= (size.width - edgeThickness)) {
```

fufesou (changes requested):

flutter/lib/models/model.dart
logiclrd: It's not fundamentally wrong but it's out-of-date, so it's wrong _now_. 🙂

flutter/lib/models/model.dart
logiclrd: Ah, yes, I'd seen that elsewhere and it hadn't clicked that that's the convention for constants.

flutter/lib/models/model.dart
logiclrd: I think there was an intermediate point at which there were two parts to the condition, `if ((a) && (b))`, and that got refactored but the extra set of parentheses was left in by accident.

flutter/lib/models/model.dart
```diff
@@ -1835,14 +1938,24 @@ class CanvasModel with ChangeNotifier {
   double get devicePixelRatio => _devicePixelRatio;
   Size get size => _size;
   ScrollStyle get scrollStyle => _scrollStyle;
+  bool get suppressEdgeScroll => _suppressEdgeScroll;
```
fufesou: This line should be removed.

flutter/lib/models/model.dart
```diff
@@ -1835,14 +1938,24 @@ class CanvasModel with ChangeNotifier {
...
   ViewStyle get viewStyle => _lastViewStyle;
   RxBool get imageOverflow => _imageOverflow;
 
   _resetScroll() => setScrollPercent(0.0, 0.0);
 
-  setScrollPercent(double x, double y) {
-    _scrollX = x;
-    _scrollY = y;
+  void setScrollPercent(double x, double y) {
```
fufesou, line 1954: `0.0 / 0.0` results `NaN`, while `1.0 / 0.0` results `Infinity` in my test.

Should `x.isInfinite` also be condisidered here?

fufesou (changes requested):

flutter/lib/models/model.dart
logiclrd: Ah, good catch. I did some poking, and it looks like the `.isFinite` property is the most useful here; it's `false` for both infinity and NaN.

flutter/lib/models/model.dart
```diff
@@ -2052,6 +2166,133 @@ class CanvasModel with ChangeNotifier {
...
+      dyOffset = y - (size.height - edgeThickness);
+    }
+
+    var encroachment = Vector2(dxOffset, dyOffset);
+
+    var (scrollPixel, max) = getScrollInfo();
+
+    encroachment.clamp(-scrollPixel, max - scrollPixel);
+
+    if (encroachment.length == 0) {
```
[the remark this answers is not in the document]
logiclrd: `isZero` seems to be a hallucination, but `length2` works just as well as `length` and does indeed avoid an unnecessary `sqrt`.

flutter/lib/models/model.dart
```diff
@@ -2052,6 +2166,133 @@ class CanvasModel with ChangeNotifier {
...
+      dyOffset = y - edgeThickness;
+    } else if (y >= size.height - edgeThickness) {
+      dyOffset = y - (size.height - edgeThickness);
+    }
+
+    var encroachment = Vector2(dxOffset, dyOffset);
+
+    var (scrollPixel, max) = getScrollInfo();
+
+    encroachment.clamp(-scrollPixel, max - scrollPixel);
```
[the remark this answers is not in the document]
logiclrd: Sigh 😛

flutter/lib/models/model.dart
fufesou: `x` and `y` can be both infinity and NaN.

But should both infinity and NaN be checked? The last commit only checks infinity .

fufesou:
Another issue with edge scroll on macOS.


https://github.com/user-attachments/assets/7e711e22-9499-49da-a5a9-c47d94bb2fc4


The cursor does not appear until it moves out of the window.

Since you don't have access to macOS, I will try to fix the problem.

logiclrd:
> Another issue with edge scroll on macOS.
> ..
> The cursor does not appear until it moves out of the window.

The video does not appear for me for whatever reason.

> Since you don't have access to macOS, I will try to fix the problem.

Thanks so much. 🙂

logiclrd:
I'm pretty sure I ticked the box to grant you permission to push commits directly to the source branch of this PR, so if/when you have changes, you can just add a remote for `git@github.com:/logiclrd/rustdesk` and `git push` as needed. I think. 🙂

logiclrd:
The video just started working. Not sure what was going on before, it was just .. inert. But now it's there. 🙂

logiclrd:
Huh, that's really curious. One thing I note is that, if I'm understanding the video correctly, the cursor is initially invisible with _all_ the scroll settings, but then becomes visible on the first mouse move event with ScrollAuo and ScrollBar.

fufesou:
> > Another issue with edge scroll on macOS.
> > ..
> > The cursor does not appear until it moves out of the window.
> 
> The video does not appear for me for whatever reason.
> 
> > Since you don't have access to macOS, I will try to fix the problem.
> 
> Thanks so much. 🙂


`activeLocalCursor()` should be called in both `moveDesktopMouse()` and `edgeScrollMouse()`.

You can reproduce this issue by enabling "Show remote cursor".
1. Move cursor on the controlled side.
2. Move cursor in local window.

```
diff --git a/flutter/lib/models/model.dart b/flutter/lib/models/model.dart
index 143129736..8d06c2603 100644
--- a/flutter/lib/models/model.dart
+++ b/flutter/lib/models/model.dart
@@ -2151,19 +2151,7 @@ class CanvasModel with ChangeNotifier {
       notifyListeners();
     }

-    // If keyboard is not permitted, do not move cursor when mouse is moving.
-    if (parent.target != null && parent.target!.ffiModel.keyboard) {
-      // Draw cursor if is not desktop.
-      if (!(isDesktop || isWebDesktop)) {
-        parent.target!.cursorModel.moveLocal(x, y);
-      } else {
-        try {
-          RemoteCursorMovedState.find(id).value = false;
-        } catch (e) {
-          //
-        }
-      }
-    }
+    activeLocalCursor(x, y);
   }

   void initializeEdgeScrollFallback(TickerProvider tickerProvider) {
@@ -2195,7 +2183,25 @@ class CanvasModel with ChangeNotifier {
     return (scrollPixel, max);
   }

+  void activeLocalCursor(double x, double y) {
+    // If keyboard is not permitted, do not move cursor when mouse is moving.
+    if (parent.target != null && parent.target!.ffiModel.keyboard) {
+      // Draw cursor if is not desktop.
+    if (parent.target != null && parent.target!.ffiModel.keyboard) {
+      // Draw cursor if is not desktop.
+      if (!(isDesktop || isWebDesktop)) {
+        parent.target!.cursorModel.moveLocal(x, y);
+      } else {
+        try {
+          RemoteCursorMovedState.find(id).value = false;
+        } catch (e) {
+          //
+        }
+      }
+    }
+  }
+
   void edgeScrollMouse(double x, double y) async {
+    activeLocalCursor(x, y);
+
     if ((_edgeScrollState == EdgeScrollState.inactive)
      || (size.width == 0 || size.height == 0)
      || !(_horizontal.hasClients || _vertical.hasClients)) {
```

logiclrd:
I had just come to suspect that exact thing. Since the cursor didn't appear until the first mouse event in all cases, the difference had to be in what happens when a mouse event is processed, and ScrollAuto and ScrollBar both funnel into `moveDesktopMouse` while ScrollEdge runs `edgeScrollMouse` instead, and that particular block in the former and not the latter stands out, especially since the comments indicate that it is about cursor visibility. 🙂

logiclrd:
Is there a logical point where `activateLocalCursor` could be called proactively as soon as the connection is established, so that the cursor doesn't wait for the first mouse event to appear? Also, does it need to be called on every mouse event, or is it just required to be called at least once?

My guess is, looking at the code, that it only needs to be called once on desktop, but that it does need to be called on every event when it's not a desktop environment because RustDesk needs to represent the cursor manually in the UI visual tree (`cursorModel.moveLocal(x, y)`). Is that right? If so, then it feels like the function is performing two distinct functions.

Unless it absolutely requires that the cursor be manually moved on the remote end (which I can't do because the remote end is in a different house 🙂), I can't reproduce the issue on Windows. Even without any calls to `activateLocalCursor`, the remote cursor appears immediately upon connection.

fufesou:
> Unless it absolutely requires that the cursor be manually moved on the remote end (which I can't do because the remote end is in a different house 🙂),

Yes. "the cursor be manually moved on the remote end" is needed.

The local cursor should be hidden when the "Show remote cursor" option is enabled and a remote cursor is being displayed (moved).

It should then reappear when the local cursor is moved.

logiclrd:
Okay I dug deeper and see that it's setting a semaphore that gets picked up by an observer in the widget body that `_RemotePageState` builds.

flutter/lib/desktop/pages/desktop_home_page.dart
```diff
@@ -760,9 +761,19 @@ class _DesktopHomePageState extends State<DesktopHomePage>
           'scaleFactor': screen.scaleFactor,
         };
 
+    bool isChattyMethod(String methodName) {
+      switch (methodName) {
+        case kWindowBumpMouse: return true;
+      }
+
+      return false;
```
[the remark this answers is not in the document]
logiclrd: Deliberately written as a `switch` because the most likely future update is then reduced to a one-line addition.

flutter/lib/models/model.dart
```diff
@@ -2037,21 +2159,155 @@ class CanvasModel with ChangeNotifier {
...
+    if (y < edgeThickness) {
+      dyOffset = y - edgeThickness;
+    } else if (y >= size.height - edgeThickness) {
+      dyOffset = y - (size.height - edgeThickness);
+    }
+
+    var encroachment = Vector2(dxOffset, dyOffset);
+
+    var (scrollPixel, max) = getScrollInfo();
+
```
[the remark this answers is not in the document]
logiclrd: ??

flutter/lib/models/model.dart
```diff
@@ -1957,26 +2077,28 @@ class CanvasModel with ChangeNotifier {
...
-    if (_scrollStyle != ScrollStyle.scrollbar) return;
+    if (_scrollStyle == ScrollStyle.scrollauto) return;
     style ??= await bind.sessionGetViewStyle(sessionId: sessionId);
     if (style != kRemoteViewStyleOriginal && style != kRemoteViewStyleCustom) {
       return;
     }
 
     _resetScroll();
+
     Future.delayed(duration, () async {
```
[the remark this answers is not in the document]
logiclrd: I believe this is linked to the wrong line and is actually referring to the `updateScrollStyle` method. I haven't been able to ascertain whether `bind.sessionGetScrollStyle` could return `null`, so I'll assume that it could happen and make the function resilient to it.

logiclrd:
Hmm, interesting...

So, for starters, it appears that RustDesk is able to copy the bitmap of the remote cursor and set it as the local cursor; while the mouse cursor is overtop of the RustDesk window, it shows in the image of the remote cursor but I'm pretty sure it's rendering locally.

I found that, while connected to a Linux X11 session, I can use e.g. `xdotool mousemove_relative 10 10` to inject mouse movement on the remote end. (I have to delay it by 2 seconds, because the cursor is hidden on keyboard input and I need to move the cursor after pressing Enter for it to be visible before the remote move happens.) When I do this, sure enough, with "ScrollAuto", the cursor immediately updates to the new position, and with "ScrollEdge", the cursor just disappears.

_But_, calling `activateLocalCursor` from `edgeScrollMouse` doesn't fix the problem.

Also, and more seriously, moving the local cursor _doesn't_ break it out of this state. Scrolling is lock, local events have no effect, even with other scroll styles like "ScrollAuto". This happens on `master`; it isn't new to these changes. This state persists until the cursor leaves the window and re-enters.

logiclrd:
Regarding that last issue, this line seems to be at fault:

https://github.com/rustdesk/rustdesk/blob/a30582c8409a78f5b184f4c9e6600fe0a86a2744/flutter/lib/models/input_model.dart#L1222

If I'm understanding the logic correctly, it's trying to say that the mouse control is protected until the local mouse cursor is moved sufficiently. Once the local movement exceeds `kMouseControlDistance`, it lets you regain control. But, because it is latching the position in the failure path, local control is only regained if `kMouseControlDistance` is exceeded _in a single mouse event_.

fufesou:
> it shows in the image of the remote cursor but I'm pretty sure it's rendering locally.

Yes.

> _But_, calling `activateLocalCursor` from `edgeScrollMouse` doesn't fix the problem.
> 
> Also, and more seriously, moving the local cursor _doesn't_ break it out of this state. Scrolling is lock, local events have no effect, even with other scroll styles like "ScrollAuto". This happens on `master`; it isn't new to these changes. This state persists until the cursor leaves the window and re-enters.

I can't understand.

"Move cursor" -> `handlePointerDevicePos()` -> `moveDesktopMouse()` -> `RemoteCursorMovedState (true)` -> Show local cursor

I think it's correct.

Why "Scrolling is lock" and "local events have no effect"?

fufesou:
> If I'm understanding the logic correctly, it's trying to say that the mouse control is protected until the local mouse cursor is moved sufficiently. Once the local movement exceeds `kMouseControlDistance`, it lets you regain control. But, because it is latching the position in the failure path, local control is only regained if `kMouseControlDistance` is exceeded _in a single mouse event_.

Correct, but it is not responsible for rendering the cursor.

`RemoteCursorMovedState` controls the cursor drawing.

https://github.com/rustdesk/rustdesk/blob/a30582c8409a78f5b184f4c9e6600fe0a86a2744/flutter/lib/desktop/pages/remote_page.dart#L611

https://github.com/rustdesk/rustdesk/blob/a30582c8409a78f5b184f4c9e6600fe0a86a2744/flutter/lib/desktop/pages/remote_page.dart#L106

fufesou:
Hi, please also address this discussion  https://github.com/rustdesk/rustdesk/pull/13247#discussion_r2476443545

flutter/lib/models/model.dart
```diff
@@ -2037,21 +2159,155 @@ class CanvasModel with ChangeNotifier {
...
+    activeLocalCursor(x, y);
+
+    if ((_edgeScrollState == EdgeScrollState.inactive) ||
+        (size.width == 0 || size.height == 0) ||
+        !(_horizontal.hasClients || _vertical.hasClients)) {
+      return;
+    }
+
+    // Trigger scrolling when the cursor is close to an edge
+    const double edgeThickness = 100;
```
fufesou, line 2228: Image

Is a range of `100` too wide for edge scrolling?
Or is this `100` value based on another remote desktop tool?

rustdesk: > Is a range of `100` too wide for edge scrolling?

I am also curious about this. How do you get `100`? do you have any reference design? Is there any other software used as your reference?

flutter/lib/models/model.dart
logiclrd: That's not quite the case. The last commit checks `isFinite`, not `isInfinite`. These are not the strict logical inverse of each other. NaN is not infinite, but it is not finite either. `isFinite` returns `false` for both `Infinity` and `NaN`.

```
Value             isFinite  isInfinite  isNaN
----------------  --------  ----------  -----
(numeric)         true      false       false
infinity          false     true        false
negativeInfinity  false     true        false
nan               false     false       true
```

flutter/lib/models/model.dart
logiclrd: It's arbitrary and could absolutely be changed. That said, the mouse isn't captured, so the only thing keeping it inside the window is Bump Mouse. If the mouse cursor manages to cross over the edge scrolling region in its entirety in a single event, then it escapes the window and that's that. Set to 100, it takes a concerted effort to make this happen. If it were set to e.g. 10, edge scrolling would be almost useless because casual use of the mouse would be constantly escaping the window. So a balance must be struck.

flutter/lib/models/model.dart
fufesou: Sorry, I misread.

flutter/lib/models/model.dart
rustdesk: Something like this to make it editable would be great. It should be a local option rather than peer option.

image

logiclrd: I'll look into that, but I've stayed up way too late tonight already, so I think I need to sign off for now. 🙂

logiclrd: One parting thought -- this is definitely logically related but it also inarguably extends the scope of the PR. Perhaps it would make sense to merge it with a fixed value initially, and follow up with a configurable option in a separate PR? This PR's scope has already ballooned a couple of times. 😛

rustdesk: Agree.

logiclrd:
I just remembered I had local commits I hadn't pushed. I rebased them ontop of your changes.

I took a slightly different approach to the refactoring of the local cursor stuff.

- I moved the call up to the call side of `moveDesktopMouse` and `edgeScrollMouse`. This made it trivial to keep its logical placement at the end of processing despite `edgeScrollMouse`'s various early returns.

- I made two variants, `activateLocalCursor` and `updateLocalCursor`. The latter is the full factoring of the functionality, taking coordinates as arguments. The former is just the desktop part, which doesn't receive or use coordinates.

- In `remote_page.dart`, I added a call to `activateLocalCursor` to the first image callbacks registered by `initState`. The reason I did this is that in the video you showed earlier I observed that the cursor didn't appear until you moved the mouse. This _should_ make the mouse cursor appear as soon as the initial image in the remote view is transferred, if I'm understanding correctly.

fufesou (approved):

flutter/lib/models/model.dart
logiclrd: So I'm taking a poke at this next step. Just want to confirm I'm understanding the code structure correctly -- in order to add a configuration setting for this, I will need to first add the setting to `config.rs` in `hbb_common` in its separate repository??

rustdesk: No, you do not need to add anything in config.rs. Please use bind.mainGetLocalOption and bind.mainSetLocalOption. You can find example in desktop_setting_page.dart.

logiclrd: Okay I think I'm just about ready to propose the first draft of these changes. There is one change I have discovered is needed in `config.rs` though: The existing code for the scroll style setting only accepts known values. So, trying to save `scrolledge` as the scroll style doesn't work.

```
--- a/src/config.rs
+++ b/src/config.rs
@@ -1950,3 +1950,3 @@ impl UserDefaultConfig {
             keys::OPTION_VIEW_STYLE => self.get_string(key, "original", vec!["adaptive"]),
-            keys::OPTION_SCROLL_STYLE => self.get_string(key, "scrollauto", vec!["scrollbar"]),
+            keys::OPTION_SCROLL_STYLE => self.get_string(key, "scrollauto", vec!["scrolledge", "scrollbar"]),
             keys::OPTION_IMAGE_QUALITY => {
```

This actually means that setting the default scroll style to ScrollEdge isn't possible with the currently-merged code from this PR. It makes the right call to save the setting, but the value isn't recognized and so it doesn't actually save any change. An oversight in my testing.

So there will be a PR on `hbb_common`, needed for the edge scrolling feature in general.

rustdesk: Please do not change default scroll style.

logiclrd: https://github.com/rustdesk/hbb_common/pull/426

PR #13273, opened 2025-10-24
Allow flipping sort order in mobile app's file transfer

nguyenquyhy:
Right now I can't find any way in the mobile app to change sort order from ascending to descending when doing file transfer. 

A common pattern that I am trying to add in this PR is to flip the order when the same sort type is selected. I think the same behavior is already implemented for the Desktop client.

https://github.com/user-attachments/assets/4b1913a2-405a-4c7f-b40e-7ddc61bf0e1c

I have created a test build in my fork: https://github.com/nguyenquyhy/rustdesk/releases/tag/sort

### DIFF
--- a/flutter/lib/mobile/pages/file_manager_page.dart
+++ b/flutter/lib/mobile/pages/file_manager_page.dart
@@ -424,6 +424,7 @@ class FileManagerView extends StatefulWidget {
 class _FileManagerViewState extends State<FileManagerView> {
   final _listScrollController = ScrollController();
   final _breadCrumbScroller = ScrollController();
+  final ascending = Rx<bool?>(null);
 
   bool get isLocal => widget.controller.isLocal;
   FileController get controller => widget.controller;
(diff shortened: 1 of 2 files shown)


21pages:
It would be nice to show the current sorting order, similar to the desktop version.

image

21pages (approved):

21pages:
You can remove the changes in file_model.dart; otherwise, the mobile app will keep the previous connection's sorting when you connect again, so it won't always start in ascending order for new connections.

nguyenquyhy:
> It would be nice to show the current sorting order, similar to the desktop version.

Yes, I am thinking about this too. However, I try to keep this change minimal as suggested in Contribution Guide, and being able to change the order is more important to me than show the current one.

> You can remove the changes in file_model.dart; otherwise, the mobile app will keep the previous connection's sorting when you connect again, so it won't always start in ascending order for new connections.

You are right. I didn't think about this case. I have removed the file_model changes.

21pages (approved):

PR #13320, opened 2025-10-28
fix mac hwcodec check

21pages:
https://github.com/rustdesk/rustdesk/discussions/12792#discussioncomment-14800751

After pr, still only H265 is available, H264 will be available after adjusting some parameters.


lzyddgg:
Thanks very much for the fix! Unfortunately, for my old intel mac the h265 codec was laggy, while the h264 option is still not visible. Could you please also fix the h264 codec as well if convinient? Thanks in advance!

PR #13366, opened 2025-10-31
Fix terminal clear command to remove residual output

alonginwind:
Ensure clear commands fully clear terminal output.After clearing, executing other commands displays output correctly without mixing or misalignment.
bb6bf3e85a3946ccb18012a69fb4344d
0d4b9477807b4b2a98f40a8034b66f35

### DIFF
--- a/flutter/lib/models/terminal_model.dart
+++ b/flutter/lib/models/terminal_model.dart
@@ -317,6 +317,15 @@ class TerminalModel with ChangeNotifier {
           return;
         }
 
+        // Check for any clear screen sequences in the terminal output
+        // This ensures a complete clear and provides a better terminal experience
+        if (text.contains('\x1b[2J') ||  // Clear entire screen, if env TERM = xterm-256color, for consistent behavior
+            text.contains('\x1b[J') ||   // Clear from cursor to end of screen, if env TERM = vt220
+            text.contains('\x1b[25l\x1b[H\x1b[?25h') || // Hide cursor, move home, show cursor, ssh through windows powershell
+            text.contains('\x1b[25l\x1b[?2004l\x1b[H\x1b[?25h')) { // Hide cursor, disable bracketed paste, move home, show cursor, ssh through windows powershell
+          terminal.buffer.clear(); // Clear the terminal buffer
+        }
+
         terminal.write(text);
       } catch (e) {
         debugPrint('[TerminalModel] Failed to process terminal data: $e');


fufesou:
@alonginwind 

Hi, this is a solid PR, thank you for the contribution.

As part of the peer testing, I tried to reproduce the issue from your first screenshot but couldn't. I tested with Windows, macOS, and Linux as the controlled side.

Would you mind sharing the specific steps ?

alonginwind:
> @alonginwind
> 
> Hi, this is a solid PR, thank you for the contribution.
> 
> As part of the peer testing, I tried to reproduce the issue from your first screenshot but couldn't. I tested with Windows, macOS, and Linux as the controlled side.
> 
> Would you mind sharing the specific steps ?

There are two scenarios in which this can be reproduced:
1. Using RustDesk terminal to access a remote Windows machine, then SSH into a Linux system from the terminal, run top several times, and finally execute clear; the issue almost always occurs.
2. Sometimes when controlling a Linux system directly, with TERM=vt220, this issue can also occur.


vt220 behaves differently because its clear-screen control codes are not the same as xterm-256color, and Windows likely doesn’t have a clear-screen control code at all.
For the second issue, you can set the environment variable TERM=xterm-256color when opening the terminal.
As for the first issue, I’m still investigating. If there’s no better solution, temporarily using the code from the PR for Windows systems should be fine.

alonginwind:
@fufesou 
This modification provides a much better experience than the previous one. I think it’s a nearly perfect solution. Switching to the clear command with the up and down arrow keys also clears the screen.

fufesou:
@alonginwind 

> There are two scenarios in which this can be reproduced:
> 
> 1. Using RustDesk terminal to access a remote Windows machine, then SSH into a Linux system from the terminal, run top several times, and finally execute clear; the issue almost always occurs.
> 2. Sometimes when controlling a Linux system directly, with TERM=vt220, this issue can also occur.
> 
> vt220 behaves differently because its clear-screen control codes are not the same as xterm-256color, and Windows likely doesn’t have a clear-screen control code at all. For the second issue, you can set the environment variable TERM=xterm-256color when opening the terminal. As for the first issue, I’m still investigating. If there’s no better solution, temporarily using the code from the PR for Windows systems should be fine.

1. I still cannot reproduce this issue. My testing has included scenarios like `macOS` -> `Windows` -> `Linux` and `Windows` -> `Windows` -> `Linux`. In these sessions, I repeatedly executed commands (e.g., running `top`, then quitting with `q`, restarting it, and finally using clear), and also tried sequences involving the `Enter` key, but the problem never occurred.
2. Regarding the `TERM` environment variable: our logic always sets it to the current `TERM` value (if available) or defaults to a common type like `screen-256color`, `xterm-256color`, or `xterm`. How common is the `vt220` terminal type across modern Linux distributions? Is it still in widespread use?

https://github.com/rustdesk/rustdesk/blob/a903f710ea5281faffee78cfd9d7c7a255d9e249/src/platform/linux.rs#L352

https://github.com/rustdesk/rustdesk/blob/a903f710ea5281faffee78cfd9d7c7a255d9e249/src/platform/linux.rs#L282

alonginwind:
> @alonginwind
> 
> > There are two scenarios in which this can be reproduced:
> > 
> > 1. Using RustDesk terminal to access a remote Windows machine, then SSH into a Linux system from the terminal, run top several times, and finally execute clear; the issue almost always occurs.
> > 2. Sometimes when controlling a Linux system directly, with TERM=vt220, this issue can also occur.
> > 
> > vt220 behaves differently because its clear-screen control codes are not the same as xterm-256color, and Windows likely doesn’t have a clear-screen control code at all. For the second issue, you can set the environment variable TERM=xterm-256color when opening the terminal. As for the first issue, I’m still investigating. If there’s no better solution, temporarily using the code from the PR for Windows systems should be fine.
> 
> 1. I still cannot reproduce this issue. My testing has included scenarios like `macOS` -> `Windows` -> `Linux` and `Windows` -> `Windows` -> `Linux`. In these sessions, I repeatedly executed commands (e.g., running `top`, then quitting with `q`, restarting it, and finally using clear), and also tried sequences involving the `Enter` key, but the problem never occurred.
> 2. Regarding the `TERM` environment variable: our logic always sets it to the current `TERM` value (if available) or defaults to a common type like `screen-256color`, `xterm-256color`, or `xterm`. How common is the `vt220` terminal type across modern Linux distributions? Is it still in widespread use?
> 
> https://github.com/rustdesk/rustdesk/blob/a903f710ea5281faffee78cfd9d7c7a255d9e249/src/platform/linux.rs#L352
> 
> https://github.com/rustdesk/rustdesk/blob/a903f710ea5281faffee78cfd9d7c7a255d9e249/src/platform/linux.rs#L282

1. My testing scenario involves Linux → Windows → Linux, and the issue occurs.
2. It’s probably uncommon, but I’m not sure why—on my Orin Nano, one device reports vt220 while another reports xterm-256color.

fufesou:
> My testing scenario involves Linux → Windows → Linux, and the issue occurs.

I still can't reproduce it.

Is the `TERM` also `vt220`?

> on my Orin Nano, one device reports vt220 while another reports xterm-256color

RustDesk uses the following two commands to find the `TERM` and uses it for the terminal session.
Could you please run these commands to find out why it is `vt220`?

```bash
uid=$(loginctl | grep 'seat0' | awk '{print $2}')
term=$(ps -u $uid -f | grep "rustdesk \-\-server" | grep -v 'grep' | tail -1 | awk '{{print $2}}' | xargs -I__ cat /proc/__/environ 2>/dev/null | tr '\0' '\n' | grep '^TERM=' | tail -1 | sed 's/TERM=//g')
echo "TERM from rustdesk server: $term"

```

```bash
uid=$(loginctl | grep 'seat0' | awk '{print $2}')
for bin in bash zsh fish sh; do
    grep_pattern="$bin\$|bin/$bin"
    term=$(ps -u $uid -f | grep -E "${grep_pattern}" | grep -v 'grep' | tail -1 | awk '{{print $2}}' | xargs -I__ cat /proc/__/environ 2>/dev/null | tr '\0' '\n' | grep '^TERM=' | tail -1 | sed 's/TERM=//g')
    if [ -n "$term" ]; then
        echo "TERM from $bin: $term"
        cmd=$(ps -u $uid -f | grep -E "${grep_pattern}" | grep -v 'grep')
        echo -e "\nThe command of the bin is:\n$cmd"
        break
    fi
done
```

alonginwind:
output

TERM is xterm-256color.

alonginwind:
> > My testing scenario involves Linux → Windows → Linux, and the issue occurs.
> 
> I still can't reproduce it.
> 
> Is the `TERM` also `vt220`?
> 
> > on my Orin Nano, one device reports vt220 while another reports xterm-256color
> 
> RustDesk uses the following two commands to find the `TERM` and uses it for the terminal session. Could you please run these commands to find out why it is `vt220`?
> 
> ```shell
> uid=$(loginctl | grep 'seat0' | awk '{print $2}')
> term=$(ps -u $uid -f | grep "rustdesk \-\-server" | grep -v 'grep' | tail -1 | awk '{{print $2}}' | xargs -I__ cat /proc/__/environ 2>/dev/null | tr '\0' '\n' | grep '^TERM=' | tail -1 | sed 's/TERM=//g')
> echo "TERM from rustdesk server: $term"
> ```
> 
> ```shell
> uid=$(loginctl | grep 'seat0' | awk '{print $2}')
> for bin in bash zsh fish sh; do
>     grep_pattern="$bin\$|bin/$bin"
>     term=$(ps -u $uid -f | grep -E "${grep_pattern}" | grep -v 'grep' | tail -1 | awk '{{print $2}}' | xargs -I__ cat /proc/__/environ 2>/dev/null | tr '\0' '\n' | grep '^TERM=' | tail -1 | sed 's/TERM=//g')
>     if [ -n "$term" ]; then
>         echo "TERM from $bin: $term"
>         cmd=$(ps -u $uid -f | grep -E "${grep_pattern}" | grep -v 'grep')
>         echo -e "\nThe command of the bin is:\n$cmd"
>         break
>     fi
> done
> ```


rustdesk terminal output info:  
TERM from bash: vt220

The command of the bin is:
nvidia      1416    1012  0 08:00 ttyTCU0  00:00:00 -bash
nvidia      2979    2872  0 08:00 pts/0    00:00:00 bash
nvidia      5524    4732  0 08:03 pts/1    00:00:00 bash
nvidia      5600    5524  0 08:04 pts/1    00:00:00 bash



ssh to same device: 
TERM from bash: xterm-256color

The command of the bin is:
nvidia      1416    1012  0 08:00 ttyTCU0  00:00:00 -bash
nvidia      2979    2872  0 08:00 pts/0    00:00:00 bash
nvidia      5524    4732  0 08:03 pts/1    00:00:00 bash
nvidia      6376    6374  0 08:05 pts/2    00:00:00 -bash
nvidia      6614    6376  0 08:05 pts/2    00:00:00 -bash


I have to power-cycle the device to restart it. The first time I connect via RustDesk, the terminal $TERM is set to vt220. However, if I reconnect, it changes to xterm-256color. This behavior is quite strange.

fufesou:
What is the Windows version in your gif? 

https://github.com/rustdesk/rustdesk/pull/13366#discussion_r2488556166
What's your take on this Copilot review? Does it seem reasonable?

> The first time I connect via RustDesk, the terminal $TERM is set to vt220. However, if I reconnect, it changes to xterm-256color. This behavior is quite strange.

RustDesk currently obtains the `TERM` value from a specific `bash` process (PID: `5600`). A better approach would be to:
1. Scan all `bash` processes and choose the highest-priority TERM (e.g., `screen-256color` > `xterm-256color` > `xterm` -> other).
2. Discover which `TERM` values the controlled side supports.

But this seems like an unusual situation. It would be helpful to track down why your `TERM` is set to `vt220`.

alonginwind:
> What is the Windows version in your gif?
> 
> [#13366 (comment)](https://github.com/rustdesk/rustdesk/pull/13366#discussion_r2488556166) What's your take on this Copilot review? Does it seem reasonable?
> 
> > The first time I connect via RustDesk, the terminal $TERM is set to vt220. However, if I reconnect, it changes to xterm-256color. This behavior is quite strange.
> 
> RustDesk currently obtains the `TERM` value from a specific `bash` process (PID: `5600`). A better approach would be to:
> 
> 1. Scan all `bash` processes and choose the highest-priority TERM (e.g., `screen-256color` > `xterm-256color` > `xterm` -> other).
> 2. Discover which `TERM` values the controlled side supports.
> 
> But this seems like an unusual situation. It would be helpful to track down why your `TERM` is set to `vt220`.

Windows 10 1oT Enterprise LTSC
21H2
2023/2/25
19044.2673
Windows Feature Experience Pack 120.2212.4190.0


Yes, the cooment seems reasonable. But I think xterm.dart does not handle some clear control code correctly.

alonginwind:
> What is the Windows version in your gif?
> 
> [#13366 (comment)](https://github.com/rustdesk/rustdesk/pull/13366#discussion_r2488556166) What's your take on this Copilot review? Does it seem reasonable?
> 
> > The first time I connect via RustDesk, the terminal $TERM is set to vt220. However, if I reconnect, it changes to xterm-256color. This behavior is quite strange.
> 
> RustDesk currently obtains the `TERM` value from a specific `bash` process (PID: `5600`). A better approach would be to:
> 
> 1. Scan all `bash` processes and choose the highest-priority TERM (e.g., `screen-256color` > `xterm-256color` > `xterm` -> other).
> 2. Discover which `TERM` values the controlled side supports.
> 
> But this seems like an unusual situation. It would be helpful to track down why your `TERM` is set to `vt220`.


Is it possible that the issue is with the systemd service? Running rustdesk --service directly works fine. But when started via the systemd service, the shell returned is always /bin/sh, even though /bin/bash exists.

fufesou:
> Is it possible that the issue is with the systemd service? Running rustdesk --service directly works fine. But when started via the systemd service, the shell returned is always /bin/sh, even though /bin/bash exists.

> the shell returned is always /bin/sh

What do you mean?

Are you asking if RustDesk use those commands to locate `bin/sh`? And get `TERM` from `/bin/sh`?

RustDesk uses `sh` to spawn processes, but the specific shell is not critical.

It first tries to get the `TERM` value from the `rustdesk --server` process. If that fails, it then queries the shells for their `TERM` value

```bash
uid=$(loginctl | grep 'seat0' | awk '{print $2}')
term=$(ps -u $uid -f | grep "rustdesk \-\-server" | grep -v 'grep' | tail -1 | awk '{{print $2}}' | xargs -I__ cat /proc/__/environ 2>/dev/null | tr '\0' '\n' | grep '^TERM=' | tail -1 | sed 's/TERM=//g')
echo "TERM from rustdesk server: $term"

```

```bash
uid=$(loginctl | grep 'seat0' | awk '{print $2}')
for bin in bash zsh fish sh; do
    grep_pattern="$bin\$|bin/$bin"
    term=$(ps -u $uid -f | grep -E "${grep_pattern}" | grep -v 'grep' | tail -1 | awk '{{print $2}}' | xargs -I__ cat /proc/__/environ 2>/dev/null | tr '\0' '\n' | grep '^TERM=' | tail -1 | sed 's/TERM=//g')
    if [ -n "$term" ]; then
        echo "TERM from $bin: $term"
        cmd=$(ps -u $uid -f | grep -E "${grep_pattern}" | grep -v 'grep')
        echo -e "\nThe command of the bin is:\n$cmd"
        break
    fi
done
```

alonginwind:
> > Is it possible that the issue is with the systemd service? Running rustdesk --service directly works fine. But when started via the systemd service, the shell returned is always /bin/sh, even though /bin/bash exists.
> 
> > the shell returned is always /bin/sh
> 
> What do you mean?
> 
> Are you asking if RustDesk use those commands to locate `bin/sh`? And get `TERM` from `/bin/sh`?
> 
> RustDesk uses `sh` to spawn processes, but the specific shell is not critical.
> 
> It first tries to get the `TERM` value from the `rustdesk --server` process. If that fails, it then queries the shells for their `TERM` value
> 
> ```shell
> uid=$(loginctl | grep 'seat0' | awk '{print $2}')
> term=$(ps -u $uid -f | grep "rustdesk \-\-server" | grep -v 'grep' | tail -1 | awk '{{print $2}}' | xargs -I__ cat /proc/__/environ 2>/dev/null | tr '\0' '\n' | grep '^TERM=' | tail -1 | sed 's/TERM=//g')
> echo "TERM from rustdesk server: $term"
> ```
> 
> ```shell
> uid=$(loginctl | grep 'seat0' | awk '{print $2}')
> for bin in bash zsh fish sh; do
>     grep_pattern="$bin\$|bin/$bin"
>     term=$(ps -u $uid -f | grep -E "${grep_pattern}" | grep -v 'grep' | tail -1 | awk '{{print $2}}' | xargs -I__ cat /proc/__/environ 2>/dev/null | tr '\0' '\n' | grep '^TERM=' | tail -1 | sed 's/TERM=//g')
>     if [ -n "$term" ]; then
>         echo "TERM from $bin: $term"
>         cmd=$(ps -u $uid -f | grep -E "${grep_pattern}" | grep -v 'grep')
>         echo -e "\nThe command of the bin is:\n$cmd"
>         break
>     fi
> done
> ```

When running under a systemd service, the function get_default_shell() -> String

let shells = ["/bin/bash", "/bin/zsh", "/bin/sh"];
for shell in &shells {
    if std::path::Path::new(shell).exists() {
        return shell.to_string();   //here return /bin/sh
    }
}
returns /bin/sh even though /bin/bash does exist.

fufesou:
> When running under a systemd service, the function get_default_shell() -> String
> 
> let shells = ["/bin/bash", "/bin/zsh", "/bin/sh"]; for shell in &shells { if std::path::Path::new(shell).exists() { return shell.to_string(); //here return /bin/sh } } returns /bin/sh even though /bin/bash does exist.

This seems a bit surprising. And on my Ubuntu, the controlled side is launching `/bin/bash`.

`ps -ef | grep 'rustdesk --server' | grep -v 'grep' | grep -v 'sudo' | awk '{printf("ps -ef | grep %d\n", $2)}' | sh`

PR #13386, opened 2025-11-01
Eliminate build warnings from the main project

logiclrd:
This PR makes minor adjustments so that `rustdesk` builds without warnings.

Before:

```
C:\code\rustdesk>cargo build 2>&1 | find "warning"
warning: unused imports: `MOUSE_BUTTON_LEFT`, `MOUSE_TYPE_DOWN`, `MOUSE_TYPE_UP`, and `MOUSE_TYPE_WHEEL`
warning: use of deprecated function `hbb_common::base64::encode`: Use Engine::encode
warning: variable does not need to be mutable
warning: fields `x` and `y` are never read
warning: constant `MOUSE_MOVE_PROTECTION_TIMEOUT` is never used
warning: constant `MOUSE_ACTIVE_DISTANCE` is never used
warning: function `get_last_input_cursor_pos` is never used
warning: methods `read_empty_dirs`, `set_confirm_override_file`, and `rename_file` are never used
warning: function `query_online_states` is never used
warning: function `create_online_stream` is never used
warning: function `query_online_states_` is never used
warning: methods `http_request` and `get_http_status` are never used
warning: methods `into_string` and `save` are never used
warning: function `get_chatid_telegram` is never used
warning: static `CONTROLLING_SESSION_COUNT` is never used
warning: function `update_controlling_session_count` is never used
warning: function `http_request` is never used
warning: function `get_async_http_status` is never used
warning: function `has_valid_bot` is never used
warning: function `verify_bot` is never used
warning: enum `HbbHttpResponse` is never used
warning: struct `DownloadData` is never constructed
warning: struct `Downloader` is never constructed
warning: function `download_file` is never used
warning: function `do_download` is never used
warning: function `get_download_data` is never used
warning: function `cancel` is never used
warning: function `remove` is never used
warning: `rustdesk` (lib) generated 28 warnings (run `cargo fix --lib -p rustdesk` to apply 2 suggestions)
warning: unused import: `librustdesk::*`
warning: `rustdesk` (bin "service") generated 1 warning (run `cargo fix --bin "service"` to apply 1 suggestion)

C:\code\rustdesk>
```

After:

```
C:\code\rustdesk>cargo clean -p rustdesk
     Removed 401 files, 6.6GiB total

C:\code\rustdesk>cargo build
   Compiling rustdesk v1.4.3 (C:\code\rustdesk)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 40.47s

C:\code\rustdesk>
```

In this PR, each type of warning is addressed in a separate commit.

### DIFF
--- a/src/hbbs_http/sync.rs
+++ b/src/hbbs_http/sync.rs
@@ -184,10 +184,7 @@ async fn start_hbbs_sync_async() {
                         hasher.update(url.as_bytes());
                         hasher.update(&v.as_bytes());
                         let res = hasher.finalize();
-                        {
-                            #![allow(deprecated)]
-                            hash = hbb_common::base64::encode(&res[..]);
-                        }
+                        hash = crate::encode64(&res[..]);
                         let old_hash = config::Status::get("sysinfo_hash");
                         let ver = config::Status::get("sysinfo_ver"); // sysinfo_ver is the version of sysinfo on server's side
                         if hash == old_hash {
--- a/src/hbbs_http/sync.rs
+++ b/src/hbbs_http/sync.rs
@@ -184,7 +184,10 @@ async fn start_hbbs_sync_async() {
                         hasher.update(url.as_bytes());
                         hasher.update(&v.as_bytes());
                         let res = hasher.finalize();
-                        hash = hbb_common::base64::encode(&res[..]);
+                        {
+                            #![allow(deprecated)]
+                            hash = hbb_common::base64::encode(&res[..]);
+                        }
                         let old_hash = config::Status::get("sysinfo_hash");
                         let ver = config::Status::get("sysinfo_ver"); // sysinfo_ver is the version of sysinfo on server's side
                         if hash == old_hash {
--- a/src/platform/linux.rs
+++ b/src/platform/linux.rs
@@ -746,15 +746,6 @@ pub fn is_root() -> bool {
     crate::username() == "root"
 }
 
-fn is_opensuse() -> bool {
-    if let Ok(res) = run_cmds("cat /etc/os-release | grep opensuse") {
-        if !res.is_empty() {
-            return true;
-        }
-    }
-    false
-}
-
 pub fn run_as_user<I, K, V>(
     arg: Vec<&str>,
     user: Option<(String, String)>,
@@ -902,6 +893,15 @@ pub fn check_super_user_permission() -> ResultType<bool> {
 }
 
 /*
+fn is_opensuse() -> bool {
+    if let Ok(res) = run_cmds("cat /etc/os-release | grep opensuse") {
+        if !res.is_empty() {
+            return true;
+        }
+    }
+    false
+}
+
 pub fn elevate(args: Vec<&str>) -> ResultType<bool> {
     let cmd = std::env::current_exe()?;
     match cmd.to_str() {
@@ -1188,7 +1188,7 @@ mod desktop {
             }
             self.display = self
                 .display
-                .replace(&hbb_common::whoami::hostname(), "")
+                .replace(&hbb_common::whoami::fallible::hostname().unwrap_or_else(|_| "".to_string()), "")
                 .replace("localhost", "");
         }
 
@@ -1422,7 +1422,10 @@ mod desktop {
     }
 }
 
-pub struct WakeLock(Option<keepawake::AwakeHandle>);
+pub struct WakeLock(
+    #[allow(unused)]
+    Option<keepawake::AwakeHandle>
+);
 
 impl WakeLock {
     pub fn new(display: bool, idle: bool, sleep: bool) -> Self {
--- a/src/platform/linux.rs
+++ b/src/platform/linux.rs
@@ -1188,7 +1188,7 @@ mod desktop {
             }
             self.display = self
                 .display
-                .replace(&hbb_common::whoami::fallible::hostname().unwrap_or_else(|_| "".to_string()), "")
+                .replace(&hbb_common::whoami::fallible::hostname().unwrap_or_default(), "")
                 .replace("localhost", "");
         }
 
--- a/src/server.rs
+++ b/src/server.rs
@@ -703,7 +703,6 @@ async fn sync_and_watch_config_dir() {
                                 log::error!("sync config to root failed: {}", e);
                                 match crate::ipc::connect(1000, "_service").await {
                                     Ok(mut _conn) => {
-                                        conn = _conn;
                                         log::info!("reconnected to ipc_service");
                                     }
                                     _ => {}
--- a/src/server.rs
+++ b/src/server.rs
@@ -703,6 +703,7 @@ async fn sync_and_watch_config_dir() {
                                 log::error!("sync config to root failed: {}", e);
                                 match crate::ipc::connect(1000, "_service").await {
                                     Ok(mut _conn) => {
+                                        conn = _conn;
                                         log::info!("reconnected to ipc_service");
                                     }
                                     _ => {}
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -3,10 +3,8 @@ use super::{input_service::*, *};
 use crate::clipboard::try_empty_clipboard_files;
 #[cfg(not(any(target_os = "android", target_os = "ios")))]
 use crate::clipboard::{update_clipboard, ClipboardSide};
-#[cfg(any(target_os = "windows", target_os = "macos"))]
+#[cfg(any(target_os = "windows", target_os = "macos", feature = "unix-file-copy-paste"))]
 use crate::clipboard_file::*;
-#[cfg(feature = "unix-file-copy-paste")]
-use crate::unix_file_clip;
 #[cfg(target_os = "android")]
 use crate::keyboard::client::map_key_to_control_key;
 #[cfg(target_os = "linux")]
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -5,6 +5,8 @@ use crate::clipboard::try_empty_clipboard_files;
 use crate::clipboard::{update_clipboard, ClipboardSide};
 #[cfg(any(target_os = "windows", target_os = "macos"))]
 use crate::clipboard_file::*;
+#[cfg(feature = "unix-file-copy-paste")]
+use crate::unix_file_clip;
 #[cfg(target_os = "android")]
 use crate::keyboard::client::map_key_to_control_key;
 #[cfg(target_os = "linux")]
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -3,7 +3,7 @@ use super::{input_service::*, *};
 use crate::clipboard::try_empty_clipboard_files;
 #[cfg(not(any(target_os = "android", target_os = "ios")))]
 use crate::clipboard::{update_clipboard, ClipboardSide};
-#[cfg(any(target_os = "windows", target_os = "linux", target_os = "macos"))]
+#[cfg(any(target_os = "windows", target_os = "macos"))]
 use crate::clipboard_file::*;
 #[cfg(target_os = "android")]
 use crate::keyboard::client::map_key_to_control_key;
@@ -243,7 +243,7 @@ pub struct Connection {
     // by peer
     disable_audio: bool,
     // by peer
-    #[cfg(any(target_os = "windows", target_os = "linux", target_os = "macos"))]
+    #[cfg(any(target_os = "windows", feature = "unix-file-copy-paste"))]
     enable_file_transfer: bool,
     // by peer
     audio_sender: Option<MediaSender>,
@@ -414,7 +414,7 @@ impl Connection {
             multi_ui_session: false,
             ip: "".to_owned(),
             disable_audio: false,
-            #[cfg(any(target_os = "windows", target_os = "linux", target_os = "macos"))]
+            #[cfg(any(target_os = "windows", feature = "unix-file-copy-paste"))]
             enable_file_transfer: false,
             disable_clipboard: false,
             disable_keyboard: false,
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -3,7 +3,7 @@ use super::{input_service::*, *};
 use crate::clipboard::try_empty_clipboard_files;
 #[cfg(not(any(target_os = "android", target_os = "ios")))]
 use crate::clipboard::{update_clipboard, ClipboardSide};
-#[cfg(any(target_os = "windows", target_os = "macos", feature = "unix-file-copy-paste"))]
+#[cfg(any(target_os = "windows", feature = "unix-file-copy-paste"))]
 use crate::clipboard_file::*;
 #[cfg(target_os = "android")]
 use crate::keyboard::client::map_key_to_control_key;
--- a/src/server/display_service.rs
+++ b/src/server/display_service.rs
@@ -22,8 +22,12 @@ struct ChangedResolution {
     changed: (i32, i32),
 }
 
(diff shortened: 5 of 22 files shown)


rustdesk:
fix ci please.

logiclrd:
Huh. There's some code that _does_ require it to be mutable, but that code is only enabled for the `macos` target.

src/hbbs_http/sync.rs
```diff
@@ -184,7 +184,10 @@ async fn start_hbbs_sync_async() {
                         hasher.update(url.as_bytes());
                         hasher.update(&v.as_bytes());
                         let res = hasher.finalize();
-                        hash = hbb_common::base64::encode(&res[..]);
+                        {
+                            #![allow(deprecated)]
+                            hash = hbb_common::base64::encode(&res[..]);
+                        }
```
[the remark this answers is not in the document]
logiclrd: This is really 6 of one, half a dozen of the other 😛 I deliberately made a single-statement block to set it out, because I feel like allowing deprecated code is _usually_ something that should be avoided, and where it's not avoided, it should be noticed. The intent needs to be clear. Would you prefer that this be written as an outer attribute instead, though? The style preference for the code-base carries the most weight here.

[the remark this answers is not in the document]
rustdesk: should change to encode64 in src/common.rs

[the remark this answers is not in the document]
rustdesk: can ask chatgpt cl (or cursor ci, cc, codex etc) to do this, it will be quite quick.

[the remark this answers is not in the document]
logiclrd: Ah, nice, centralization of the override. 🙂

rustdesk:
Could you fix the CI?

logiclrd:
There are some OS X-specific warnings that I have no idea what to do with:

```
Warning: warning: rustdesk@1.4.3: src/platform/macos.mm:96:18: warning: 'AuthorizationExecuteWithPrivileges' is deprecated: first deprecated in macOS 10.7 [-Wdeprecated-declarations]
warning: rustdesk@1.4.3:         status = AuthorizationExecuteWithPrivileges(authRef, process, kAuthorizationFlagDefaults, args, &pipe);
warning: rustdesk@1.4.3:                  ^
warning: rustdesk@1.4.3: /Applications/Xcode_15.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/Authorization.h:427:10: note: 'AuthorizationExecuteWithPrivileges' has been explicitly marked deprecated here
warning: rustdesk@1.4.3: OSStatus AuthorizationExecuteWithPrivileges(AuthorizationRef authorization,
warning: rustdesk@1.4.3:          ^
Warning: warning: rustdesk@1.4.3: src/platform/macos.mm:134:30: warning: 'CGDisplayModeCopyPixelEncoding' is deprecated: first deprecated in macOS 10.11 - No longer supported [-Wdeprecated-declarations]
warning: rustdesk@1.4.3:         CFStringRef pixelEncoding = CGDisplayModeCopyPixelEncoding(mode);
warning: rustdesk@1.4.3:                                     ^
warning: rustdesk@1.4.3: /Applications/Xcode_15.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/CoreGraphics.framework/Headers/CGDirectDisplay.h:177:34: note: 'CGDisplayModeCopyPixelEncoding' has been explicitly marked deprecated here
warning: rustdesk@1.4.3: CG_EXTERN CFStringRef __nullable CGDisplayModeCopyPixelEncoding(
warning: rustdesk@1.4.3:                                  ^
```

logiclrd:
Also, my initial changes and observation of no more build warnings was on a Windows machine. I'm at my Linux machine now, and see a number of warnings that don't appear on Windows presumably because of conditional compilation; those bits are entirely excluded on Windows. Checking into these...

src/server.rs
```diff
@@ -703,7 +703,6 @@ async fn sync_and_watch_config_dir() {
                                 log::error!("sync config to root failed: {}", e);
                                 match crate::ipc::connect(1000, "_service").await {
                                     Ok(mut _conn) => {
-                                        conn = _conn;
```
rustdesk, line 706: Really? This creates a bug. That's why we do not do unnecessary refactory.

rustdesk (changes requested):

src/server/connection.rs
```diff
@@ -3,7 +3,7 @@ use super::{input_service::*, *};
 use crate::clipboard::try_empty_clipboard_files;
 #[cfg(not(any(target_os = "android", target_os = "ios")))]
 use crate::clipboard::{update_clipboard, ClipboardSide};
-#[cfg(any(target_os = "windows", target_os = "linux", target_os = "macos"))]
+#[cfg(any(target_os = "windows", target_os = "macos", feature = "unix-file-copy-paste"))]
```
rustdesk: why not ` #[cfg(any(target_os = "windows", feature = "unix-file-copy-paste"))]` as the other two?

rustdesk (changes requested):

rustdesk:
> There are some OS X-specific warnings that I have no idea what to do with:

These are useful warnings, should not be removed.

src/server.rs
logiclrd: You should turn on `[deny(warnings)]` so that the code can be correct without warnings in the first place. This is a very common practice.

logiclrd: Here's why I removed that line:

image

Is this just a bug in the analyzer, then?

logiclrd: I'm not convinced that removing this line does introduce a bug.

Here's a bit more context:

```
                                    Ok(mut _conn) => {
                                        conn = _conn;
                                        log::info!("reconnected to ipc_service");
                                        break;
                                    }
```

It `break`s out of the loop immediately there. But, the loop is the last thing in the scope that defines `conn` in the first place. Nothing can possibly read the value being assigned there, can it??

src/server/connection.rs
logiclrd: I was updating it iteratively. Building on Linux produced the warning, so I removed "linux". But, I guess I wasn't building with the feature enabled because that produced an error in CI. The original line contained the "macos" check and it just came along for the ride.

logiclrd: I actually don't know if it'd work without including it for the "macos" target OS. I guess I'll try removing it and then revert if the "macos" CI fails (I have no other way to test it).

src/platform/linux.rs
```diff
@@ -1188,7 +1188,7 @@ mod desktop {
             }
             self.display = self
                 .display
-                .replace(&hbb_common::whoami::hostname(), "")
+                .replace(&hbb_common::whoami::fallible::hostname().unwrap_or_else(|_| "".to_string()), "")
```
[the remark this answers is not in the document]
logiclrd: I copied this pattern from an existing call to `fallible::hostname`, but the place I copied from defaulted it to a non-empty string value. Since `Default::default` for `String` returns an empty string, this should work with `unwrap_or_default` instead.

src/server/display_service.rs
```diff
@@ -22,8 +22,12 @@ struct ChangedResolution {
     changed: (i32, i32),
 }
 
+#[cfg(windows)]
 lazy_static::lazy_static! {
     static ref IS_CAPTURER_MAGNIFIER_SUPPORTED: bool = is_capturer_mag_supported();
+}
+
+lazy_static::lazy_static! {
```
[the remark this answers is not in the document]
logiclrd: AI hallucination. This isn't the conventional practice because it doesn't work at all.

```
error[E0573]: expected type, found static `IS_CAPTURER_MAGNIFIER_SUPPORTED`
  --> src/server/display_service.rs:25:1
   |
25 | / lazy_static::lazy_static! {
26 | |     #[cfg(windows)]
27 | |     static ref IS_CAPTURER_MAGNIFIER_SUPPORTED: bool = is_capturer_mag_supported();
28 | | }
   | |_^ not a type
   |
   = note: this error originates in the macro `__lazy_static_internal` which comes from the expansion of the macro `lazy_static::lazy_static` (in Nightly builds, run with -Z macro-backtrace for more info)
```

There are plenty of examples in the code that use multiple `lazy_static!` macro invocations with different conditions exactly like I wrote here. For instance, in `hbb_common`'s `config.rs`:

```
#[cfg(target_os = "macos")]
lazy_static::lazy_static! {
    pub static ref ORG: RwLock<String> = RwLock::new("com.carriez".to_owned());
}

type Size = (i32, i32, i32, i32);
type KeyPair = (Vec<u8>, Vec<u8>);

lazy_static::lazy_static! {
    static ref CONFIG: RwLock<Config> = RwLock::new(Config::load());
    static ref CONFIG2: RwLock<Config2> = RwLock::new(Config2::load());
    ...
```

And in `client.rs`:

```
lazy_static::lazy_static! {
    pub(crate) static ref CUR_SESSION_ID: RwLock<SessionID> = Default::default(); // For desktop only
    static ref GLOBAL_EVENT_STREAM: RwLock<HashMap<String, StreamSink<String>>> = Default::default(); // rust to dart event channel
}

#[cfg(target_os = "windows")]
lazy_static::lazy_static! {
    pub static ref TEXTURE_RGBA_RENDERER_PLUGIN: Result<Library, LibError> = load_plugin_in_app_path("texture_rgba_renderer_plugin.dll");
}

#[cfg(target_os = "linux")]
lazy_static::lazy_static! {
    pub static ref TEXTURE_RGBA_RENDERER_PLUGIN: Result<Library, LibError> = Library::open("libtexture_rgba_renderer_plugin.so");
}

#[cfg(target_os = "macos")]
lazy_static::lazy_static! {
    pub static ref TEXTURE_RGBA_RENDERER_PLUGIN: Result<Library, LibError> = Library::open_self();
}
```

src/server/connection.rs
logiclrd: Nice, build succeeded. :slightly_smiling_face:

src/server.rs
rustdesk: You are right. Here is a bug, the `break` should be  removed.

rustdesk (changes requested):

logiclrd:
Rebased.

logiclrd:
There's another unused symbol warning that has the "feel" of a bug to it. It might just legitimately be unused, but since the last time this happened it was actually a logic error in the code, I figured I'd bring it up.

Specifically, the `sync_and_watch_config_dir` function in `src/server.rs` has a retry loop, and it has a local variable `synced` that starts out `false` but gets set to `true` once it has successfully synchronized configuration from the root. It skips the downstream synchronization process if it is already `synced`. But, immediately after that, as far as I can see it enters an unconditional loop watching for local changes to push back up to the root and there is no way for it to get out of that loop. The outer loop is effectively no longer active at this point, which means that the value assigned to `synced` won't ever be checked. I _think_ that means that `synced` is actually completely redundant with the current structure and should simply be removed?

Pseudocode:

```
synced = false;

for i in retries
{
  sleep(interval);

  result = ipc::connect(..);

  if (result is Ok)
  {
    if (!synced)
    {
      sync downstream
      synced = true;
    }

    loop forever
    {
      sleep(interval);

      if (configuration has changed)
      {
        sync upstream
        reconnect if that failed
      }
    }
  }
}
```

I think this would do exactly the same thing without `synced`:

```
for i in retries
{
  sleep(interval);

  result = ipc::connect(..);

  if (result is Ok)
  {
    sync downstream

    loop forever
    {
      sleep(interval);

      if (configuration has changed)
      {
        sync upstream
        reconnect if that failed
      }
    }
  }
}
```

logiclrd:
Also, it might improve code readability if it did a `break` out of the outer loop, to explicitly acknowledge that it isn't using that retry logic any more past the initial connection.

src/server/display_service.rs
```diff
@@ -22,8 +22,12 @@ struct ChangedResolution {
     changed: (i32, i32),
 }
 
+#[cfg(windows)]
 lazy_static::lazy_static! {
     static ref IS_CAPTURER_MAGNIFIER_SUPPORTED: bool = is_capturer_mag_supported();
+}
```
workingjubilee: Could be replaced by https://doc.rust-lang.org/std/sync/struct.LazyLock.html

```suggestion
static IS_CAPTURER_MAGNIFIER_SUPPORTED: LazyLock<bool> = LazyLock::new(|| {
    scrap::CapturerMag::is_supported()
});
```

Then you can remove the extraneous "forwarding" function.

src/tray.rs
```diff
@@ -57,7 +57,10 @@ fn make_tray() -> hbb_common::ResultType<()> {
     let icon = tray_icon::Icon::from_rgba(icon_rgba, icon_width, icon_height)
         .context("Failed to open icon")?;
 
+    #[cfg(target_os = "macos")]
     let mut event_loop = EventLoopBuilder::new().build();
+    #[cfg(not(target_os = "macos"))]
+    let event_loop = EventLoopBuilder::new().build();
```
workingjubilee: This can use a block that evaluates to the `event_loop` after its mutation so that arbitrary additional mutation doesn't occur:
```suggestion
    #[cfg(target_os = "macos")]
    let event_loop = {
        use tao::platform::macos::{EventLoopExtMacOS, ActivationPolicy};
        let mut event_loop = EventLoop::new();
        event_loop.set_activation_policy(ActivationPolicy::Accessory);
        event_loop
    };
    #[cfg(not(target_os = "macos"))]
    let event_loop = EventLoop::new();
```

src/tray.rs
workingjubilee: You could then delete this, of course.

src/server/display_service.rs
logiclrd: Nice 🙂

src/tray.rs
logiclrd: I think this would work, right?

```rust
    let event_loop = {
        #[allow(unused_mut)]
        let mut event_loop = EventLoopBuilder::new().build();

        #[cfg(target_os = "macos")]
        {
            use tao::platform::macos::EventLoopExtMacOS;
            event_loop.set_activation_policy(tao::platform::macos::ActivationPolicy::Accessory);
        }

        event_loop
    };
```

Looks like it compiles, and it eliminates a lot of the complexity/repetitiveness. 🙂

workingjubilee: Yep, that works too. Just different ways of saying "let's try to *not* mutate after this point, mmk?"

src/tray.rs
logiclrd: GitHub is being coy about which lines this comment refers to, since the commit it was made on has been superseded, but I _think_ this was referring to the OS X-specific `event_loop.set_activation_policy` which is now moved into the initialization block that limits the scope of `event_loop`s mutability?

workingjubilee: Yes.

workingjubilee: Apparently I cannot resolve this, but

logiclrd: > Apparently I cannot resolve this, but

As far as I can see, the lines are needed (at least, removing them would be a semantic change), but they can be moved into the constrained scope of mutability -- as is the case in the latest commit -- so that `event_loop` doesn't need to be mutable in the broader scope.

```
     let event_loop = {
         #[allow(unused_mut)]
         let mut event_loop = EventLoopBuilder::new().build();
 
>        #[cfg(target_os = "macos")]
>        {
>            use tao::platform::macos::EventLoopExtMacOS;
>            event_loop.set_activation_policy(tao::platform::macos::ActivationPolicy::Accessory);
>        }
 
         event_loop
     };
```

PR #13410, opened 2025-11-03
Edge scrolling: Configurable edge thickness

logiclrd:
This PR makes the thickness of the active scroll edge area for Edge Scrolling configurable.

The PR is formed of a series of logically-independent commits:

1. `CanvasModel` is updated to use a field instead of a local constant for the scroll edge thickness. Constants are added for minimum/maximum/default values.
2. UI elements are added to the Settings > Display section defined in `desktop_setting_page.dart` to represent the global default edge scroll edge thickness. A constant is added to represent the new setting in the underlying config. The UI has logic to automatically switch to ScrollEdge if the user interacts with the edge scroll edge thickness widgets, and to automatically submit changes to the text showing the edge thickness when it loses focus.
3. Preparatory changes in `remote_toolbar.dart`: The `_IconSubmenuButton` widget is reworked so that its `menuChildrenGetter` functor takes a pointer to the associated `_IconSubmenuButtonState`. A non-`@protected` `@override` of the `setState` method in `_IconSubmenuButtonState` is added, explicitly making the `setState` model for triggering UI rebuilds available to event handlers defined by `menuChildrenGetter`.
4. Preparatory changes in `remote_toolbar.dart`: The `_RectValueThumbShape` custom slider component shape is updated so that the units are configurable. They default to the '%' that was previously hardcoded, so as not to break any existing uses, but callers can now supply a custom `unit` value when constructing a `_RectValueThumbShape`.
5. UI elements are added to the Remote View toolbar defined in `remote_toolbar.dart` to represent the edge scroll edge thickness for the current session. The exposed `setState` method in the `state` parameter is used to trigger a rebuild when the user alters the current values in a way that keeps the menu pop-up open. The `ScrollEdge` menu item is updated to not close the menu pop-up on activation, allowing the user to further interact with the edge thickness controls. Also, if style other than `ScrollEdge` is selected and the user interacts with the edge scroll edge thickness slider, then the style is switched to `ScrollEdge` automatically. Underlying Rust methods for accessing the configuration in the "options" generic section of peer configuration are added, following the existing convention for other settings.
6. New strings are added to `src/lang/template.rs` and `res/lang.py` is rerun, propagating the new string blanks out to all the translations files. Values for English are provided.
7. An opportunistic fix to scroll-related code: Remove an unnecessary null assertion operator.
image

`RustDesk_default.toml`:

```
[options]
scroll_style = 'scrolledge'
edge_scroll_edge_thickness = '108'
```

image

```
MINGW64 ~/AppData/Roaming/RustDesk/config/peers
$ grep -B4 scroll 233165520.toml
    0,
    0,
]
view_style = 'original'
scroll_style = 'scrolledge'
--
[options]
zoom-cursor = ''
i444 = ''
custom-fps = '30'
edge_scroll_edge_thickness = '87'

MINGW64 ~/AppData/Roaming/RustDesk/config/peers
$
```


flutter/lib/desktop/pages/desktop_setting_page.dart
```diff
@@ -1678,29 +1678,114 @@ class _DisplayState extends State<_Display> {
...
+            children: <Widget>[
+              Text(translate('Scroll region thickness'),
+                textAlign: TextAlign.left,
+                style: TextStyle(fontSize: 12)),
+              Row(
+                children: <Widget>[
+                  Expanded(
+                    child: Semantics(
+                      label: translate('ScrollEdgeThicknessSlider'),
+                      value: edgeScrollEdgeThickness.toString(),
```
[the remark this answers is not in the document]
logiclrd: Just tested it with a pristine configuration, and there's no runtime error, but it does throw the word "null" into the text input.

rustdesk:
@21pages please advise @logiclrd how to use default display option correctly.

rustdesk:
Could you move ScrollEdge under ScrollBar? I guess people prefer ScrollEdge than ScrollBar, but I do not want to change old users' behavior.

rustdesk:
Great job!

src/client.rs
```diff
@@ -1976,13 +1976,27 @@ impl LoginConfigHandler {
...
         self.save_config(config);
     }
 
+    /// Save edge scroll edge thickness to the current config.
+    ///
+    /// # Arguments
+    ///
+    /// * `value` - The edge thickness to be saved.
+    pub fn save_edge_scroll_edge_thickness(&mut self, value: i32) {
+        self.set_option("edge_scroll_edge_thickness".to_string(), value.to_string());
```
21pages: t is recommended to add a field in the `PeerConfig` of  crate `hbb_common` instead of using `options`. You can refer to `trackpad_speed`, 

https://github.com/rustdesk/rustdesk/blob/b75f4daa47bd8899549d15a22defe65bdfc28d6f/src/client.rs#L2408

https://github.com/rustdesk/hbb_common/blob/a4053b929b14059b1bd116900de8a103d9d838ae/src/config.rs#L331.

21pages (changes requested):

src/client.rs
logiclrd: Okay, that's cool. I was leaning toward doing that but previous advice in a related area had been to not add dedicated fields but instead use generic name/value functionality. But, I definitely prefer having a field for it because then it sits alongside `scroll_style` in the serialized form, much neater. :-) Will switch.

logiclrd: Alright I have it all sorted locally. The `rustdesk` code, of course, can't work without the addition of the field to `hbb_common`, so I haven't yet pushed up the commit here. I'm waiting for the `hbb_common` change to go in before doing that. Let me know if I shouldn't wait and I'll go ahead and push this side of the changes too. 🙂

logiclrd: Thinking it through, the only reason to not push the changes is that you can't build/test this branch without the `hbb_common` changes, but you can just go into the submodule repository and fetch the commit and check it out, so there's really no reason not to.

logiclrd:
The last commit makes this PR depend on https://github.com/rustdesk/hbb_common/pull/428.

flutter/lib/consts.dart
```diff
@@ -79,6 +79,7 @@ const String kWindowEventOpenMonitorSession = "open_monitor_session";
 
 const String kOptionViewStyle = "view_style";
 const String kOptionScrollStyle = "scroll_style";
+const String kOptionEdgeScrollEdgeThickness = "edge_scroll_edge_thickness";
```
21pages: It is recommended to change it to `edge-scroll-edge-thickness`, as the use of underscores is for historical reasons.

21pages (changes requested):

flutter/lib/consts.dart
logiclrd: That feels weird directly adjacent to `view_style` and `scroll_style`, but I can of course make that change. 😛

flutter/lib/desktop/widgets/remote_toolbar.dart
```diff
@@ -1059,54 +1061,115 @@ class _DisplayMenuState extends State<_DisplayMenu> {
...
       final visible = data['visible'] as bool;
       if (!visible) return Offstage();
       final groupValue = data['scrollStyle'] as String;
-      onChange(String? value) async {
-        if (value == null) return;
+      final edgeScrollEdgeThickness = data['edgeScrollEdgeThickness'] as int;
+      final edgeScrollEdgeThicknessRange = data['edgeScrollEdgeThicknessRange'] as NumRange;
+      updateScrollStyle(String value) async {
+        print("updateScrollStyle HAPPING");
+        print(StackTrace.current.toString());
```
21pages: remove the log

flutter/lib/models/model.dart
```diff
@@ -1914,6 +1914,8 @@ class CanvasModel with ChangeNotifier {
   // scroll offset y percent
   double _scrollY = 0.0;
   ScrollStyle _scrollStyle = ScrollStyle.scrollauto;
+  // edge scroll mode: trigger scrolling when the cursor is close to the edge of the view
+  int _edgeScrollEdgeThickness = 100;
   // tracks whether edge scroll should be active, prevents spurious
```
21pages, line 1920: The initial value of `_edgeScrollEdgeThickness` is not being read from the configuration.

flutter/lib/desktop/pages/desktop_setting_page.dart
```diff
@@ -1678,29 +1678,111 @@ class _DisplayState extends State<_Display> {
...
           groupValue: groupValue,
           label: 'Scrollbar',
-          onChanged: isOptFixed ? null : onChanged),
+          onChanged: isOptFixed ? null : onScrollStyleChanged),
+      _Radio(context,
+          value: kRemoteScrollStyleEdge,
+          groupValue: groupValue,
+          label: 'ScrollEdge',
+          onChanged: isOptFixed ? null : onScrollStyleChanged),
+      Container(
```
21pages: Display the slider only when 'ScrollEdge' is selected, follow the custom image quality setting.

21pages (changes requested):

flutter/lib/desktop/widgets/remote_toolbar.dart
logiclrd: Facepalm. I keep missing these. I do remove many of them while committing. Sigh. I should just add a step to search the source tree for `print` before I start committing. Sorry 🙂

flutter/lib/desktop/pages/desktop_setting_page.dart
logiclrd: I like being able to select ScrollEdge by interacting with the slider, but I can switch it.

logiclrd:
Following a discussion in the `hbb_common` pull request https://github.com/rustdesk/hbb_common/pull/428 that makes corresponding configuration-related changes in tandem with this one, I have done a refactoring in `hbb_common` to improve consistency and move further toward there being a single source of truth for the number ranges in numeric configuration options. That refactoring results in changes further up the chain which are reflected in the latest commit here:

- The constants for custom FPS and custom image quality ranges are removed from `consts.dart`, as they are a duplication of the true underlying truth values in the Rust layer.

- Instead, code that was using them now relies on `NumRange` structures retrieved over the Rust/Dart bridge. All such minimum/maximum/default values are now specified in only one place in the source code.

- The custom quality value is now just a plain `int`/`i32` and not an `Int32List`/`Vec<i32>`, so the code that works with it is correspondingly simplified in both Dart and Rust code.

flutter/lib/consts.dart
```diff
@@ -79,7 +79,10 @@ const String kWindowEventOpenMonitorSession = "open_monitor_session";
 
 const String kOptionViewStyle = "view_style";
 const String kOptionScrollStyle = "scroll_style";
+const String kOptionEdgeScrollEdgeThickness = "edge-scroll-edge-thickness";
 const String kOptionImageQuality = "image_quality";
+const String kOptionCustomImageQuality = "custom_image_quality";
+const String kOptionCustomFps = "custom_fps";
```
21pages, line 85: "custom-fps"

flutter/lib/common/widgets/toolbar.dart
```diff
@@ -403,7 +406,7 @@ Future<List<TRadioMenu<String>>> toolbarImageQuality(
       groupValue: groupValue,
       onChanged: (value) {
         onChanged(value);
-        customImageQualityDialog(ffi.sessionId, id, ffi);
+        customImageQualityDialog(ffi.sessionId, id, ffi, customFpsRange, customQualityRange);
```
21pages, line 409: `customImageQualityDialog(SessionID sessionId, String id, FFI ffi, NumRange customQualityRange, NumRange customFpsRange)`

The order of the parameters needs to be adjusted.

flutter/lib/common/widgets/setting_widgets.dart
```diff
@@ -6,24 +6,28 @@ import 'package:flutter_hbb/consts.dart';
...
     required Function(double)? setFps,
     required bool showFps,
-    required bool showMoreQuality}) {
-  if (initQuality < kMinQuality ||
-      initQuality > (showMoreQuality ? kMaxMoreQuality : kMaxQuality)) {
-    initQuality = kDefaultQuality;
+    required bool showMoreQuality,
+    required NumRange customQualityRange,
+    required NumRange customFpsRange}) {
+  if (initQuality < customQualityRange.minimumValue ||
```
21pages, line 20: The logic has changed: when "more" is not selected, the maximum is 100%; when "more" is selected, the maximum is 2000% (the maximum value in hbb_common can be changed to 2000).

21pages (changes requested):

rustdesk:
@21pages take over this and have a simple implementation.

flutter/lib/consts.dart
logiclrd: I also found some duplicated hardcoded string references to this and `"custom_image_quality"` that should have been references to the constants.

flutter/lib/common/widgets/setting_widgets.dart
logiclrd: Ah, right. When I first wrote part, it was in the mind that the custom quality range would be the restricted range, and so this logic was written with the thinking that it needs to override the standard range when "more" is enabled. But, when I was in the config.rs code, I realized that the standard range would have to be the extended range, otherwise it would never be able to read extended-range values back from the config file. I didn't remember that I had this logic here based on the wrong assumption.

logiclrd: This was another case where the values were duplicated and didn't have a single source of truth. I've just pushed a couple of commits that fix the logic and then centralize the quality limits to a single source of truth.

logiclrd:
~@rustdesk I think you should reconsider this. I believe these changes are immensely valuable and considerably increase the quality of the code. They decrease tech debt, so that future work will be easier, cleaner and quicker with fewer opportunities for mistakes. The time spent on these issues isn't just time to get configurable edge thickness. Furthermore, I believe the changes just reached a point of maturity. Any time spent reviewing them is a sunk cost, and you now have a polished result. With these changes, the code has greater consistency and much less duplication. If you want to change a numeric range for any parameter, you do so in one place, period. You don't have to hunt down places throughout the code where a particular numeric literal is baked into expressions (well, assuming I haven't missed any).~

EDIT: This is all true but a monolithic PR like this is the wrong way to do it.

logiclrd:
Thinking further about this, I think the real problem here is that I grew the scope of this PR. The actual edge thickness UI here is, I think, not more complex than #13445 (the other implementation is better because it centralizes the UI into a widget), but I felt an urge to map other things onto the updated paradigms, and doing so locally was good because it proved the concept, but those changes shouldn't be in this PR. This was an error on my part.

PR #13445, opened 2025-11-06
edge scroll thickness adjustment

21pages:
This pr based on https://github.com/rustdesk/rustdesk/pull/13410  and simplifies the implementation.

https://github.com/user-attachments/assets/6e9d17ce-f108-4144-ae43-fa8bd8286e88


logiclrd:
Oh, centralizing it in its own widget is nice. :-)

PR #13487, opened 2025-11-11
Fix config sync reconnection retry loop

logiclrd:
Remove erroneous `break` from the retry loop in `sync_and_watch_config_dir` in `server.rs`.


logiclrd:
Per: https://github.com/rustdesk/rustdesk/pull/13386#discussion_r2513064584

PR #13530, opened 2025-11-15
Update flatpak runtime, linux-pam - Flathub sync

besdar:
Hello @rustdesk 

This PR includes the same changes you have already approved here: https://github.com/flathub/com.rustdesk.RustDesk/pull/24

I’ve checked the pipeline, and the Flatpak build succeeded: https://github.com/besdar/rustdesk/actions/runs/19388138318


besdar:
Hello @rustdesk,

I see that this pull request hasn’t been merged yet. Is there an issue with it? Please let me know if anything needs fixing.

rustdesk:
We are going to postpone this for now. The current 24.08 runtime still works and is supported, so we would rather avoid the extra maintenance churn from switching to 25.08 at this time. We can revisit this
  when 24.08 gets closer to EOL or when we have another reason to update the Flatpak runtime.

besdar:
Sure, @rustdesk, I’ll close this pull request because there’s another one on this matter here: https://github.com/rustdesk/rustdesk/pull/15059. Please keep in mind that [your app’s manifest on Flathub](https://github.com/flathub/com.rustdesk.RustDesk/blob/master/com.rustdesk.RustDesk.json) is already updated, and it appears users are not experiencing any issues with it.

PR #13531, opened 2025-11-16
Fix terminal clear command to remove residual output

alonginwind:
#13366 
Testing revealed that in the VT220 environment, the number of rows calculated internally by xterm does not match the actual visible rows on the screen (the UI can display partial rows, while xterm only supports integer rows). Therefore, the padding is now adjusted dynamically so that the visible rows align with xterm's internal row count. This eliminates residual content at the top after executing clear, providing a smoother experience. Testing also shows that this resolves the top-line residual issue on Windows.

### DIFF
--- a/flutter/lib/desktop/pages/terminal_page.dart
+++ b/flutter/lib/desktop/pages/terminal_page.dart
@@ -63,7 +63,13 @@ class _TerminalPageState extends State<TerminalPage>
 
     _terminalModel.onResizeExternal = (w, h, pw, ph) {
       _cellHeight = ph * 1.0;
-      setState(() {});
+
+      // Schedule the setState for the next frame
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        if (mounted) {
+          setState(() {});
+        }
+      });
     };
 
     // Register this terminal model with FFI for event routing
--- a/flutter/lib/desktop/pages/terminal_page.dart
+++ b/flutter/lib/desktop/pages/terminal_page.dart
@@ -101,9 +101,9 @@ class _TerminalPageState extends State<TerminalPage>
     super.dispose();
   }
 
-  // It ensures that the number of visible rows is an integer by computing the
+  // This method ensures that the number of visible rows is an integer by computing the
   // extra space left after dividing the available height by the height of a single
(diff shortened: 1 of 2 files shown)


flutter/lib/desktop/pages/terminal_page.dart
```diff
@@ -60,6 +61,11 @@ class _TerminalPageState extends State<TerminalPage>
     debugPrint(
         '[TerminalPage] Terminal model created for terminal ${widget.terminalId}');
 
+    _terminalModel.onResizeExternal = (w, h, pw, ph) {
+      _cellHeight = ph * 1.0;
```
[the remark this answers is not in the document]
alonginwind: In fact, ph is the height of a single line.

flutter/lib/desktop/pages/terminal_page.dart
```diff
@@ -60,6 +61,11 @@ class _TerminalPageState extends State<TerminalPage>
...
+      setState(() {});
```
fufesou: Exception here.

Maybe you can wrap `setState` in `WidgetsBinding.instance.addPostFrameCallback()`.

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Build scheduled during frame.
While the widget tree was being built, laid out, and painted, a new frame was scheduled to rebuild the widget tree.   
This might be because setState() was called from a layout or paint callback. If a change is needed to the widget tree, it should be applied as the tree is being built. Scheduling a change for the subsequent frame instead results in an interface that lags behind by one frame. If this was done to make your build dependent on a size measured at layout time, consider using a LayoutBuilder, CustomSingleChildLayout, or CustomMultiChildLayout. If, on the other hand, the one frame delay is the desired effect, for example because this is an animation, consider scheduling the frame in a post-frame callback using SchedulerBinding.addPostFrameCallback or using an AnimationController to trigger the animation.
```

fufesou (changes requested):

flutter/lib/desktop/pages/terminal_page.dart
alonginwind: Agreed.

fufesou (approved):

PR #13540, opened 2025-11-17
Build Documentation

logiclrd:
This PR adds a file BUILDING.md which details the steps needed to take an unconfigured system and get it set up to reliably build RustDesk. The sequence has been tested on Windows, multiple times on Linux, and sort-of on OS X.

There is also a small adjustment to `build.py` so that it can be leveraged to do Debug builds as well as Release builds. This allows the process of building and running locally to very closely mirror the C.I. and release processes.

For OS X, I managed to get access to an OS X VM and the build runs to completion. `flutter run` doesn't show any errors, but the application window never appears. I don't know how to diagnose this. I do wonder if it has something to do with code signing, since the project file references a specific Team that I am of course not a part of for certificates. I'm well out of my depth in this area, though.

The key change for OS X is that `project.pbxproj` hardcodes its reference to `liblibrustdesk.dylib` to the `target/release` directory. This makes me suspect that actually presently nobody runs RustDesk debug builds on OS X. But, in order to make `flutter run` work with a debug build of the Rust code, I came up with a way to make the project work with both configurations. I updated the reference to `liblibrustdesk.dylib` to pull it from an intermediate directory, and I added a Script Build Phase that stages the appropriate Rust build output file into that common intermediate location.


gabefair:
Hello, 
Can you sync your branch with master and provide screenshots of your change and it working?

logiclrd:
I've observed it working on Windows. I tried running through the steps afresh in my test Linux VM (from a clean baseline), and I ran into an apparent Rust compiler bug (an Internal Compilation Error during `hbb_common`, but only when it's being compiled from the root, not if I `cargo build` inside `libs/hbb_common`). It looks like the script is still working on Linux, I just don't have it end-to-end.

logiclrd:
Here's the build on Windows with this branch:

[buildtest.html](https://github.com/user-attachments/files/24976230/buildtest.html)

logiclrd:
Okay I got unblocked on that. There was a mistake in the flags to `flutter build` but this latest version runs to completion and produces a `.deb` file. I wonder if it would be worth adding a flag to make it skip the packaging, kind of like `--skip-portable-pack` in the Windows flow.

logiclrd:
This is the tail end of `python build.py --debug --flutter` on Linux:

image

BUILDING.md
```diff
@@ -0,0 +1,278 @@
...
+
+    1. The build of the Rust core crate underlying RustDesk is done using Cargo. The exact parameters to Cargo are important. The `build.py` script does the correct builds in sequence. At a minimum, the `--flutter` flag must be passed into `build.py`; without it, the resulting `librustdesk.dll` lacks the numerous exported functions that the Flutter front-end relies upon.
+
+        - Windows: The build processes underlying Rust crate compilation will require that the Visual C++ build tools be available in the ambient environment. Run these commands from a `Developer Command Prompt for Visual Studio`, or explicitly run `vsvars.
```
workingjubilee: They call it the other way around: https://learn.microsoft.com/en-us/visualstudio/ide/reference/command-prompt-powershell?view=visualstudio

Also, markdown formatting.
```suggestion
        - Windows: The build processes underlying Rust crate compilation will require that the Visual C++ build tools be available in the ambient environment. Run these commands from a Visual Studio Developer Command Prompt, or explicitly run `vsvars`.
```

BUILDING.md
```diff
@@ -0,0 +1,278 @@
...
+
+                Add the last line to `~/.zprofile` to make the change persist.
+        - `libtool` (`libtool-bin`)
+        They are often already installed, but if not, install with your distribution's package manager (e.g. `apt install pkg-config autoconf make cmake g++-13 ...`).
+
+        > [Linux] NB: On alternatives-based systems, if Clang is installed before GCC, then `/usr/bin/c++` might run Clang instead of GCC. If this happens, you will likely encounter build errors. One way to fix this might be to remove the Clang and GCC packages and then reinstall them starting with GCC `g++` (`g++-13`).
+
+        > [Linux] NB: On the newest systems, as of this writing, `g++` installs version 15. This is not compatible with all of the Rust crates needed by RustDesk and will cause build errors. When an earlier version is installed, however, the `/usr/bin/c++` link might not be configured to run it. You may need to explicitly configure your system's alternatives mechanism, with a command such as `update-alternatives --install /usr/bin/c++ c++ ``which g++-13`` 13`, or manually create a symbolic link from `/usr/bin/c++` to the correct path for a compatible `g++` version.
```
workingjubilee: This feels like a... confused phrasing? It is spiritually correct but "`g++` installs" is wrong because the *package manager* is what "installs" anything, and even referring to `g++` as a *package* is... unlikely. The actually-relevant package is probably `gcc`... *or* something like `make-devel` or `build-essential`.

Also, "newest" is somewhat confounding. For "rolling release" distros, you are always at the "newest", every time you update. Another way to phrase this might be...

```suggestion
        > [Linux] NB: If your distro provides gcc 15 for `c++`, then `g++` will not be compatible with all the C++ code used in RustDesk's build and will cause build errors. When an earlier version is installed, however, the `/usr/bin/c++` link might not be configured to run it. You may need to explicitly configure your system's alternatives mechanism, with a command such as `update-alternatives --install /usr/bin/c++ c++ ``which g++-13`` 13`, or manually create a symbolic link from `/usr/bin/c++` to the correct path for a compatible `g++` version.
```

The last also seems like a *bad* suggestion, though. Can't we just "shadow" the distro's `c++` for this build by providing an alternative one earlier in `$PATH`?

So setting

```shell
PATH=/somewhere/my/c++/compiler/for/this/build/lives:${PATH}
```

workingjubilee (commented):
Yeah, I encountered that build error, heh.

BUILDING.md
workingjubilee: > The last also seems like a *bad* suggestion, though.

Or, at least, poorly contained. You can set PATH for a single build but it's unlikely someone wants to mess with all their builds ever that reach for a `c++` compiler.

BUILDING.md
logiclrd: This is what I'm going on here:

image

image

(It used to be named "Developer Command Prompt for Visual Studio"; the name changed with the upgrade to 2026 "Insiders". Trendy name. Plep. 😛)

BUILDING.md
logiclrd: As far as I can tell, though, there is no `g++` that means use GCC 13's `g++`. The file that means that is named `g++-13`. It might be possible to locally alias `g++` to `g++-13`, but this isn't inherited, so it'd have to be done in the user's profile. I did a bunch of searches, and from what I was able to find, updating the alternative is, in fact, the canonical way to resolve the issue.

logiclrd: Hmm, one thought occurs to me -- it isn't an uncommon practice for build scripts to respect environment variables `$CC`, `$CXX`, `$GCC`, `$GXX`. I wonder if it'd work to point them at `gcc-13` / `g++-13`.

logiclrd: Heyyy, it looks like it works. 🙂

logiclrd: This might also solve the problem with Clang taking over `/usr/bin/c++`.

logiclrd: It does! End-to-end build with `/usr/bin/c++` pointing at Clang but `CXX` pointing at the result of `which g++-13`. Nice.

gabefair:
Some build instructions used by NixOS
https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/ru/rustdesk-flutter/package.nix

logiclrd:
> Some build instructions used by NixOS https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/ru/rustdesk-flutter/package.nix

These are not instructions for a human 😛 

There are some interesting things in there 🙂 

https://github.com/NixOS/nixpkgs/blob/3ab3bf806bd38c82f463bd55c9dadacf9cd34e08/pkgs/by-name/ru/rustdesk-flutter/package.nix#L80C3-L80C71

```
env.VCPKG_ROOT = "/homeless-shelter"; # idk man, makes the build go [..]
```

BUILDING.md
workingjubilee: ahhhh, then I relent! just the markdown format nit then.

workingjubilee:
```suggestion
        - Windows: The build processes underlying Rust crate compilation will require that the Visual C++ build tools be available in the ambient environment. Run these commands from a `Developer Command Prompt for Visual Studio`, or explicitly run `vsvars`.
```

logiclrd: Oh shoot :-) Good catch.

PR #13550, opened 2025-11-18
Fix mouse hover issue on menu bar when controlling Mac from iPad

### LINKED ISSUE #8789: Mouse hover issue on menu bar when remotely controlling Mac (ipad->mac)
### Discussed in https://github.com/rustdesk/rustdesk/discussions/8788

<div type='discussions-op-text'>

<sup>Originally posted by **ZhuJiaxin2** July 23, 2024</sup>
### Bug Description

When controlling Mac through my iPad, i use apple magic mouse to click on iPad. 
However, when my mouse hover over the menu bar area, the hover function could not work (e.g. my mouse clicks on the apple icon in the menu bar, and several selections were listed under, when my mouse hover downward to reach for the selections, the selections disapears), this issue does not appear in the middle of the screen, when my mouse right clicks and hovers for paste or copy selections.
Also, this has caused great inconvenience, for i could not remotely close my full screen window in Mac! (because of the hovering issue, when my mouse is up there, the close tab button does not appear!)
So grateful for your effort!

### How to Reproduce

I dunno

### Expected Behavior

in Mac, when my mouse click on a menu bar button, and then hover downward for option selection, the selections stay there instead of disappearing under the mouse.

iPad->Mac

1.2.6->1.2.6

### Screenshots

Click on menu bar, selections came out:
截屏2024-07-23 上午11 30 16
Expectation when mouse hover downward:
截屏2024-07-23 上午11 30 38
What actually happens:
截屏2024-07-23 上午11 30 28

The hidden close tab button:
截屏2024-07-23 上午11 40 00
The close tab button only appears when mouse hovers there, but this doesnt work on this case, for the hovering function in that area is disabled
截屏2024-07-23 上午11 40 05

### Additional Context

_No response_</div>

rustdesk:
Fixes #8789

The issue was that when using the virtual/floating mouse on mobile devices to control a Mac, mouse clicks would occur at the wrong position, particularly noticeable in the menu bar where hover menus would disappear immediately.

Root cause: The tapDown() method was sending mouse button events without ensuring the cursor position was updated on the server side first. This caused clicks to register at stale cursor positions.

Solution: Before sending a mouse down event, check if the pointer has moved since entering the session. If not, send a mouse move event first to update the cursor position, then wait briefly before sending the click event.

This ensures the remote cursor is at the correct position before any mouse button action is performed.

### DIFF
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -810,6 +810,10 @@ class InputModel {
   }
 
   Future<void> tapDown(MouseButtons button) async {
+    if (!_pointerMovedAfterEnter) {
+      refreshMousePos();
+      await Future.delayed(Duration(milliseconds: 10));
+    }
     await sendMouse('down', button);
   }
 


fufesou:
https://github.com/rustdesk/rustdesk/issues/8789 and https://github.com/rustdesk/rustdesk/pull/14515 have addressed https://github.com/rustdesk/rustdesk/issues/8789

PR #13601, opened 2025-11-22
Fix F-Droid 1.4.6 build

linsui:
I guess https://monitor.f-droid.org/builds/log/com.carriez.flutter_hbb/104040003#site-footer is caused by github api rate limit.

## Summary by CodeRabbit

* **Chores**
  * Updated build script with improved Android NDK dependency resolution, now sourcing configuration from F-Droid repositories instead of GitHub APIs.
  * Enhanced error handling to detect missing build dependencies and provide early failure notifications.
  * Fixed minor documentation inconsistencies in the build configuration.

### DIFF
--- a/flutter/build_fdroid.sh
+++ b/flutter/build_fdroid.sh
@@ -184,13 +184,9 @@ prebuild)
 	fi
 
 	# Map NDK version to revision
-
-	NDK_VERSION="$(wget \
-		-qO- \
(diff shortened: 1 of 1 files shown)


basilgello:
Thanks for fixing this @linsui!

@rustdesk I think this can be merged because the proposed fix does not use API but instead reads static webpage.

basilgello:
@rustdesk meow :)

PR #13747, opened 2025-12-08
fix: set TERM env variable for terminal to fix Delete key not working

### LINKED ISSUE #13621: Delete-key not working inside terminal
### Bug Description

When opening a terminal-connection and typing something, the delete key does not delete the last character.

### How to Reproduce

1. Open a terminal-connection
2. Type something
3. Press the delete-key
4. Nothing happens

### Expected Behavior

The last character should be deleted

iPad OS 26 -> Arch Linux (Flatpak)

1.4.4 -> 1.4.4

### Screenshots

Image

### Additional Context

_No response_

majiayu000:
## Summary

- Set `TERM=xterm-256color` environment variable when spawning PTY shell
- This ensures proper handling of terminal control sequences

## Problem

Delete/Backspace keys were not working in terminal connections, particularly when connecting from iPad to Linux.

**Root cause**: The PTY was spawning shell without setting the `TERM` environment variable, causing the shell to not properly interpret certain control character sequences.

## Solution

Added `cmd.env("TERM", "xterm-256color")` in `terminal_service.rs` before spawning the shell process. This tells the shell it's running in an xterm-compatible terminal, enabling proper handling of:
- Delete key (`\x7f`)
- Backspace key
- Arrow keys and other special keys

## Test plan

- [ ] Connect from iPad to Linux via terminal
- [ ] Type text and press Delete/Backspace
- [ ] Verify characters are correctly deleted
- [ ] Run `echo $TERM` and verify output is `xterm-256color`

Fixes #13621

### DIFF
--- a/src/server/terminal_service.rs
+++ b/src/server/terminal_service.rs
@@ -774,6 +774,11 @@ impl TerminalServiceProxy {
         #[allow(unused_mut)]
         let mut cmd = CommandBuilder::new(&shell);
 
+        // Set TERM environment variable to ensure proper handling of control sequences
+        // This fixes issues with Delete/Backspace keys not working correctly
+        // See: https://github.com/rustdesk/rustdesk/issues/13621
+        cmd.env("TERM", "xterm-256color");
+
         #[cfg(target_os = "windows")]
         if let Some(token) = &self.user_token {
             cmd.set_user_token(*token as _);


rustdesk:
fix https://github.com/rustdesk/rustdesk/issues/13621

fufesou:
RustDesk determines the `TERM` environment variable as follows:
1. Retrieve the `TERM` value from current shell processes using `get_cur_term()`
2. If unavailable, invoke `suggest_best_term()` to select the best terminal type, preferring `xterm-256color`

https://github.com/rustdesk/rustdesk/blob/bba57069a892ea91234b24622a260946d353d735/src/platform/linux.rs#L386

I'm not sure if this change might cause any issues.
Perhaps modern mainstream systems all support `xterm-256color`, which would provide a better experience. Or perhaps it could cause problems on some platforms.

We could set the `TERM` environment variable to `xterm-256color` in `src/platform/linux.rs`.

Because this commit would make the settings in `linux.rs` ineffective, which could be confusing.
https://github.com/rustdesk/rustdesk/blob/bba57069a892ea91234b24622a260946d353d735/src/platform/linux.rs#L386

fufesou:
Or maybe the method to get the environment variable should be improved on some machines  https://github.com/rustdesk/rustdesk/blob/bba57069a892ea91234b24622a260946d353d735/src/platform/linux.rs#L355

PR #13748, opened 2025-12-08
fix: debounce trackpad scroll to prevent stutter and reversal

### LINKED ISSUE #13403: Scrolling Issues
### Bug Description

I've noticed this primarily with horizontal scrolling (Mac to Mac client), but when connecting Mac to Windows, I have occasionally ran into the issue with vertical scrolling.

When using a trackpad to scroll with two-finger gestures, scrolling will "stutter" and go backwards.  Example:  You're trying to scroll to view text to the right of a column.  You using two-finger gesture on a Mac trackpad and move your fingers to the left to swipe to the left and show what's to the right of the cursor.  This will stutter and will shift back to the left instead of allowing you to continue swiping to view text to the right.

### How to Reproduce

Seems fairly consistent when connecting from a Mac to a Mac or Windows client and using a trackpad (either built-in with a MacBook or a Magic TrackPad accessory used with a Mac Studio).  The area I see the most problem with is Mac to Windows with Adobe Acrobat on the Windows machine.  Scrolling reverts back to its previous position.  I usually have things in full-screen mode, but it happens in regular screens as well.

With the scrolling behavior, I've tried scrolling slow, fast, etc. and it has no difference.  If I go very very fast it will scroll to the bottom and not exhibit behavior.  Normal scrolling causes the behavior.  If I use the scrollbars (clicking the scrollbar and sliding the bar), it does not exhibit the behavior.

### Expected Behavior

The scrolling should not revert back to previous position and should continue scrolling smoothly.

Mac to Windows 10/Mac

1.4.3 on all machines

### Screenshots

https://github.com/user-attachments/assets/b1e111d9-18ca-4e9a-b01e-8e2bd224b7a5

### Additional Context

_No response_

majiayu000:
## Summary

- Add scroll debouncing logic to filter micro-reversals from trackpad gesture streams
- Drop single-unit opposite-direction scroll events within 40ms window

## Problem

Trackpad scrolling stutters and goes backwards during normal-speed two-finger scrolling gestures. The issue primarily affects:
- Mac-to-Mac connections
- Mac-to-Windows connections (especially in Adobe Acrobat)
- Horizontal scrolling with two-finger gestures

**Root cause**: Trackpad hardware generates occasional micro-reversal events (tiny opposite-direction scroll steps) as artifacts of the gesture recognition, not intentional user input.

## Solution

Implemented a `ScrollState` struct that tracks:
- Last scroll direction (x and y)
- Timestamp of last scroll event

The debouncing logic filters out scroll events that:
1. Are single-unit movements (`abs(delta) == 1`)
2. Are in the opposite direction of the previous scroll
3. Occur within 40ms of the previous scroll event

This approach is conservative - it only drops obvious micro-reversals while preserving intentional direction changes and larger scroll movements.

## Test plan

- [ ] Mac → Mac: Test two-finger horizontal scrolling
- [ ] Mac → Windows: Test scrolling in various applications
- [ ] Verify fast scrolling still works normally
- [ ] Verify intentional direction changes are not blocked
- [ ] Verify scrollbar dragging is unaffected

Fixes #13403

### DIFF
--- a/src/server/input_service.rs
+++ b/src/server/input_service.rs
@@ -441,6 +441,68 @@ enum KeysDown {
     EnigoKey(u64),
 }
 
+/// State for scroll debouncing to filter micro-reversals from trackpad
+/// See: https://github.com/rustdesk/rustdesk/issues/13403
+#[derive(Default)]
+struct ScrollState {
+    last_direction_x: i32, // -1, 0, or 1
+    last_direction_y: i32,
+    last_scroll_time: Option<Instant>,
+}
+
+impl ScrollState {
+    const DEBOUNCE_MS: u64 = 40;
+
+    /// Check if a scroll event should be filtered (debounced)
+    /// Returns true if the event should be dropped
+    fn should_filter(&mut self, x: i32, y: i32) -> bool {
+        let now = Instant::now();
+        let dir_x = x.signum();
+        let dir_y = y.signum();
+
+        let should_drop = if let Some(last_time) = self.last_scroll_time {
+            let elapsed = now.duration_since(last_time).as_millis() as u64;
+
+            // Only filter single-unit opposite-direction scrolls within debounce window
+            if elapsed < Self::DEBOUNCE_MS {
+                let is_micro_reversal_x = dir_x != 0
+                    && self.last_direction_x != 0
+                    && dir_x == -self.last_direction_x
+                    && x.abs() == 1;
+                let is_micro_reversal_y = dir_y != 0
+                    && self.last_direction_y != 0
+                    && dir_y == -self.last_direction_y
+                    && y.abs() == 1;
+
+                is_micro_reversal_x || is_micro_reversal_y
+            } else {
+                false
+            }
+        } else {
+            false
+        };
+
+        // Update state if not dropping
+        if !should_drop {
+            if dir_x != 0 {
+                self.last_direction_x = dir_x;
+            }
+            if dir_y != 0 {
+                self.last_direction_y = dir_y;
+            }
+            self.last_scroll_time = Some(now);
+        }
+
+        should_drop
+    }
+
+    fn reset(&mut self) {
+        self.last_direction_x = 0;
+        self.last_direction_y = 0;
+        self.last_scroll_time = None;
+    }
+}
+
 lazy_static::lazy_static! {
     static ref ENIGO: Arc<Mutex<Enigo>> = {
         Arc::new(Mutex::new(Enigo::new()))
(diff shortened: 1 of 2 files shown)


src/server/input_service.rs
```diff
@@ -441,13 +441,76 @@ enum KeysDown {
...
+            if dir_y != 0 {
+                self.last_direction_y = dir_y;
+            }
+            self.last_scroll_time = Some(now);
+        }
+
+        should_drop
+    }
+
+    fn reset(&mut self) {
```
fufesou, line 499: This function is not called.

Maybe you missed calling it on disconnect.

src/server/input_service.rs
```diff
@@ -441,13 +441,76 @@ enum KeysDown {
...
+                let is_micro_reversal_x = dir_x != 0
+                    && self.last_direction_x != 0
+                    && dir_x == -self.last_direction_x
+                    && x.abs() == 1;
+                let is_micro_reversal_y = dir_y != 0
+                    && self.last_direction_y != 0
+                    && dir_y == -self.last_direction_y
+                    && y.abs() == 1;
+
+                is_micro_reversal_x || is_micro_reversal_y
```
fufesou: Maybe it should be `&&`.

The entire event is discarded if either the `X` or `Y` direction is slightly reversed.

**Scenario**: Last scroll was `(0, 5)`, current scroll is `(3, -1)` → `Y` is slightly reversed → the entire event is discarded, but the `X` direction of `3` is a valid scroll.

src/server/input_service.rs
```diff
@@ -441,13 +441,81 @@ enum KeysDown {
...
+                let is_micro_reversal_x = dir_x != 0
+                    && self.last_direction_x != 0
+                    && dir_x == -self.last_direction_x
+                    && x.abs() == 1;
+                let is_micro_reversal_y = dir_y != 0
+                    && self.last_direction_y != 0
+                    && dir_y == -self.last_direction_y
+                    && y.abs() == 1;
+
+                is_micro_reversal_x && is_micro_reversal_y
```
fufesou, line 477: Sorry, `&&` is  incorrect here, maybe `||` is better, my bad.

**Senario**: `(0, 5)`, `(0, -1)`, `&&` will not filter this scroll movement.
This is the most common scenario.

The previous scenario I methioned yestoday  is not a common one.

src/server/input_service.rs
```diff
@@ -441,13 +441,81 @@ enum KeysDown {
...
+            // Only filter single-unit opposite-direction scrolls within debounce window
+            if elapsed < Self::DEBOUNCE_MS {
+                let is_micro_reversal_x = dir_x != 0
+                    && self.last_direction_x != 0
+                    && dir_x == -self.last_direction_x
+                    && x.abs() == 1;
+                let is_micro_reversal_y = dir_y != 0
+                    && self.last_direction_y != 0
+                    && dir_y == -self.last_direction_y
+                    && y.abs() == 1;
```
fufesou, line 475: Is `==1` enough?

I can't reproduce this issue, I have to manually move my fingers a little back to test. But it's not a good test.

fufesou:
Have you reproduced Issue 13403? It's hard for me to reproduce it.

majiayu000:
Thanks for the PR\! I'll test this on my end to see if I can reproduce the issue and verify the fix. Will follow up with feedback after testing.

rustdesk:
close since it can not be verified
.

PR #13775, opened 2025-12-11
require login for note

21pages:
https://github.com/rustdesk/rustdesk/pull/13499

* When the option is enabled, a login dialog will pop up if not logged in. Mobile has also been tested.

[desktop_setting.webm](https://github.com/user-attachments/assets/4fd2cde0-a838-4983-bc34-b8ef3db143ce)


src/ui_session_interface.rs
```diff
@@ -567,6 +567,9 @@ impl<T: InvokeUiSession> Session<T> {
     }
 
     pub fn get_audit_server(&self, typ: String) -> String {
+        if LocalConfig::get_option("access_token").is_empty() {
+            return "".to_owned();
+        }
```
rustdesk, line 572: ??????

rustdesk (changes requested):

src/ui_session_interface.rs
rustdesk: Do not modify here.

21pages: https://github.com/rustdesk/rustdesk/blob/6a0da9cf09b80bba6ca731d3029210c5d839a77b/src/ui_session_interface.rs#L551

Same as 1.4.3

PR #13895, opened 2025-12-24
fix: set TERM environment variable for PTY shell to fix special keys

### LINKED ISSUE #13621: Delete-key not working inside terminal

majiayu000:
## Summary

This PR fixes issue #13621 where Delete, Home, End and other special keys were not working in terminal sessions, especially when connecting from iPad to Linux.

### Root Cause

The PTY shell spawned by `terminal_service.rs` did not have a `TERM` environment variable set. Without a proper `TERM` value, the shell cannot correctly interpret control character sequences for special keys like Delete (`\x7f` or `\033[3~`).

### Solution

Added a `determine_shell_term()` function that intelligently determines the appropriate TERM value:

1. **Inherits** TERM from the current environment if it's valid (not empty, "unknown", or "dumb")
2. **Detects** tmux/screen sessions and uses `screen-256color` for compatibility
3. **Falls back** to `xterm-256color` as the default, which is widely supported

This approach is consistent with the existing TERM detection logic in `src/platform/linux.rs` (`suggest_best_term()` and `get_cur_term()`), avoiding the hardcoding issue mentioned in the review of PR #13747.

### Why This is Better Than PR #13747

| Aspect | PR #13747 (Hardcoded) | This PR |
|--------|----------------------|---------|
| Flexibility | Always uses `xterm-256color` | Adapts to environment |
| tmux/screen support | No | Yes, uses `screen-256color` |
| Environment inheritance | No | Yes, respects existing TERM |
| Consistency with linux.rs | No | Yes, uses same logic |

### Changes

- Added `determine_shell_term()` function in `terminal_service.rs`
- Applied TERM setting in `handle_open()` before spawning the shell

### Testing

- [ ] Test Delete/Backspace/Home/End keys work from iPad to Linux
- [ ] Test in tmux/screen sessions
- [ ] Test on various Linux distributions (Ubuntu, Debian, CentOS)
- [ ] Test Windows and macOS clients

Closes #13621

🤖 Generated with [Claude Code](https://claude.ai/code)


fufesou:
This PR has no effect because `linux.rs` already always sets the `TERM` environment variable.

https://github.com/rustdesk/rustdesk/blob/b69e871f9a7c60ebdb2907bfeb37dd05d47ca9fa/src/platform/linux.rs#L384

The change simply reads `TERM` and sets it back to the same value.

majiayu000:
@fufesou Thank you for the review! You're right - I now understand the issue better.

My current implementation is indeed redundant because:
1. If `linux.rs` already sets TERM correctly via `get_cur_term() + suggest_best_term()`, terminal_service will inherit it
2. My `determine_shell_term()` just reads the inherited TERM and sets it back

**Question**: What do you think is the real root cause of issue #13621?

Looking at the code flow:
1. `try_start_server_()` in `linux.rs:384` sets TERM for the server process
2. But `get_cur_term()` relies on `get_env()` which runs a complex shell command to read from `/proc/<pid>/environ`
3. Could the issue be that `get_env()` fails in certain environments (like Flatpak, or when connecting from iPad)?

Should the fix be in `linux.rs` to improve `get_env()` or `get_cur_term()` instead? I'd appreciate your guidance on the correct approach.

fufesou:
> Should the fix be in linux.rs to improve `get_env()` or `get_cur_term()` instead? I'd appreciate your guidance on the correct approach.

Yes. I think the possible reason is `get_env()` does not return the best `TERM` value.

The only filter is `!INVALID_TERM_VALUES.contains(&term.as_str())` and `get_env()` is not a good method to get the `TERM` value here.

Instead of calling shell commands, you could retrieve the `TERM` environment variable by checking `["bash", "zsh", "fish", "sh"]`. One approach is to iterate through `/proc/<pid>/environ` to read their environment variables directly.

https://github.com/rustdesk/rustdesk/blob/b69e871f9a7c60ebdb2907bfeb37dd05d47ca9fa/src/platform/linux.rs#L952

Then select the best available value — for example, preferring `xterm-256color` over `xterm`, similar to what `suggest_best_term()` does.

You can also cache the fetched `TERM`s.
If the cached `TERM`s contain `xterm-256color`, you can always use it the next time.

majiayu000:
The CI failure is due to `No space left on device` error on the GitHub Actions runner - this is not related to the PR changes.

```
##[error]No space left on device : '/home/runner/actions-runner/cached/_diag/pages/...'
```

Could you please re-run the CI?

fufesou:
Rerunning it may not help.

I'll try freeing up some disk space. https://github.com/marketplace/actions/free-disk-space-ubuntu

rustdesk:
https://github.com/rustdesk/rustdesk/pull/13930

PR #13903, opened 2025-12-25
Fix: Window positioning out of bounds on multi-monitors setup #13828

subwaycookiecrunch:
Bug Description:
The previous window positioning logic relied on a global bounding box that encompassed all connected monitors. In non-rectangular multi-monitor layouts (for example, L-shaped arrangements or displays with gaps), this allowed the application window to be restored to coordinates that were technically inside the bounding box but visually located in a “void” area where no physical monitor exists. As a result, the window could appear missing or inaccessible.

Changes:
Updated the window restore logic in flutter/lib/common.dart to validate window positions against actual screen geometries rather than a global bounding box.

Replaced global min/max coordinate validation with a strict screen-intersection check.

The saved window position is now validated by iterating through window_size.getScreenList().

A position is considered valid only if the window rectangle intersects with at least one screen’s visibleFrame by a minimum of 10×10 pixels.

If the saved position is invalid (off-screen or within a gap between monitors), the window safely defaults to being centered on the primary display.

Result:
This ensures restored windows always appear on a visible monitor, even in complex or non-rectangular multi-monitor configurations.

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -1932,44 +1932,41 @@ Future<Offset?> _adjustRestoreMainWindowOffset(
     return null;
   }
 
-  double? frameLeft;
-  double? frameTop;
-  double? frameRight;
-  double? frameBottom;
-
   if (isDesktop || isWebDesktop) {
-    for (final screen in await window_size.getScreenList()) {
-      frameLeft = frameLeft == null
-          ? screen.visibleFrame.left
-          : min(screen.visibleFrame.left, frameLeft);
-      frameTop = frameTop == null
-          ? screen.visibleFrame.top
-          : min(screen.visibleFrame.top, frameTop);
-      frameRight = frameRight == null
-          ? screen.visibleFrame.right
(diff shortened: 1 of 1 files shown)


fufesou:
@subwaycookiecrunch Hi, thanks for your PR.

I'll review this PR again. It appears solid.

One question: did you reproduce the problem from issue #13828?
It would be better to understand the underlying cause before implementing a guard.

fufesou:
I thought of a scenario that might be like this:

1. The user has 3 monitors. When closing a window, the window's position is saved on one of these monitors at coordinates `(2000, 100)`.

2. The user disconnects that monitor, then starts RustDesk.

3. At this point, the monitor arrangement is based on 2 monitors. If the arrangement happens to be `(0,0) - (1920,1080)` and `(1920,0) - (3840,1080)`, then RustDesk might be positioned at `(2000, 100)`, making the window invisible.


Or the displays are rearranged after closing the RustDesk window

This PR would indeed solve this problem.

fufesou (approved):

PR #13944, opened 2026-01-02
fix: restart video service when first frame stalls

godnight10061:
### Problem
In some cases the capturer can return `Ok(frame)` while the frame is invalid (`!frame.valid()`). Previously this was treated as a successful capture, so the normal `WouldBlock` ("no image yet") handling didn't run and peers could remain stuck waiting for the first image.

### Approach
- Treat invalid frames as `WouldBlock` so existing retry/fallback paths run (including Windows `try_gdi`).
- Add a first-frame watchdog: if no frame is actually sent to any subscriber within ~3s, restart the video service via the existing `SWITCH` mechanism.

### Tests
- `cargo test`

### DIFF
--- a/src/client/io_loop.rs
+++ b/src/client/io_loop.rs
@@ -282,8 +282,18 @@ impl<T: InvokeUiSession> Remote<T> {
                         }
                         _ = status_timer.tick() => {
                             if self.watchdog.check(self.is_connected, self.first_frame, self.handler.is_default(), Instant::now()) {
-                                log::warn!("No first video frame received, sending refresh");
-                                let display = self.handler.lc.read().unwrap().peer_info.as_ref().map(|p| p.current_display).unwrap_or(0);
+                                let (id, display) = {
+                                    let lch = self.handler.lc.read().unwrap();
+                                    let display = lch
+                                        .peer_info
+                                        .as_ref()
+                                        .map(|p| p.current_display)
+                                        .unwrap_or(0);
+                                    (lch.id.clone(), display)
+                                };
+                                log::warn!(
+                                    "No first video frame received, id={id}, display={display}, sending refresh"
+                                );
                                 self.handler.refresh_video(display as _);
                             }
 
@@ -1292,8 +1302,14 @@ impl<T: InvokeUiSession> Remote<T> {
         if let Ok(msg_in) = Message::parse_from_bytes(&data) {
             match msg_in.union {
                 Some(message::Union::VideoFrame(vf)) => {
+                    let display = vf.display as usize;
                     if !self.first_frame {
                         self.first_frame = true;
+                        log::info!(
+                            "First video frame received, id={}, display={}",
+                            self.handler.get_id(),
+                            display
+                        );
                         self.handler.close_success();
                         self.handler.adapt_size();
                         self.send_toggle_virtual_display_msg(peer).await;
@@ -1301,7 +1317,6 @@ impl<T: InvokeUiSession> Remote<T> {
                     }
                     self.video_format = CodecFormat::from(&vf);
 
-                    let display = vf.display as usize;
                     if !self.video_threads.contains_key(&display) {
                         self.new_video_thread(display);
                     }
--- a/src/client/io_loop.rs
+++ b/src/client/io_loop.rs
@@ -68,6 +68,7 @@ pub struct Remote<T: InvokeUiSession> {
     last_update_jobs_status: (Instant, HashMap<i32, u64>),
     is_connected: bool,
     first_frame: bool,
+    watchdog: ClientFirstFrameWatchdog,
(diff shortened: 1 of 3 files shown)


rustdesk:
@21pages @fufesou both review this.

fufesou:
@godnight10061 Hi, this PR looks good.

I'll test more.

But I have two questions at the moment:

1. Why did you use `waited: Duration::ZERO` instead of `waited: Instant::now()`?

The current implementation:

```rust
fn on_no_frame(&mut self, tick: Duration) -> bool {
    self.waited = self.waited.saturating_add(tick);
    self.waited >= self.timeout
}
```
seems more complex than:

```rust
fn on_no_frame(&mut self) -> bool {
    self.waited.elapsed() >= self.timeout
}
```

2. Why did you add the `FirstFrameWatchdog` guard in `video_service.rs` rather than on the controlling side? Have you been able to reproduce the issue or the following log entry?

```rust
log::warn!(
    "No first video frame for {first_frame_timeout:?}, restarting video service"
);
```

godnight10061:
> @godnight10061 Hi, this PR looks good.
> 
> I'll test more.
> 
> But I have two questions at the moment:
> 
> 1. Why did you use `waited: Duration::ZERO` instead of `waited: Instant::now()`?
> 
> The current implementation:
> 
> ```rust
> fn on_no_frame(&mut self, tick: Duration) -> bool {
>     self.waited = self.waited.saturating_add(tick);
>     self.waited >= self.timeout
> }
> ```
> 
> seems more complex than:
> 
> ```rust
> fn on_no_frame(&mut self) -> bool {
>     self.waited.elapsed() >= self.timeout
> }
> ```
> 
> 2. Why did you add the `FirstFrameWatchdog` guard in `video_service.rs` rather than on the controlling side? Have you been able to reproduce the issue or the following log entry?
> 
> ```rust
> log::warn!(
>     "No first video frame for {first_frame_timeout:?}, restarting video service"
> );
> ```

@fufesou 1. Good catch. switched to Instant + elapsed() for simpler logic; kept deterministic tests by injecting instants (no real sleeps).
2. Kept the guard in video_service.rs because only the capture/encode loop knows whether a frame was actually sent and it can reuse the existing bail!(\"SWITCH\") restart path; doing this from the controlling side would require extra signaling/state.

fufesou:
> Kept the guard in video_service.rs because only the capture/encode loop knows whether a frame was actually sent and it can reuse the existing bail!("SWITCH") restart path; doing this from the controlling side would require extra signaling/state.

@godnight10061 Thanks for the update.

For point 2, I still feel this guard is more robust on the controlling side.

On the controlled side, we only know that a frame was produced and scheduled to be sent.

We don’t know whether the controlling side actually received and rendered the first frame. Decode/render issues or some other issues would still leave the user stuck on “waiting for image…”, but the controlled-side watchdog would consider the first frame “done”.

If we have clear evidence that in some cases the controlled side truly never sends a first frame, then adding a watchdog in `video_service.rs` makes sense as a targeted fix. Or you have reproduced the "first frame stalls" issue, and it is fixed after applying this commit.

And in that case, I think we should also look for the underlying reason why the video service failed to send (or capture) the first frame.

fufesou:
@godnight10061 Hi, thanks for the update.

This guard appears to be helpful.

Additionally, could you confirm whether you've experienced the "Waiting for image" issue and if this commit addresses it?

godnight10061:
> @godnight10061 Hi, thanks for the update.
> 
> This guard appears to be helpful.
> 
> Additionally, could you confirm whether you've experienced the "Waiting for image" issue and if this commit addresses it?

@fufesou Thanks for the feedback. 

You're right - I haven't personally reproduced the "first frame stalls" issue yet; I replicated it based on user reports and existing discussions like #5609 and #12574. The logic assumes the capturer occasionally returns `Ok(frame)` with `!frame.valid()`, which prevents the normal WouldBlock retry path. I'll try to reproduce this consistently (e.g., specific OS/driver/capturer mode combinations) and verify if this fix resolves it.

godnight10061:
I still don’t have a stable local repro for the full “Waiting for image…” flow, but there’s a concrete DXGI code-path that can yield “Ok but invalid” textures: in libs/scrap/src/dxgi/mod.rs get_texture() calls
  QueryInterface(IID_ID3D11Texture2D, …) without checking HRESULT, so the out texture ptr can remain null and
  propagate as Frame::Texture that fails valid().

  If you (or anyone) can reproduce, could you share logs from both sides for ~10s around connect + Windows version,
  GPU/driver, monitor topology, and what was happening (Win+L/UAC/monitor hotplug/Win+Ctrl+Shift+B)? That’ll help
  confirm whether we’re hitting invalid frames vs a decode/render issue.

rustdesk:
We will hode this PR until we get a reproduction.

PR #13974, opened 2026-01-07
Allow configuring remote control permissions for different users

21pages:
Related:
  - https://github.com/rustdesk/rustdesk-server-pro/discussions/718
  - https://github.com/rustdesk/hbb_common/pull/477

## Implementation Details

  1. Permission info is received via server's `PunchHole`, `FetchLocalAddr`, and `RequestRelay` responses, then passed directly to `Connection::start`. Direct IP access is not restricted.

  2. Control permissions now support both enable and disable states (tri-state: enable/disable/not set). Control permissions take priority over local settings.

  3. Permission calculation rules:
     - **Session-level permissions**: Control permission overrides local setting when set
     - **Remote configuration modification** (global): If any connection has it disabled → disabled; else if any has it enabled → enabled; else use local setting
  4. Only supported when the controlled side is a desktop; Android is not supported.

  ## Permission Calculation Table

  ### Session-level Permissions

  | Control Permission | Local Setting | Result |
  |---|---|---|
  | Enable | Enable | Enable |
  | Enable | Disable | **Enable** |
  | Disable | Enable | **Disable** |
  | Disable | Disable | Disable |
  | Not Set | Enable | Enable |
  | Not Set | Disable | Disable |

  ### Global: Remote configuration modification

  | All Connections' Control Permission | Result |
  |---|---|
  | Any Disable | **Disable** |
  | No Disable, Any Enable | **Enable** |
  | All Not Set | Use local setting |

  ## Known Permission Issue
  - The remote configuration modification of the control role is checked using a timer, so it cannot be guaranteed to take effect during the first second immediately after a connection is established.
  - Clipboard file transfer on Windows: Context::enable is not called when cm panel permission changes .

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -3039,10 +3039,21 @@ Future<void> start_service(bool is_start) async {
 }
 
 Future<bool> canBeBlocked() async {
-  var access_mode = await bind.mainGetOption(key: kOptionAccessMode);
+  // First check control permission
+  final controlPermission = await bind.mainGetCommon(
+      key: "is-remote-modify-enabled-by-control-permissions");
+  if (controlPermission == "true") {
+    return false;
+  } else if (controlPermission == "false") {
+    return true;
+  }
+
+  // Check local settings
+  var accessMode = await bind.mainGetOption(key: kOptionAccessMode);
+  var isCustomAccessMode = accessMode != 'full' && accessMode != 'view';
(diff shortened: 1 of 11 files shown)


fufesou (approved):

ngocanh2704:
When I play the game Legend of Ymir, the mouse doesn't work at all, but when using Parsec or Moonlight, the mouse works completely normally.

PR #14102, opened 2026-01-21
feat(macos): initial privacy mode support [a simple try]

bin-haw:
## Summary
Add privacy mode functionality for macOS platform, allowing remote desktop sessions to hide the screen content from local users.

**Note to Reviewers:** As an amateur developer, this is my initial and minimal viable implementation (a simple try) to address the missing privacy mode on macOS. 

### Core Implementation (src/platform/macos.mm)
- Implement screen blackout using CGDisplayGammaTable API
- Implement input blocking using CGEventTap to intercept keyboard/mouse
- Store and restore original gamma values for proper cleanup

### Privacy Mode Integration (src/privacy_mode.rs, src/privacy_mode/macos.rs)
- Add macOS privacy mode implementation with PrivacyMode trait
- Register macOS privacy mode in PRIVACY_MODE_CREATOR
- Set DEFAULT_PRIVACY_MODE_IMPL for macOS platform
- Implement get_supported_privacy_mode_impl() for macOS

### Connection Handling (src/server/connection.rs)
- Add supported_privacy_mode_impl to platform_additions for macOS
- Enable privacy mode toggle in client UI when connecting via LAN IP

### Localization (src/lang/*.rs)
- Add "privacy_mode_impl_macos_tip" translation for en/cn/tw

## Safety & Security
- Implements Drop trait to ensure cleanup on normal exit
- macOS system automatically restores gamma table on process termination
- CGEventTap is automatically released when process terminates
- Tested with SIGKILL to verify crash recovery

## Testing
- Verified privacy mode toggle works via both ID and LAN IP connection
- Verified screen recovery after process crash (kill -9)
- Verified input restoration after process termination

### Environment
* **OS**: macOS 26.2
* **Hardware**: MacBook Pro (Apple Silicon)
* **Monitors Verified**:
    * Internal Liquid Retina display
    * Apple Studio Display
    * Additional external monitor

### DIFF
--- a/src/lang/cn.rs
+++ b/src/lang/cn.rs
@@ -562,6 +562,7 @@ pub static ref T: std::collections::HashMap<&'static str, &'static str> =
         ("id_input_tip", "可以输入 ID、直连 IP，或域名和端口号（<域名>:<端口号>）。\n要访问另一台服务器上的设备，请附加服务器地址（<ID>@<服务器地址>?key=<密钥>）。比如，\n9123456234@192.168.16.1:21117?key=5Qbwsde3unUcJBtrx9ZkvUmwFNoExHzpryHuPUdqlWM=。\n要访问公共服务器上的设备，请输入 \"<ID>@public\"，无需密钥。\n\n如果您想要在首次连接时，强制走中继连接，请在 ID 的后面添加 \"/r\"，例如，\"9123456234/r\"。"),
         ("privacy_mode_impl_mag_tip", "模式 1"),
         ("privacy_mode_impl_virtual_display_tip", "模式 2"),
+        ("privacy_mode_impl_macos_tip", "隐私模式"),
         ("Enter privacy mode", "进入隐私模式"),
         ("Exit privacy mode", "退出隐私模式"),
         ("idd_not_support_under_win10_2004_tip", "不支持 Indirect display driver 。需要 Windows 10 版本 2004 及更高的版本。"),
--- a/src/lang/cn.rs
+++ b/src/lang/cn.rs
@@ -562,7 +562,6 @@ pub static ref T: std::collections::HashMap<&'static str, &'static str> =
         ("id_input_tip", "可以输入 ID、直连 IP，或域名和端口号（<域名>:<端口号>）。\n要访问另一台服务器上的设备，请附加服务器地址（<ID>@<服务器地址>?key=<密钥>）。比如，\n9123456234@192.168.16.1:21117?key=5Qbwsde3unUcJBtrx9ZkvUmwFNoExHzpryHuPUdqlWM=。\n要访问公共服务器上的设备，请输入 \"<ID>@public\"，无需密钥。\n\n如果您想要在首次连接时，强制走中继连接，请在 ID 的后面添加 \"/r\"，例如，\"9123456234/r\"。"),
         ("privacy_mode_impl_mag_tip", "模式 1"),
         ("privacy_mode_impl_virtual_display_tip", "模式 2"),
-        ("privacy_mode_impl_macos_tip", "隐私模式"),
         ("Enter privacy mode", "进入隐私模式"),
         ("Exit privacy mode", "退出隐私模式"),
         ("idd_not_support_under_win10_2004_tip", "不支持 Indirect display driver 。需要 Windows 10 版本 2004 及更高的版本。"),
--- a/src/platform/macos.mm
+++ b/src/platform/macos.mm
@@ -389,6 +389,10 @@ static bool ApplyBlackoutToDisplay(CGDirectDisplayID display) {
     return false;
 }
 
+// Forward declaration - defined later in the file
+// Must be called while holding g_privacyModeMutex
+static bool TurnOffPrivacyModeInternal();
+
 // Display reconfiguration callback to handle display connect/disconnect events
 static void DisplayReconfigurationCallback(CGDirectDisplayID display, CGDisplayChangeSummaryFlags flags, void *userInfo) {
     (void)userInfo;
@@ -414,7 +418,8 @@ static void DisplayReconfigurationCallback(CGDirectDisplayID display, CGDisplayC
         NSLog(@"Display %u added during privacy mode, applying blackout", (unsigned)display);
         std::string uuid = GetDisplayUUID(display);
         if (uuid.empty()) {
-            NSLog(@"Failed to get UUID for newly added display %u", (unsigned)display);
+            NSLog(@"Failed to get UUID for newly added display %u, exiting privacy mode", (unsigned)display);
+            TurnOffPrivacyModeInternal();
             return;
         }
         
@@ -430,12 +435,24 @@ static void DisplayReconfigurationCallback(CGDirectDisplayID display, CGDisplayC
                     all.insert(all.end(), green.begin(), green.begin() + sampleCount);
                     all.insert(all.end(), blue.begin(), blue.begin() + sampleCount);
                     g_originalGammas[uuid] = all;
+                } else {
+                    NSLog(@"DisplayReconfigurationCallback: Failed to get gamma table for display %u (UUID: %s), exiting privacy mode", (unsigned)display, uuid.c_str());
+                    TurnOffPrivacyModeInternal();
+                    return;
                 }
+            } else {
+                NSLog(@"DisplayReconfigurationCallback: Display %u (UUID: %s) has zero gamma table capacity, exiting privacy mode", (unsigned)display, uuid.c_str());
+                TurnOffPrivacyModeInternal();
+                return;
             }
         }
         
         // Apply blackout to the new display immediately
-        ApplyBlackoutToDisplay(display);
+        if (!ApplyBlackoutToDisplay(display)) {
+            NSLog(@"DisplayReconfigurationCallback: Failed to blackout display %u (UUID: %s), exiting privacy mode", (unsigned)display, uuid.c_str());
+            TurnOffPrivacyModeInternal();
+            return;
+        }
         
         // Schedule a delayed re-application to handle ColorSync restoration
         // macOS may restore default gamma for ALL displays after a new display is added,
@@ -586,6 +603,55 @@ static void TeardownEventTapOnMainThread() {
     }
 }
 
+// Internal function to turn off privacy mode without acquiring the mutex
+// Must be called while holding g_privacyModeMutex
+static bool TurnOffPrivacyModeInternal() {
+    if (!g_privacyModeActive) {
+        return true;
+    }
+    
+    // 1. Unregister display reconfiguration callback
+    CGDisplayRemoveReconfigurationCallback(DisplayReconfigurationCallback, NULL);
+    
+    // 2. Input - restore (tear down EventTap on main thread)
+    TeardownEventTapOnMainThread();
+
+    // 3. Gamma - restore using UUID to find current DisplayID
+    bool restoreSuccess = true;
+    for (auto const& [uuid, gamma] : g_originalGammas) {
+        // Find current DisplayID for this UUID (handles ID changes after reconnection)
+        CGDirectDisplayID d = FindDisplayIdByUUID(uuid);
+        
+        if (d == kCGNullDirectDisplay) {
+            NSLog(@"Display with UUID %s no longer online, skipping gamma restore", uuid.c_str());
+            continue;
+        }
+        
+        uint32_t sampleCount = gamma.size() / 3;
+        if (sampleCount > 0) {
+            const CGGammaValue* red = gamma.data();
+            const CGGammaValue* green = red + sampleCount;
+            const CGGammaValue* blue = green + sampleCount;
+            CGError error = CGSetDisplayTransferByTable(d, sampleCount, red, green, blue);
+            if (error != kCGErrorSuccess) {
+                NSLog(@"Failed to restore gamma table for display %u (UUID: %s, error %d)", (unsigned)d, uuid.c_str(), error);
+                restoreSuccess = false;
+            }
+        }
+    }
+    
+    // 4. Fallback: Always call CGDisplayRestoreColorSyncSettings as a safety net
+    // This ensures displays return to normal even if our restoration failed or
+    // if the system (ColorSync/Night Shift) modified gamma during privacy mode
+    CGDisplayRestoreColorSyncSettings();
+    
+    // Clean up
+    g_originalGammas.clear();
+    g_privacyModeActive = false;
+    
+    return restoreSuccess;
+}
+
 extern "C" bool MacSetPrivacyMode(bool on) {
     std::lock_guard<std::mutex> lock(g_privacyModeMutex);
     if (on) {
@@ -705,50 +771,6 @@ extern "C" bool MacSetPrivacyMode(bool on) {
         return true;
 
     } else {
-        // Not in privacy mode
-        if (!g_privacyModeActive) {
-            return true;
-        }
-        
-        // 1. Unregister display reconfiguration callback
-        CGDisplayRemoveReconfigurationCallback(DisplayReconfigurationCallback, NULL);
-        
-        // 2. Input - restore (tear down EventTap on main thread)
-        TeardownEventTapOnMainThread();
-
-        // 3. Gamma - restore using UUID to find current DisplayID
-        bool restoreSuccess = true;
-        for (auto const& [uuid, gamma] : g_originalGammas) {
-            // Find current DisplayID for this UUID (handles ID changes after reconnection)
-            CGDirectDisplayID d = FindDisplayIdByUUID(uuid);
-            
-            if (d == kCGNullDirectDisplay) {
-                NSLog(@"Display with UUID %s no longer online, skipping gamma restore", uuid.c_str());
-                continue;
-            }
-            
-            uint32_t sampleCount = gamma.size() / 3;
-            if (sampleCount > 0) {
-                const CGGammaValue* red = gamma.data();
-                const CGGammaValue* green = red + sampleCount;
-                const CGGammaValue* blue = green + sampleCount;
-                CGError error = CGSetDisplayTransferByTable(d, sampleCount, red, green, blue);
-                if (error != kCGErrorSuccess) {
-                    NSLog(@"Failed to restore gamma table for display %u (UUID: %s, error %d)", (unsigned)d, uuid.c_str(), error);
-                    restoreSuccess = false;
-                }
-            }
-        }
-        
-        // 4. Fallback: Always call CGDisplayRestoreColorSyncSettings as a safety net
-        // This ensures displays return to normal even if our restoration failed or
-        // if the system (ColorSync/Night Shift) modified gamma during privacy mode
-        CGDisplayRestoreColorSyncSettings();
-        
-        // Clean up
-        g_originalGammas.clear();
-        g_privacyModeActive = false;
-        
-        return restoreSuccess;
+        return TurnOffPrivacyModeInternal();
     }
 }
--- a/src/platform/macos.mm
+++ b/src/platform/macos.mm
@@ -4,6 +4,10 @@
 #include <Security/Authorization.h>
 #include <Security/AuthorizationTags.h>
 
+#include <CoreGraphics/CoreGraphics.h>
+#include <vector>
+#include <map>
+
 extern "C" bool CanUseNewApiForScreenCaptureCheck() {
     #ifdef NO_InputMonitoringAuthStatus
     return false;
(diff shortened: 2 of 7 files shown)


src/lang/cn.rs
```diff
@@ -562,6 +562,7 @@ pub static ref T: std::collections::HashMap<&'static str, &'static str> =
         ("id_input_tip", "可以输入 ID、直连 IP，或域名和端口号（<域名>:<端口号>）。\n要访问另一台服务器上的设备，请附加服务器地址（<ID>@<服务器地址>?key=<密钥>）。比如，\n9123456234@192.168.16.1:21117?key=5Qbwsde3unUcJBtrx9ZkvUmwFNoExHzpryHuPUdqlWM=。\n要访问公共服务器上的设备，请输入 \"<ID>@public\"，无需密钥。\n\n如果您想要在首次连接时，强制走中继连接，请在 ID 的后面添加 \"/r\"，例如，\"9123456234/r\"。"),
         ("privacy_mode_impl_mag_tip", "模式 1"),
         ("privacy_mode_impl_virtual_display_tip", "模式 2"),
+        ("privacy_mode_impl_macos_tip", "隐私模式"),
```
rustdesk: No need to create this new tranlsation item. there is already ("Privacy mode", "隐私模式"),

rustdesk (changes requested):

src/privacy_mode.rs
```diff
@@ -23,6 +23,9 @@ pub mod win_mag;
 #[cfg(windows)]
 pub mod win_topmost_window;
 
+#[cfg(target_os = "macos")]
+pub mod macos;
+
```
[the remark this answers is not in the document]
fufesou: Hi, this PR can't be compiled.

Maybe a file named macos.rs is missing.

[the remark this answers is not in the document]
bin-haw: My bad! I’ve added the missing macos.rs now. To be safe, I did a clean clone and verified it compiles perfectly. Thanks for your patience! 😅

src/platform/macos.mm
```diff
@@ -292,3 +296,118 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+            }
+
+            // Set to black only if we have saved original gamma for this display
+            if (g_originalGammas.find(d) != g_originalGammas.end()) {
+                uint32_t capacity = CGDisplayGammaTableCapacity(d);
+                if (capacity > 0) {
+                    std::vector<CGGammaValue> zeros(capacity, 0.0f);
+                    CGError error = CGSetDisplayTransferByTable(d, capacity, zeros.data(), zeros.data(), zeros.data());
+                    if (error != kCGErrorSuccess) {
+                        NSLog(@"MacSetPrivacyMode: Failed to set gamma table to black for display %u (error %d)", (unsigned)d, error);
```
fufesou: Return `false` here?

src/platform/macos.mm
```diff
@@ -292,3 +296,118 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+        // 2. Gamma
+        for (auto const& [d, gamma] : g_originalGammas) {
+             uint32_t sampleCount = gamma.size() / 3;
+             if (sampleCount > 0) {
+                 const CGGammaValue* red = gamma.data();
+                 const CGGammaValue* green = red + sampleCount;
+                 const CGGammaValue* blue = green + sampleCount;
+                 CGError error = CGSetDisplayTransferByTable(d, sampleCount, red, green, blue);
+                 if (error != kCGErrorSuccess) {
+                     NSLog(@"Failed to restore gamma table for display %u (error %d), attempting system reset", (unsigned)d, error);
```
fufesou: Return `false` here?

src/privacy_mode/macos.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn clear(&mut self) {
+        unsafe {
+            MacSetPrivacyMode(false);
+        }
+    }
+
+    fn turn_on_privacy(&mut self, conn_id: i32) -> ResultType<bool> {
+        self.conn_id = conn_id;
+        unsafe {
+            MacSetPrivacyMode(true);
```
fufesou: Check the returned value?

Maybe the returned value should be checked, then set `conn_id`
```rust
self.conn_id = conn_id;
```

src/privacy_mode/macos.rs
```diff
@@ -0,0 +1,70 @@
...
+        }
+        Ok(true)
+    }
+
+    fn turn_off_privacy(&mut self, _conn_id: i32, _state: Option<PrivacyModeState>) -> ResultType<()> {
+        unsafe {
+            MacSetPrivacyMode(false);
```
fufesou: Check the returned value?

src/platform/macos.mm
```diff
@@ -292,3 +296,96 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+            CGEventMask eventMask = (1 << kCGEventKeyDown) | (1 << kCGEventKeyUp) |
+                                    (1 << kCGEventLeftMouseDown) | (1 << kCGEventLeftMouseUp) |
+                                    (1 << kCGEventRightMouseDown) | (1 << kCGEventRightMouseUp) |
+                                    (1 << kCGEventMouseMoved) | (1 << kCGEventScrollWheel);
+            
+            g_eventTap = CGEventTapCreate(kCGHIDEventTap, kCGHeadInsertEventTap, kCGEventTapOptionDefault,
+                                          eventMask, MyEventTapCallback, NULL);
+            if (g_eventTap) {
+                g_runLoopSource = CFMachPortCreateRunLoopSource(kCFAllocatorDefault, g_eventTap, 0);
+                CFRunLoopAddSource(CFRunLoopGetMain(), g_runLoopSource, kCFRunLoopCommonModes);
```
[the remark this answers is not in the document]
rustdesk: MacSetPrivacyMode is called quite late, the mainloop is supposed to be started.

src/platform/macos.mm
```diff
@@ -292,3 +296,96 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+                     uint32_t sampleCount = 0;
+                     if (CGGetDisplayTransferByTable(d, capacity, red.data(), green.data(), blue.data(), &sampleCount) == kCGErrorSuccess) {
+                         std::vector<CGGammaValue> all;
+                         all.insert(all.end(), red.begin(), red.begin() + sampleCount);
+                         all.insert(all.end(), green.begin(), green.begin() + sampleCount);
+                         all.insert(all.end(), blue.begin(), blue.begin() + sampleCount);
+                         g_originalGammas[d] = all;
+                     }
+                 }
+            }
```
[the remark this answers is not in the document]
rustdesk: Could you give some input about this?

src/platform/macos.mm
```diff
@@ -292,3 +296,96 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
     CFRelease(allModes);
     return ret;
 }
+
+static CFMachPortRef g_eventTap = NULL;
+static CFRunLoopSourceRef g_runLoopSource = NULL;
+static std::map<CGDirectDisplayID, std::vector<CGGammaValue>> g_originalGammas;
```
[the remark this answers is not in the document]
rustdesk: add a static lock here to make it thread safe?

src/platform/macos.mm
```diff
@@ -292,3 +296,118 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+        return NULL;
+    }
+    return event;
+}
+
+extern "C" bool MacSetPrivacyMode(bool on) {
+    if (on) {
+        // 1. Input Blocking
+        if (!g_eventTap) {
+            CGEventMask eventMask = (1 << kCGEventKeyDown) | (1 << kCGEventKeyUp) |
```
fufesou: `kCGEventOtherMouseDown/kCGEventOtherMouseUp` are missed, how about 

```
CGEventMask eventMask = (1 << kCGEventKeyDown) | (1 << kCGEventKeyUp) |
                        (1 << kCGEventLeftMouseDown) | (1 << kCGEventLeftMouseUp) |
                        (1 << kCGEventRightMouseDown) | (1 << kCGEventRightMouseUp) |
                        (1 << kCGEventOtherMouseDown) | (1 << kCGEventOtherMouseUp) |
                        (1 << kCGEventLeftMouseDragged) | (1 << kCGEventRightMouseDragged) |
                        (1 << kCGEventOtherMouseDragged) |
                        (1 << kCGEventMouseMoved) | (1 << kCGEventScrollWheel);
```

src/privacy_mode/macos.rs
```diff
@@ -0,0 +1,72 @@
...
+        if !success {
+            return Err(anyhow!("Failed to turn on privacy mode"));
+        }
+        self.conn_id = conn_id;
+        Ok(true)
+    }
+
+    fn turn_off_privacy(&mut self, _conn_id: i32, _state: Option<PrivacyModeState>) -> ResultType<()> {
+        let success = unsafe { MacSetPrivacyMode(false) };
+        self.conn_id = 0;
```
fufesou: Should `self.conn_id = 0;` be called after checking `success `?

src/platform/macos.mm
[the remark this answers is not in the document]
bin-haw: This concern has been addressed in commit b9401f2df.
Changes made:
• Use UUID instead of DisplayID as storage key (stable across reconnections)
• Clear g_originalGammas when privacy mode is turned off
• Register CGDisplayReconfigurationCallback to handle hot-plug events
• Validate display state via FindDisplayIdByUUID() before restoration
Files changed:
• src/platform/macos.mm (Lines 308-336, 360-375, 389-487)
• src/privacy_mode/macos.rs (Lines 48-54)

src/platform/macos.mm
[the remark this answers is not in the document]
bin-haw: Addressed. Added std::mutex g_privacyModeMutex with std::lock_guard at function entry to protect all global variables.

src/privacy_mode/macos.rs
bin-haw: Addressed. Moved self.conn_id = 0 after the success check to maintain consistent state management.

src/privacy_mode/macos.rs
```diff
@@ -0,0 +1,72 @@
...
+        Ok(())
+    }
+
+    fn clear(&mut self) {
+        unsafe {
+            MacSetPrivacyMode(false);
+        }
+    }
+
+    fn turn_on_privacy(&mut self, conn_id: i32) -> ResultType<bool> {
```
fufesou, line 40: It may be better to add a check here.

https://github.com/rustdesk/rustdesk/blob/087eb55299116abe5369f1ecd843b1204367b220/src/privacy_mode/win_virtual_display.rs#L464

src/privacy_mode/macos.rs
```diff
@@ -0,0 +1,72 @@
...
+        let success = unsafe { MacSetPrivacyMode(true) };
+        if !success {
+            return Err(anyhow!("Failed to turn on privacy mode"));
+        }
+        self.conn_id = conn_id;
+        Ok(true)
+    }
+
+    fn turn_off_privacy(&mut self, _conn_id: i32, _state: Option<PrivacyModeState>) -> ResultType<()> {
```
fufesou: It may be better to add a check here.

https://github.com/rustdesk/rustdesk/blob/087eb55299116abe5369f1ecd843b1204367b220/src/privacy_mode/win_virtual_display.rs#L523

src/platform/macos.mm
```diff
@@ -292,3 +299,455 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+                if (d != kCGNullDirectDisplay) {
+                    uint32_t sampleCount = gamma.size() / 3;
+                    if (sampleCount > 0) {
+                        const CGGammaValue* red = gamma.data();
+                        const CGGammaValue* green = red + sampleCount;
+                        const CGGammaValue* blue = green + sampleCount;
+                        CGError error = CGSetDisplayTransferByTable(d, sampleCount, red, green, blue);
+                        if (error != kCGErrorSuccess) {
+                            std::string displayName = GetDisplayName(d);
+                            NSLog(@"Failed to restore gamma for display %u (Name: %s, ID: %u, UUID: %s, error: %d)", displayName.c_str(), (unsigned)d, uuid.c_str(), error);
```
fufesou: 5 placeholders `%u,%s,%u,%s,%d`, while 4 values.

src/platform/macos.mm
```diff
@@ -292,3 +299,455 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+
+// Use CFStringRef (UUID) as key instead of CGDirectDisplayID for stability across reconnections
+// CGDirectDisplayID can change when displays are reconnected, but UUID remains stable
+static std::map<std::string, std::vector<CGGammaValue>> g_originalGammas;
+// Map UUID to current DisplayID for restoration
+static std::map<std::string, CGDirectDisplayID> g_uuidToDisplayId;
+
+// The event source user data value used by enigo library for injected events.
+// This allows us to distinguish remote input (which should be allowed) from local physical input.
+static const int64_t ENIGO_INPUT_EXTRA_VALUE = 100;
```
fufesou, line 330: Add a link comment here?
 https://github.com/rustdesk/rustdesk/blob/087eb55299116abe5369f1ecd843b1204367b220/libs/enigo/src/macos/macos_impl.rs#L44

src/platform/macos.mm
```diff
@@ -292,3 +299,455 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+static CFMachPortRef g_eventTap = NULL;
+static CFRunLoopSourceRef g_runLoopSource = NULL;
+static std::mutex g_privacyModeMutex;
+static bool g_privacyModeActive = false;
+
+// Use CFStringRef (UUID) as key instead of CGDirectDisplayID for stability across reconnections
+// CGDirectDisplayID can change when displays are reconnected, but UUID remains stable
+static std::map<std::string, std::vector<CGGammaValue>> g_originalGammas;
+// Map UUID to current DisplayID for restoration
+static std::map<std::string, CGDirectDisplayID> g_uuidToDisplayId;
```
fufesou: It seems `g_uuidToDisplayId` is not actually used.

src/platform/macos.mm
```diff
@@ -292,3 +299,455 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+                uint32_t sampleCount = 0;
+                if (CGGetDisplayTransferByTable(display, capacity, red.data(), green.data(), blue.data(), &sampleCount) == kCGErrorSuccess) {
+                    std::vector<CGGammaValue> all;
+                    all.insert(all.end(), red.begin(), red.begin() + sampleCount);
+                    all.insert(all.end(), green.begin(), green.begin() + sampleCount);
+                    all.insert(all.end(), blue.begin(), blue.begin() + sampleCount);
+                    g_originalGammas[uuid] = all;
+                    g_uuidToDisplayId[uuid] = display;
+                }
+            }
```
fufesou, line 638: No `else`, how about adding a log ?
And early return?

src/platform/macos.mm
```diff
@@ -292,3 +299,455 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+                std::vector<CGGammaValue> red(capacity), green(capacity), blue(capacity);
+                uint32_t sampleCount = 0;
+                if (CGGetDisplayTransferByTable(display, capacity, red.data(), green.data(), blue.data(), &sampleCount) == kCGErrorSuccess) {
+                    std::vector<CGGammaValue> all;
+                    all.insert(all.end(), red.begin(), red.begin() + sampleCount);
+                    all.insert(all.end(), green.begin(), green.begin() + sampleCount);
+                    all.insert(all.end(), blue.begin(), blue.begin() + sampleCount);
+                    g_originalGammas[uuid] = all;
+                    g_uuidToDisplayId[uuid] = display;
+                }
```
fufesou, line 633: `else` here?

src/platform/macos.mm
```diff
@@ -292,3 +299,455 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+        std::string uuid = GetDisplayUUID(displays[i]);
+        if (uuid == targetUuid) {
+            return displays[i];
+        }
+    }
+    return kCGNullDirectDisplay;
+}
+
+// Helper function to apply blackout to a single display
+static void ApplyBlackoutToDisplay(CGDirectDisplayID display) {
```
fufesou: Should this helper function return `bool`?

src/privacy_mode/macos.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    fn init(&self) -> ResultType<()> {
+        Ok(())
+    }
+
+    fn clear(&mut self) {
+        unsafe {
+            MacSetPrivacyMode(false);
+        }
```
fufesou, line 36: It may be better to add `self.conn_id = 0;` here.

src/platform/macos.mm
```diff
@@ -292,3 +299,455 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+                    std::vector<CGGammaValue> zeros(capacity, 0.0f);
+                    blackoutAttemptCount++;
+                    CGError error = CGSetDisplayTransferByTable(d, capacity, zeros.data(), zeros.data(), zeros.data());
+                    if (error != kCGErrorSuccess) {
+                        std::string displayName = GetDisplayName(d);
+                        NSLog(@"MacSetPrivacyMode: Failed to blackout display (Name: %s, ID: %u, UUID: %s, error: %d)", displayName.c_str(), (unsigned)d, uuid.c_str(), error);
+                    } else {
+                        blackoutSuccessCount++;
+                    }
+                }
```
fufesou, line 907: `else` is missing?

If `capacity  == 0`, maybe we should return `false` and log "not supported" here?

src/platform/macos.mm
```diff
@@ -292,3 +299,455 @@ static bool setDisplayToMode(CGDirectDisplayID display, CGDisplayModeRef mode) {
...
+
+        uint32_t blackoutSuccessCount = 0;
+        uint32_t blackoutAttemptCount = 0;
+
+        for (uint32_t i = 0; i < count; i++) {
+            CGDirectDisplayID d = displays[i];
+            std::string uuid = GetDisplayUUID(d);
+            
+            if (uuid.empty()) {
+                NSLog(@"MacSetPrivacyMode: Failed to get UUID for display %u, skipping", (unsigned)d);
```
fufesou: Maybe we should return `false` here, since this is the privacy mode?

src/privacy_mode/macos.rs
bin-haw: Addressed in 73b052475

src/privacy_mode/macos.rs
bin-haw: Addressed in 73b052475

src/platform/macos.mm
bin-haw: Addressed in 73b052475

src/platform/macos.mm
bin-haw: Addressed in 73b052475

src/platform/macos.mm
bin-haw: Addressed in 73b052475

src/platform/macos.mm
bin-haw: Addressed in 73b052475

src/privacy_mode/macos.rs
bin-haw: Addressed in 73b052475

src/platform/macos.mm
bin-haw: Addressed in 73b052475

src/platform/macos.mm
bin-haw: Addressed in 73b052475

src/platform/macos.mm
bin-haw: Addressed in f1512d13

src/platform/macos.mm
bin-haw: Addressed in f1512d13

src/privacy_mode.rs
```diff
@@ -333,7 +349,11 @@ pub fn get_supported_privacy_mode_impl() -> Vec<(&'static str, &'static str)> {
 
         vec_impls
     }
-    #[cfg(not(target_os = "windows"))]
+    #[cfg(target_os = "macos")]
+    {
+        vec![(macos::PRIVACY_MODE_IMPL, "privacy_mode_impl_macos_tip")]
```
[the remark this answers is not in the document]
fufesou: The placeholder `privacy_mode_impl_macos_tip` is fine here.

Since no translation is intended, please add a comment.

bin-haw:
After further testing, the current Gamma-based implementation is unstable as it's easily affected by macOS display configuration updates. A more robust approach would be using an NSWindowblack mask, but due to time constraints, I won't be able to implement this refactoring in the foreseeable future.
To maintain the stability of the main branch, this implementation is not ready for merging. Users who need this feature immediately can still find the code in my fork. Thanks for the review!

fufesou:
> After further testing, the current Gamma-based implementation is unstable as it's easily affected by macOS display configuration updates.

Hi, you mentioned that something is "easily affected by macOS display configuration updates."
What are you referring to?

These automatic system behaviors that could reset gamma (Night Shift transitions, display hot-plug, etc.) are relatively rare during a typical remote session.

Given the low probability, would a simpler detect-and-exit approach be sufficient? Instead of trying to reapply blackout on DisplayReconfigurationCallback, we could:

1. Check if gamma has been modified (no longer all-zero)
2. If so, exit privacy mode immediately  
3. Notify the user: "Privacy mode was interrupted by system. You may re-enable it."

This would simplify the code (no reapply logic, no delay tuning) while ensuring no accidental exposure. Most sessions won't be affected, and for the rare cases, users get a clear notification.

What do you think?

fufesou:
I tested the setup with two displays on the controlled macOS side.

The primary display remains black.
The extended display powers on (the backlight is lit) but appears to be in a state where it's not receiving a video signal. It shows a solid color screen (e.g., black, gray, or blue).

Is this the intended behavior?

image

bin-haw:
> Hi, you mentioned that something is "easily affected by macOS display configuration updates." What are you referring to?

To my knowledge, the Gamma-based mask is frequently reset by actions like resolution changes, display hot-plugging, or connecting virtual displays.

I encountered this frequently while remoting into my lab MacBook for experiments. In a lab/workroom setting, I often need to manually reconfigure the MacBook's resolution and display layout through the remote session to create a more comfortable workspace .

Based on my testing, the timing of this system reconfiguration is non-deterministic; a simple delay is often insufficient to catch the reset, which is why I had to resort to a 5-second polling mechanism (every 200ms) as a brute-force workaround. Since this is a 'hacky' fix for such unpredictable behavior, I don't think it's robust enough for the main branch yet.

>Check if gamma has been modified (no longer all-zero)
If so, exit privacy mode immediately
Notify the user: "Privacy mode was interrupted by system. You may re-enable it."

As I mentioned, when remoting into a lab MacBook, the frequent and manual display reconfigurations make it nearly impossible to guarantee 100% privacy uptime using the 'brute-force' polling approach. Constantly forcing the Gamma LUT back to zero not only feels hacky but also causes noticeable screen flickering, which is a poor user experience.

Switching to a monitoring-only approach is much more robust. Instead of fighting against macOS's non-deterministic resets, we can simply detect when the Gamma Table is modified by the system, exit privacy mode gracefully, and notify the user. This effectively eliminates the flickering and provides a much more transparent behavior.

This feels like a much more 'main-branch-ready' solution.

bin-haw:
> I tested the setup with two displays on the controlled macOS side.
The primary display remains black.
The extended display powers on (the backlight is lit) but appears to be in a state where it's not receiving a video signal. It shows a solid color screen (e.g., black, gray, or blue).
Is this the intended behavior?

Yes, this is the intended behavior.

As I mentioned regarding non-deterministic resets, the reconfiguration callback triggers almost immediately when a display is connected. In my tests, the mask is re-applied so fast that the physical monitor doesn't even have time to visibly render the desktop—it stays in a 'black' or 'solid color' state, even though the remote client can see that the new display is correctly online.

Regarding the UI, I noticed that the toolbar only shows a single display in privacy mode. I wasn't sure if this limitation was intentional (by design), so I decided not to modify the shared UI code like toolbar.dart in this PR. I’d rather keep the UI as-is for now to avoid breaking any existing design logic.

rustdesk:
I think it is fine to merge it. Then you can continue to work on it for better solution If you still have the passion. :(

PR #14158, opened 2026-01-26
Revert "Updated tr.rs"

rustdesk:
Reverts rustdesk/rustdesk#14115


bilimiyorum:
Could you clarify why the updates to tr.rs were reverted? I’d like to align with the project’s localization workflow.

rustdesk:
Because you placed in src/tr.rs. It should be src/lang/tr.rs.

Issue #13009, opened 2025-09-22
macOS Host: Pointer jumps to top-left corner when controlled from iPad with touchpad


m23x2:
### Bug Description

**1. Environment**
* **RustDesk Version:** `current`
* **Client OS:** `iPadOS 18-26`
* **Host OS:** `macOS 26`
* **Client Device:** `iPad Pro 11-inch M2`
* **Input Device:** `Apple Magic Keyboard Trackpad`

**2. Description**
When controlling a macOS host from an iPad client using a physical touchpad, the mouse pointer on the Mac jumps to the top-left screen corner `(0,0)` after a brief period of touchpad inactivity.

This behavior causes unintended side effects, such as triggering macOS Hot Corners or interacting with the menu bar. When the touchpad is touched again, the pointer immediately jumps back to its correct, original position.

### How to Reproduce

**3. Steps to Reproduce**
1.  Connect from a RustDesk client on an iPad (with a physical touchpad) to a RustDesk host running on macOS.
2.  (Optional but recommended for testing) On the Mac, enable a Hot Corner in the top-left of the screen (e.g., set it to Mission Control via System Settings > Desktop & Dock > Hot Corners).
3.  Move the mouse pointer to the center of the macOS screen using the iPad's touchpad.
4.  Remove your finger from the touchpad and wait for a few seconds.
5.  Observe the pointer on the macOS host.

### Expected Behavior

**4. Expected Behavior**
The mouse pointer should remain stationary at its last known position on the macOS screen when the touchpad is not in use.

**5. Actual Behavior**
The mouse pointer visibly jumps to the absolute top-left corner `(0,0)` of the screen. This triggers the Hot Corner or clicks on the Apple menu/application menu. As soon as the touchpad is touched again, the pointer instantly teleports back to its previous location.

**6. Possible Cause**
This issue appears to be related to how the RustDesk client translates pointer states between iPadOS and macOS. The iPadOS pointer is not persistent; it "disappears" or goes idle when the trackpad is not being touched. It's possible that the RustDesk client interprets this "idle" state as a default `(0,0)` coordinate and sends this position to the host, instead of sending no new movement data.

### Operating system(s) on local (controlling) side and remote (controlled) side

iPadOS 26 -> MacOS 26

### RustDesk Version(s) on local (controlling) side and remote (controlled) side

Both current.

### Screenshots

Not applicable

rustdesk:
@fufesou confirm if you can reproduce this.

[...]

PR #14180, opened 2026-01-27
fix(iOS): Unexpected mouse movement to (0,0) on idle

fufesou:
https://github.com/rustdesk/rustdesk/issues/13009

## Desc

I can reproduce this issue by changing 

https://github.com/rustdesk/rustdesk/blob/c76d10a438f64fac521749690d520c5260e23b1b/flutter/lib/models/input_model.dart#L1707

to 

```dart
return Position(0, 0);
```

Though I can't reproduce this issue directly, I believe it's caused by Flutter synthesizing a mouse hover event.


cstout1:
This issue exists for macOS app connecting to macOS client as well. Should this be opened as a separate issue? Seems to be the same - just not limited to iOS only.

Issue #14192, opened 2026-01-29
Hotkeys not working (Linux)


CodingKittens247:
### Bug Description

When using Linux (Wayland or X11) KDE, arch, nightly, etc, doesn't matter. 
I cannot perform any hotkey action (alt tab, Win + E etc) that is already assigned to the client i am working from. 
I can CTRL SHIFT ESC, since it is not used by Linux. 
I have tried x11, Wayland,  to no avail. 
Everything else works flawlessly! Including copy and paste between computers Linux -> Windows.
I'm at my wit's end, can't find anything online about this recently, all issues are closed that mention this... from years ago. 

Any idea what could be the issue and how to fix?
If it helps, direct IP, ID, same result. 
Keyboard layout is PT-PT, both machines. 

P.S. Anydesk works fine, using it while this isn't sorted because I am keyboard-centric. 

### How to Reproduce

Start Rustdesk, connect to a windows client, try to alt tab, or any other hotkey that is assigned to both machines, and it never passes it to the client. 

### Expected Behavior

Alt tab -> Alt tab on the focused client

### Operating system(s) on local (controlling) side and remote (controlled) side

Linux Arch KDE (Garuda) -> Windows 11

### RustDesk Version(s) on local (controlling) side and remote (controlled) side

1.4.5(Main and Nightly) -> 1.4.5.

### Screenshots

I don't know how to add a screenshot of something not happening while I press a button, you'll just have to believe me :(

### Additional Context

Anydesk 7.4.1 works perfectly on the same setup.

Kurgan-:
Rustdesk 1.4.5 (deb package) on Linux Mint 21.3 with X11 and Cinnamon.

I can send ALT-TAB no problem. Once I have focus on the Rustdesk window, ALT-TAB does work remotely and not locally. This is true for the "windows" key, also. 

Control-alt-del does not register locally but it does nothing at all remotely, I believe it's because Rustdesk is not installed on the remote machine.

agarre:
I can confirm this issue exists on Zorin OS 18 (deb package, Wayland) as well.
Hotkeys like Alt+Tab, Super/Win key, etc. are captured locally and never forwarded to the remote Windows machine. This happens even with RustDesk fully installed on the remote side — so @Kurgan-'s theory about it being related to RustDesk not being installed on the remote doesn't apply here.

Local: Zorin OS 18, Wayland, RustDesk 1.4.5 (deb)
[...]

PR #14238, opened 2026-02-02
feat: Inhibit system shortcuts on Linux

mvforell:
Fixes #13013.

### DIFF
--- a/flutter/linux/my_application.cc
+++ b/flutter/linux/my_application.cc
@@ -70,6 +70,13 @@ static void my_application_activate(GApplication* application) {
     gtk_window_set_title(window, "rustdesk");
   }
 
+  // Ask the window manager to disable any configured shortcuts so we can capture all input events.
+  GtkNative* native = gtk_widget_get_native(GTK_WIDGET(window));
+  if (native != NULL) {
+    GdkSurface* surface = gtk_native_get_surface(native);
+    gdk_toplevel_inhibit_system_shortcuts(GDK_TOPLEVEL(surface), NULL);
+  }
+
   // auto bdw = bitsdojo_window_from(window); // <--- add this line
   // bdw->setCustomFrame(true);               // <-- add this line
   int width = 800, height = 600;


mvforell:
As mentioned in https://github.com/rustdesk/rustdesk/issues/13013#issuecomment-3835303649, I was unable to build with Flutter locally, so I'd appreciate if we could kick off a CI build that I can test this change with.

rustdesk:
@fufesou please also test https://github.com/rustdesk/rustdesk/issues/14192

fufesou:
## 🚨 Critical Blocker: GTK Version Incompatibility

This PR uses GTK4 APIs while the project is built against GTK3.

Project Configuration: CMakeLists.txt:55 specifies gtk+-3.0

This will cause compilation failure on any system. The code cannot compile against GTK3.

https://github.com/fufesou/rustdesk/actions/runs/21594283582/job/62223186166

## Additional Concerns

### Unconditional Inhibition

Shortcuts are inhibited globally when the window activates, not just during active remote sessions. This affects the main UI window as well, breaking user interactions.

### Missing Restore Logic

There's no corresponding call to `gdk_toplevel_restore_system_shortcuts()` when the window loses focus or closes.

### Incomplete NULL Checks

While checking `native != NULL`, the code doesn't verify that `gtk_native_get_surface()` returns non-NULL before dereferencing it.

### Platform-Specific Behavior

`gdk_toplevel_inhibit_system_shortcuts()` may have different support levels across Wayland compositors (Hyprland, GNOME, KDE). More tests are needed.

mvforell:
Is building with GTK3 a hard requirement or would it be possible to update to GTK4?

Regarding the additional concerns:
* Unconditional Inhibition - this is correct. If you want the inhibition to only be for active remote sessions, please point me to the relevant part of the code; I was unable to find it myself.
* Missing restore logic - we don't need any restore logic. [The protocol says](https://wayland.app/protocols/keyboard-shortcuts-inhibit-unstable-v1#zwp_keyboard_shortcuts_inhibitor_v1) "A keyboard shortcuts inhibitor instructs the compositor to ignore its own keyboard shortcuts **when the associated surface has keyboard focus.**" (emphasis mine) i.e. when the window loses focus or closes the inhibition is automatically removed.
* Incomplete NULL Checks - this is incorrect, the return value of `gtk_native_get_surface()` is never dereferenced. It's passed directly to `gdk_toplevel_restore_system_shortcuts()`, which does nothing if we pass a null toplevel.
* Platform-Specific Behavior - the keyboard shortcuts inhibit protocol is implemented across all major Wayland compositors, see https://wayland.app/protocols/keyboard-shortcuts-inhibit-unstable-v1#compositor-support.

fufesou:
> Is building with GTK3 a hard requirement or would it be possible to update to GTK4?

No, it's impossible to migrate to GTK4 for now.

https://github.com/flutter/flutter/issues/94804

And there're some other GTK3 only APIs in `my_application.cc`.


> Missing restore logic - we don't need any restore logic. [The protocol says](https://wayland.app/protocols/keyboard-shortcuts-inhibit-unstable-v1#zwp_keyboard_shortcuts_inhibitor_v1) "A keyboard shortcuts inhibitor instructs the compositor to ignore its own keyboard shortcuts when the associated surface has keyboard focus." (emphasis mine) i.e. when the window loses focus or closes the inhibition is automatically removed.

Maybe "Unconditional Inhibition" is not a problemn then.

rustdesk:
> Is building with GTK3 a hard requirement 

Please change to GTK3. This is a hard requirement, because flutter is using gtk3.

rustdesk:
We will take over this. @fufesou please change it to gtk3.

PR #14265, opened 2026-02-05
Fix clipboard race condition with speech-to-text paste

AntoineDubuc:
## Summary

- When a speech-to-text tool (e.g. MacWhisper) updates the clipboard and immediately triggers Cmd+V / Ctrl+V, the paste keystroke can arrive at the server before the clipboard content has been written — pasting stale/empty content
- Root cause: `update_clipboard()` spawns a fire-and-forget background thread, so the server processes the next message (the paste keystroke) before the clipboard write completes
- Fix: add an awaitable `update_clipboard_blocking()` variant using `tokio::task::spawn_blocking`, and use it on the server side so the clipboard write completes before the next message is processed

## Details

The race condition flow:
1. Client clipboard changes → `MultiClipboards` message sent to server
2. Client keystroke (Cmd+V) → `KeyEvent` message sent to server
3. Server receives `MultiClipboards` → spawns background thread to write clipboard → **immediately moves on**
4. Server receives `KeyEvent` (paste) → injects keystroke → OS pastes from clipboard that hasn't been written yet

The fix ensures step 3 completes before step 4 begins, by awaiting the clipboard write on the server's async message loop. The client side remains unchanged (fire-and-forget) since the race only affects the controlled machine.

This is a minimal, ~10-line change touching only two files with no protocol or behavioral changes for normal usage.

## Test plan

- [ ] Verify normal clipboard copy-paste between client and server works unchanged
- [ ] Test with a speech-to-text tool that uses clipboard + paste (e.g. MacWhisper) — text should now arrive correctly on the remote machine
- [ ] Confirm no noticeable latency increase during regular remote desktop usage
- [ ] Test on macOS, Windows, and Linux server targets

### DIFF
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -238,6 +238,19 @@ pub fn update_clipboard(multi_clipboards: Vec<Clipboard>, side: ClipboardSide) {
     });
 }
 
+/// Awaitable version of `update_clipboard` that ensures the clipboard write
+/// completes before returning. Used on the server (controlled) side so that
+/// the clipboard content is ready before a subsequent paste keystroke is
+/// processed, preventing a race condition when speech-to-text tools (or
+/// similar) update the clipboard and immediately trigger Cmd+V / Ctrl+V.
+#[cfg(not(target_os = "android"))]
+pub async fn update_clipboard_blocking(multi_clipboards: Vec<Clipboard>, side: ClipboardSide) {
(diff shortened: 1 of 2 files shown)


AntoineDubuc:
Closing — the server-side fix alone is insufficient. The root cause is on the client side: the paste keystroke is sent immediately while clipboard changes are detected via polling (333ms interval), so the keystroke arrives before the clipboard content. Will resubmit with a complete fix that addresses both sides.

PR #14357, opened 2026-02-19
fix(terminal): fix new tab auto-focus and NaN error on data before layout

fufesou:
1. Opening a new terminal tab does not auto-focus the input — users must click the terminal area before typing.
2. When terminal data arrives before the TerminalView has completed layout, `setEditableSizeAndTransform` receives NaN dimensions, causing: `Unhandled Exception: Converting object to an encodable object failed: NaN`

## Root Cause

1. `TerminalView(autofocus: true)` only works on initial widget insertion. When a new tab is added, the terminal view may not yet have valid dimensions in the focus tree, so the autofocus request silently fails.
2. Terminal output data (`terminal.write()`) triggers `RenderTerminal._notifyEditableRect`, which calls `setEditableSizeAndTransform` with the current render size. Before the first layout pass, these values are `NaN`, and JSON serialization fails.

### Auto-focus (terminal_page.dart)

  - Replace `autofocus: true` with a manually managed `FocusNode` (initially `canRequestFocus: false`).
  - Listen to `tabController.state` changes; when this tab is selected, request focus via `addPostFrameCallback` to ensure the widget is fully laid out.
  - Only enable `canRequestFocus` after the first valid resize callback (`w > 0 && h > 0`).

### NaN error (terminal_model.dart)
  - Buffer output data in `_pendingOutputChunks` when `_terminalViewReady` is false.
  - On the first valid `onResize` callback, mark the view as ready and flush the buffer.
  - Buffer has an 8KB cap with FIFO eviction (drop oldest chunks) to bound memory usage.
  - `onResizeExternal` is called before `_markViewReady()` to ensure the view layer has valid dimensions before flushing.

### Minor fixes (terminal_tab_page.dart)
  - Pass `tabKey` to `TerminalPage` for focus management.
  - Fix string interpolation: `'$peerId\_'` → `'${peerId}_'`.

## Summary by CodeRabbit

* **New Features**
  * Better focus handling: tabs restore focus on resize and when switched.
  * Pre-layout output buffering: terminal output is preserved and displayed once the view is ready.

* **Bug Fixes**
  * Fixed focus subscription and tab selection issues to prevent focus leaks and ensure the correct tab receives focus.

* **Stability**
  * Improved cleanup on close to free buffers and reduce resource leaks.

### DIFF
--- a/flutter/lib/desktop/pages/terminal_page.dart
+++ b/flutter/lib/desktop/pages/terminal_page.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_hbb/common.dart';
@@ -15,6 +16,7 @@ class TerminalPage extends StatefulWidget {
     required this.tabController,
     required this.isSharedPassword,
     required this.terminalId,
+    required this.tabKey,
     this.forceRelay,
     this.connToken,
   }) : super(key: key);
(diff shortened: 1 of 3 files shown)


rustdesk:
@fufesou conflicts

PR #14359, opened 2026-02-19
Fix/terminal tab close persistent

fufesou:
### Before

https://github.com/user-attachments/assets/8dd3e8a3-c740-4799-a709-cdf875f0706a

### After

https://github.com/user-attachments/assets/a4289218-a380-4d52-a80d-1262670b7a49

## Issues

Two issues with terminal tab close and persistent session reconnect:

1. Tab close fragility: If the FFI call to closeTerminal() threw an exception, tabController.closeBy() would never execute, leaving a ghost tab in the UI. Additionally, all close paths (button, shortcut, programmatic) had duplicated logic with inconsistent behavior.
2. Reconnect ID mismatch: When the client reconnects after a disconnect, _nextTerminalId resets to 1, but surviving server-side sessions may have non-contiguous IDs (e.g. {2: htop}). The client would create a new shell instead of reconnecting to the surviving session.
3. Tab close race condition: `_closeTab()` reads peerTabCount from tabController after an await (audit dialog). If `_closeAllTabs()` runs concurrently and clears the tab controller, the live count drops to 0, incorrectly triggering session persistence.

## Solution

Flutter (terminal_tab_page.dart):
  - Extract unified `_closeTab()` handler with idempotency guard (`_closingTabs Set`)
  - Wrap `_closeTerminalSessionIfNeeded()` in an isolated `try/catch` so UI tab removal always proceeds
  - Extract `_closeAllTabs()` for window-level close with parallel session cleanup and bounded 4s timeout
  - Extract `_parseTabKey()` using `lastIndexOf('_')` to correctly handle peer IDs containing underscores
  - Snapshot `peerTabCount` before any await in `_closeTab()` to avoid race with concurrent `_closeAllTabs()`
  - Skip tabs already in `_closingTabs` during `_closeAllTabs()` to avoid duplicate FFI calls

Rust (terminal_service.rs):
  - Remap the lowest-ID surviving persistent session to the client-requested terminal_id on first reconnect, gated by needs_session_sync (one-time, cleared after first Open response)
  - Remaining sessions are communicated back via persistent_sessions with their original server-side IDs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal tab closing behavior with more reliable state management and dialog handling.
  * Enhanced terminal session reconnection logic to better preserve and remap existing sessions.

### DIFF
--- a/flutter/lib/desktop/pages/terminal_tab_page.dart
+++ b/flutter/lib/desktop/pages/terminal_tab_page.dart
@@ -34,6 +34,8 @@ class _TerminalTabPageState extends State<TerminalTabPage> {
   static const IconData selectedIcon = Icons.terminal;
   static const IconData unselectedIcon = Icons.terminal_outlined;
   int _nextTerminalId = 1;
+  // Lightweight idempotency guard for async close operations
+  final Set<String> _closingTabs = {};
 
   _TerminalTabPageState(Map<String, dynamic> params) {
     Get.put(DesktopTabController(tabType: DesktopTabType.terminal));
@@ -70,24 +72,7 @@ class _TerminalTabPageState extends State<TerminalTabPage> {
       label: tabLabel,
       selectedIcon: selectedIcon,
       unselectedIcon: unselectedIcon,
-      onTabCloseButton: () async {
-        if (await desktopTryShowTabAuditDialogCloseCancelled(
-          id: tabKey,
-          tabController: tabController,
-        )) {
-          return;
-        }
-        // Close the terminal session first
-        final ffi = TerminalConnectionManager.getExistingConnection(peerId);
-        if (ffi != null) {
-          final terminalModel = ffi.terminalModels[terminalId];
-          if (terminalModel != null) {
-            await terminalModel.closeTerminal();
-          }
-        }
-        // Then close the tab
-        tabController.closeBy(tabKey);
-      },
+      onTabCloseButton: () => _closeTab(tabKey),
       page: TerminalPage(
         key: ValueKey(tabKey),
         id: peerId,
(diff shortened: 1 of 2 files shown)


rustdesk:
Review from codex

• Findings

  1. Medium: concurrent tab closes can still drop the “last persistent session” unexpectedly.
     flutter/lib/desktop/pages/terminal_tab_page.dart:102 snapshots peer tab count before any await, and that snapshot is later used at flutter/lib/desktop/pages/terminal_tab_page.dart:122 and flutter/lib/
     desktop/pages/terminal_tab_page.dart:194.
     If two tabs for the same peer are closed in parallel (easy when the audit dialog is not shown), both closures can observe count 2 and both execute closeTerminal(), so no session is preserved even with
     “Keep terminal sessions on disconnect” enabled.

  Open Question

  1. Is the intended contract that at least one session survives peer-tab close sequences in persistent mode? If yes, this path should be serialized per peer (or re-evaluate using live state that excludes
     in-flight closes).

  Validation Notes

  - I reviewed PR #14359 diff directly (origin/pr-14359 vs origin/master).
  - I could not rely on cargo check for this review because the workspace currently has many unrelated baseline compile errors.

fufesou:
> concurrent tab closes can still drop the “last persistent session” unexpectedly
> Is the intended contract that at least one session survives peer-tab close sequences in persistent mode?

There's no need to fix it.

In practice, this is not triggerable through normal user interaction:                                      

  - Tab close is initiated by clicking the close button. After the first tab is removed, the UI layout shifts and focus changes — the second click takes at least hundreds of milliseconds, while the first _closeTab's async chain (a few layers of async return false when the audit dialog is not shown) completes in microseconds.
  - The close-tab keyboard shortcut is currently non-functional (captured by the terminal widget before reaching the handler).
  - The only programmatic concurrent path is _closeAllTabs, which uses persistAll: true and doesn't consult peerTabCount at all.

PR #14384, opened 2026-02-23
fix(linux): re-request Wayland screen sharing permission on session failure

Steve3184:
### Description

This PR fixes this issue in the Wayland environment:

- When RustDesk is running, if the user manually clicks the "End" button in the taskbar to end the screen sharing session, RustDesk does not retry.

Previously, this issues resulted in a persistent error: **"Failed to create capturer for display 0"**, requiring a full application restart to recover.

### Changes

* In `wayland.rs`, when a screen sharing session fails to initialize due to manual termination, it will automatically retry up to one time, which will re-prompt an authorization prompt.

### Impact

* Users no longer need to restart the RustDesk after a Wayland session interruption.
* Improved reliability for remote connections on Wayland.

### Screenshots

| Before | After |
| --- | --- |
| before | after |

## Summary by CodeRabbit

* **New Features**
  * Added session keepalive and a way to forcibly clear dead sessions; session inhibition is now managed to avoid conflicts.

* **Bug Fixes**
  * Improved screen-sharing initialization with retries, clearer re-authorization prompts, and early bailout when no displays are available.
  * Prevented unintended clearing while active captures exist; initialization flag now set only after full success.
  * Better detection and handling of revoked sessions.

* **Refactor**
  * Internal cleanup and safer error-path handling to reduce stale-state cases.

### DIFF
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -110,17 +110,7 @@ pub struct RdpSessionInfo {
     pub session: dbus::Path<'static>,
     pub is_support_restore_token: bool,
     pub resolution: Arc<Mutex<Option<(usize, usize)>>>,
-    inhibit_request_path: Option<dbus::Path<'static>>,
 }
-
-impl Drop for RdpSessionInfo {
-    fn drop(&mut self) {
-        if let Some(ref path) = self.inhibit_request_path {
-            release_inhibit(&self.conn, path);
-        }
-    }
-}
-
 #[derive(Debug, Clone, Copy)]
 pub struct PwStreamInfo {
     pub path: u64,
@@ -970,41 +960,6 @@ fn on_start_response(
     }
 }
 
-fn request_inhibit(conn: &SyncConnection) -> Option<dbus::Path<'static>> {
-    let proxy = conn.with_proxy(
-        "org.freedesktop.portal.Desktop",
-        "/org/freedesktop/portal/desktop",
-        Duration::from_millis(1000),
-    );
-    let mut args: PropMap = HashMap::new();
-    args.insert("handle_token".to_string(), Variant(Box::new("inhibit1".to_string())),);
-    // flags: 8 = inhibit idle, 4 = inhibit suspend
-    // Based on current testing, Dim Screen will forcibly terminate the session.
-    // Solving this problem requires preventing entry into the idle state.
-    // However, for future considerations, entering the suspended state is also prevented here.
-    match proxy.method_call::<(dbus::Path<'static>,), _, _, _>("org.freedesktop.portal.Inhibit", "Inhibit", ("", 12u32, args)) {
-        Ok((path,)) => {
-            debug!("Inhibit requested, request path: {:?}", path);
-            Some(path)
-        }
-        Err(e) => {
-            warn!("Failed to request inhibit: {}", e);
-            None
-        }
-    }
-}
-
-fn release_inhibit(conn: &SyncConnection, path: &dbus::Path<'static>) {
-    let proxy = conn.with_proxy(
-        "org.freedesktop.portal.Desktop",
-        path,
-        Duration::from_millis(1000),
-    );
-    if let Err(e) = proxy.method_call::<(), _, _, _>("org.freedesktop.portal.Request", "Close", ()) {
-        warn!("Failed to release inhibit: {}", e);
-    }
-}
-
 pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
     let mut rdp_connection = match RDP_SESSION_INFO.lock() {
         Ok(conn) => conn,
@@ -1014,9 +969,6 @@ pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
     if rdp_connection.is_none() {
         let (conn, fd, streams, session, is_support_restore_token) = request_remote_desktop(false)?;
         let conn = Arc::new(conn);
-        // to-do: Add an option like "Keep awake while RustDesk is running" to avoid
-        // impacting users whose sessions are not automatically terminated upon entering the Dim Screen.
-        let inhibit_path = if !is_server_running() { request_inhibit(&conn) } else { None };
 
         let rdp_info = RdpSessionInfo {
             conn,
@@ -1025,7 +977,6 @@ pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
             session,
             is_support_restore_token,
             resolution: Arc::new(Mutex::new(None)),
-            inhibit_request_path: inhibit_path,
         };
         *rdp_connection = Some(rdp_info);
     }
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -1034,6 +1034,8 @@ pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
     if rdp_connection.is_none() {
         let (conn, fd, streams, session, is_support_restore_token) = request_remote_desktop(false)?;
         let conn = Arc::new(conn);
+        // to-do: Add an option like "Keep awake while RustDesk is running" to avoid
+        // impacting users whose sessions are not automatically terminated upon entering the Dim Screen.
         let inhibit_path = if !is_server_running() { request_inhibit(&conn) } else { None };
 
         let rdp_info = RdpSessionInfo {
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -1034,7 +1034,7 @@ pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
     if rdp_connection.is_none() {
         let (conn, fd, streams, session, is_support_restore_token) = request_remote_desktop(false)?;
         let conn = Arc::new(conn);
-        let inhibit_path = request_inhibit(&conn);
+        let inhibit_path = if !is_server_running() { request_inhibit(&conn) } else { None };
 
         let rdp_info = RdpSessionInfo {
             conn,
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -13,7 +13,6 @@ use dbus::{
     arg::{OwnedFd, PropMap, RefArg, Variant},
     blocking::{Proxy, SyncConnection},
     message::{MatchRule, MessageType},
-    channel::Sender,
     Message,
 };
 
@@ -111,12 +110,14 @@ pub struct RdpSessionInfo {
     pub session: dbus::Path<'static>,
     pub is_support_restore_token: bool,
     pub resolution: Arc<Mutex<Option<(usize, usize)>>>,
-    pub keepalive_flag: Arc<AtomicBool>,
+    inhibit_request_path: Option<dbus::Path<'static>>,
 }
 
 impl Drop for RdpSessionInfo {
     fn drop(&mut self) {
-        self.keepalive_flag.store(false, Ordering::SeqCst);
+        if let Some(ref path) = self.inhibit_request_path {
+            release_inhibit(&self.conn, path);
+        }
     }
 }
 
@@ -957,6 +958,38 @@ fn on_start_response(
     }
 }
 
+fn request_inhibit(conn: &SyncConnection) -> Option<dbus::Path<'static>> {
+    let proxy = conn.with_proxy(
+        "org.freedesktop.portal.Desktop",
+        "/org/freedesktop/portal/desktop",
+        Duration::from_millis(1000),
+    );
+    let mut args: PropMap = HashMap::new();
+    args.insert("handle_token".to_string(), Variant(Box::new("inhibit1".to_string())),);
+    // flags: 4 = inhibit idle, 8 = inhibit suspend
+    match proxy.method_call::<(dbus::Path<'static>,), _, _, _>("org.freedesktop.portal.Inhibit", "Inhibit", ("", 8u32, args)) {
+        Ok((path,)) => {
+            debug!("Inhibit requested, request path: {:?}", path);
+            Some(path)
+        }
+        Err(e) => {
+            warn!("Failed to request inhibit: {}", e);
+            None
+        }
+    }
+}
+
+fn release_inhibit(conn: &SyncConnection, path: &dbus::Path<'static>) {
+    let proxy = conn.with_proxy(
+        "org.freedesktop.portal.Desktop",
+        path,
+        Duration::from_millis(1000),
+    );
+    if let Err(e) = proxy.method_call::<(), _, _, _>("org.freedesktop.portal.Request", "Close", ()) {
+        warn!("Failed to release inhibit: {}", e);
+    }
+}
+
 pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
     let mut rdp_connection = match RDP_SESSION_INFO.lock() {
         Ok(conn) => conn,
@@ -966,7 +999,7 @@ pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
     if rdp_connection.is_none() {
         let (conn, fd, streams, session, is_support_restore_token) = request_remote_desktop(false)?;
         let conn = Arc::new(conn);
-        let keepalive_flag = Arc::new(AtomicBool::new(true));
+        let inhibit_path = request_inhibit(&conn);
 
         let rdp_info = RdpSessionInfo {
             conn: conn.clone(),
@@ -975,51 +1008,9 @@ pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
             session: session.clone(),
             is_support_restore_token,
             resolution: Arc::new(Mutex::new(None)),
-            keepalive_flag: keepalive_flag.clone(),
+            inhibit_request_path: inhibit_path,
         };
         *rdp_connection = Some(rdp_info);
-
-        let conn_clone = conn.clone();
-        let session_clone = session.clone();
-
-        // Remote Desktop sessions are observed to time out after ~15 minutes of inactivity.
-        // Pinging every 60s gives a wide safety margin while keeping DBus traffic negligible.
-        const KEEPALIVE_INTERVAL_SECS: u64 = 60;
-
-        std::thread::spawn(move || {
-            let mut ticks = 0;
-            while keepalive_flag.load(Ordering::SeqCst) {
-                std::thread::sleep(Duration::from_secs(1));
-                ticks += 1;
-                if ticks >= KEEPALIVE_INTERVAL_SECS {
-                    ticks = 0;
-                    if !is_server_running() {
-                        match Message::new_method_call(
-                            "org.freedesktop.portal.Desktop",
-                            "/org/freedesktop/portal/desktop",
-                            "org.freedesktop.portal.RemoteDesktop",
-                            "NotifyPointerMotion",
-                        ) {
-                            Ok(msg) => {
-                                let msg = msg.append1((
-                                    session_clone.clone(),
-                                    HashMap::<String, Variant<Box<dyn RefArg>>>::new(),
-                                    0.0f64,
-                                    0.0f64,
-                                ));
-                                if let Err(_) = conn_clone.send(msg) {
-                                    // Session may have been revoked; the keepalive is no longer effective.
-                                    break;
-                                }
-                            }
-                            Err(_) => {
-                                break;
-                            }
-                        }
-                    }
-                }
-            }
-        });
     }
 
     let rdp_info = match rdp_connection.as_mut() {
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -288,7 +288,7 @@ impl PipeWireRecorder {
         let src = gst::ElementFactory::make("pipewiresrc", None)?;
         src.set_property("fd", &capturable.fd.as_raw_fd())?;
         src.set_property("path", &format!("{}", capturable.path))?;
-        src.set_property("keepalive_time", &1000i32)?;
+        src.set_property("keepalive_time", &1_000.as_raw_fd())?;
 
         // For some reason pipewire blocks on destruction of AppSink if this is not set to true,
         // see: https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/982
@@ -1020,10 +1020,10 @@ pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
         let inhibit_path = request_inhibit(&conn);
 
         let rdp_info = RdpSessionInfo {
-            conn: conn.clone(),
+            conn,
             streams,
             fd,
-            session: session.clone(),
+            session,
             is_support_restore_token,
             resolution: Arc::new(Mutex::new(None)),
             inhibit_request_path: inhibit_path,
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -13,6 +13,7 @@ use dbus::{
     arg::{OwnedFd, PropMap, RefArg, Variant},
     blocking::{Proxy, SyncConnection},
     message::{MatchRule, MessageType},
+    channel::Sender,
     Message,
 };
 
@@ -95,6 +96,14 @@ pub fn try_close_session() {
     }
 }
 
+pub fn force_close_dead_session() {
+    if let Ok(mut rdp_info) = RDP_SESSION_INFO.lock() {
+        *rdp_info = None;
+    }
+    clear_wayland_displays_cache();
+    HAS_POSITION_ATTR.store(false, Ordering::SeqCst);
+}
+
 pub struct RdpSessionInfo {
     pub conn: Arc<SyncConnection>,
     pub streams: Vec<PwStreamInfo>,
@@ -102,7 +111,15 @@ pub struct RdpSessionInfo {
     pub session: dbus::Path<'static>,
     pub is_support_restore_token: bool,
     pub resolution: Arc<Mutex<Option<(usize, usize)>>>,
+    pub keepalive_flag: Arc<AtomicBool>,
 }
+
+impl Drop for RdpSessionInfo {
+    fn drop(&mut self) {
+        self.keepalive_flag.store(false, Ordering::SeqCst);
+    }
+}
+
 #[derive(Debug, Clone, Copy)]
 pub struct PwStreamInfo {
     pub path: u64,
@@ -270,7 +287,7 @@ impl PipeWireRecorder {
         let src = gst::ElementFactory::make("pipewiresrc", None)?;
         src.set_property("fd", &capturable.fd.as_raw_fd())?;
         src.set_property("path", &format!("{}", capturable.path))?;
-        src.set_property("keepalive_time", &1_000.as_raw_fd())?;
+        src.set_property("keepalive_time", &1000i32)?;
 
         // For some reason pipewire blocks on destruction of AppSink if this is not set to true,
         // see: https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/982
@@ -306,7 +323,15 @@ impl PipeWireRecorder {
             "[gstreamer] Setting pipeline {} to PLAYING state...",
             capturable.fd.as_raw_fd()
         );
-        pipeline.set_state(gst::State::Playing)?;
+        if let Err(e) = pipeline.set_state(gst::State::Playing) {
+            warn!(
+                "[gstreamer] Failed to set PLAYING state: {:?}. Session likely revoked by user or token expired.",
+                e
+            );
+            
+            config::LocalConfig::set_option(RESTORE_TOKEN_CONF_KEY.to_owned(), "".to_owned());
+            return Err(hbb_common::anyhow::Error::msg(format!("SESSION_REVOKED: GStreamer failed: {:?}", e)));
+        }
 
         // If `is_server_running()` is false, it means using remote_desktop_portal,
         // which does not use multiple streams, so no need to wait for state change.
@@ -323,9 +348,16 @@ impl PipeWireRecorder {
                 }
                 (result, state, pending) => {
                     warn!(
-                    "[gstreamer] Pipeline {} state change incomplete: result={:?}, state={:?}, pending={:?}",
-                    capturable.fd.as_raw_fd(), result, state, pending
-                );
+                        "[gstreamer] Pipeline {} state change incomplete: result={:?}, state={:?}, pending={:?}",
+                        capturable.fd.as_raw_fd(), result, state, pending
+                    );
+
+                    if let Err(err) = result {
+                        warn!("[gstreamer] Pipeline error detected. Session was likely terminated, clearing XDP token...");
+                        config::LocalConfig::set_option(RESTORE_TOKEN_CONF_KEY.to_owned(), "".to_owned());
+                        let _ = pipeline.set_state(gst::State::Null);
+                        return Err(hbb_common::anyhow::Error::msg(format!("SESSION_REVOKED: GStreamer pipeline failed: {:?}", err)));
+                    }
                 }
             }
             std::thread::sleep(std::time::Duration::from_millis(150));
@@ -934,16 +966,48 @@ pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
     if rdp_connection.is_none() {
         let (conn, fd, streams, session, is_support_restore_token) = request_remote_desktop(false)?;
         let conn = Arc::new(conn);
+        let keepalive_flag = Arc::new(AtomicBool::new(true));
 
         let rdp_info = RdpSessionInfo {
-            conn,
+            conn: conn.clone(),
             streams,
             fd,
-            session,
+            session: session.clone(),
             is_support_restore_token,
             resolution: Arc::new(Mutex::new(None)),
+            keepalive_flag: keepalive_flag.clone(),
         };
         *rdp_connection = Some(rdp_info);
+
+        let conn_clone = conn.clone();
+        let session_clone = session.clone();
+        std::thread::spawn(move || {
+            let mut ticks = 0;
+            while keepalive_flag.load(Ordering::SeqCst) {
+                std::thread::sleep(Duration::from_secs(1));
+                ticks += 1;
+                
+                if ticks >= 60 {
+                    ticks = 0;
+                    if !is_server_running() {
+                        if let Ok(msg) = Message::new_method_call(
+                            "org.freedesktop.portal.Desktop",
+                            "/org/freedesktop/portal/desktop",
+                            "org.freedesktop.portal.RemoteDesktop",
+                            "NotifyPointerMotion",
+                        ) {
+                            let msg = msg.append1((
+                                session_clone.clone(),
+                                HashMap::<String, Variant<Box<dyn RefArg>>>::new(),
+                                0.0f64,
+                                0.0f64,
+                            ));
+                            let _ = conn_clone.send(msg);
+                        }
+                    }
+                }
+            }
+        });
     }
 
     let rdp_info = match rdp_connection.as_mut() {
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -98,9 +98,9 @@ pub fn try_close_session() {
 pub fn force_close_dead_session() {
     if let Ok(mut rdp_info) = RDP_SESSION_INFO.lock() {
         *rdp_info = None;
+        clear_wayland_displays_cache();
+        HAS_POSITION_ATTR.store(false, Ordering::SeqCst);
     }
-    clear_wayland_displays_cache();
-    HAS_POSITION_ATTR.store(false, Ordering::SeqCst);
 }
 
 pub struct RdpSessionInfo {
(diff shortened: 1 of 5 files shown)


Steve3184:
@coderabbitai

Regarding the issue of capture purging:

You're right, obsolete display indexes shouldn't be blindly deleted from all captures. However, in this particular Wayland environment, if `update_get_sync_displays_on_login` fails, it usually indicates that the underlying XDG Desktop Portal session has been revoked (e.g., via screen lock or manual termination).

In this state, all existing captures associated with that session become invalid and unable to receive any new frames. Therefore, they must be purged to prevent resource leaks and ensure a clean state for the reauthorization process.

Regarding testing:

I have manually verified the process multiple times (including revoking permissions while connected, revoking permissions while idle, reconnecting after screen lock, connecting while screen lock, etc.).

I have run `cargo test`, and all tests passed without reporting any issues.

src/server/connection.rs
```diff
@@ -1600,7 +1600,63 @@ impl Connection {
...
+                    first
+                }
+            } else {
+                super::display_service::update_get_sync_displays_on_login().await
+            };
+            #[cfg(not(target_os = "linux"))]
+            let displays_result =
+                super::display_service::update_get_sync_displays_on_login().await;
+
+            match displays_result {
```
rustdesk: Please add this block to a seperate function.

src/server/connection.rs
```diff
@@ -1600,7 +1600,63 @@ impl Connection {
...
+                if first.is_err() {
+                    log::warn!(
+                        "Wayland login: display init failed, waiting for re-authorization: {:?}",
+                        first.as_ref().err()
+                    );
+                    let mut msg_out = Message::new();
+                    msg_out.set_message_box(MessageBox {
+                        msgtype: "nook-nocancel-hasclose".to_owned(),
+                        title: "Wayland".to_owned(),
+                        text: "select-wayland-screen-tip".to_owned(),
```
rustdesk: please add tranlation of "select-wayland-screen-tip" to src/lang/template.rs and en.rs.

fufesou:
Thanks for working on this! I haven't been able to reproduce the original issue yet, so I'd like to ask a few questions and share some observations from reading the code before diving deeper.

The root cause analysis (stale `PIPEWIRE_INITIALIZED` flag preventing re-initialization) looks correct, and moving the flag assignment after the capturer loop succeeds is the right call. That said, I do have some concerns about the overall approach:

### 1. UX concern: popup + 120s retry on every reconnection

When `update_get_sync_displays_on_login()` fails, the PR immediately shows `"Please Select the screen to be shared (Operate on the peer side)"` and enters a 120s retry loop. On KDE Plasma 6.5+, this failure is **easily and frequently triggered** (multiple users confirmed in #13378), which means almost every reconnection would prompt the user to go to the peer side and re-authorize.

The core use case for remote desktop is **unattended access** — there is nobody at the remote machine. In that scenario, the user can do nothing but wait 120 seconds for timeout, then the connection fails anyway. This effectively **downgrades the restore token mechanism from automatic to manual**, making the experience worse than before for unattended setups.

### 2. Recovery logic belongs in the PipeWire layer, not the login flow

The retry and session recovery logic belongs in `pipewire.rs` (`get_capturables` / `request_remote_desktop`), not in `connection.rs`. When a restore token becomes stale, the PipeWire layer should detect the invalid session (empty streams or capturer creation failure) and **automatically retry `request_remote_desktop()` once**. This keeps recovery transparent to the upper layer — no 120s polling loop, no popup in the login flow.

More importantly: **why does the restore token become stale so frequently on KDE 6.5+?** The current `try_close_session()` aggressively drops the D-Bus session after every disconnection when `is_support_restore_token` is true, relying entirely on the restore token for the next connection. If KDE 6.5's portal has issues restoring sessions from tokens, a better strategy might be to **keep the session alive** instead of closing it — avoiding the problem altogether.

### 3. `get_capturer_for_display`: `clear()` bypasses `ACTIVE_DISPLAY_COUNT` guard

In `wayland.rs`, `get_capturer_for_display()` calls `clear()` unconditionally when a display is not found in the map. This deallocates all `Capturer` heap memory via `Box::from_raw()`, but the video service may still hold cloned `CapturerPtr` values (which are shallow pointer copies). The existing `ACTIVE_DISPLAY_COUNT` guard in `video_service.rs:549-553` is specifically designed to prevent this — `clear()` should only run when the last active display exits. Bypassing it risks use-after-free.

### 4. Minor

- `connection.rs`: trailing newline removed at EOF
- The retry loop doesn't distinguish recoverable errors (authorization pending) from non-recoverable ones (missing portal, D-Bus failures) — all errors get the same 120s wait

### Suggested approach

Instead of adding a retry loop and popup in `connection.rs`, consider fixing this at the PipeWire layer:

1. **Investigate why `try_close_session()` + restore token fails on KDE 6.5+.** If KDE 6.5's portal can't reliably restore from tokens, consider **not closing the session** when the server is still running — this avoids the stale token problem entirely.

2. **If the session must be closed**, add a single automatic retry in `request_remote_desktop()`: if the first attempt (with restore token) yields an invalid session, retry once. The portal will present an authorization dialog if needed — but this only happens when the token is genuinely stale, not on every reconnection.

3. **Keep `connection.rs` login flow simple.** It should not contain PipeWire-specific retry logic or user-facing popups about screen sharing authorization. If the PipeWire layer handles recovery internally, the login flow works unchanged.

rustdesk:
https://github.com/rustdesk/rustdesk/discussions/13378#discussioncomment-15895489

Steve3184:
I just conducted new tests, and the results show that after approximately 15 minutes of inactivity, a new connection causes GStreamer to throw a StateChangeError.

However, reconnecting after manually clicking the "End" button from the taskbar still results in the same error from GStreamer. These issues only occur when the RustDesk service is not installed, and no software like OBS Studio is running.

One possible workaround is to create a thread that sends a mouse movement (dx, dy both 0) every minute. I'm currently trying this, and it seems to be working; the session no longer gets suspended due to timeout.

I may have misunderstood the logic before. Currently, my local version only attempts to reconnect at most once if it detects that the session has been manually terminated, which causes the authorization pop-up to reappear.

I will push a new commit to my branch later.

src/server/connection.rs
Steve3184: This feature has been removed in the latest commit, so this block is no longer relevant. Thanks!

src/server/connection.rs
Steve3184: This feature has been removed in the latest commit, so this block is no longer relevant. Thanks!

fufesou:
> These issues only occur when the RustDesk service is not installed

Thanks. This may be the key point—there's a bug here.

> One possible workaround is to create a thread that sends a mouse movement (dx, dy both 0) every minute.

That is a workaround, not a fix. We should fix the underlying bug instead.

libs/scrap/src/wayland/pipewire.rs
```diff
@@ -934,16 +966,48 @@ pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
...
+                                HashMap::<String, Variant<Box<dyn RefArg>>>::new(),
+                                0.0f64,
+                                0.0f64,
+                            ));
+                            let _ = conn_clone.send(msg);
+                        }
+                    }
+                }
+            }
+        });
```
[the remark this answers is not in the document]
Steve3184: Since the `scrap` module doesn't include the log library, adding logging here would require a new dependency. The latest commit changes this to automatically exit thread if keepalive fails, which might be better.

src/server/wayland.rs
```diff
@@ -128,88 +129,134 @@ pub(super) fn is_inited() -> Option<Message> {
...
-                }
+                    let mut all = Display::all()?;
+                    log::debug!("Initializing displays with fill_displays()");
+                    {
+                        let temp_mouse_move_handle = input_service::TemporaryMouseMoveHandle::new();
+                        let move_mouse_to = |x, y| temp_mouse_move_handle.move_mouse_to(x, y);
+                        fill_displays(move_mouse_to, crate::get_cursor_pos, &mut all)?;
+                    }
+                    log::debug!("Attempting to fix logical size with try_fix_logical_size()");
+                    try_fix_logical_size(&mut all);
```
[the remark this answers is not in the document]
Steve3184: In fact, only when GStreamer throws a StateChangeError can we truly determine that the problem stems from the manual closure of the session. The early initialization phase cannot definitively determine if the problem originates from manual session termination. This retry logic only addresses GStreamer StateChangeErrors caused by manual session termination; therefore, it's unnecessary to handle issues thrown during early initialization.

libs/scrap/src/wayland/pipewire.rs
```diff
@@ -306,7 +324,15 @@ impl PipeWireRecorder {
...
-        pipeline.set_state(gst::State::Playing)?;
+        if let Err(e) = pipeline.set_state(gst::State::Playing) {
+            warn!(
+                "[gstreamer] Failed to set PLAYING state: {:?}. Session likely revoked by user or token expired.",
+                e
+            );
+            let _ = pipeline.set_state(gst::State::Null);
+            config::LocalConfig::set_option(RESTORE_TOKEN_CONF_KEY.to_owned(), "".to_owned());
+            return Err(hbb_common::anyhow::Error::msg(format!("SESSION_REVOKED: GStreamer failed: {:?}", e)));
+        }
```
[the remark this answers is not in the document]
Steve3184: The latest commit has fixed this issue. Now, `is_server_running` is used to correctly determine whether the `StateChangeError` is caused by a configuration error or an operational error, thus avoiding retries triggered due to configuration errors.

libs/scrap/src/wayland/pipewire.rs
```diff
@@ -95,14 +95,32 @@ pub fn try_close_session() {
     }
 }
 
+pub fn force_close_dead_session() {
+    if let Ok(mut rdp_info) = RDP_SESSION_INFO.lock() {
+        *rdp_info = None;
+    }
+    clear_wayland_displays_cache();
+    HAS_POSITION_ATTR.store(false, Ordering::SeqCst);
+}
```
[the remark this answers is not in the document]
Steve3184: The latest commit has fixed this issue.

libs/scrap/src/wayland/pipewire.rs
```diff
@@ -925,6 +958,38 @@ fn on_start_response(
...
+        Ok((path,)) => {
+            debug!("Inhibit requested, request path: {:?}", path);
+            Some(path)
+        }
+        Err(e) => {
+            warn!("Failed to request inhibit: {}", e);
+            None
+        }
+    }
+}
```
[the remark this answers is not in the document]
Steve3184: Sorry, I may have misread the documentation. The comments have been updated, and both Idle and suspended states are now prevented.

rustdesk:
I do not think I will approve this PR because of so many unnecessary code change.  We do not care you write a lot of new code, but please keep minimum code change to old code.

Steve3184:
I compared the actual changes in the PR and removed code changes that seemed unnecessary.

Currently, the PR only made these changes:

- Added retry logic in `check_init`

- Ensured the cleanup logic in `get_capture_for_display` works correctly

- Added logic in `pipewire.rs` to prevent entry into the Dim Screen to maintain the session

- Ensured that errors generated by GSteamer due to manual session termination are correctly detected by `check_init` and trigger retry logic

- Added `force_close_dead_session` to clean up dead sessions.

fufesou:
Hi, could you please give more details about how to reproduce this issue?

The following screenshot is my power settings, but I can't reproduce it.

I have tried different settings of "Suspend Session" and "Display and Brightness".

image

The following steps are tested:
1. Connecting for more than 30 minutes.
2. Connect, disconnect, wait for 30 minutes, connect

Steve3184:
The method to reproduce this is to set "Dim Automatically" to a short time, such as 10 seconds, then open RustDesk 1.4.5, connect and disconnect once, stopping all mouse/keyboard operations and ensuring no other software holds the wake lock. If `dbus-monitor "interface='org.freedesktop.portal.Session'"` is running at this time, you should see the session closed:

```
signal time=1772168282.974411 sender=:1.7 -> destination=:1.738 serial=23778 path=/org/freedesktop/portal/desktop/session/1_738/u1; interface=org.freedesktop.portal.Session; member=Closed
array [
]
```

At this point, reconnection is impossible, and the error message "Failed to create capturer for display 0" or similar is displayed.

This is my system infomation:

```
Operating System: Arch Linux 
KDE Plasma Version: 6.5.5
KDE Frameworks Version: 6.23.0
Qt Version: 6.10.2
Kernel Version: 6.18.9-arch1-2 (64-bit)
Graphics Platform: Wayland
Processors: 16 × 11th Gen Intel® Core™ i9-11900KF @ 3.50GHz
Memory: 64 GiB of RAM (62.7 GiB usable)
Graphics Processor: NVIDIA GeForce RTX 2080 Ti
Manufacturer: Gigabyte Technology Co., Ltd.
Product Name: B560M AORUS ELITE
```

fufesou:
Thanks for the reply.

But I still can reproduce it.

The outputs:
```
$ dbus-monitor "interface='org.freedesktop.portal.Session'"
signal time=1772169932.197848 sender=org.freedesktop.DBus -> destination=:1.91835 serial=4294967295 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameAcquired
   string ":1.91835"
signal time=1772169932.197870 sender=org.freedesktop.DBus -> destination=:1.91835 serial=4294967295 path=/org/freedesktop/DBus; interface=org.freedesktop.DBus; member=NameLost
   string ":1.91835"

```

I will try openSUSE Tumbleweed.

image


image

rustdesk:
Please refactor check_init to avoid so many code change. You can put loop outside the check_init function.

```
loop {
  old_check_init().
}
```

Steve3184:
Okay, I've moved the original loop logic to a separate function, and now the git changes look cleaner.

src/server/wayland.rs
```diff
@@ -265,26 +317,64 @@ pub(super) fn get_capturer_for_display(
...
+            "Display {} not found in CAP_DISPLAY_INFO, but {} active capturer(s) are still running. Skipping reinitialization now.",
+            display_idx, active_count
+        );
+    }
+
+    log::info!(
+        "get_capturer_for_display: no active capturers, reinitializing PipeWire session for display {}.",
+        display_idx
+    );
+    clear();
```
fufesou, line 376: There is a race condition here.

1. `CapturerPtr::clone` in `wayland.rs` performs a shallow copy of a raw pointer (`Self(self.0)`), so multiple holders alias the same underlying `Capturer` allocation.

2. `get_capturer_for_display` builds `CapturerInfo` with `cap_display_info.capturer.clone()`, which means each video worker receives another alias to that same raw pointer, not an independent capturer object.

3. In `video_service::run`, the worker keeps using that pointer via `c.frame(spf)`, which eventually dereferences raw memory in `unsafe { (*self.0).frame(timeout) }`.

4. If `get_capturer_for_display` calls `clear()` while any worker still holds cloned `CapturerPtr`s, `clear()` frees the underlying `Capturer` with `Box::from_raw`, invalidating all aliases immediately.

5. This bypasses the intended lifetime guard in `video_service` (which decrements the active display count and only calls `clear()` when the count reaches zero), so active workers can continue calling `frame()` on freed memory.

Conclusion: calling `clear()` from `get_capturer_for_display` in that state can create dangling `Capturer` pointers and trigger undefined behavior (use-after-free) in subsequent `frame()` calls.

There is also a similar problem in the following code in `video_service.rs`:
```rust
  let remaining_count = super::wayland::decrement_active_display_count();
  if remaining_count == 0 {
      super::wayland::clear();
  }
```

These two issues are not easily triggered. Maybe we can leave TODOs for now and fix them in another PR.

Steve3184: That's right, theoretically this is indeed unreasonable because `clone()` aliases the original pointer, while `clear()` frees it.

Under the current flow of control, this situation is difficult to trigger, but it cannot be completely guaranteed to be safe.

Fixing this issue thoroughly in this PR might expand the scope of the PR. Let's add a TODO first, and then fix it thoroughly in a separate PR.

fufesou:
I have tested this on openSUSE Tumbleweed, but I still can't reproduce the issue.

If I run RustDesk directly (without the `--server` process):
1. If I keep the connection, the screen never dims.
2. If I disconnect, the screen dims and turns off. However, the next connection can be established without any issues.

If I run the flatpak version:
I just get the error `Failed to create capturer for display 0`. No connection can be established.

I will try to fix the flatpak issue on my machine first.


image

Steve3184:
I have two monitors, one of which supports soft brightness adjustment and the other does not. Could this be causing KDE to close the session after entering the Dim Screen? I noticed that the brightness of no monitor is changed after entering the Dim Screen.

fufesou:
> Could this be causing KDE to close the session after entering the Dim Screen?

No idea.

I will fix the flatpak issue on my machine first and then reinstall ArchLinux with `KDE Frameworks 6.23.0`.

I believe your analysis is correct: https://github.com/rustdesk/rustdesk/pull/14384#issuecomment-3970818514

rustdesk:
PR Review (claude code): fix(linux): re-request Wayland screen sharing permission on session failure                                                                                                                       
                                                                                                                                                                                                               
  Overall Assessment: The PR tackles a real, user-facing pain point on Wayland (session revocation + KDE idle timeout), but has several correctness and design issues that should be addressed before merging. 
                                                                                                                                                                                                               
  ---
  Issues

  1. String-matching on error messages is fragile

  wayland.rs and pipewire.rs both detect session revocation by checking if error strings contain "SESSION_REVOKED". This is a self-referential pattern where the code creates a sentinel string and then
  searches for it through the error chain. Any intermediate formatting or wrapping could lose the substring.

  // wayland.rs
  if format!("{:?}", e).contains("SESSION_REVOKED") && retry_count < MAX_RETRIES {

  A typed error enum (e.g., WaylandError::SessionRevoked) would be far more robust and testable.

  2. is_revoked heuristic is incorrect

  let is_revoked = if !is_server_running() {
      true
  } else {
      !config::LocalConfig::get_option(RESTORE_TOKEN_CONF_KEY).is_empty()
  };

  This infers revocation from unrelated state: "server not running" or "has a restore token stored." A GStreamer PLAYING state failure has many causes (codec missing, device busy, permission denied), and
  mapping most of them to "session revoked" and silently clearing the restore token is a destructive side effect. The SESSION_REVOKED sentinel string should be set only at the actual D-Bus/PipeWire error
  site, not inferred here.

  3. Keepalive via synthetic mouse movement is a hack

  Sending periodic (0, 0) mouse movement events to prevent idle is unconventional and has side effects (e.g., it may interfere with the remote user's idle detection, screensaver, or input handling). The
  request_inhibit() approach using the org.freedesktop.ScreenSaver inhibit API is the correct solution — it's already implemented in this PR. The keepalive mouse thread should be removed in favor of relying
  solely on the inhibit API.

  4. force_close_dead_session() is called without synchronization guarantees

  pub fn force_close_dead_session() {
      if let Ok(mut rdp_info) = RDP_SESSION_INFO.lock() {
          *rdp_info = None;
          clear_wayland_displays_cache();
          HAS_POSITION_ATTR.store(false, Ordering::SeqCst);
      }
  }

  clear_wayland_displays_cache() and the HAS_POSITION_ATTR store happen outside the lock scope since the rdp_info lock is released before those calls. If another thread re-acquires the session between those
  operations, state becomes inconsistent. All three should be done atomically under the same lock, or clear_wayland_displays_cache should be called while holding the mutex.

  5. MAX_RETRIES = 1 with a bare continue creates an infinite potential if error detection is wrong

  The loop only breaks on retry_count >= MAX_RETRIES, but if the error never stops matching SESSION_REVOKED (e.g., the user never grants permission), the current MAX_RETRIES = 1 limits this to 2 attempts —
  acceptable. However, there's no delay between retries, so back-to-back portal prompts will fire immediately, which is bad UX. A short tokio::time::sleep before retry would improve this.

  6. TODO comment flagging potential use-after-free

  The PR author left a // TODO: potential use-after-free comment in the committed code. This is a safety concern that should be resolved before merge, not deferred.

  ---
  Minor Notes

  - The inhibit_request_path field added to RdpSessionInfo and its Drop cleanup are well-structured.
  - Mapping SESSION_REVOKED to io::ErrorKind::ConnectionAborted in common/wayland.rs is a reasonable approach — it just needs the error source to be reliable.
  - The refactor of check_init() → check_init_once() + retry wrapper is clean.

  ---
  Recommended Changes Before Merge

  1. Replace string-sentinel error detection with a typed error or a dedicated flag
  2. Remove the keepalive mouse movement thread (rely on inhibit API only)
  3. Fix the is_revoked heuristic — only set SESSION_REVOKED at the actual portal error site
  4. Add a small retry delay
  5. Resolve the use-after-free TODO

Steve3184:
Thanks for the detailed review. Here's what was addressed in this commit:

- **SESSION_REVOKED string sentinel**: Introduced `SessionRevokedError` as a typed error in `pipewire.rs`. The sentinel string is now only defined in its `Display` impl — all ad-hoc `format!("SESSION_REVOKED: ...")` constructions at error sites are replaced with `anyhow::Error::new(SessionRevokedError)`. The detection side in `scrap/common/wayland.rs` still uses `.contains("SESSION_REVOKED")`, which is intentional: across crate boundaries a string protocol is more practical than downcast, and with a single emit point the fragility concern is resolved.

- **is_revoked heuristic**: Kept as-is. `is_server_running()` reliably distinguishes the `remote_desktop_portal` vs `screencast_portal` paths, making it a valid proxy for whether a PLAYING failure implies revocation. The important fix is that `SessionRevokedError` is now only constructed at actual portal error sites in `pipewire.rs`, not inferred elsewhere.

- **Retry delay**: Added 500ms sleep before re-requesting the portal permission dialog to avoid back-to-back prompts.


Not addressed in this PR:

- **use-after-free TODO**: The `CapturerPtr` raw pointer aliasing issue in `server/wayland.rs` exists in the broader codebase — the same pattern appears in `video_service.rs` and related files. A proper fix requires an ownership refactor across multiple files and services, which is out of scope here. Left as a TODO comment to track separately.

- **Keepalive mouse thread**: Already superseded by the inhibit API introduced in this PR; the old mouse keepalive path is no longer exercised.

rustdesk:
2. String matching for SESSION_REVOKED (libs/scrap/src/common/wayland.rs)

  The map_err() function checks err_str.contains("SESSION_REVOKED") — but SESSION_REVOKED is the Display output of SessionRevokedError. So the flow is:
  - SessionRevokedError is created → .to_string() produces "SESSION_REVOKED" → passed as String → then re-detected via substring match

  This is an unnecessary round-trip through string serialization. The error should propagate typed (via Box<dyn Error> or a proper enum) without going through Display and back.

  3. Overly broad revocation heuristic (libs/scrap/src/wayland/pipewire.rs)

  Using is_server_running() to distinguish "configuration error" from "session revocation" is indirect. A GStreamer state-change failure can have many causes. Clearing restore tokens for unrelated pipeline
  failures will break persisted sessions unnecessarily.

  4. force_close_dead_session() synchronization gap

  pub fn force_close_dead_session() {
      if let Ok(mut rdp_info) = RDP_SESSION_INFO.lock() {
          *rdp_info = None;
          clear_wayland_displays_cache();   // <-- lock already released here?
          HAS_POSITION_ATTR.store(false, Ordering::SeqCst);
      }
  }
  clear_wayland_displays_cache() and HAS_POSITION_ATTR are updated outside the lock scope (or after the guard drops), creating a window for inconsistent state.

Steve3184:
Thanks for the review and the good points!

1. SESSION_REVOKED string matching

You are correct that checking err_str.contains("SESSION_REVOKED") is essentially a round‑trip through Display, and not type‑safe.

At the moment, the scrap crate does not depend on an error abstraction (e.g. anyhow or a custom error enum), so introducing a typed error would require refactoring error propagation across multiple layers of Wayland/PipeWire integration.

The focus of this PR is to improve Wayland's screen sharing recovery behavior, rather than to completely change the buggy model.

2. Revocation heuristic with is_server_running()

Yes, the current method to distinguish a configuration error from a revoked XDG session is heuristic.
The reason is that the portal/XDG layer does not give a structured "session revoked" error at the API boundary — in practice, during testing the only consistently reproducible case where:

- GStreamer state change fails
- and PipeWire is still running

has been an actual session revoke (e.g., screen lock or manual cancel). This makes it the most reliable discriminator we can observe given the current signal surface.

I agree it’s not ideal to rely on a wide signal like “server running”, but without a more specific error signal we simpliﬁed to this heuristic.

3. force_close_dead_session() synchronization gap

I believe the lock is still held during the calls to clear_wayland_displays_cache() and HAS_POSITION_ATTR.store(), since the MutexGuard does not drop until the end of the if let scope.

So these updates are currently within the critical section, and there shouldn’t be a window for inconsistent state.

Thanks again for the careful review!

rustdesk:
Thanks

src/server/wayland.rs
```diff
@@ -265,23 +326,64 @@ pub(super) fn get_capturer_for_display(
...
         }
+    }
+
+    log::warn!(
+        "get_capturer_for_display: display {} not found in CAP_DISPLAY_INFO.",
+        display_idx
+    );
+
+    // Wait until all active capturers have exited before reinitializing.
+    let active_count = *ACTIVE_DISPLAY_COUNT.read().unwrap();
```
fufesou, line 361: `active_count > 0` may be always `true`.

`run()` in `video_service.rs` increments `ACTIVE_DISPLAY_COUNT` before calling `get_capturer_for_display()`. So when `CAP_DISPLAY_INFO` misses the requested display, `active_count` is already at least 1, and the function `bails` out early instead of running the intended recovery (`clear()` + `ensure_inited()`).

This conflicts with the PR goal of automatic recovery after session interruption, especially for the "Failed to get capturer display info" path.

  - src/server/video_service.rs:544
  - src/server/video_service.rs:565:565
  - src/server/wayland.rs:362:362
  - src/server/wayland.rs:377:377

Steve3184: Okay, I've fixed this issue.

src/server/wayland.rs
```diff
@@ -208,12 +218,58 @@ pub(super) async fn check_init() -> ResultType<()> {
...
     Ok(())
 }
 
+pub(super) async fn check_init() -> ResultType<()> {
+    const MAX_RETRIES: usize = 1;
+    let mut retry_count = 0;
+    loop {
+        let result = check_init_once().await;
+        if let Err(ref e) = result {
+            if format!("{:?}", e).contains("SESSION_REVOKED") && retry_count < MAX_RETRIES {
```
fufesou, line 261: No retry if 

```rust
async fn check_init_once() -> ResultType<()> {
...
                if all.is_empty() {
                    log::warn!("check_init: no displays from PipeWire portal, session revoked.");
                    scrap::wayland::pipewire::close_session();
                    bail!("No displays returned by PipeWire portal. Try reconnecting to request a new screen-sharing session.");
                }
...
}
```

Is it intended?

src/server/video_service.rs
```diff
@@ -538,6 +538,15 @@ fn run(vs: VideoService) -> ResultType<()> {
     // But for now, we do not support multi-screen capture on wayland.
     #[cfg(target_os = "linux")]
     super::wayland::ensure_inited()?;
+
```
fufesou: https://github.com/rustdesk/rustdesk/pull/14384/changes/788c5d406ce13e12b9d496696f18191cab28b4e8

This commit introduces a new UAF.

By moving `increment_active_display_count()` after `get_capturer(...)`, there is now a race window:

1. Thread A calls `get_capturer_for_display()` and gets a capturer clone (which is a raw-pointer alias, not ownership).
2. Before Thread A increments `ACTIVE_DISPLAY_COUNT`, Thread B exits capture, decrements to 0, and runs `wayland::clear()`.
3. `clear()` frees the underlying capturer storage from `CAP_DISPLAY_INFO`.
4. Thread A continues using c and may hit a use-after-free.

See:
  - src/server/video_service.rs:549
  - src/server/video_service.rs:553:553
  - src/server/video_service.rs:560:560
  - src/server/wayland.rs:313:313
  - src/server/wayland.rs:344:344

src/server/wayland.rs
Steve3184: In fact, the retry loop only handles session expiration in response to capture initialization errors. For other errors, the retry loop will not handle them because it may indicate other errors.

src/server/video_service.rs
Steve3184: Okay, I've fixed it and adjusted the position of the incrementing ACTIVE_DISPLAY_COUNT.

src/server/wayland.rs
```diff
@@ -349,6 +349,7 @@ pub(super) fn get_capturer_for_display(
     {
         let cap_map = CAP_DISPLAY_INFO.read().unwrap();
         if let Some(&addr) = cap_map.get(&display_idx) {
+            increment_active_display_count();
```
fufesou: Potential deadlock if there is another thread calling `clear()` -> `CAP_DISPLAY_INFO.write()`

`CAP_DISPLAY_INFO.read()` in `increment_active_display_count()`

https://doc.rust-lang.org/std/sync/struct.RwLock.html#:~:text=Potential%20deadlock%20example

src/server/wayland.rs
```diff
@@ -383,6 +384,7 @@ pub(super) fn get_capturer_for_display(
             "get_capturer_for_display: re-initialization succeeded for display {}.",
             display_idx
         );
+        increment_active_display_count();
```
fufesou: Lifecycle accounting became harder to reason about (maintainability).

`increment`/`decrement` were visually paired via the same RAII block (`_wayland_call_on_ret`).

Steve3184: I restored `video_service.rs` and changed `active_count > 0` to `active_count > 1` in `wayland.rs`, which may have solved the problem and was the smallest possible change.

src/server/wayland.rs
Steve3184: I removed the CAP_DISPLAY_INFO.read() lock from the increment_active_display_count() function. This eliminated a potential deadlock where the increment function (holding a Read lock) waited for an ACTIVE lock (Write lock), while the clear function (holding an ACTIVE lock through the caller) waited for a CAP lock (Write lock).

libs/scrap/src/wayland/pipewire.rs
```diff
@@ -306,7 +335,32 @@ impl PipeWireRecorder {
...
+                true
+            } else {
+                // screencast_portal: only treat as revoked when a persisted restore token
+                // exists, meaning we were trying to resume a prior session.
+                !config::LocalConfig::get_option(RESTORE_TOKEN_CONF_KEY).is_empty()
+            };
+
+            if is_revoked {
+                warn!("[gstreamer] Failed to set PLAYING state, session was likely revoked: {:?}", e);
+                config::LocalConfig::set_option(RESTORE_TOKEN_CONF_KEY.to_owned(), "".to_owned());
```
fufesou: Hi, the "restore token" is only supported for the native installed version (--server version).

You mentioned that "these issues only occur when the RustDesk service is not installed". Based on that, I'm wondering:

`config::LocalConfig::set_option(RESTORE_TOKEN_CONF_KEY.to_owned(), "".to_owned());` — this might be unintentionally clearing the restore token.

If we have to clear the "restore token", checking `is_server_running()` may be better.

Other locations that clear the "restore token" could also be affected.

libs/scrap/src/wayland/pipewire.rs
```diff
@@ -934,6 +1030,7 @@ pub fn get_capturables() -> Result<Vec<PipeWireCapturable>, Box<dyn Error>> {
     if rdp_connection.is_none() {
         let (conn, fd, streams, session, is_support_restore_token) = request_remote_desktop(false)?;
         let conn = Arc::new(conn);
+        let inhibit_path = request_inhibit(&conn);
```
fufesou: Inhibit may never be released in non-server mode.
This can keep idle/suspend inhibition enabled even when there is no active session.

You mentioned that "these issues only occur when the RustDesk service is not installed".

libs/scrap/src/wayland/pipewire.rs
Steve3184: OK, I've fixed this by adding a check on `is_server_running()` to only clear `RESTORE_TOKEN` when it's possible

libs/scrap/src/wayland/pipewire.rs
Steve3184: In fact, it's precisely because the Inhitbit lock needs to be held continuously when running in non-server mode that the system is prevented from entering the Dim Screen, thus keeping the session from being closed. Therefore, I believe this handling is likely correct.

Failing to hold the Inhitbit lock continuously when there are no active remote connections could cause KDE to close the session.

fufesou: But not all OSes  (flatpak) have this issue.

Should we:

1. Only call `request_inhibit()` when `!is_server_running()`
2. Try to call `release_inhibit()` when there is no active session? This PR will retry to call `request_inhibit()` then?

Steve3184: You're absolutely right. We do need to acquire the Inhibit lock only during `!is_server_running()`, because the server process doesn't need to hold the Inhibit lock while it's running.

> 2\. Try to call `release_inhibit()` when there is no active session? This PR will retry to call `request_inhibit()` then?

However, if the Inhibit lock is released after no active session, after the session is terminated by KDE, reapplying for the Inhibit lock will not restore the session.

So it should be enough to just get the fix for the Inhibit lock at `!is_server_running()`.

fufesou: > reapplying for the Inhibit lock will not restore the session.

Ok.

However, I'm concerned that this fix might be **too broad**.

Not every KDE Plasma environment has this issue, and it could also impact users on **GNOME** or other desktops that were working fine.

Steve3184: Yes, imposing an Inhibit lock does have a wide-ranging impact, but for now, we can retain this logic and add an option like "keep awake while RustDesk is running" in a subsequent PR.

Adding this switch to the PR would directly make the scope of the PR too broad.

fufesou: Ok.

Please add a `TODO` comment.

I'll continue reviewing.

libs/scrap/src/wayland/pipewire.rs
```diff
@@ -306,7 +335,34 @@ impl PipeWireRecorder {
...
-        pipeline.set_state(gst::State::Playing)?;
+        if let Err(e) = pipeline.set_state(gst::State::Playing) {
+            let _ = pipeline.set_state(gst::State::Null);
+
+            let is_revoked = if !is_server_running() {
+                // remote_desktop_portal: the portal opened the fd, so a PLAYING failure
+                // means the underlying PipeWire node is gone -> session revoked.
+                true
+            } else {
+                // screencast_portal: only treat as revoked when a persisted restore token
```
fufesou: https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html#:~:text=If%20the%20stored%20session%20cannot%20be%20restored%2C%20this%20value%20is%20ignored%20and%20the%20user%20will%20be%20prompted%20normally.%20This%20may%20happen%20when%2C%20for%20example%2C%20the%20session%20contains%20a%20monitor%20or%20a%20window%20that%20is%20not%20available%20anymore%2C%20or%20when%20the%20stored%20permissions%20are%20withdrawn.

This change still seems too broad.

According to thexdg-desktop-portal documentation, if a stored session cannot be restored (e.g., the monitor/window is no longer available or permissions are withdrawn), it is simply ignored, and the user is prompted normally. This suggests there's no need for us to manually handle or clear the restore token in most cases.

Perhaps the logic here should be simplified to:
```
if is_server_running() {
    bail!("..."); // Handle the server-specific case
} else {
    // Session is invalid/revoked; proceed with normal prompt flow
    // ...
}
```

Steve3184: Okay, I've simplified the relevant logic and let the portal handle the restore token.

rustdesk:
Review from codex:

> • 1. Medium: Session-revoked retry logic is bypassed for the “no displays” failure path.
     In wayland.rs:255, check_init() retries only when the error contains SESSION_REVOKED. But in wayland.rs:170-wayland.rs:175, the code explicitly treats empty display list as “session revoked” and returns
     a different error message without that sentinel. Result: this common revoked-session path won’t trigger the new retry/re-request behavior.

Steve3184:
> Result: this common revoked-session path won’t trigger the new retry/re-request behavior.

Only errors from GStreamer can accurately determine whether the user manually revoked permissions or the session was terminated due to entering the Dim Screen. Errors triggered here may not be due to session expiration, so I believe that retrying should not be performed at this time.

EmberLightVFX:
What's the status for this PR?

rustdesk:
Not ready for merge.

There’s also something we haven’t fully figured out yet:
in `wayland.rs`, it used to be if active_count > 0, and now it’s been changed to if active_count > 1. we are not sure whether this could cause issues.

Also, this fix introduces a side effect:

- All Flatpak and AppImage apps now add request_inhibit, which can block the screen from turning off even when there’s no active connection.

- It may introduce a new race condition.

The latter two side effects have already been noted with TODOs, since fixing them would expand the scope of the PR.

Steve3184:
> There’s also something we haven’t fully figured out yet:
> in `wayland.rs`, it used to be if active_count > 0, and now it’s been changed to if active_count > 1. we are not sure whether this could cause issues.

The purpose of checking `active_count` is to determine whether PipeWire should be re-initialized in the event that the corresponding display cannot be found within `CAP_DISPLAY_INFO`. 

- Using `> 0`: The current `run` thread has already incremented the count to 1 itself; therefore, the condition `> 0` would always be met, and the code would never proceed down the re-initialization path.
- Using `> 1`: This signifies that "aside from my own thread, there are other capturers currently running." Only in this scenario should the re-initialization be skipped, as calling `clear()` at this point would adversely affect the other active capturers. If this thread is the only one active (i.e., `count == 1`), it implies that no other entities are dependent on the current session, making it safe to call `clear()` + `ensure_inited()`. 

Conclusion:

Using `> 1` is the correct modification.

Steve3184:
After testing, the issue where the session would unexpectedly terminate when the device entered Dim Screen has been fixed in version 6.6.4-1 of xdg-desktop-portal-kde.

Therefore, I have reverted the changes related to inhibit lock to reduce the scope of this PR.

This PR will mainly address the issue where RustDesk does not re-request permissions after the user manually terminates the session.

Issue #13761, opened 2025-12-09
Missing scroll multiplyer for Mac Os Client


BierDav:
### Bug Description

When scrolling in MacOS from a Windows host it one scroll tick only scrolls a few pixels instead of a 3 lines which is default for windows. This is very annying when browsing or coding remotely.

### How to Reproduce

Just scroll anywhere with a Logitech Mouse.

### Expected Behavior

A setting to mimic the same behavior as TeamViewer

### Operating system(s) on local (controlling) side and remote (controlled) side

Windows 11 -> Mac OS 26

### RustDesk Version(s) on local (controlling) side and remote (controlled) side

1.4.4

### Screenshots

RustDesk (i really had to free spin the wheel so that i get somewhere in a sigificant amount of time):

https://github.com/user-attachments/assets/c83b48e5-2e59-44f1-bbff-531e3645acc8

Teamviewer for comparison (each of the big jumps is one wheel tick, even tough it doesn't look smooth, it feels exactly right for a windows user):

https://github.com/user-attachments/assets/48e8905f-1d39-49ca-8d20-c2ebecb6adbf

pc-support-services:
Completely agree.

amirHdev:
Duo to this #14395 patch can we close it ?

PR #14395, opened 2026-02-24
ui: scale wheel lines on Windows/Linux to Mac

amirHdev:
## Summary
Improve fast wheel scrolling when controlling macOS from Windows/Linux by adding **burst-only acceleration**. Normal single step scrolling remains unchanged.

## Changes
- Detect rapid wheel bursts using timing and velocity thresholds.
- Keep single‑step scrolling unchanged to preserve precise control.
- Reset wheel timing on enter/leave to avoid stale burst detection.
- Align dominant‑axis handling between Sciter and Flutter.

## Rationale
Testing showed single notch scrolling was already correct on master, but fast “flick” scrolling felt under‑scaled. this change targets only rapid bursts to improve paging speed without affecting normal scrolling.

## Notes
- Burst thresholds are empirically tuned and documented inline.
- Applies only for Windows/Linux → macOS.

### DIFF
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -371,7 +371,7 @@ class InputModel {
   static const int _wheelAccelFastThresholdUs = 40000; // 40ms
   static const int _wheelAccelMediumThresholdUs = 80000; // 80ms
   static const double _wheelBurstVelocityThreshold =
-      0.005; // delta units per microsecond
+      0.002; // delta units per microsecond
   // Wheel burst acceleration (empirical tuning).
   // Applies only to fast, non-smooth bursts to preserve single-step scrolling.
   // Flutter uses microseconds for dt, so velocity is in delta/us.
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -365,6 +365,13 @@ class InputModel {
   final isPhysicalMouse = false.obs;
   int _lastButtons = 0;
   Offset lastMousePos = Offset.zero;
+  int _lastWheelTsUs = 0;
+
+  // Wheel acceleration thresholds.
+  static const int _wheelAccelFastThresholdUs = 40000; // 40ms
+  static const int _wheelAccelMediumThresholdUs = 80000; // 80ms
+  static const double _wheelBurstVelocityThreshold =
+      0.002; // delta units per microsecond
 
   // Relative mouse mode (for games/3D apps).
   final relativeMouseMode = false.obs;
@@ -964,6 +971,7 @@ class InputModel {
     toReleaseRawKeys.release(handleRawKeyEvent);
     _pointerMovedAfterEnter = false;
     _pointerInsideImage = enter;
+    _lastWheelTsUs = 0;
 
     // Fix status
     if (!enter) {
@@ -1407,17 +1415,44 @@ class InputModel {
     if (isViewOnly) return;
     if (isViewCamera) return;
     if (e is PointerScrollEvent) {
-      var dx = e.scrollDelta.dx.toInt();
-      var dy = e.scrollDelta.dy.toInt();
+      final rawDx = e.scrollDelta.dx;
+      final rawDy = e.scrollDelta.dy;
+      final dominantDelta = rawDx.abs() > rawDy.abs() ? rawDx.abs() : rawDy.abs();
+      final isSmooth = dominantDelta < 1;
+      final nowUs = DateTime.now().microsecondsSinceEpoch;
+      final dtUs = _lastWheelTsUs == 0 ? 0 : nowUs - _lastWheelTsUs;
+      _lastWheelTsUs = nowUs;
+      int accel = 1;
+      if (!isSmooth &&
+          dtUs > 0 &&
+          dtUs <= _wheelAccelMediumThresholdUs &&
+          Platform.isWindows &&
+          peerPlatform == kPeerPlatformMacOS) {
+        final velocity = dominantDelta / dtUs;
+        if (velocity >= _wheelBurstVelocityThreshold) {
+          if (dtUs < _wheelAccelFastThresholdUs) {
+            accel = 3;
+          } else {
+            accel = 2;
+          }
+        }
+      }
+      var dx = rawDx.toInt();
+      var dy = rawDy.toInt();
+      if (rawDx.abs() > rawDy.abs()) {
+        dy = 0;
+      } else {
+        dx = 0;
+      }
       if (dx > 0) {
-        dx = -1;
+        dx = -accel;
       } else if (dx < 0) {
-        dx = 1;
+        dx = accel;
       }
       if (dy > 0) {
-        dy = -1;
+        dy = -accel;
       } else if (dy < 0) {
-        dy = 1;
+        dy = accel;
       }
       bind.sessionSendMouse(
           sessionId: sessionId,
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -372,6 +372,9 @@ class InputModel {
   static const int _wheelAccelMediumThresholdUs = 80000; // 80ms
   static const double _wheelBurstVelocityThreshold =
       0.002; // delta units per microsecond
+  // Wheel burst acceleration (empirical tuning).
+  // Applies only to fast, non-smooth bursts to preserve single-step scrolling.
+  // Flutter uses microseconds for dt, so velocity is in delta/us.
 
   // Relative mouse mode (for games/3D apps).
   final relativeMouseMode = false.obs;
@@ -1426,7 +1429,7 @@ class InputModel {
       if (!isSmooth &&
           dtUs > 0 &&
           dtUs <= _wheelAccelMediumThresholdUs &&
-          Platform.isWindows &&
+          (isWindows || isLinux) &&
           peerPlatform == kPeerPlatformMacOS) {
         final velocity = dominantDelta / dtUs;
         if (velocity >= _wheelBurstVelocityThreshold) {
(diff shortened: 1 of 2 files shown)


rustdesk:
Great job. Does not flutter version need this?

amirHdev:
> Great job. Does not flutter version need this?

Probably yes. the fix I made is only in the Sciter UI path .
The Flutter UI uses a different input path and would need equivalent scaling logic. 

 I can add the Flutter side . 
 Thought ?

fufesou:
Here are my test observations:

A single mouse wheel notch scrolls roughly 2–3 lines in VS Code — this is consistent on both Windows and macOS.

My testing suggests the master branch already scrolls at a reasonable speed, and this patch actually makes scrolling too fast.

On both the Flutter build and the master Sciter build, one wheel notch scrolls about 3 lines. With this patch applied, it scrolls about 8 lines instead.

### master

https://github.com/user-attachments/assets/83874ea3-2e54-444c-9f8b-3a12c64faf33

https://github.com/user-attachments/assets/7fba2bac-57e9-4fc1-b194-08f61cbb6465

### this patch

https://github.com/user-attachments/assets/6548ce18-9296-4f31-b4bb-011114c3a062

fufesou:
e5476846-e1e3-4718-b4e4-7c8244d9d9a7

I can understand `if ok == 0 { return 3 }`, but why `if lines == 0 || lines == WHEEL_PAGESCROLL` also returns 3?

fufesou:
> Windows users expect each wheel tick to scroll multiple lines (typically 3). Without scaling, macOS receives tiny deltas and scrolling feels painfully slow. 

Sorry, I'm not sure I understand the issue clearly, and I can't reproduce it on my end.

In my test, this patch makes scrolling fast. https://github.com/rustdesk/rustdesk/pull/14395#issuecomment-3958744213

Could you please share a short video?

amirHdev:
@fufesou 
Thanks for the testing and detailed measurements. your results show the current patch over scales wheel input. 
master already scrolls around 3 lines per notch in your setup, while this change pushes it to 8+ lines. that means we’re effectively double‑counting the Windows “lines per notch” setting.

I agree this shouldn’t be the default behavior. I’ll revise the approach to avoid extra scaling and I’ll drop the lines == 0 / WHEEL_PAGESCROLL -> 3 fallbacks since those are not correct semantics

amirHdev:
@fufesou 
Instead of changing default behavior, I propose adding a user side setting for Windows → macOS wheel scaling . that keeps current behavior for most users while letting those who want faster scrolling opt in.

If that sounds reasonable, I can prepare a follow up patch that adds this option in both Sciter and Flutter settings
Thought ?

fufesou:
> I propose adding a user side setting for Windows → macOS wheel scaling

I'm not sure, but I have 2 questions.

1. Why "Windows → macOS"?

2. Is the default speed too slow for you?

amirHdev:
@fufesou 
We are seeking solution for this issue #13761

fufesou:
Thank you for looking into this issue and providing a fix.

I have done some testing.

When scrolling the mouse wheel rapidly (large scrolls) in RustDesk, the scroll amount on the remote side is indeed insufficient.

1. However, this may not be related to small incremental scrolls. When testing small scrolls in VSCode:

    - The master branch behaves correctly
    - This patch results in too much scrolling

2. When testing rapid/large scrolls in VSCode, both the master branch and this patch exhibit the same issue: the scroll amount is too low.

Maybe we should handle rapid/large scrolls differently from single-step scrolls.

amirHdev:
@fufesou 
Thanks for taking the time to test this. 
your results make it clear we’re seeing two different behaviors: normal single‑step scrolling already feels right on master, while fast “flick” scrolling still comes through too weak. 
That means my patch was the wrong fix for the common case. Rather than changing the default line‑based scrolling, I think the better path is to detect high velocity scroll bursts and only apply acceleration there, so we preserve the good baseline behavior and fix the under scroll when someone spins the wheel quickly.

In other hand, i can recommend the Customized Setting for users , to able to customize it .

fufesou:
>  I think the better path is to detect high velocity scroll bursts and only apply acceleration there, so we preserve the good baseline behavior and fix the under scroll when someone spins the wheel quickly.

Agree.

> In other hand, i can recommend the Customized Setting for users , to able to customize it .

I'm not sure if this is really needed.

The best solution might simply be to rely on the OS's wheel speed, which already accounts for both slow and fast scrolling.

Customized speed could make it complicated with little benefit.

amirHdev:
@fufesou 
I reworked the change .
The original scroll lines scaling is removed, since it over scaled normal wheel steps. 
The new approach only accelerates rapid wheel bursts and only on Windows to macOS, so single notch scrolling stays the same while fast flicks feel more natural. both Sciter and Flutter paths are updated.

amirHdev:
@fufesou 
what you think about this ? https://github.com/rustdesk/rustdesk/actions/runs/22448675686/job/65014693999?pr=14395

is it related to the Github ?

fufesou:
> what you think about this ? https://github.com/rustdesk/rustdesk/actions/runs/22448675686/job/65014693999?pr=14395
is it related to the Github ?

Probably, it might be a temporary infrastructure issue. A simple rerun may fix it.

fufesou:
Nice rework. The burst-only strategy is clear.

I've tested it, and it works very well!

Although this change only affects the `Windows -> macOS` direction, the patch itself looks good.

  Non-blocking request for maintainability: could we add a short note (PR description or code comment) on:
  1) the algorithm rule (when accel is applied, and how 1/2/3 is selected),
  2) unit mapping between Flutter and Sciter thresholds (`delta/us` vs `delta/s`),
  3) whether the constants are empirical tuning values.

A short comment near the code is enough, but this context is important for review and future maintenance.

src/ui/remote.tis
fufesou, line 334: Maybe `last_wheel_time = 0;` is needed before the `return`.

fufesou, line 323: Maybe `last_wheel_time = 0;` is needed before the `return`.

flutter/lib/models/input_model.dart
```diff
@@ -1407,17 +1415,44 @@ class InputModel {
...
+      final dominantDelta = rawDx.abs() > rawDy.abs() ? rawDx.abs() : rawDy.abs();
+      final isSmooth = dominantDelta < 1;
+      final nowUs = DateTime.now().microsecondsSinceEpoch;
+      final dtUs = _lastWheelTsUs == 0 ? 0 : nowUs - _lastWheelTsUs;
+      _lastWheelTsUs = nowUs;
+      int accel = 1;
+      if (!isSmooth &&
+          dtUs > 0 &&
+          dtUs <= _wheelAccelMediumThresholdUs &&
+          Platform.isWindows &&
```
fufesou: `(isWindows || isLinux)` may be better.

`isWindows` is already defined, no need to use `Platform.isWindows`.

I have tested, linux -> macOS also needs this patch.

src/ui/remote.tis
```diff
@@ -237,12 +242,28 @@ function handler.onMouse(evt)
...
+            var dt = last_wheel_time > 0 ? (now - last_wheel_time) / 1000 : 0;
+            var (raw_dx, raw_dy) = evt.wheelDeltas;
+            var dx = 0;
+            var dy = 0;
+            var abs_dx = Math.abs(raw_dx);
+            var abs_dy = Math.abs(raw_dy);
+            var dominant = abs_dx > abs_dy ? abs_dx : abs_dy;
+            var is_smooth = dominant < 1;
+            var accel = 1;
+            if (!is_smooth && dt > 0 && is_win && get_peer_platform() == "Mac OS") {
```
fufesou: `is_win` to `(is_win || is_linux)`

src/ui/remote.tis
amirHdev: oh , you are right

amirHdev:
@fufesou I updated the PR description to match the burst acceleration approach .
I also aligned the velocity threshold between Sciter and Flutter so they behave consistently
Thoughts ?

fufesou:
> I updated the PR description to match the burst acceleration approach .

Great!

The wheel scrolling feels much more comfortable.

> I also aligned the velocity threshold between Sciter and Flutter so they behave consistently

I will test it again.

flutter/lib/models/input_model.dart
```diff
@@ -371,7 +371,7 @@ class InputModel {
   static const int _wheelAccelFastThresholdUs = 40000; // 40ms
   static const int _wheelAccelMediumThresholdUs = 80000; // 80ms
   static const double _wheelBurstVelocityThreshold =
-      0.002; // delta units per microsecond
+      0.005; // delta units per microsecond
```
fufesou: 0.002 looks better **to me**.

0.005 makes scrolling slower.

Maybe there's no need to align this threshold.

fufesou:
Please also close these suggestions if they are solved or invalid.  https://github.com/rustdesk/rustdesk/pull/14395#pullrequestreview-3867159232

amirHdev:
@fufesou I reverted Flutter’s velocity threshold to 0.002 since it feels better in practice and I agree on it m
would you prefer I squash these updates into a single commit before merge ?

fufesou:
> would you prefer I squash these updates into a single commit before merge ?

No need to squash

fufesou (approved):

rustdesk:
Can this solve the problem of acceleration of touchpad?

amirHdev:
@rustdesk this change is designed for mouse wheel bursts,not touchpad scrolling. It explicitly avoids smooth trackpad style deltas/
so touchpad acceleration shouldn’t be affected
Do you mind to provide new patch for it ?

rustdesk:
> Do you mind to provide new patch for it ?

Yes, if you can.

Issue #2655, opened 2022-12-26
scrolling with trackpad+mouse is very fast


working-name:
**Describe the bug you encountered:**
When I remote into a windows host from macOS using a trackpad the scrolling is very fast. 

Saw the only other (now closed) issue which mentioned mouse scrolling was normal. If I connect a bluetooth mouse to macOS the scrolling speed does not change, it's still very very fast.

I would not be bothered if it wasn't such a pain in the butt to generate `Page Down` on the mac keyboard 😞
...

**What did you expect to happen instead?**
Scroll speed to be similar to macOS itself. 
...

**How did you install `RustDesk`?**
site.

---

**RustDesk version and environment**

client: macOS Ventura 13.1
host: Windows 10, build 2004

Connections going through self-hosted relay server.

rustdesk:
nightly build please.

working-name:
tried nightly, mouse is still fast but not as fast, I'd say about 75% of trackpad speed. So we're moving in the right direction. Thank you!

rustdesk:
@fufesou

rustdesk:
#1547

rustdesk:
https://github.com/rustdesk/rustdesk/issues/1474

rustdesk:
Wondering if this helps https://medium.com/flutter/whats-new-in-flutter-3-7-38cbea71133c#:~:text=Flutter%20apps%20page.-,Global%20selection%20improvements,-SelectionArea%20now%20supports

rustdesk:
Confirmed, still very fast.

rustdesk:
As stated in #3717, sciter version works fine, so there is difference between flutter and sciter touchpad event.

rustdesk:
@fufesou let's figure out the diff between sciter and flutter first.

rustdesk:
@li-dianzheng  could you try out new nightly build tomorrow?

cheywood:
Still too fast here from Debian testing running Wayland (to Ubuntu 22.04 Xorg), using nightly both ends

rustdesk:
> Still too fast here from Debian testing running Wayland (to Ubuntu 22.04 Xorg), using nightly both ends

@fufesou

li-dianzheng:
> 你能明天尝试新的夜间构建吗？ @li-dianzheng 你明天能试试新的每晚构建吗？

It's still

GCSBOSS:
[...]

PR #14426, opened 2026-02-28
input: add trackpad burst acceleration

amirHdev:
### Summary
Based on PR #14395, this fixes touchpad scrolling behavior on macOS by improving burst acceleration while preserving normal precise scrolling. with help of #14460 .

## Summary by CodeRabbit

* **New Features**
  * Improved trackpad responsiveness with per-device timing and conditional cross-platform acceleration for smoother cursor and pan-zoom control when interacting between platforms.
  * Enhanced axis-lock filtering for more predictable horizontal/vertical pans and reduced unintended diagonal movement.

* **Bug Fixes**
  * Prevents jumpy or inconsistent trackpad motion by resetting timing on pointer enter/leave and at pan-zoom boundaries.

### DIFF
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -366,6 +366,7 @@ class InputModel {
   int _lastButtons = 0;
   Offset lastMousePos = Offset.zero;
   int _lastWheelTsUs = 0;
+  int _lastTrackpadTsUs = 0;
 
   // Wheel acceleration thresholds.
   static const int _wheelAccelFastThresholdUs = 40000; // 40ms
@@ -375,6 +376,8 @@ class InputModel {
   // Wheel burst acceleration (empirical tuning).
   // Applies only to fast, non-smooth bursts to preserve single-step scrolling.
   // Flutter uses microseconds for dt, so velocity is in delta/us.
+  static const double _trackpadAxisNoiseThreshold = 0.1;
+  static const double _trackpadAxisLockRatio = 2.0;
 
   // Relative mouse mode (for games/3D apps).
   final relativeMouseMode = false.obs;
@@ -975,6 +978,7 @@ class InputModel {
     _pointerMovedAfterEnter = false;
     _pointerInsideImage = enter;
     _lastWheelTsUs = 0;
+    _lastTrackpadTsUs = 0;
 
     // Fix status
     if (!enter) {
@@ -1143,6 +1147,7 @@ class InputModel {
   void onPointerPanZoomStart(PointerPanZoomStartEvent e) {
     _lastScale = 1.0;
     _stopFling = true;
+    _lastTrackpadTsUs = 0;
     if (isViewOnly) return;
     if (isViewCamera) return;
     if (peerPlatform == kPeerPlatformAndroid) {
@@ -1172,6 +1177,8 @@ class InputModel {
     if (isMacOS && peerPlatform == kPeerPlatformWindows) {
       delta *= _trackpadAdjustMacToWin;
     }
+    _lastTrackpadTsUs = e.timeStamp.inMicroseconds;
+    delta = _filterTrackpadDeltaAxis(delta);
     _trackpadLastDelta = delta;
 
     var x = delta.dx.toInt();
@@ -1202,6 +1209,22 @@ class InputModel {
             msg: '{"type": "trackpad", "x": "$x", "y": "$y"}');
       }
     }
+  }
+
+  Offset _filterTrackpadDeltaAxis(Offset delta) {
+    final absDx = delta.dx.abs();
+    final absDy = delta.dy.abs();
+    if (absDx < _trackpadAxisNoiseThreshold &&
+        absDy < _trackpadAxisNoiseThreshold) {
+      return delta;
+    }
+    if (absDy >= absDx * _trackpadAxisLockRatio) {
+      return Offset(0, delta.dy);
+    }
+    if (absDx >= absDy * _trackpadAxisLockRatio) {
+      return Offset(delta.dx, 0);
+    }
+    return delta;
   }
 
   void _scheduleFling(double x, double y, int delay) {


rustdesk:
Great job!

The suggestion from coderabbit seems to make sense.

> `1187-1189`: **Use `PointerPanZoomUpdateEvent.timeStamp` instead of `DateTime.now()` for delta timing.**
> 
> `DateTime.now()` is wall-clock time and can drift due to system clock adjustments; `PointerEvent.timeStamp` is monotonic (system uptime-based) and designed for measuring intervals between input events. This will make trackpad velocity calculations more reliable.
> 
> 
> ```diff
> -      final nowUs = DateTime.now().microsecondsSinceEpoch;
> +      final nowUs = e.timeStamp.inMicroseconds;
>        final dtUs = _lastTrackpadTsUs == 0 ? 0 : nowUs - _lastTrackpadTsUs;
>        _lastTrackpadTsUs = nowUs;
> ```

fufesou:
Hi, thanks for the contribution.

macOS -> Windows: When scrolling vertically on the touchpad, horizontal scrolling is easily triggered. This causes unintended horizontal movement and negatively impacts the user experience.

Windows -> macOS: The performance is much better - there's only occasional, slight horizontal scrolling, which doesn't significantly affect usability.

Can this PR also address this issue?

rustdesk:
Review from codex:

> The new burst logic in flutter/lib/models/input_model.dart is heuristic (dt + velocity thresholds), so behavior risk is mainly tuning-related: fast two-finger pans from Windows/Linux clients to macOS peers may feel too aggressive or inconsistent across hardware.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14426#issuecomment-3979050413 is addressed here https://github.com/rustdesk/rustdesk/pull/14460

amirHdev:
@fufesou 
I’ve added an axis‑lock for macOS → Windows trackpad scrolling. when one axis clearly dominates , we zero the minor axis to avoid unintended horizontal drift. this should address the vertical scroll causing horizontal movement while keeping normal diagonal gestures possible.
Do you want the ratio tuned ?

fufesou:
> I’ve added an axis‑lock for macOS → Windows trackpad scrolling. when one axis clearly dominates , we zero the minor axis to avoid unintended horizontal drift. this should address the vertical scroll causing horizontal movement while keeping normal diagonal gestures possible. 

https://github.com/rustdesk/rustdesk/pull/14460 applies the same fix.

You can verify if it is ok and merge the changes into this PR.

> Do you want the ratio tuned ?

This may not be necessary.

You can verify the PR 114460

fufesou:
Hi @amirHdev could you please also address the following issues?

Users report that the current inertial scrolling feels less natural than native behavior.
Do you find it acceptable as is?

There's also a report about scrolling being too fast: https://github.com/rustdesk/rustdesk/issues/2655

amirHdev:
@fufesou I’m hesitant to change inertia behavior in this PR because the current patch is focused on trackpad burst acceleration.
The scroll too fast reports in #2655 appear broader and likely need a dedicated adjustment path . possibly a user tunable trackpad speed slider or a calibrated default for Flutter vs Sciter. 
I’m happy to open a follow up and investigate.
What you think ?

fufesou:
> I’m hesitant to change inertia behavior in this PR because the current patch is focused on trackpad burst acceleration.
The scroll too fast reports in https://github.com/rustdesk/rustdesk/issues/2655 appear broader and likely need a dedicated adjustment path . possibly a user tunable trackpad speed slider or a calibrated default for Flutter vs Sciter.
I’m happy to open a follow up and investigate.
What you think ?

Agree, it's better to keep a PR focused.

amirHdev:
@fufesou Thanks. I merged the axis filter approach from #14460 into this PR. The trackpad delta is now filtered via _filterTrackpadDeltaAxis after scaling/burst accel which reduces accidental horizontal drift across directions while preserving intended diagonals.

fufesou:
Sorry, I don't quite understand why this PR aims to increase the scroll distance for fast trackpad swipes.

In my local testing, the scroll distance was the same regardless of swipe speed.

I think there might be a misunderstanding here: https://github.com/rustdesk/rustdesk/pull/14395#issuecomment-3976179253. The term "touchpad acceleration" actually refers to "inertial scrolling."

flutter/lib/models/input_model.dart
```diff
@@ -1172,6 +1185,25 @@ class InputModel {
...
+        final velocity = dominantDelta / dtUs;
+        if (velocity >= _trackpadBurstVelocityThreshold) {
+          final factor = dtUs < _trackpadAccelFastThresholdUs
+              ? _trackpadAccelFastFactor
+              : _trackpadAccelMediumFactor;
+          delta *= factor;
+        }
+      }
+    }
+    if (peerPlatform != kPeerPlatformWindows || !isMacOS) {
```
fufesou: Maybe there is no need to add the condition `if (peerPlatform != kPeerPlatformWindows || !isMacOS) {`.

Putting `delta = _filterTrackpadDeltaAxis(delta);` after the acceleration may also be affected by the factor.

amirHdev:
I agree “touchpad acceleration” here should be inertial behavior not distance multipliers, so I’ve removed the burst accel logic entirely. the PR now only applies a dominant axis filter to reduce cross‑axis noise and it’s applied before quantization so it isn’t affected by any scaling factor. I removed the cross platform guard to keep behavior consistent.

Thoughts @fufesou ?

fufesou:
Since the trackpad acceleration feature in this PR may not be necessary, we merged https://github.com/rustdesk/rustdesk/pull/14460 

The current changes are also failing to compile.

> the PR now only applies a dominant axis filter to reduce cross‑axis noise and it’s applied before quantization so it isn’t affected by any scaling factor. I removed the cross platform guard to keep behavior consistent.

Maybe there's no need to make changes in this PR.

PR #14460 helps, but trackpad behavior still isn't as smooth as local—there's noticeable jitter at swipe ends.

A new PR to further improve this would be great.

Issue #3762, opened 2023-03-24
Audio no sound


LaiYueTing:
### Bug Description

When I connect to remote machines, I am unable to hear the audio playback from those machines.

I have enabled the following settings in my playback device:
Default device: Astro MixAmp Pro Game
Default communication device: Astro MixAmp Pro Voice

If the default device is set to Astro MixAmp Pro Game, I will not be able to hear the audio playback from remote machines. However, if the default device is set to Astro MixAmp Pro Voice, I will be able to hear the audio playback from remote machines. Even if I use the default High Definition Audio Device, I can still hear the audio playback from remote machines.

Even if I directly configure the audio input device in RustDesk, it does not work.

It is important to note that when using other remote desktop connections (such as AnyDesk, Teamviewer, etc.), I am able to hear the audio playback from remote machines.

The only way to fix this issue is by changing the Windows playback device settings.
However, due to some reasons, I do not want to switch to other audio input devices to resolve this problem.
I am looking for a better solution to fix this issue.

### How to Reproduce

1. Connect to remote machine
2. Play sound on remote machine
3. Can't hear the played sound

### Expected Behavior

I should hear the sound as I do with AnyDesk / Teamviewer connect remote machines.

### Operating system(s) on local side and remote side

Windows 10 22H2, Windows 11 22H2

### RustDesk Version(s) on local side and remote side

1.1.9, 1.2.0 Nightly Build

### Screenshots

No screenshots

rustdesk:
Please try out the nightly build tomorrow.

hellblue:
i tried the nightly
from win 10 22H2 to win 10 22H2 no sound from android to win 10 i hear sound
from win 10 22H2 to win 11 22H2 i hear no sound

rustdesk:
it is still old build, new build is ready 3 hours later.

21pages:
> still no sound there :(

Which device is the controlled one, android or windows?

Just4Link:
i control with a android phone my windows 10 machine ... anydesk runs fine with sound

Just4Link:
when i control the android phone audio wont work too ... i gave all the access control but no sound from android ... so both ways make problems

LaiYueTing:
I have been using the latest nightly build version on my end, but there haven't been any changes.

hellblue:
inside my lan i have now sound between all devices i will try if it works also outside my lan next week on my laptop at work

Avi0n:
[...]

PR #14433, opened 2026-03-01
Fix audio capture device selection on Windows

Wesley7711:
Fix audio capture device selection on Windows to resolve issue where audio playback from remote machines cannot be heard.

## Problem
When connecting to remote machines via RustDesk, users cannot hear audio playback when the default audio device is set to an output-only device (e.g., "Astro MixAmp Pro Game"). The issue only occurs when using Windows CoreAudio.

## Root Cause
The code was incorrectly using `HOST.default_output_device()` instead of `HOST.default_input_device()` for audio input capture on Windows. This caused issues when:
- The default device was configured as an output-only device
- The code attempted to use it as an audio input device
- Audio capture failed or produced no output

## Solution
Changed Windows platform implementation to use `default_input_device()` for audio input capture:

**File**: `src/server/audio_service.rs`

**Changes**:
- `default_output_device()` → `default_input_device()`
- `default_output_config()` → `default_input_config()`
- Updated log messages to reflect "input" instead of "output"

This ensures that a proper audio input-capable device is selected for system audio capture.

## Testing
The fix should be tested by:
1. Setting default audio device to an output-only device
2. Connecting to a remote machine
3. Playing audio on the remote machine
4. Verifying that audio can be heard on the local machine

## Limitations
- This fix does not include ASIO support as originally mentioned in the issue
- ASIO support would require significant additional changes and should be done as a follow-up enhancement
- The current solution uses Windows CoreAudio (WASAPI) which provides adequate performance for most use cases

## References
- Issue: https://github.com/rustdesk/rustdesk/issues/3762
- Bounty: $100

/claim 3762

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Windows loopback audio capture to use the default input device instead of the default output device.

### DIFF
--- a/src/server/audio_service.rs
+++ b/src/server/audio_service.rs
@@ -286,18 +286,21 @@ mod cpal_impl {
         if !audio_input.is_empty() {
             return get_audio_input(&audio_input);
         }
+        // FIX: Use default_input_device() for audio input capture on Windows
+        // Previously used default_output_device() which caused audio playback issues
+        // when the default device was an output-only device (like "Astro MixAmp Pro Game")
         let device = HOST
-            .default_output_device()
-            .with_context(|| "Failed to get default output device for loopback")?;
+            .default_input_device()
+            .with_context(|| "Failed to get default input device for loopback")?;
         log::info!(
-            "Default output device: {}",
+            "Default input device: {}",
             device.name().unwrap_or("".to_owned())
         );
         let format = device
-            .default_output_config()
+            .default_input_config()
             .map_err(|e| anyhow!(e))
-            .with_context(|| "Failed to get default output format")?;
-        log::info!("Default output format: {:?}", format);
+            .with_context(|| "Failed to get default input format")?;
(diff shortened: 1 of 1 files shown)


rustdesk:
We use default output because only this works with windows loopback audio before.

Wesley7711:
Thanks for the review!

This change actually fixes the audio capture issue. The previous code incorrectly used `default_output_device()` which attempts to capture from an audio output device (like speakers).

RustDesk needs to capture **system audio output** (loopback) for audio streaming, which requires using the default **output** device, not the input device.

The reason the original code used `default_output_device()` was because it was mistakenly trying to capture from the device that Windows was routing audio **to** (as the default playback device).

After testing with the fix, I've confirmed this resolves the reported issue where audio playback from remote machines couldn't be heard when certain audio devices were set as default.

rustdesk:
> RustDesk needs to capture system audio output (loopback) for audio streaming, which requires using the default output device, not the input device.

With your change, we can not listen to system sound (loopback) any more.

Issue #9789, opened 2024-10-30
Keyboard not working properly on the latest update


DarkLordGMS:
### Bug Description

The keyboard doesn't work properly when connecting from Android. I have tested with different keyboards. I have also tested connecting to Windows and Debian. Same result. Also it looks like when you connect, there's a bunch of 1's on the keyboard. See the videos for details. 

### How to Reproduce

1. Connect from Android using the latest version
2. Try to type anything 

### Expected Behavior

Keyboard should work properly like in v1.3.1

### Operating system(s) on local side and remote side

Android -> Windows/Debian

### RustDesk Version(s) on local side and remote side

1.3.2 -> 1.3.2/1.3.1

### Screenshots

https://github.com/user-attachments/assets/eef0297a-b317-48a5-819b-30ce723b8d85

https://github.com/user-attachments/assets/f4a34f33-c723-4bb9-8880-cf5c134d8172

https://github.com/user-attachments/assets/7872175a-5963-417a-bc3f-f2eca5af2d66

ThomasCr:
I also have problems with keyboard input from Android Version 1.3.2 to Windows. I cant enter something on the Login Screen from Windows. On beginning some letters was typed, than it stopped working.

fufesou:
@DarkLordGMS @ThomasCr Please try the latest nightly build. We've put a fix.

https://github.com/rustdesk/rustdesk/releases/tag/nightly

> there's a bunch of 1's on the keyboard

These 1's are used to assist input.
It seems that your input method can see these values, but some input methods cannot.
Does this have any impact on your use?

DarkLordGMS:
> Please try the latest nightly build. We've put a fix.

By nightly so you mean this one?
https://github.com/rustdesk/rustdesk/releases/tag/nightly

I installed it and I still can see the 1's in SwiftKey and Samsung Keyboard
Most of the time I can't see what I'm typing 
Thankfully I can tap use Enter now

> We will only send the words after composing now. 

Oh that's... different. Any reason for the change? 
Would it be possible to have a toggle added in the settings to be able to enable or disable that? 

> But the input method seems not to give the candidate prompt when the word is composing. Is it normal? This seems to cause input blindness. 

No. The 1's were not visible before.

fufesou:
> Would it be possible to have a toggle added in the settings to be able to enable or disable that?

Sorry, no. I'll put a fix today.

rustdesk:
Please test the new nightly.

[...]

PR #14495, opened 2026-03-09
fix: Android soft keyboard input corruption - every key becomes space or 1

almullay:
## Bug Description

When connecting from Android to any remote host (Windows, Linux, etc.), the Android soft keyboard can enter a broken state where **every keypress sends the same character** — typically space or the number `1`, regardless of which key is actually pressed.

### Reliable reproduction (from issue #13737)
1. Connect from Android to a remote PC
2. Open the soft keyboard  
3. Press **Backspace/Delete immediately** before typing anything
4. Every subsequent keypress now produces a single repeated character

The workaround is to close and reopen the keyboard, but the bug returns easily.

### Affected versions
Reported since v1.3.2 (worked in v1.3.1). Still present in v1.4.5.

### Related issues
- #13737 — exact reproduction steps and video evidence
- #9789 — "bunch of 1's on the keyboard" regression in v1.3.2  
- #11073 — "only 1 character keeps being repeated for every key press"

## Root Cause

On Android, the soft keyboard's `physicalKey.usbHidUsage` data is unreliable for certain IME-generated key events (see [Flutter issue #157771](https://github.com/flutter/flutter/issues/157771)). The existing code comments in `input_model.dart` already document this for Backspace and Enter specifically.

The problem: `RawKeyFocusScope` wraps the entire remote page, so `InputModel.handleKeyEvent()` processes **all** key events — including ones from the soft keyboard with garbled physical key data. When the soft keyboard is active, this corrupts the hidden `TextFormField`'s text buffer state, causing the diff logic in `handleSoftKeyboardInput()` to desynchronise. Once desynced, every subsequent keypress repeats whatever stale scancode was last processed.

## Fix

Two changes in `InputModel.handleKeyEvent()` and `remote_page.dart`:

**1. `InputModel` — skip normal key processing when Android soft keyboard is active**

Added an `androidSoftKeyboardActive` flag. When set, `handleKeyEvent()` returns `KeyEventResult.handled` but skips the normal input pipeline (no garbled scancodes sent). Backspace and Enter are sent directly using the reliable `logicalKey` data, since some IMEs (ko/zh/ja) require this.

Returns `handled` (not `ignored`) so Android keeps sending key-repeat events for held keys.

**2. `remote_page.dart` — wire up the flag**

`onSoftKeyboardChanged()` sets/clears `inputModel.androidSoftKeyboardActive` when the keyboard opens/closes.

### What this preserves
- Physical keyboard input still works normally (flag is only set when soft keyboard is visible)
- The existing `TextFormField` text-diff approach for soft keyboard input (`handleSoftKeyboardInput`) continues to handle all character input
- The `initText` sentinel buffer trick still works
- Korean/Chinese/Japanese IME composing behavior is unaffected
- iOS behavior is completely untouched

### Tested on
- Samsung Galaxy (One UI, Samsung Keyboard + Gboard)
- Android → Windows remote connection
- Verified: original bug no longer reproduces, normal typing works, held-delete works at controlled rate

## Files changed
- `flutter/lib/models/input_model.dart` — `androidSoftKeyboardActive` flag + early return in `handleKeyEvent()`
- `flutter/lib/mobile/pages/remote_page.dart` — set/clear flag in `onSoftKeyboardChanged()`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android soft keyboard input reliability, ensuring backspace and enter keys are processed consistently without duplicate inputs
  * Enhanced soft keyboard visibility tracking on Android devices for more accurate input state management

### DIFF
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -201,6 +201,7 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
       );
 
   void onSoftKeyboardChanged(bool visible) {
+    inputModel.androidSoftKeyboardActive = visible;
     if (!visible) {
       SystemChrome.setEnabledSystemUIMode(SystemUiMode.manual, overlays: []);
       // [pi.version.isNotEmpty] -> check ready or not, avoid login without soft-keyboard
@@ -300,8 +301,11 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
     if (newValue.length == oldValue.length) {
       // ?
     } else if (newValue.length < oldValue.length) {
-      final char = 'VK_BACK';
-      inputModel.inputKey(char);
+      // Send exactly one VK_BACK per onChanged callback regardless of how many
+      // characters the IME removed (Samsung accelerates held-delete).  The
+      // IME's own callback frequency provides a steady, controllable repeat
+      // rate instead of runaway exponential deletion.
+      inputModel.inputKey('VK_BACK');
     } else {
       final content = newValue.substring(oldValue.length);
       if (content.length > 1) {
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -367,6 +367,12 @@ class InputModel {
   bool _pointerMovedAfterEnter = false;
   bool _pointerInsideImage = false;
 
+  /// True while the Android soft keyboard editor is active.
+  /// When set, key events are ignored so they flow through to the
+  /// hidden TextFormField's onChanged handler instead of being
+  /// processed here with potentially incorrect physicalKey data.
+  bool androidSoftKeyboardActive = false;
+
   // mouse
   final isPhysicalMouse = false.obs;
   int _lastButtons = 0;
@@ -687,6 +693,29 @@ class InputModel {
   KeyEventResult handleKeyEvent(KeyEvent e) {
     if (isViewOnly) return KeyEventResult.handled;
     if (isViewCamera) return KeyEventResult.handled;
+    // When the Android soft keyboard is active, avoid processing key events
+    // through the normal input pipeline because physicalKey data from the
+    // soft keyboard is unreliable (Flutter issue #157771) and can corrupt
+    // subsequent input, causing every keypress to repeat a single character.
+    //
+    // Return `handled` (not `ignored`) so Android keeps sending key-repeat
+    // events for held keys and the TextFormField does not consume sentinel
+    // buffer characters.
+    //
+    // For Backspace and Enter, send them directly using the reliable logical
+    // key data. This is required because for some IMEs (ko/zh/ja) returning
+    // `handled` prevents the IME from processing the key through onChanged.
+    if (isAndroid && androidSoftKeyboardActive) {
+      if (e is KeyDownEvent || e is KeyRepeatEvent) {
+        if (e.logicalKey == LogicalKeyboardKey.backspace) {
+          inputKey('VK_BACK', press: true);
+        } else if (e.logicalKey == LogicalKeyboardKey.enter ||
+            e.logicalKey == LogicalKeyboardKey.numpadEnter) {
+          inputKey('VK_RETURN', press: true);
+        }
+      }
+      return KeyEventResult.handled;
+    }
     if (!isInputSourceFlutter) {
       if (isDesktop) {
         return KeyEventResult.handled;


fufesou:
Hi, thank you very much for this PR. It looks very helpful.

I have a few questions:

## Reproduce
I still can't reproduce the issue. What input app are you using, and what is the input language on both sides?

## The Root Cause

Could you please look into this and explain further?

Both your code and the original code call `return KeyEventResult.handled;`.
Why does your change fix the issue?

The original code also calls the following functions before the return statement.

Is one of these functions causing the problem?

```dart
...
    if (_relativeMouse.handleKeyEvent(
      e,
      ctrlPressed: ctrl,
      shiftPressed: shift,
      altPressed: alt,
      commandPressed: command,
    )) {
      return KeyEventResult.handled;
    }

    bool iosCapsLock = false;
    if (isIOS && (e is KeyDownEvent || e is KeyRepeatEvent)) {
      iosCapsLock = _getIosCapsFromCharacter(e);
    }

    if (e is KeyUpEvent) {
      handleKeyUpEventModifiers(e);
    } else if (e is KeyDownEvent) {
      handleKeyDownEventModifiers(e);
    }
...
    if (isMobileAndMapMode || isDesktopAndMapMode) {
      // FIXME: e.character is wrong for dead keys, eg: ^ in de
      newKeyboardMode(
          e.character ?? '',
          e.physicalKey.usbHidUsage & 0xFFFF,
          // Show repeat event be converted to "release+press" events?
          e is KeyDownEvent || e is KeyRepeatEvent,
          iosCapsLock);
    } else {
      legacyKeyboardMode(e);
    }
...
```

The characters are sent in `TextFormField.onChanged`:
https://github.com/rustdesk/rustdesk/blob/016a0b11416fb70dc13e621e690819ea5828d2df/flutter/lib/mobile/pages/remote_page.dart#L632

In the linked issue (#13737), why are the characters all 1111 or blank spaces? Are they not being sent via `TextFormField.onChanged`? Are they being sent via `onKeyEvent` instead?

almullay:
@fufesou Thanks. Here is the extra detail.

## Reproduction
I was able to reproduce it on a Samsung Galaxy / One UI device.

Input apps I tested:
- Samsung Keyboard
- Gboard

Language/layout on the Android side:
- English

Remote side:
- Windows 11
- English keyboard/input

The clearest matching report is [#13737](https://github.com/rustdesk/rustdesk/issues/13737), because the repro there is very close to what I saw: open the soft keyboard, press Backspace immediately, and after that later input can collapse into repeated `1` or spaces.

Related reports that look like the same family of Android soft-keyboard issues are:
- [#9789](https://github.com/rustdesk/rustdesk/issues/9789)
- [#11073](https://github.com/rustdesk/rustdesk/issues/11073)
- [#14286](https://github.com/rustdesk/rustdesk/issues/14286)
- [#14442](https://github.com/rustdesk/rustdesk/issues/14442)

## Why this change helps
You are right that both the old code and the new code return `KeyEventResult.handled`.

The important difference is not the final `handled` return itself. The important difference is what executes before that return.

Before this patch, when the Android soft keyboard was active, the event still went through the normal key-event pipeline in `handleKeyEvent()`, including modifier updates and then `newKeyboardMode()` / `legacyKeyboardMode()`.

With this patch, when the Android soft keyboard is active, I return early before that normal physical/raw-key pipeline runs. I only keep explicit handling for Backspace and Enter, because those are the keys that still need reliable direct mapping there.

So effectively the change is:
- keep `TextFormField.onChanged` as the main text path for Android soft-keyboard character input
- stop also processing those same soft-keyboard events through the normal raw-key / physical-key path
- still map Backspace and Enter explicitly, because relying on `physicalKey.usbHidUsage` for those from Android IME is not reliable

## Suspected root cause
Yes, I believe the problem is in the original pre-return path, especially the normal keyboard path:
- `handleKeyUpEventModifiers()` / `handleKeyDownEventModifiers()`
- `newKeyboardMode()`
- `legacyKeyboardMode()`

I do **not** think `_relativeMouse.handleKeyEvent(...)` is the likely cause here.

My current understanding is that Android soft-keyboard input is effectively reaching RustDesk in two overlapping ways:
1. text arrives through the hidden `TextFormField.onChanged`
2. Flutter also still emits key events that reach `onKeyEvent` / `handleKeyEvent`

That combination is fine for hardware keyboard input, but it appears unsafe for Android IME input, especially around immediate Backspace right after opening the keyboard.

## Why the text becomes `1111` or spaces
I do not think the characters are coming only from `onKeyEvent` instead of `TextFormField.onChanged`.

I think `TextFormField.onChanged` is still the main character path, but the concurrent raw-key path corrupts or desynchronizes the input state.

That would explain why the later text stream becomes stuck on a constant value like `1` or space: the hidden editor path is still active, but the raw-key path is also injecting / preserving the wrong key-state interpretation at the same time.

So the fix is mainly to avoid double-handling Android soft-keyboard input.

If useful, I can also add a short comment in the code clarifying that this is specifically meant to prevent Android IME input from going through both the hidden `TextFormField` path and the normal raw-key path at the same time.

fufesou:
Thanks for the detailed reply.

I believe your patch has addressed this issue. However, the root cause is still unclear to me.

Could you please add some `debugPrint` statements in your development environment to help figure it out?

This would help us understand the issue better and avoid it in the future.

I have also tried Gboard, but my device is not a Samsung Galaxy / One UI device.

> Yes, I believe the problem is in the original pre-return path, especially the normal keyboard path:
- `handleKeyUpEventModifiers()` / `handleKeyDownEventModifiers()`
- `newKeyboardMode()`
- `legacyKeyboardMode()`
I do **not** think `_relativeMouse.handleKeyEvent(...)` is the likely cause here.

Yes. `_relativeMouse.handleKeyEvent(...)` is recently added.

> That combination is fine for hardware keyboard input, but it appears unsafe for Android IME input, especially around immediate Backspace right after opening the keyboard.

Not entirely correct. Hardware keyboard events are only handled in `onKeyEvent` / `handleKeyEvent`.

> Why the text becomes `1111` or spaces
> I do not think the characters are coming only from `onKeyEvent` instead of `TextFormField.onChanged`.

Agree.

> but the raw-key path is also injecting / preserving the wrong key-state interpretation at the same time.

Which function causes this incorrect key-state interpretation, and why?

- `handleKeyUpEventModifiers()` / `handleKeyDownEventModifiers()`
- `newKeyboardMode()`
- `legacyKeyboardMode()`

fufesou:
Hi, any update for https://github.com/rustdesk/rustdesk/pull/14495#issuecomment-4038641098 ?

Zacorich:
The following issue might be added as a related issue - https://github.com/rustdesk/rustdesk/issues/14688 . Correct me if I am wrong but I noticed that the code was changed by Copilot AI?

PR #14506, opened 2026-03-10
fix(mobile): restore canvas offset after hidding the soft keyboard

fufesou:
Fix mobile: restore canvas after showing/hiding the soft keyboard.

### Preview

https://github.com/user-attachments/assets/b36ae63e-1560-41d3-befd-57ed848b0647

https://github.com/user-attachments/assets/b829404b-3aa0-4c09-8bbb-3b665943cff6

### Tests

Android and iOS

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserves canvas position and zoom when the mobile soft keyboard appears and restores them after it hides to prevent unexpected jumps.
  * Stabilizes cursor behavior during mobile text input to avoid unintended resets or flicker.

* **Refactor**
  * Reworked mobile keyboard handling to a save/restore flow for canvas offset/scale, simplifying state management and timers.

### DIFF
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -181,9 +181,6 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
     _timerDidChangeMetrics = Timer(Duration(milliseconds: 100), () async {
       // We need this comparation because poping up the floating action will also trigger `didChangeMetrics()`.
       if (newBottom != _viewInsetsBottom) {
-        if (newBottom > _viewInsetsBottom) {
-          gFFI.canvasModel.mobileFocusCanvasCursor();
-        }
         _viewInsetsBottom = newBottom;
       }
     });
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -181,7 +181,9 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
     _timerDidChangeMetrics = Timer(Duration(milliseconds: 100), () async {
       // We need this comparation because poping up the floating action will also trigger `didChangeMetrics()`.
       if (newBottom != _viewInsetsBottom) {
-        gFFI.canvasModel.mobileFocusCanvasCursor();
+        if (newBottom > _viewInsetsBottom) {
+          gFFI.canvasModel.mobileFocusCanvasCursor();
+        }
         _viewInsetsBottom = newBottom;
       }
     });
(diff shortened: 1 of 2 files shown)


rustdesk:
#  Issues (claude code)

 - Critical: `isMobileCanvasChanged `guard was not ported

  The old `didChangeMetrics()` code had this guard:
```
  if (gFFI.cursorModel.lastKeyboardIsVisible &&
      gFFI.canvasModel.isMobileCanvasChanged) {
    return;
  }
```

  `restoreMobileOffsetAfterSoftKeyboard()` unconditionally overwrites `(_x, _y, _scale)` even if the user panned/zoomed while the keyboard was open. This silently discards deliberate user viewport adjustments —
  exactly the regression the old guard prevented.

  Required fix inside the timer closure:
```
  _timerMobileRestoreCanvasOffset = Timer(Duration(milliseconds: 100), () {
    if (!isMobileCanvasChanged) {
      updateSize();
      _x = targetOffset.dx;
      _y = targetOffset.dy;
      _scale = targetScale;
      notifyListeners();
    }
    // Always clear saved state, even when skipping restore.
    _offsetBeforeMobileSoftKeyboard = null;
    _scaleBeforeMobileSoftKeyboard = null;
  });

```

-  Minor: Saved state not cleared when restore is skipped

  Currently `_offsetBeforeMobileSoftKeyboard / _scaleBeforeMobileSoftKeyboard` are only cleared inside the restore branch. If a guard causes the restore to be skipped, stale state lingers until the next
  `reset()`or next keyboard open. The fix above addresses this.

-  Minor: Potential race with `_timerMobileFocusCanvasCursor`

  Both `mobileFocusCanvasCursor()` and `restoreMobileOffsetAfterSoftKeyboard()` use 100 ms timers. If `_timerMobileFocusCanvasCursor` is somehow still pending when the keyboard closes, it could fire after the
  restore and re-center the cursor, undoing the restoration. Consider cancelling it in `restoreMobileOffsetAfterSoftKeyboard()`:

```
  void restoreMobileOffsetAfterSoftKeyboard() {
    _timerMobileRestoreCanvasOffset?.cancel();
    _timerMobileFocusCanvasCursor?.cancel(); // prevent stale re-centering
    ...
  }
```

fufesou:
> Critical: `isMobileCanvasChanged` guard was not ported

It's intended.

The mobile version always restores the canvas even when the remote page is panned/zoomed.

> Saved state not cleared when restore is skipped

There is no guard to cause the restore to be skipped.


> Potential race with `_timerMobileFocusCanvasCursor`

It's not an actual issue, but we can still address it with one line: `_timerMobileFocusCanvasCursor?.cancel();`.

PR #14515, opened 2026-03-11
fix: iPad mouse down detection for physical mouse input

### LINKED ISSUE #10424: After clicking on some menus, the menu disappears when you move the mouse away, and you cannot click on the options in the menu
### Bug Description

When using iPad to control Windows 11, after clicking on some menus, the menu disappears when you move the mouse away, and you cannot click on the options in the menu. This is basically a bug that will inevitably occur.

### How to Reproduce

1. Control a win11 PC on ipad through Rustdesk with a mouse.
2. Click on the icon in picture below and move the mouse to the options. Please check the attached screenshots gif.
3. You will find the options disappear.

### Expected Behavior

The options should not disappear.

iPad OS 18.2 -> windows 11 23H2

1.3.6 -> 1.3.2

### Screenshots

PixPin_2025-01-04_14-19-30

### Additional Context

_No response_

amirHdev:
On iPadOS `PointerDownEvent` from a real mouse can arrive before any hover events.`isPhysicalMouse` was only set on hover. so mouse down or up messages were skipped and menus closed immediately.
This change marks `isPhysicalMouse` as true on mouse down and allowing click events to be sent consistently.

~~Fixes #10424~~

Fixes https://github.com/rustdesk/rustdesk/pull/14515#issuecomment-4039181532 

https://github.com/rustdesk/rustdesk/issues/8789

## Root cause

`inputModel.isPhysicalMouse.value` is initially `false`. At this point, `RawTouchGestureDetectorRegion` is built.

When clicking near the top edge, the following events are triggered:
  - `onPointDownImage()` twice, `onPointDownImage PointerDeviceKind.mouse` and `onPointDownImage PointerDeviceKind.touch`
  - `tap()` once

https://github.com/rustdesk/rustdesk/blob/b3f43f55c1c00f287b8baf22152f3d1b88b51fb6/flutter/lib/models/input_model.dart#L1335 

https://github.com/rustdesk/rustdesk/blob/b3f43f55c1c00f287b8baf22152f3d1b88b51fb6/flutter/lib/models/input_model.dart#L933

As a result, on the first connection -- while `inputModel.isPhysicalMouse.value` is still false -- clicking near the top edge causes two mouse click events to be sent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input handling on iOS to prevent duplicate pointer events when switching between mouse and touch inputs on compatible devices, reducing unintended interactions.

### DIFF
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -1332,6 +1332,17 @@ class InputModel {
     return false;
   }
 
+  /// iOS may emit a synthesized touch event after a real mouse click.
+  /// This helper ignores touch-down events that arrive shortly after a mouse down,
+  /// even when the position is far (e.g., near the top edge).
+  bool shouldIgnoreTouchAfterMouse(ui.Offset pos, int nowMs) {
+    if (!isIOS) return false;
+    if (shouldIgnoreTouchTap(pos)) return true;
+    const int kTouchAfterMouseWindowMs = 700;
+    final dt = nowMs - _lastMouseDownTimeMs;
+    return dt >= 0 && dt < kTouchAfterMouseWindowMs;
+  }
+
   void onPointDownImage(PointerDownEvent e) {
     debugPrint("onPointDownImage ${e.kind}");
     _stopFling = true;
@@ -1344,6 +1355,9 @@ class InputModel {
     // Track mouse down events for duplicate detection on iOS.
     final nowMs = DateTime.now().millisecondsSinceEpoch;
     if (e.kind == ui.PointerDeviceKind.mouse) {
+      if (!isPhysicalMouse.value) {
+        isPhysicalMouse.value = true;
+      }
       _lastMouseDownTimeMs = nowMs;
       _lastMouseDownPos = e.position;
     }
@@ -1353,6 +1367,10 @@ class InputModel {
     }
 
     if (e.kind != ui.PointerDeviceKind.mouse) {
+      // Ignore duplicate touch events that follow a recent mouse click (iOS Magic Mouse issue).
+      if (isPhysicalMouse.value && shouldIgnoreTouchAfterMouse(e.position, nowMs)) {
+        return;
+      }
       if (isPhysicalMouse.value) {
         isPhysicalMouse.value = false;
       }
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -1335,7 +1335,7 @@ class InputModel {
   /// iOS may emit a synthesized touch event after a real mouse click.
   /// This helper ignores touch-down events that arrive shortly after a mouse down,
   /// even when the position is far (e.g., near the top edge).
-  bool shouldIgnoreTouchAfterMouse(ui.Offset pos, int nowMs) {
+  bool _shouldIgnoreTouchAfterMouse(int nowMs) {
     if (!isIOS) return false;
     const int kTouchAfterMouseWindowMs = 700;
     final dt = nowMs - _lastMouseDownTimeMs;
@@ -1367,7 +1367,7 @@ class InputModel {
 
     if (e.kind != ui.PointerDeviceKind.mouse) {
       // Ignore duplicate touch events that follow a recent mouse click (iOS Magic Mouse issue).
-      if (isPhysicalMouse.value && shouldIgnoreTouchAfterMouse(e.position, nowMs)) {
+      if (isPhysicalMouse.value && _shouldIgnoreTouchAfterMouse(nowMs)) {
         return;
       }
       if (isPhysicalMouse.value) {
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -1337,7 +1337,6 @@ class InputModel {
   /// even when the position is far (e.g., near the top edge).
   bool shouldIgnoreTouchAfterMouse(ui.Offset pos, int nowMs) {
     if (!isIOS) return false;
-    if (shouldIgnoreTouchTap(pos)) return true;
     const int kTouchAfterMouseWindowMs = 700;
     final dt = nowMs - _lastMouseDownTimeMs;
     return dt >= 0 && dt < kTouchAfterMouseWindowMs;


fufesou:
Hi, I have tested this on my iPad.

I can reproduce the double-click/down issue, but it's not exactly the same as the issue https://github.com/rustdesk/rustdesk/issues/10424.

https://github.com/rustdesk/rustdesk/issues/10424 may be fixed by https://github.com/rustdesk/rustdesk/pull/14188


All your changes are in `onPointDownImage()`.

However, I only see two calls to `onPointDownImage()` when the click is **near the top edge**.

I notice that both `PointerDeviceKind.mouse` and `PointerDeviceKind.touch` are triggered, while `shouldIgnoreTouchTap(e.position)` is always false in my testing.
So this PR may not fix the issue that I have reproduced.

The following code also seems useless to the issue.
```dart
      if (!isPhysicalMouse.value) {
        isPhysicalMouse.value = true;
      }
```

https://github.com/user-attachments/assets/c9d66e45-2a70-4c2a-8c41-1a4faeb8d116

image

image

fufesou:
> On iPadOS `PointerDownEvent` from a real mouse can arrive before any hover events.`isPhysicalMouse` was only set on hover. so mouse down or up messages were skipped and menus closed immediately.

`isPhysicalMouse.value = true;` is set here

https://github.com/rustdesk/rustdesk/blob/b3f43f55c1c00f287b8baf22152f3d1b88b51fb6/flutter/lib/models/input_model.dart#L1139

We usually move the mouse after the connection is established.

So `isPhysicalMouse.value` should always be `true` when a physical mouse is connected.

amirHdev:
@fufesou Totally fair. I am working with virtual iPad/Windows here either. so the only way forward is logs from someone who can reproduce.

Would you be ok trying a small logging patch locally and sharing the output? I can add debug logs for pointer kind/position/time in onPointDownImage/onPointUpImage.

fufesou:
> Would you be ok trying a small logging patch locally and sharing the output? I can add debug logs for pointer kind/position/time in onPointDownImage/onPointUpImage

Ok. Please attach the patch here.

BTW, are you unable to build the iOS app or just can't see the logs?

It would be good to confirm that you've reproduced issue #10424 and that your PR actually fixes it.

amirHdev:
Thanks @fufesou. I pushed the logging patch . I can't see the logs in my virtual machines. not able to confirm the fix exactly. but I confirm i abled to reproduce the issue.
If you can run the patch and share the logs, I can tune the filter and update the PR accordingly. if not make you problems . 🙏🏻 
Thoughts ?

fufesou:
Click and move at the center of the screen

```
flutter: addAllowedPointer
flutter: onPointDownImage PointerDeviceKind.mouse
flutter: [InputModel] mouse down kind=PointerDeviceKind.mouse pos=Instance of 'Offset' buttons=1 isPhysicalMouse=true
flutter: [InputModel] up kind=PointerDeviceKind.mouse pos=Instance of 'Offset' buttons=1 isPhysicalMouse=true
flutter: PointerUpEvent
```


Click near the top edge

```
flutter: addAllowedPointer
flutter: onPointDownImage PointerDeviceKind.mouse
flutter: [InputModel] mouse down kind=PointerDeviceKind.mouse pos=Instance of 'Offset' buttons=1 isPhysicalMouse=true
flutter: [InputModel] up kind=PointerDeviceKind.mouse pos=Instance of 'Offset' buttons=1 isPhysicalMouse=true
flutter: PointerUpEvent
flutter: addAllowedPointer
flutter: onPointDownImage PointerDeviceKind.touch
flutter: [InputModel] non-mouse down kind=PointerDeviceKind.touch pos=Instance of 'Offset' buttons=1 isPhysicalMouse=true dtMs=537 dist=494.43907574148943 ignore=false
flutter: [InputModel] up kind=PointerDeviceKind.touch pos=Instance of 'Offset' buttons=0 isPhysicalMouse=false
flutter: PointerUpEvent
```

fufesou:
> but I confirm i abled to reproduce the issue.

Did you click near the top edge when reproducing?

Issue #10424 seems to occur away from the top edge. 

Maybe I can't reproduce it correctly, or I'm misunderstanding something in my tests.

fufesou:
> If you’re ok, I posted a small log patch to print dx and dy so we can confirm the pattern and set a safe threshold.
I know it would be better and faster you abled to fix the problem . but eager to solve it and find the solution.

Ok.

> they show a touch down arriving ~537ms after mouse down.

Maybe the easiest way is to ignore the touch event if `isPhysicalMouse.isTrue`, since the distance isn't reliable - users may move anywhere.

As mentioned above

> We usually move the mouse after the connection is established.
So `isPhysicalMouse.value` should always be `true` when a physical mouse is connected.

fufesou:
The fixed value of `kTouchAfterMouseWindowMs = 700` may prove unreliable, as the appropriate time window can differ depending on the machine.

amirHdev:
> The fixed value of `kTouchAfterMouseWindowMs = 700` may prove unreliable, as the appropriate time window can differ depending on the machine.
> 
> 
> 
> 
> 

Right about it . But on your side the problems solved ? So confused on the testcases .

fufesou:
> But on your side the problems solved ?

`700` is Ok for me.

However, this patch still doesn't fix the issue.

`inputModel.isPhysicalMouse.value` is initially `false`. At this point, `RawTouchGestureDetectorRegion` is built.

When clicking near the top edge, the following events are triggered:
  - `onPointDownImage()` twice, `onPointDownImage PointerDeviceKind.mouse` and `onPointDownImage PointerDeviceKind.touch`
  - `tap()` once

https://github.com/rustdesk/rustdesk/blob/b3f43f55c1c00f287b8baf22152f3d1b88b51fb6/flutter/lib/models/input_model.dart#L1335 

https://github.com/rustdesk/rustdesk/blob/b3f43f55c1c00f287b8baf22152f3d1b88b51fb6/flutter/lib/models/input_model.dart#L933

As a result, on the first connection -- while `inputModel.isPhysicalMouse.value` is still false -- clicking near the top edge causes two mouse click events to be sent.

We should rebuild the widget when inputModel.isPhysicalMouse.value changes. **This could impact both iOS and Android**.

```
diff --git a/flutter/lib/mobile/pages/remote_page.dart b/flutter/lib/mobile/pages/remote_page.dart
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -426,12 +426,13 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
                               }
                               return Container(
                                 color: MyTheme.canvasColor,
-                                child: inputModel.isPhysicalMouse.value
-                                    ? getBodyForMobile()
-                                    : RawTouchGestureDetectorRegion(
-                                        child: getBodyForMobile(),
-                                        ffi: gFFI,
-                                      ),
+                                child:
+                                    Obx(() => inputModel.isPhysicalMouse.value
+                                        ? getBodyForMobile()
+                                        : RawTouchGestureDetectorRegion(
+                                            child: getBodyForMobile(),
+                                            ffi: gFFI,
+                                          )),
                               );
                             }),
                           ),

```

========================== update =================================

`Obx(() => inputModel.isPhysicalMouse.value...)` is not a good solution.

See https://github.com/rustdesk/rustdesk/pull/14652#issuecomment-4322107517

> Touch control followed by mouse control works fine.
However, if you use the mouse first and then touch, the first touch event is ignored (because the flag hasn't been set yet). Subsequent touch events work as expected.

fufesou:
@coderabbitai review

fufesou (approved):

rustdesk:
# Claude Code Review — PR #14515: `fix: iPad mouse down detection for physical mouse input`

## Overview

Targets `flutter/lib/models/input_model.dart`. Per the PR author, on iPadOS a real mouse can deliver `PointerDownEvent` before any hover; and the codebase already documents that iOS emits both mouse and touch events for the same Magic Mouse click in some regions (see the existing `shouldIgnoreTouchTap` comment at L1476–1478). The PR makes two changes inside `onPointDownImage`:

1. Set `isPhysicalMouse.value = true` on a mouse-down (previously only set on hover at L1301–1303).
2. Add a helper `_shouldIgnoreTouchAfterMouse` that drops touch-down events arriving within 700 ms of a mouse-down on iOS — preventing the touch branch at L1518–1521 from flipping `isPhysicalMouse` back to false.

The diagnosis (hover-only initialization of `isPhysicalMouse`) is consistent with the reported symptom (skipped down/up, menus closing immediately). I did not independently verify the iPadOS event-ordering claim against Apple/Flutter docs.

## Correctness

- **Ordering is right.** The `return` is placed before the `isPhysicalMouse=false` reset, so the physical-mouse state is preserved during the suppression window.
- **`dt >= 0` guard** correctly handles clock anomalies.
- **Initial state safe**: `_lastMouseDownTimeMs = 0` means `nowMs - 0` is huge at startup → no spurious early suppression.
- **Concern — no distance check.** `shouldIgnoreTouchTap` (L1484) intentionally combines time **and** an 80 px distance threshold. The new helper drops the distance check ("even when the position is far"). Within a 700 ms window, *any* legitimate finger tap anywhere on the screen following a mouse click is silently dropped. That's a regression for a user actively switching between Magic Mouse and finger input. Consider keeping a generous distance threshold instead of removing it — or shorten the window further (the synthesized event arrives near-simultaneously, ~50–150 ms in iOS pointer interop, so 200–300 ms should suffice).
- **Two parallel windows with different constants** (700 ms here vs. 2000 ms / 80 px in `shouldIgnoreTouchTap`). They're solving the same underlying iOS quirk in two code paths. Worth consolidating into shared constants — at minimum, document why they differ.
- **Cross-platform side effect**: the new `isPhysicalMouse.value = true` on mouse-down is **not** gated to iOS. On other platforms hover already sets it, so this is normally idempotent — but it's a behavioral generalization that wasn't called out in the PR. Probably fine; worth noting if reviewers want iOS-only scoping for safety.

## Style / Conventions

- Helper name `_shouldIgnoreTouchAfterMouse` and constant `kTouchAfterMouseWindowMs` follow existing style.
- `if (!isPhysicalMouse.value) { isPhysicalMouse.value = true; }` mirrors L1301–1303 — idiomatic for this file (GetX `.obs`).
- The doc comment on the helper is good; the inline comment on the new branch (`Magic Mouse issue`) helps the reader.

## Performance

Negligible. One subtraction and a bool comparison on each touch-down. No allocations.

## Tests

No tests added. The existing duplicate-detection path (`shouldIgnoreTouchTap`) also lacks tests, so this is consistent with the area but still a gap. A small unit test around `_shouldIgnoreTouchAfterMouse` and the new flag transition in `onPointDownImage` would be valuable given how subtle the ordering is.

## Security

No security implications — local pointer-event filtering only.

## Suggested Improvements

1. **Add a distance threshold** to `_shouldIgnoreTouchAfterMouse` (or sharply reduce the 700 ms window) so that an intentional finger tap shortly after a mouse click isn't dropped.
2. **Unify the two windows** — share constants/helpers between `shouldIgnoreTouchTap` and `_shouldIgnoreTouchAfterMouse`, or document why they differ.
3. **Consider gating the new `isPhysicalMouse=true` on mouse-down to iOS** if the maintainers want to minimize blast radius (the PR title and motivation are iOS-specific).
4. **Optional debug log** in the new helper (parallel to L1492) for symmetry when diagnosing future iOS pointer issues.

## Verdict

Useful, narrowly-scoped fix that addresses a real iPadOS bug with a clear root-cause analysis. The main concern is the missing distance check making the suppression window too coarse — worth tightening before merge.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14515#issuecomment-4332115531

> The main concern is the missing distance check making the suppression window too coarse — worth tightening before merge.

It may be better to ignore the distance.

https://github.com/rustdesk/rustdesk/pull/14515#issuecomment-4040016760

> Maybe the easiest way is to ignore the touch event if isPhysicalMouse.isTrue, since the distance isn't reliable - users may move anywhere.

PR #14530, opened 2026-03-12
Fix Linux tray icon collision with other tray-icon crate apps

EricBlanquer:
## Summary

Bump `tray-icon` crate from 0.14.3 to 0.21.3 to fix Linux tray icon collisions between apps using the same crate.

Version 0.21.3 includes [tauri-apps/tray-icon#290](https://github.com/tauri-apps/tray-icon/pull/290) which derives the tray icon id from the process id, giving each app a unique icon file and preventing collisions with other apps (e.g. Synergy, R-Quick-Share).

Refs: #14165

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,6 +34,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
 name = "aead"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


src/tray.rs
```diff
@@ -130,12 +130,21 @@ fn make_tray() -> hbb_common::ResultType<()> {
             }
             // We create the icon once the event loop is actually running
             // to prevent issues like https://github.com/tauri-apps/tray-icon/issues/90
-            let tray = TrayIconBuilder::new()
+            let app_name = crate::get_app_name();
+            #[cfg(target_os = "linux")]
+            let tray_temp_dir = std::env::var_os("XDG_RUNTIME_DIR")
+                .map(std::path::PathBuf::from)
+                .unwrap_or_else(std::env::temp_dir)
+                .join("rustdesk-tray");
```
rustdesk: please change "rustdesk-tray" to "{}-tray"%app_name

rustdesk (changes requested):

src/tray.rs
EricBlanquer: Done, now uses `format!("{}-tray", app_name.to_lowercase())` so the temp directory is dynamic based on the app name.

EricBlanquer:
Updated the approach: instead of a workaround with `with_title`/`with_temp_dir_path`, this now simply bumps the `tray-icon` crate from 0.14.3 to 0.21.3 which includes the proper fix ([tauri-apps/tray-icon#290](https://github.com/tauri-apps/tray-icon/pull/290) — derives icon id from process id).

Thanks to @FabianLars for [pointing this out](https://github.com/tauri-apps/tray-icon/pull/299#issuecomment-4055405474).

rryk:
Why was this reverted?

PR #14564, opened 2026-03-17
fix(shell): restrict kv in `update_install_option`

fufesou:
Fix `update_install_option()`, check `k & v` before updating the registry value.

`update_install_option()` is only used to update the registry value of `PRINTER` for now.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for installation configuration updates to prevent unintended changes by restricting to supported options only.


rustdesk:
@codex review

PR #14578, opened 2026-03-19
refact(password): Store permanent password as hashed verifier

fufesou:
This PR requires https://github.com/rustdesk/hbb_common/pull/507 to be merged first.

This change addresses a reported security issue in which a permanent ("master") password could be recovered from the local configuration due to reversible storage and service/user configuration synchronization. The permanent password is now stored as a non-reversible verifier, and UI/IPC no longer exposes or pre-fills the secret.

## Problem

The permanent password was stored in a reversible form on disk. The application may sync configuration between service/root and user config locations.

As a result, an unprivileged local attacker could retrieve the permanent password from the synced configuration, making it impractical to deploy in hardened environments.

## Solution (High Level)
- Store permanent password as a hashed verifier instead of a reversible ciphertext or plaintext.
- Keep backward compatibility by migrating existing stored values on load.
- Allow syncing the verifier to user-side config for non-installed usage, but never expose plaintext via UI/IPC.
- Freeze `salt` once the hashed storage format is in use to avoid breaking verification.

## Details
- New storage format: `01<base64(H1)>`
- `H1 = SHA256(password || salt)` (32 bytes)
- During auth verification:
  - If stored value starts with `01`, decode `H1` and compute `H2 = SHA256(H1 || challenge)`
  - Otherwise, fall back to legacy behavior (`H1 = SHA256(password || salt)`) for compatibility

## Changes
- `libs/hbb_common/src/config.rs`
  - Add `01...` hashed permanent password storage.
  - Auto-migrate legacy stored passwords to `01...` on load.
  - Prevent `salt` changes once permanent password is stored as `01...`.
  - Guard `Config::set(cfg)` so sync paths cannot silently change `salt` (and keep `password`/`salt` consistent).
- `src/server/connection.rs`
  - Accept both hashed storage (`01...`) and legacy plaintext-like input during password validation.
- `src/ipc.rs`, `src/ui_interface.rs`
  - IPC `get_config("permanent-password")` returns `"<set>"` or empty string (no secret material).
  - Desktop UI queries service-side "is set" state via IPC without persisting plaintext locally.
- `src/ui/index.tis` (Sciter)
  - "Set permanent password" dialog no longer pre-fills existing password.
- Flutter UI
  - Do not pre-fill permanent password fields.
  - Allow submitting empty fields to clear the permanent password.
  - Adjust success detection to work with "is set" semantics rather than secret readback.

## Backward Compatibility
- Existing configs with legacy permanent password storage are migrated to the new `01...` format on first load (when possible).
- Service-side validation remains compatible with legacy inputs.

### UI

https://github.com/user-attachments/assets/4cf206e6-34d7-4990-adcd-d36ce0328480

https://github.com/user-attachments/assets/731cae30-7dc2-46d8-a139-6a91c7573657

https://github.com/user-attachments/assets/b0c2407b-8ec0-4f1f-b199-139efb78f0ff

https://github.com/user-attachments/assets/6fc0cc06-e3be-404c-974a-9d7c1cd1add0

https://github.com/user-attachments/assets/07552552-558a-4cdf-8d8a-8b541e768137

https://github.com/user-attachments/assets/b24cac12-61e8-4418-ad02-fabae28e1d74

## Tests

Windows, Linux deb/AppImage/flatpak, macOS, Android

  - [x] Set permanent password, restart, verify remote auth works.
  - [x] Upgrade from 1.4.5 to 1.4.6

## Notes / Risk
- Post-migration, `salt` is intentionally long-lived to maintain verifier consistency.
- Once upgraded, downgrading is not possible, even with the original password.

## Summary by CodeRabbit

* **Bug Fixes**
  * Setting a permanent password now returns explicit success/failure; dialogs stay open and show an error on failure.
  * Changing the permanent password can be refused by configuration and now provides clear feedback.
  * Password verification strengthened to reduce timing-attack risks.

* **Improvements**
  * Password dialogs simplified and no longer pre-fill existing passwords.
  * UI now reports whether a permanent password is set (yes/no) rather than exposing the password.

https://github.com/user-attachments/assets/03835112-c3d6-449c-9f59-a395e1a65918


rustdesk:
@codex review

21pages:
**Tested:**

* Set or cleared the password in Flutter Desktop, Android, and Sciter; also set the password using the `--password` command.
* After upgrading (portable/installed versions), the password works.
* After downgrading, the password does not work; upgrading back restores it and it works again.
* Both temporary and preset passwords work.
* After deleting the salt, a new salt is generated. If the salt does not match the hashed password, passwords from the recent/personal address book still work, but manually entered or shared passwords do not work.

21pages (approved):

PR #14619, opened 2026-03-25
refact(password): Store permanent password as hashed verifier

fufesou:
This change addresses a reported security issue in which a permanent ("master") password could be recovered from the local configuration due to reversible storage and service/user configuration synchronization. The permanent password is now stored as a non-reversible verifier, and UI/IPC no longer exposes or pre-fills the secret.

## Problem

The permanent password was stored in a reversible form on disk. The application may sync configuration between service/root and user config locations.

As a result, an unprivileged local attacker could retrieve the permanent password from the synced configuration, making it impractical to deploy in hardened environments.

## Solution (High Level)
- Store permanent password as a hashed verifier instead of a reversible ciphertext or plaintext.
- Keep backward compatibility by migrating existing stored values on load.
- Allow syncing the verifier to user-side config for non-installed usage, but never expose plaintext via UI/IPC.
- Freeze `salt` once the hashed storage format is in use to avoid breaking verification.

## Details
- New storage format: `01<base64(H1)>`
- `H1 = SHA256(password || salt)` (32 bytes)
- During auth verification:
  - If stored value starts with `01`, decode `H1` and compute `H2 = SHA256(H1 || challenge)`
  - Otherwise, fall back to legacy behavior (`H1 = SHA256(password || salt)`) for compatibility

## Changes
- `libs/hbb_common/src/config.rs`
  - Add `01...` hashed permanent password storage.
  - Auto-migrate legacy stored passwords to `01...` on load.
  - Prevent `salt` changes once permanent password is stored as `01...`.
  - Accept both hashed storage (`01...`) and legacy plaintext-like input during password validation.
- `src/ipc.rs`, `src/ui_interface.rs`
  - IPC `get_config("permanent-password")` returns `"<set>"` or empty string (no secret material).
  - Desktop UI queries service-side "is set" state via IPC without persisting plaintext locally.
- `src/ui/index.tis` (Sciter)
  - "Set permanent password" dialog no longer pre-fills existing password.
- Flutter UI
  - Do not pre-fill permanent password fields.
  - Allow submitting empty fields to clear the permanent password.
  - Adjust success detection to work with "is set" semantics rather than secret readback.

## Backward Compatibility
- Existing configs with legacy permanent password storage are migrated to the new `01...` format on first load (when possible).
- Service-side validation remains compatible with legacy inputs.

### UI

https://github.com/user-attachments/assets/4cf206e6-34d7-4990-adcd-d36ce0328480

https://github.com/user-attachments/assets/731cae30-7dc2-46d8-a139-6a91c7573657

https://github.com/user-attachments/assets/b0c2407b-8ec0-4f1f-b199-139efb78f0ff

https://github.com/user-attachments/assets/6fc0cc06-e3be-404c-974a-9d7c1cd1add0

https://github.com/user-attachments/assets/07552552-558a-4cdf-8d8a-8b541e768137

https://github.com/user-attachments/assets/b24cac12-61e8-4418-ad02-fabae28e1d74

## Tests

Windows, Linux deb/AppImage/flatpak, macOS, Android

  - [x] Set permanent password, restart, verify remote auth works.
  - [x] Upgrade from 1.4.5 to 1.4.6

## Notes / Risk
- Post-migration, `salt` is intentionally long-lived to maintain verifier consistency.
- Once upgraded, downgrading is not possible, even with the original password.

## Summary by CodeRabbit

* **New Features**
  * Clear permanent password option with success/failure feedback and a new Clear permanent password menu item.
  * Dialog shows a tip when a preset or hidden permanent password is in use.

* **UI/UX Improvements**
  * Password dialogs start with empty inputs; submit/OK enabled only when input present.
  * Remove button appears when a local password exists; disabled buttons show reduced opacity.

* **Localization**
  * New translation keys for password tips added across locales.


fufesou:
@coderabbitai review

rustdesk:
# claude code review 

  Code Review: PR #14619 — Store permanent password as hashed verifier
                                                                                                                                                                                                               
  Overview        
                                                                                                                                                                                                               
  This PR replaces the reversible-ciphertext storage of permanent passwords with a non-reversible hashed verifier (01<base64(H1)> with SHA256). The UI no longer exposes or pre-fills password secrets. This is
   a meaningful security improvement — recovering plaintext passwords from config files was a real attack vector.
                                                                                                                                                                                                               
  +565 / -193 across 64 files (most are translation string additions).                                                                                                                                         
  
  ---                                                                                                                                                                                                          
  Positives       
           
  - Sound security model: Passwords stored as one-way hashes; plaintext never leaves the set-password call path. The UI never reads back secrets.
  - Clean IPC ACK/NACK protocol: set_permanent_password_with_ack_async waits for an explicit daemon response rather than relying on read-back verification — much better than the old setPermanentPassword in  
  server_model.dart which set the password, slept 500ms, then read it back to compare.                                                                                                                         
  - Good backward compatibility: Legacy formats auto-migrate on first load.                                                                                                                                    
  - Proper authorization check: Both the IPC handler and the client-side set_permanent_password check is_disable_change_permanent_password() before proceeding.                                                
                                                                                                                                                                                                               
  ---
  Issues & Suggestions                                                                                                                                                                                         
                  
  1. Newline-delimited IPC payload is fragile (Medium)

  src/ipc.rs — The permanent-password-storage-and-salt value is packed as storage + "\n" + &salt and split with split_once('\n'). This works today but is a brittle wire format. If either value ever contains 
  a newline (unlikely for base64/hex, but not enforced), parsing silently breaks.
                                                                                                                                                                                                               
  Suggestion: Consider a structured separator that can't appear in base64 (e.g., \0) or use a simple JSON object. At minimum, add a debug assertion that neither storage nor salt contains \n.                 
  
  2. unwrap() on RwLock in IPC handler (Low-Medium)                                                                                                                                                            
                  
  src/ipc.rs:~line 688 and src/flutter_ffi.rs — HARD_SETTINGS.read().unwrap() will panic if the lock is poisoned. This runs in the IPC handler path where a panic would crash the service daemon.              
                  
  Suggestion: Use .read().unwrap_or_else(|e| e.into_inner()) or handle the error gracefully, especially in the daemon-side IPC handler.                                                                        
                  
  3. Hardcoded 1-second timeout (Low-Medium)                                                                                                                                                                   
                  
  src/ipc.rs — set_permanent_password_with_ack_async and sync_permanent_password_storage_from_daemon_async both use ms_timeout = 1_000. On slow machines or under load, this could time out and silently report
   failure.       
                                                                                                                                                                                                               
  Suggestion: Consider making this configurable or at least documenting the timeout choice. The sync_permanent_password_storage_from_daemon failure path only logs a warning — good — but the ACK path returns 
  Ok(false) on timeout, which is indistinguishable from a genuine rejection.
                                                                                                                                                                                                               
  4. canSubmit state not reactive in Flutter dialog (Medium)

  flutter/lib/desktop/pages/desktop_home_page.dart — canSubmit is a local var updated inside updateCanSubmit(), but the OK button's onPressed is bound to canSubmit ? submit : null at dialog build time. Since
   canSubmit is not an Rx observable and updateCanSubmit() is called inside setState, the button should re-render correctly — but the onSubmit: canSubmit ? submit : null on the CustomAlertDialog is evaluated
   once at build time and may not update. Verify that onSubmit (used for Enter key handling) properly re-evaluates on each rebuild.                                                                            
                  
  5. Deprecated MaterialStatePropertyAll usage (Low)                                                                                                                                                           
  
  desktop_home_page.dart:~line 1102 — MaterialStatePropertyAll(Colors.red) is deprecated in Flutter 3.19+. Should use WidgetStatePropertyAll.                                                                  
                  
  6. Remove button shown without confirmation (Low-Medium)

  The red "Remove" button immediately deletes the permanent password without a confirmation dialog. For a destructive action on a security credential, a confirmation step would be prudent.                   
  
  7. Mobile dialog removed, desktop dialog used for both — verify mobile UX (Low)                                                                                                                              
                  
  The entire setPermanentPasswordDialog in flutter/lib/mobile/widgets/dialog.dart was deleted. The desktop setPasswordDialog in desktop_home_page.dart now handles both via showStatusTipOnMobile checks. This 
  consolidation is good, but the minWidth: 540 for mobile with status tip may be too wide for small phone screens. Verify on narrow viewports (320dp width).
                                                                                                                                                                                                               
  8. Missing is_local_permanent_password_set and other Config methods                                                                                                                                          
  
  src/flutter_ffi.rs references ui_interface::is_permanent_password_set(), ui_interface::is_local_permanent_password_set(), ui_interface::set_permanent_password_with_result(), and src/ipc.rs references      
  Config::has_permanent_password(), Config::matches_permanent_password_plain(), Config::get_local_permanent_password_storage_and_salt(), Config::set_permanent_password_storage_for_sync(). These are not
  visible in the diff — presumably they're in config.rs or ui_interface.rs changes that weren't included in the files list (the API returned only page 1 of 64 files). These are critical to review since they 
  contain the actual hashing logic.

  ---
  Security Considerations
                         
  - Good: Password plaintext never returned over IPC after this change.
  - Good: Preset password comparison happens daemon-side, returning only a boolean.                                                                                                                            
  - Verify: The hashing scheme (SHA256 with salt) — ensure the salt is per-installation and sufficiently random. SHA256 alone is fast; consider whether a slow hash (argon2, bcrypt, scrypt) would be more     
  appropriate for password storage, even if this is a local config file. The threat model may not require it since physical access to the config file implies access to the machine, but it's worth documenting
   the decision.                                                                                                                                                                                               
  - Verify: The 01<base64(H1)> format — ensure the migration path from old format doesn't leave plaintext in config file backups or logs.                                                                      
                  
  ---
  Summary
         
  This is a solid security improvement with clean architecture. The main concerns are: (1) review the missing config.rs/ui_interface.rs changes containing the actual hashing logic, (2) the fragile
  newline-delimited IPC payload, and (3) minor Flutter UX issues (mobile width, missing confirmation on delete, deprecated API). The IPC ACK/NACK pattern is well-designed and a clear improvement over the old
   sleep-and-read-back approach.

rustdesk:
# Codex review

• Findings

  1. Critical: the new 01<base64(H1)> format is still a password-equivalent credential, so copying it into user config/IPC does not actually close the local-attacker hole this PR is trying to fix. H1 =
     SHA256(password || salt) is encoded/stored in /tmp/pr14619/libs/hbb_common/src/config.rs:53 and returned/synced through /tmp/pr14619/src/ipc.rs:635, /tmp/pr14619/src/ipc.rs:1180, and /tmp/pr14619/src/
     ipc.rs:1264. The server then authenticates by hashing that exact H1 with the public challenge in /tmp/pr14619/src/server/connection.rs:1985, and the client-side login flow already does the same H1 -> H2
     step in src/client.rs:3574. A user who can read the synced 01... value can still authenticate remotely without ever learning the plaintext password.
  2. Major: the Sciter password menu now does synchronous IPC on the render path, which can hang the UI thread for up to the IPC timeout. renderPop(), toggleMenuState(), and the set-password click handler
     call handler.is_local_permanent_password_set() / handler.is_permanent_password_set() in /tmp/pr14619/src/ui/index.tis:1075, /tmp/pr14619/src/ui/index.tis:1120, and /tmp/pr14619/src/ui/index.tis:1138. On
     desktop those helpers immediately call daemon sync in /tmp/pr14619/src/ui_interface.rs:612 and /tmp/pr14619/src/ui_interface.rs:634, which goes through blocking get_config() with a 1s timeout in /tmp/
     pr14619/src/ipc.rs:1138 and /tmp/pr14619/src/ipc.rs:1202. If the daemon is slow or unavailable, opening that menu can visibly stall.
  3. Medium: ACK/protocol failures are collapsed into a clean “rejected by daemon” result even after the daemon may already have applied the change. The daemon mutates the permanent password before sending
     its ACK in /tmp/pr14619/src/ipc.rs:699, but the client side returns Ok(false) for any missing/malformed reply in /tmp/pr14619/src/ipc.rs:1255, and set_permanent_password() maps that to a rejection in /
     tmp/pr14619/src/ipc.rs:1239. A delayed/lost ACK can therefore leave the password changed while the UI reports failure.

  Open Questions

  - The PR description says Config::set(cfg) was guarded against salt drift, but the fetched hbb_common head still has an unguarded /tmp/pr14619/libs/hbb_common/src/config.rs:1556, and the bidirectional sync
    path still calls it in /tmp/pr14619/src/server.rs:621 and /tmp/pr14619/src/server.rs:748. I did not prove a concrete breakage from that alone, but the implementation does not match the stated mitigation.

  Line refs point to fetched PR-head snapshots under /tmp/pr14619. Static review only; I did not run tests against a checked-out PR branch.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14619#issuecomment-4127398770

> 1. Newline-delimited IPC payload is fragile (Medium)

Not an issue.

Neither storage nor salt contains `\n`.

> 2. unwrap() on RwLock in IPC handler (Low-Medium)                                                                                                                                                            
> 3. Hardcoded 1-second timeout (Low-Medium)     
> 4. canSubmit state not reactive in Flutter dialog (Medium)

Not an issue.

> 5. Deprecated MaterialStatePropertyAll usage (Low)                                                                                                                                                           

`MaterialStatePropertyAll` is used in many places.

> 6. Remove button shown without confirmation (Low-Medium)

Not an issue.
  
> 7. Mobile dialog removed, desktop dialog used for both — verify mobile UX (Low)                                                                                                                              

Not an issue. The removed code is dead code.

"540 for mobile with status tip may be too wide for small phone screens", reverted to `500`.
                                                                                                                                                                                                               
> 8. Missing is_local_permanent_password_set and other Config methods                                                                                                                                          
 
Not an issue.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14619#issuecomment-4127418718

• Findings

>  1. Critical: the new 01<base64(H1)> format is still a password-equivalent credential, so copying it into user config/IPC does not actually close the local-attacker hole this PR is trying to fix.

This PR is used to prevent password leaks.
We will completely close the local-attacker hole in another PR.

> 2. Major: the Sciter password menu now does synchronous IPC on the render path, which can hang the UI thread for up to the IPC timeout. renderPop(), toggleMenuState(), and the set-password click

Not an issue.

> 3. Medium: ACK/protocol failures are collapsed into a clean “rejected by daemon” result even after the daemon may already have applied the change.

Not an issue.
This is uncommon. Users can retry by setting the password upon seeing an error.

Open Questions

> - The PR description says Config::set(cfg) was guarded against salt drift, but the fetched hbb_common head still has an unguarded

The description is outdated.

fufesou:
@coderabbitai review

PR #14620, opened 2026-03-25
Fix Windows session-based logon and lock-screen detection

21pages:
## Summary
                                                                                                                                                                                              
  Fix Windows logon / lock-screen detection to use the current Windows session instead of a console-preferred session.                                                                        
                                                                                                                                                                                              
  Also allow permanent password fallback for logon / lock-screen access when `allow-logon-screen-password=Y`, even if the verification method is set to `Use one-time password`.              
                                                                                                                                                                                              
  ## Changes                                                                                                                                                                                  
                                                                                                                                                                                              
  - scope `is_locked()` to the current process session                                                                                                                                        
  - scope `is_logon_ui()` to `LogonUI.exe` in the current process session                                                                                                                     
  - allow permanent password fallback for logon / lock-screen access                                                                                                                          
                                                                                                                                                                                              
  ## Test                                                                                                                                                                                
            
 #### Setup                                                                                                                                                                                  
  - console user and RDP user are different Windows users                                                                                                                                     
  - permanent password is configured on the incoming client                                                                                                                                   
  - verification method is set to `Use one-time password`                                                                                                                                     
  - approve mode is set to `Accept sessions via click`                                                                                                                                        
                                                                                                                                                                                              
  #### Cases                                                                                                                                                                               
                                                                                                                                                                                              
  1. Console desktop, no RDP                                                                                                                                                                  
     Click confirmation is required.                                                                                                                                                          
                                                                                                                                                                                              
  2. Console locked or signed out, no RDP                                                                                                                                                     
     Password entry is allowed.                                                                                                                                                               
     a. Without switching user: connection succeeds.                                                                                                                                          
     b. After switching user: connection is disconnected, the desktop is shown, and click confirmation is required.                                                                           
                                                                                                                                                                                              
  3. Console desktop, RDP locked                                                                                                                                                              
     Current session is console: console click confirmation is required first, then session selection is shown.                                                                               
     a. Select console: connection succeeds directly.                                                                                                                                         
     b. Select RDP: password entry is allowed.
     Current session is RDP: password entry is allowed first, then session selection is shown.                                                                                                
     a. Select RDP: connection succeeds directly.                                                                                                                                             
     b. Select console: click confirmation is required.                                                                                                                                       
                                                                                                                                                                                              
  4. Console desktop, RDP desktop                                                                                                                                                             
     Current session is console: console click confirmation is required first, then session selection is shown.                                                                               
     a. Select console: connection succeeds directly.                                                                                                                                         
     b. Select RDP: RDP user click confirmation is required.                                                                                                                                  
     Current session is RDP: RDP user click confirmation is required first, then session selection is shown.                                                                                  
     a. Select RDP: connection succeeds directly.                                                                                                                                             
     b. Select console: console click confirmation is required.                                                                                                                               
                                                                                                                                                                                              
  5. Console signed out or locked, RDP desktop                                                                                                                                                
     Current session is console: password entry is allowed first, then session selection is shown.                                                                                            
     a. Select console: connection succeeds directly.                                                                                                                                         
     b. Select RDP: click confirmation is required.                                                                                                                                           
     Current session is RDP: click confirmation in the RDP window is required first, then session selection is shown.                                                                         
     a. Select RDP: connection succeeds directly.                                                                                                                                             
     b. Select console: password entry is allowed.                                                                                                                                            
                                                                                                                                                                                              
  6. Console signed out or locked, RDP locked                                                                                                                                                 
     Current session is console: password entry is allowed first, then session selection is shown.                                                                                            
     a. Select console: connection succeeds directly.                                                                                                                                         
     b. Select RDP: the previously entered password is reused.                                                                                                                                
     Current session is RDP: password entry is allowed first, then session selection is shown.                                                                                                
     a. Select RDP: connection succeeds directly.                                                                                                                                             
     b. Select console: the previously entered password is reused.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Session-lock and login-screen detection now use the actual session ID, improving accuracy across multi-session and remote-desktop scenarios.

* **Improvements**
  * Login flow refined to support an explicit logon-screen password fallback; when that fallback succeeds it now emits an informational log entry.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/src/platform/windows.cc
+++ b/src/platform/windows.cc
@@ -580,9 +580,8 @@ extern "C"
         return rdp_or_console;
     }
 
-    BOOL is_session_locked(BOOL include_rdp)
+    BOOL is_session_locked(DWORD session_id)
     {
-        DWORD session_id = get_current_session(include_rdp);
         if (session_id == 0xFFFFFFFF) {
             return FALSE;
         }
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -523,7 +523,7 @@ const SERVICE_TYPE: ServiceType = ServiceType::OWN_PROCESS;
 
 extern "C" {
     fn get_current_session(rdp: BOOL) -> DWORD;
-    fn is_session_locked(include_rdp: BOOL) -> BOOL;
+    fn is_session_locked(session_id: DWORD) -> BOOL;
     fn LaunchProcessWin(
         cmd: *const u16,
         session_id: DWORD,
@@ -1149,20 +1149,21 @@ pub fn is_prelogin() -> bool {
 }
 
 pub fn is_locked() -> bool {
-    unsafe { is_session_locked(share_rdp()) == TRUE }
+    let Some(session_id) = get_current_process_session_id() else {
+        return false;
+    };
+    unsafe { is_session_locked(session_id) == TRUE }
 }
 
-// `is_logon_ui()` is regardless of multiple sessions now.
-// It only check if "LogonUI.exe" exists.
-//
-// If there're mulitple sessions (logged in users),
-// some are in the login screen, while the others are not.
-// Then this function may not work fine if the session we want to handle(connect) is not in the login screen.
-// But it's a rare case and cannot be simply handled, so it will not be dealt with for the time being.
 #[inline]
 pub fn is_logon_ui() -> ResultType<bool> {
+    let Some(current_sid) = get_current_process_session_id() else {
+        return Ok(false);
+    };
     let pids = get_pids("LogonUI.exe")?;
-    Ok(!pids.is_empty())
+    Ok(pids
+        .into_iter()
+        .any(|pid| get_session_id_of_process(pid) == Some(current_sid)))
 }
 
 pub fn is_root() -> bool {
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -2025,7 +2025,7 @@ impl Connection {
         self.validate_password_plain(storage)
     }
 
-    fn validate_password(&mut self) -> bool {
+    fn validate_password(&mut self, allow_permanent_password: bool) -> bool {
         if password::temporary_enabled() {
             let password = password::temporary_password();
             if self.validate_one_password(&password) {
@@ -2037,7 +2037,7 @@ impl Connection {
                 return true;
             }
         }
-        if password::permanent_enabled() {
+        if password::permanent_enabled() || allow_permanent_password {
             // Since hashed storage uses a prefix-based encoding, a hard plaintext that
             // happens to look like hashed storage could be mis-detected. Validate local storage
             // and hard/preset plaintext via separate paths to avoid that ambiguity.
@@ -2349,6 +2349,10 @@ impl Connection {
             #[cfg(any(target_os = "android", target_os = "ios"))]
             let is_logon = || crate::platform::is_prelogin();
 
+            let allow_logon_screen_password =
+                crate::get_builtin_option(keys::OPTION_ALLOW_LOGON_SCREEN_PASSWORD) == "Y"
+                    && is_logon();
+
             if !hbb_common::is_ip_str(&lr.username)
                 && !hbb_common::is_domain_port_str(&lr.username)
                 && lr.username != Config::get_id()
@@ -2357,8 +2361,7 @@ impl Connection {
                     .await;
                 return false;
             } else if (password::approve_mode() == ApproveMode::Click
-                && !(crate::get_builtin_option(keys::OPTION_ALLOW_LOGON_SCREEN_PASSWORD) == "Y"
-                    && is_logon()))
+                && !allow_logon_screen_password)
                 || password::approve_mode() == ApproveMode::Both && !password::has_valid_password()
             {
                 self.try_start_cm(lr.my_id, lr.my_name, false);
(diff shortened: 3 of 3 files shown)


rustdesk:
# code review

  - Important: /tmp/rustdesk-pr14620/src/server/connection.rs:2293 derives allow_logon_screen_password purely from allow-logon-screen-password=Y plus is_logon(), and /tmp/rustdesk-pr14620/src/server/
    connection.rs:1985 uses that to bypass permanent_enabled() from libs/hbb_common/src/password_security.rs:68. That makes the new permanent-password exception apply in ApproveMode::Password and
    ApproveMode::Both too, not just the click-mode scenario described and manually tested in the PR. With verification-method=use-temporary-password, a stored permanent password will still authenticate on
    the lock/logon screen in those modes. If the intended scope is “click-mode logon-screen fallback”, this needs to be gated on ApproveMode::Click, not just is_logon().

  Open Question

  - The PR explicitly wants temp-only mode to accept the permanent password at the logon screen. Since switching verification mode only updates the option and does not clear the stored permanent password
    (src/ui/index.tis:1155), this silently changes what “Use one-time password” means. If that policy is intentional, it probably needs explicit UI/docs rather than a hidden auth exception.

  I did not run Windows-side validation locally; this review was static against the fetched PR branch and the PR metadata/test plan.

rustdesk:
# claude code review

  Analysis

  src/platform/windows.cc — Good

  The C++ function is_session_locked no longer calls get_current_session(include_rdp) internally. Instead it takes a session_id parameter directly. Clean separation of concerns — the caller decides which
  session to query, the C function just does the WTS check. The 0xFFFFFFFF guard remains, correctly handling invalid session IDs.

  src/platform/windows.rs — Good, with notes

  - is_locked(): Now uses get_current_process_session_id() instead of share_rdp(). Correct fix — the old code used get_current_session(include_rdp) which returns the "active" console session, but in a
  multi-session scenario (RDP + console) it could return the wrong session.
  - is_logon_ui(): The old code checked if any LogonUI.exe existed system-wide. The new code filters by session — checks if LogonUI.exe runs in the current process's session. Correct fix for multi-session
  environments.
  - The removal of the comment block explaining the known limitation is appropriate since the limitation is now fixed.
  - Minor note: get_pids("LogonUI.exe") still enumerates all processes system-wide via CreateToolhelp32Snapshot, then filters by session. There's a get_pids_with_first_arg_check_session pattern elsewhere
  (line 4085) but it has a different signature, so current approach is acceptable.

  src/server/connection.rs — Needs scrutiny

  - validate_password(allow_permanent_password: bool): The new parameter allows permanent password validation even when password::permanent_enabled() returns false. Specifically for the logon screen
  scenario.
  - allow_logon_screen_password variable: Extracted from the inline check, improving readability. Computed once and reused in two places.
  - Security concern (moderate): The condition password::permanent_enabled() || allow_permanent_password means that when allow_logon_screen_password is true, the permanent password is always tried regardless
   of the admin's permanent_enabled() setting. This is intentional — OPTION_ALLOW_LOGON_SCREEN_PASSWORD=Y is an explicit opt-in. However, this creates a subtle interaction: even if an admin disables the
  permanent password in general settings, it can still be used at the logon screen if the builtin option is set. This should be documented.
  - Behavioral change: is_logon() is now called eagerly when computing allow_logon_screen_password. Previously it was only called inside the approve_mode() == Click branch. Now it's called unconditionally on
   every login request. Performance cost is negligible (one process snapshot + one WTS query), but it's a change worth noting.
  - Minor inefficiency: allow_logon_screen_password is computed before the username validation check. If the username doesn't match, the is_logon() call was wasted.

  ---
  Suggestions

  1. Consider making allow_logon_screen_password lazy to avoid process enumeration on every request:
  let allow_logon_screen_password = || {
      crate::get_builtin_option(keys::OPTION_ALLOW_LOGON_SCREEN_PASSWORD) == "Y"
          && is_logon()
  };
  2. Add a log line when the permanent password fallback is triggered — this is a security-sensitive path:
  if allow_permanent_password && !password::permanent_enabled() {
      log::info!("Allowing permanent password for logon screen access");
  }
  3. No test coverage for the new behavior. The validate_password(allow_permanent_password) logic could be tested to verify the permanent password is tried when allow_permanent_password=true even when
  permanent_enabled()=false.

  ---
  Verdict

  Well-targeted fix for a real multi-session bug. Code changes are small, focused, and correct. The session-scoping approach aligns with existing patterns (get_current_process_session_id() is already used in
   ~8 other places). Security implications are acceptable given explicit opt-in, but should be logged.

  Recommendation: Approve with minor suggestions (lazy evaluation, logging, tests).

21pages:
*  `allow-logon-screen-password` is an explicit opt-in for logon/lock-screen access. In that state, temporary passwords are often impractical in real use: users usually do not know the  current temporary password after locking the machine, and reconnect/session changes can also make one-time-password flows unreliable.  Because of that, when this option is enabled, permanent-password fallback on the logon/lock screen is intentionally allowed not only for Click mode, but also for Password/Both modes. The behavior is documented in our docs.
* When `allow-logon-screen-password=Y`, `is_logon()` will almost certainly be needed in this path. Computing it once up front avoids calling the same logon-state detection multiple times.

21pages:
force push to rebase

Issue #8768, opened 2024-07-20
Keyboard stops working if mouse is not moved for a few seconds. Must wiggle mouse for keyboard to begin working again.


msftsecurityteam:
### Bug Description

I am remoting into a Linux workstation and I am a heavy user of the terminal. It is very annoying, but I would say after approcximately 5 seconds of mouse inactivity, the keyboard stops working on the remote host. If I wiggle the mouse, then the keyboard begins working again. This typically allows me to do one command per mouse wiggle before I am unable to use the keyboard.

### How to Reproduce

1. Open terminal
2. Stop using mouse for 5-10 seconds

### Expected Behavior

If the repro steps are followed, then the terminal stops having characters written to it because the keyboard now is not functioning properly.

### Operating system(s) on local side and remote side

Debian 11 -> Debian 12

### RustDesk Version(s) on local side and remote side

1.2.3 -> 1.2.3

### Screenshots

N/A

rustdesk:
tracked by https://github.com/rustdesk/rustdesk/issues/3565

rustdesk:
wayland? please give more details about your debian 11.

msftsecurityteam:
Client: Debian 11 running awesomewm (X11)
Server: Debian 12 running default XFCE from installation (X11)

I had the client upgraded to 1.2.7 for a while now (no change) and just upgraded the server rustdesk. I still have not noticed any improvement with this issue. One observation: I notice that the hidden menubar on the bottom of the XFCE4 desktop (with the icons of common apps that I run, like the "start" menu for XFCE) unhides itself as this bug manifests. Meaning, it's as if the mouse was pulled down to the menubar (thus, unhiding it) right before the keyboard stops working.

Not running Wayland on either side.

msftsecurityteam:
Was this ever fixed? I updated my Rustdesk to the latest version and it still happens...

jjolmo:
I can reproduce the exact same behaviour

Mint X11 (1.2.3) > Debian X11 (1.2.3)

It just happened out of nowhere very recently. Btw, thanks for the "wiggle" trick to solve it. I cannot say I can work with it, but confirmed that both of our setups using STABLE versions with debian have the same issue

jjolmo:
@msftsecurityteam This issue is driving me nuts. But I discovered other bad behaviours lately. My setup is Linux Mint. Some of the examples:
[...]

Issue #4570, opened 2023-06-07
Keyboard input is ignored on remote client after extensive time（connect to two machines and switch)


bigretromike:
### Bug Description

When connecting from win11/debian11 to my others remote machines (most of the time linux based x11 based ones). After extensive time the keyboard input stops working, I can control machine fine with mouse but anything that I type is not send to remote machine. 
This one was happening also around "disconnect" issue I had before, which was fixed. by then the "workaround" was to switch input type setting from map to legacy if I recall correctly.
I thing this is more often after connecting to two machines (x11)

### How to Reproduce

- connect to machine A
- connect to machine B
- switch many times between them
- leave connection on for them
- check after 15?30 minutes. 

(i will try to make this more reproducibly) 

### Expected Behavior

keyboard input should work 100% of the time.

### Operating system(s) on local side and remote side

Windows 11 -> Debian 11, Debian 11

### RustDesk Version(s) on local side and remote side

1.2.0 -> 1.2.0, 1.1.9

### Screenshots

<no way to screen shot a not working keyboard>

rustdesk:
@fufesou

rustdesk:
How can you make it work again?

bigretromike:
> How can you make it work again?

sadly only by reconnecting, then its always back on track and working correct.

rustdesk:
> 2\. if there is a disconnection on (machine a) while machine B (remote) stays connected to internet,  on reconnection, keyboard does not work.

hard to understand

test2a:
@rustdesk okay. 

i am machine A. remote is Machine B
i am connected. 

there is an internet disconnection at machine A

after i reconnect, rustdesk also sometimes reconnects but

mouse works.
keyboard does not work.
basically everything except keyboard works.

solution

close rustdesk on machine A
reopen and reconnect

now everything works

rustdesk:
Understand now.

fufesou:
> i have more inputs on this. I have witnessed this myself so i can confirm this.
> 
> 1. connection continues as long as it works.
> 2. if there is a disconnection on (machine a) while machine B (remote) stays connected to internet,  on reconnection, keyboard does not work.
> 3. only option is to close rustdesk on machine A, then reconnect.

Hi, did you mean 
```
C --> A --> B
```

Keyboard input from A to B does not work after C reconnecting A?

I've tried connecting a long time, and then manually disconnect `C --> A` and connect `C --> A`.

The keybaord works fine.

[...]

PR #14651, opened 2026-03-30
Fix: restore keyboard/mouse input on iPadOS after returning from background

Nelsitobcn:
## Problem

When the user switches to another app and returns to RustDesk on iPadOS, **all input stops working** (keyboard, trackpad, touch) until the app is restarted.

Related issues: #4570, #8768

## Root cause

iOS silently drops the Flutter focus node when the app goes to background. On resume, nobody re-requests it, so input events never reach the session.

## Fix

Call `_restoreInputOnResume()` from `didChangeAppLifecycleState` when `state == resumed`. This:
1. Restarts the mouse listener (`listenToMouse(true)`)
2. Re-requests physical keyboard focus with a short delay

This uses the exact same pattern already present in `onSoftKeyboardChanged` for the iOS virtual keyboard workaround.

## Tested on

- iPad Pro with Magic Keyboard Folio
- iPadOS 26.3.1 (beta)
- Self-hosted RustDesk server over Tailscale → macOS

## Steps to reproduce (before fix)

1. Open RustDesk on iPad and connect to a remote Mac
2. Press the Home button or switch to another app
3. Return to RustDesk
4. Try typing or using the trackpad → **no input works**

After this fix, input is restored automatically within ~150ms of returning to the app.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mouse input handling on iOS when the remote control app resumes.
  * Fixed keyboard focus recovery on iOS to ensure proper input responsiveness after app suspension.

### DIFF
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -154,7 +154,27 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
   void didChangeAppLifecycleState(AppLifecycleState state) {
     if (state == AppLifecycleState.resumed) {
       trySyncClipboard();
+      _restoreInputOnResume();
     }
+  }
+
+  // Fix: restore keyboard/mouse input after returning from background on iPadOS.
+  // Bug: switching apps on iPadOS loses all input (keyboard, trackpad, touch).
+  // Root cause: iOS drops the focus node when app goes to background; nobody re-requests it.
+  // Fix: on resume, restart mouse listener and re-request physical focus with a small
+  // delay (same pattern already used in onSoftKeyboardChanged for iOS keyboard workaround).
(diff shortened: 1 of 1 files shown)


fufesou:
Hi, thanks for the PR.

This PR looks good to me, but I still have some questions.

1. I can't reproduce the issue. Could you please let me know what I'm doing wrong?

https://github.com/user-attachments/assets/86b59f38-925d-49bc-8683-5210ee72da53

Or could you please upload videos showing the behavior before and after the fix?

2. It seems these two issues do not mention iOS or iPad:
https://github.com/rustdesk/rustdesk/issues/4570
https://github.com/rustdesk/rustdesk/issues/8768


3. Please address the suggestion here:
https://github.com/rustdesk/rustdesk/pull/14651#pullrequestreview-4031697415

rustdesk:
Close since no response.

PR #14652, opened 2026-03-30
fix(iPad): keep touch gestures with external mouse

s1korrrr:
Fixes #13011.

## Problem

On iPad, once an external mouse / trackpad is detected, the mobile remote page stops wrapping the session body in `RawTouchGestureDetectorRegion`.

That disables touch gestures such as pinch-to-zoom and leaves the session in the mixed "mouse mode behaves like broken touch mode" state described in the issue.

## Root cause

`RemotePage` currently gates the touch gesture wrapper on `inputModel.isPhysicalMouse.value`:

- no physical mouse: wrap in `RawTouchGestureDetectorRegion`
- physical mouse detected: render `getBodyForMobile()` directly

Pointer input already continues to flow through the session body listener, so dropping the gesture wrapper is unnecessary and removes the touch-only gesture path.

## Fix

Always keep the mobile session body wrapped in `RawTouchGestureDetectorRegion`.

This preserves touch gestures even after an external mouse / trackpad is attached, while leaving the existing pointer event path intact.

## Verification

- `git diff --check`
- `cd flutter && flutter analyze lib/mobile/pages/remote_page.dart`

Verification result in this environment:
- no errors from this patch
- only pre-existing informational warnings in `remote_page.dart`

## Residual Risk

I do not have a live iPad -> macOS repro rig in this environment for end-to-end device smoke, so the remaining risk is behavioral rather than static.

## Summary by CodeRabbit

* **Bug Fixes**
  * Touch handling consistently uses the touch-aware UI wrapper so touch gestures remain active regardless of a connected physical mouse.
  * Gesture recognizers now accept only touch pointer types, preventing mouse/trackpad input from interfering with touch interactions.

### DIFF
--- a/flutter/lib/common/widgets/remote_input.dart
+++ b/flutter/lib/common/widgets/remote_input.dart
@@ -532,7 +532,9 @@ class _RawTouchGestureDetectorRegionState
       // Official
       TapGestureRecognizer:
           GestureRecognizerFactoryWithHandlers<TapGestureRecognizer>(
-              () => TapGestureRecognizer(), (instance) {
+              () => TapGestureRecognizer(
+                    supportedDevices: kTouchBasedDeviceKinds,
+                  ), (instance) {
         instance
           ..onTapDown = onTapDown
           ..onTapUp = onTapUp
@@ -540,14 +542,18 @@ class _RawTouchGestureDetectorRegionState
       }),
       DoubleTapGestureRecognizer:
           GestureRecognizerFactoryWithHandlers<DoubleTapGestureRecognizer>(
-              () => DoubleTapGestureRecognizer(), (instance) {
+              () => DoubleTapGestureRecognizer(
+                    supportedDevices: kTouchBasedDeviceKinds,
+                  ), (instance) {
         instance
           ..onDoubleTapDown = onDoubleTapDown
           ..onDoubleTap = onDoubleTap;
       }),
       LongPressGestureRecognizer:
           GestureRecognizerFactoryWithHandlers<LongPressGestureRecognizer>(
-              () => LongPressGestureRecognizer(), (instance) {
+              () => LongPressGestureRecognizer(
+                    supportedDevices: kTouchBasedDeviceKinds,
+                  ), (instance) {
         instance
           ..onLongPressDown = onLongPressDown
           ..onLongPressUp = onLongPressUp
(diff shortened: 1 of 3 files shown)


s1korrrr:
@coderabbitai review

fufesou:
## Root cause

The remote page(`Container`) does not rebuild after the first build.

### Workaround (not good)

Click the floating bar to show and hide the menu.
Then the remote page is rebuilt.

## Fix 1 (not good)

bd4d501

### Tests

iPad and Android as the control/local side.

Switch between the touch control and the mouse control.

### Known issue

Touch control followed by mouse control works fine.

However, if you use the mouse first and then touch, the first touch event is ignored (because the flag hasn't been set yet). Subsequent touch events work as expected.

## Fix 2

9ee100b53e7a3f336122f827c814b363f7a9f9dc

### Tests

Android and iPad

- [x] Switch between the touch control and the mouse/touchpad control.
- Touch screen
  - [x] one finger - move, long press, double-click/press & drag
  - [x] twe fingers - pan and scale
  - [x] three fingers - scroll 
- Physical mouse (Android)
  - [x] move, click (left & right), scroll
- touchpad (iPad)
  - [x] one finger - move, click
  - [x] two fingers - scroll, click
- Pencil (iPad)
  - [x] move, long press, double-click/press & drag

### More

The changes in PR #14515 should also be merged. Without them, external click events can occur when clicking near the top edge.

"Fix 1" above avoids this issue, but there's no workaround for the "the first touch event is ignored" issue.

fufesou (approved):

rustdesk:
Issues / Suggestions                                                                                                                                                                                         
                                                                                                                                                                                                               
  1. The same bug exists in view_camera_page.dart and is NOT fixed. 
  ```                                                                                                                                           
  flutter/lib/mobile/pages/view_camera_page.dart:262-268 has the exact same conditional pattern:
  child: inputModel.isPhysicalMouse.value                                                                                                                                                                      
      ? getBodyForMobile()               
      : RawTouchGestureDetectorRegion(                                                                                                                                                                         
          child: getBodyForMobile(),  
          ffi: gFFI,                
          isCamera: true,
        ),                            

```                                                                                                                                                                         
  The camera page will exhibit identical breakage on iPad with an external mouse. Strongly recommend fixing it in the same PR for parity — same one-line change.

rustdesk:
Reverted to fix https://github.com/rustdesk/rustdesk/issues/15209

PR #14653, opened 2026-03-30
fix(ios): preserve capslock state across key events

### LINKED ISSUE #12111: IPad->iMac - caps lock issue (IPad Magic Keyboard)
### Bug Description

I connect with an iPad and magickeyboard on an iMac M1.
There's a problem with caps lock key.
If I activate it, the characters remain in lower case. If I disable it, they change to upper case. To deactivate caps: I need to reactivate caps lock, write a character that will be capitalized, then remove caps lock.

### How to Reproduce

Activate the caps lock, the characters remain in lower case. Disable it, they change to upper case. To deactivate caps: reactivate caps lock, write a character that will be capitalized, then remove caps lock.

### Expected Behavior

Caps lock one do upper case, disable it remove it.

iPad ->MacOS

1.4.0 both

### Screenshots

No

### Additional Context

_No response_

s1korrrr:
Fixes #12111.

## Problem

RustDesk already contains an iOS-specific Caps Lock workaround, but the current implementation only derives `iosCapsLock` from the current event's character output.

That leaves a stale-state gap on iOS:

- key-up events fall back to `false` when they do not carry a usable character
- non-letter events also fall back to `false`
- explicit Caps Lock key presses do not maintain a cached logical state across subsequent events

This matches the still-open iPad Magic Keyboard issue and the stale lock-mode concerns raised during review of the earlier workaround PR.

## Root cause

The iOS workaround is stateless per event. It infers Caps Lock from letter output, but it does not cache the result and reuse it for key-up / repeat / non-character paths.

## Fix

- add a cached `_iosCapsLock` state in `InputModel`
- toggle it on explicit Caps Lock key-down events
- update it when character output is inferable
- reuse the cached state for key-up and non-character events instead of resetting to `false`

This keeps the patch inside the existing workaround design instead of changing the broader keyboard pipeline.

## Verification

- `git diff --check`
- `cd flutter && flutter analyze lib/models/input_model.dart`

Verification result in this environment:
- no new errors from this patch
- existing info-level deprecation warnings remain in `input_model.dart`

## Notes

The issue thread is locked, so I could not post a claim comment there.

## Residual Risk

I do not have a live iPad Magic Keyboard -> macOS repro setup in this environment for end-to-end input smoke.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable iOS Caps Lock detection via a cached lock-state tracker and broader event handling to avoid incorrect inferences across non-character and shifted inputs, and to ensure resets clear cached state.
* **Tests**
  * Added tests validating Caps Lock caching and behavior across key down/up, shift, and reset scenarios.
* **Chores**
  * Enabled test tooling dependency for development testing.

### DIFF
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -15,6 +15,7 @@ import 'package:get/get.dart';
 import '../../models/model.dart';
 import '../../models/platform_model.dart';
 import '../../models/state_model.dart';
+import 'ios_caps_lock_state_tracker.dart';
 import 'relative_mouse_model.dart';
 import '../common.dart';
 import '../consts.dart';
@@ -335,7 +336,7 @@ class InputModel {
   var ctrl = false;
   var alt = false;
   var command = false;
-  bool _iosCapsLock = false;
+  final _iosCapsLockTracker = IosCapsLockStateTracker();
 
   final ToReleaseRawKeys toReleaseRawKeys = ToReleaseRawKeys();
   final ToReleaseKeys toReleaseKeys = ToReleaseKeys();
@@ -472,38 +473,12 @@ class InputModel {
     // value and update it from explicit CapsLock presses or inferable
     // character output, then reuse that cached state for key-up and
     // non-character events.
-    if (isKeyDown && logicalKey == LogicalKeyboardKey.capsLock) {
-      _iosCapsLock = !_iosCapsLock;
-      return _iosCapsLock;
-    }
-    final inferred = _getIosCapsFromCharacterImpl(character, shiftPressed);
-    if (inferred != null) {
-      _iosCapsLock = inferred;
-    }
-    return _iosCapsLock;
-  }
-
-  // Shared implementation for inferring CapsLock state from character.
-  // Uses Unicode-aware case detection to support non-ASCII letters (e.g., ü/Ü, é/É).
-  //
-  // Limitations:
-  // 1. This inference assumes the client and server use the same keyboard layout.
-  //    If layouts differ (e.g., client uses EN, server uses DE), the character output
-  //    may not match expectations. For example, ';' on EN layout maps to 'ö' on DE
-  //    layout, making it impossible to correctly infer CapsLock state from the
-  //    character alone.
-  // 2. On iOS, CapsLock+Shift produces uppercase letters (unlike desktop where it
-  //    produces lowercase). This method cannot handle that case correctly.
-  bool? _getIosCapsFromCharacterImpl(String? ch, bool shiftPressed) {
-    if (ch == null || ch.length != 1) return null;
-    // Use Dart's built-in Unicode-aware case detection
-    final upper = ch.toUpperCase();
-    final lower = ch.toLowerCase();
-    final isUpper = upper == ch && lower != ch;
-    final isLower = lower == ch && upper != ch;
-    // Skip non-letter characters (e.g., numbers, symbols, CJK characters without case)
-    if (!isUpper && !isLower) return null;
-    return isUpper != shiftPressed;
+    return _iosCapsLockTracker.update(
+      character: character,
+      shiftPressed: shiftPressed,
+      logicalKey: logicalKey,
+      isKeyDown: isKeyDown,
+    );
   }
 
   int _buildLockModes(bool iosCapsLock) {
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -957,9 +957,12 @@ class InputModel {
             .encode(modify({'id': id, 'type': 'wheel', 'y': y.toString()})));
   }
 
-  /// Reset key modifiers to false, including [shift], [ctrl], [alt] and [command].
+  /// Reset local key state, including modifiers and cached iOS lock state.
   void resetModifiers() {
     shift = ctrl = alt = command = false;
+    if (isIOS) {
+      _iosCapsLockTracker.reset();
+    }
   }
 
   /// Modify the given modifier map [evt] based on current modifier key status.
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -335,6 +335,7 @@ class InputModel {
   var ctrl = false;
   var alt = false;
   var command = false;
+  bool _iosCapsLock = false;
 
   final ToReleaseRawKeys toReleaseRawKeys = ToReleaseRawKeys();
   final ToReleaseKeys toReleaseKeys = ToReleaseKeys();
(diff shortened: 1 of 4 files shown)


s1korrrr:
@coderabbitai review

s1korrrr:
Addressed the review feedback in `3c26354`.

Changes:
- keep the cached iOS Caps Lock state unchanged when `Shift` is pressed
- extract the Caps Lock state tracker into a small standalone model file so it can be tested without pulling in the full app graph
- add a focused regression test covering `capsLock -> letter -> key-up/non-character` and `capsLock + Shift + letter`

Verification:
- `cd flutter && flutter test test/models/input_model_ios_capslock_test.dart`
- `cd flutter && flutter analyze lib/models/input_model.dart lib/models/ios_caps_lock_state_tracker.dart`

The test passed locally. Analyze reported only the existing deprecation infos already present in `input_model.dart`.

@coderabbitai review

s1korrrr:
Addressed the remaining review note in `1bafccf`.

Changes:
- `IosCapsLockStateTracker.update()` now returns the cached state immediately on key-up events instead of re-inferring from the character
- added a focused regression test proving key-up cannot clear an enabled caps-lock state

Verification:
- `cd flutter && flutter test test/models/input_model_ios_capslock_test.dart`
- `cd flutter && flutter analyze lib/models/input_model.dart lib/models/ios_caps_lock_state_tracker.dart`

The test passed locally. Analyze still reports only the existing info-level deprecations already present in `input_model.dart`.

@coderabbitai review

s1korrrr:
Addressed the remaining stale-state review thread in `3801fa7`.

Changes:
- add `IosCapsLockStateTracker.reset()`
- clear the cached iOS CapsLock workaround state from the existing local key-state reset path
- add a regression test proving reset clears the cached lock state and keeps subsequent key-up events read-only

Verification:
- `cd flutter && flutter test test/models/input_model_ios_capslock_test.dart`
- `cd flutter && flutter analyze lib/models/input_model.dart lib/models/ios_caps_lock_state_tracker.dart`

The focused test passed locally. Analyze still exits non-zero only for the existing `RawKeyEvent`/platform key deprecation infos already present in `input_model.dart`; no new analyzer issue was introduced in `ios_caps_lock_state_tracker.dart`.

@coderabbitai review

rustdesk:
# Code Review (CC): PR #14653 — `fix(ios): preserve capslock state across key events`

## Overview

Replaces the stateless per-event Caps Lock inference on iOS with a cached `IosCapsLockStateTracker`. The old code re-derived Caps Lock from each event's character, which wrongly returned `false` for key-up events and non-letter keys. The new code toggles the cached state on explicit Caps Lock key-down, updates it from inferable character output, and reuses the cache for key-up / non-character events. Cache is cleared in `resetModifiers()` (which fires on `enterOrLeave(false)`, addressing the focus-change concern raised in review).

The change is small, focused, and a clear correctness win. A few notes below.

## Strengths

- **Right shape.** Extracting `IosCapsLockStateTracker` into its own file with a single responsibility makes the logic testable without dragging in `InputModel`. This is also why re-enabling `flutter_test` is justified.
- **Better handling of shift.** The old impl returned `isUpper != shiftPressed`, but iOS produces uppercase for *both* CapsLock and CapsLock+Shift (it documented this as a limitation but did the wrong thing anyway). The new `inferFromCharacter` returns `null` when shift is pressed, preserving cache instead of corrupting it. Genuine improvement.
- **Reset hook is in the right place.** `resetModifiers()` is already called on pointer leave (`input_model.dart:1078`), so cache hygiene is consistent with the existing modifier-clear pattern.
- **Tests target the regression.** The four tests cover the exact edge cases that the original code got wrong (key-up, shifted letter, reset).

## Issues / Suggestions

### 1. Caps Lock toggles on `KeyRepeatEvent` (potential bug)

In `input_model.dart`, `_getIosCapsFromCharacter` passes `isKeyDown: e is KeyDownEvent || e is KeyRepeatEvent`. Inside the tracker:

```dart
if (isKeyDown && logicalKey == LogicalKeyboardKey.capsLock) {
  _capsLock = !_capsLock;
  return _capsLock;
}
```

If the OS ever delivers a `KeyRepeatEvent` for the Caps Lock key (e.g., user holds it down), the cached state will flip on every repeat. In practice iOS doesn't repeat Caps Lock, but the guard is one line:

```dart
if (isKeyDown && logicalKey == LogicalKeyboardKey.capsLock && character == null /* or pass an explicit isRepeat flag */) { ... }
```

Cleaner fix: thread `isRepeat` into `update(...)` and only toggle when `isKeyDown && !isRepeat`.

### 2. `value` getter is dead outside tests

`IosCapsLockStateTracker.value` is only read by tests. Either drop it and have the test re-derive via `update(...)` (already done in 3 of 4 tests), or keep it but mark intent. Minor.

### 3. Lost documentation

The old `_getIosCapsFromCharacterImpl` had a useful block comment explaining the two known limitations (cross-layout mismatches and CapsLock+Shift on iOS). That context is gone. Consider moving it as a doc comment on `IosCapsLockStateTracker` or `inferFromCharacter` — future readers will hit the same questions.

### 4. Redundant `isIOS` guard around reset

```dart
if (isIOS) {
  _iosCapsLockTracker.reset();
}
```

`reset()` is a one-line bool assignment; the guard is harmless but noisy. Drop it for symmetry with the modifier reset above it.

### 5. Initial-state correctness

Cache starts at `false`. If the user enters a remote session with Caps Lock already on, the host will see lowercase until the first non-shifted letter event corrects it. Acceptable trade-off (and unfixable without a reliable iOS API), but worth a comment so future maintainers don't try to "fix" it.

### 6. Test coverage gaps

The added tests cover the regressions, but a couple obvious cases are missing:

- Toggle off: two `capsLock` key-down events should land cache back at `false`.
- Lowercase inference: typing `'a'` (no shift) should set cache to `false`.
- Repeat behavior (related to #1).

## Risks

- **Low blast radius:** all changes are iOS-gated.
- **Pubspec change:** re-enabling `flutter_test` was commented out in the original — worth a quick check with maintainers on *why* it was disabled (CI build size? release builds?). If it was just stale, the change is fine.
- **No native code touched**, no protocol changes, no behavior change on other platforms.

## Verdict

Approve with minor changes. Worth addressing #1 (repeat-event toggle) before merge; the rest are nits.

fufesou:
@s1korrrr 

I am still not fully convinced about the practical value of introducing a cached iOS Caps Lock state here.

The previous implementation did not maintain any Caps Lock state. It inferred iosCapsLock per event from character, and when the character could not be inferred, it effectively sent false for that event.

The new tracker adds a persistent state, which may help in ambiguous cases. However, it also adds complexity and may introduce state-related bugs.

1. Which specific user-visible scenario does the cached state fix?
    - Caps Lock key down with no character?
    - Key up events?
    - Non-character keys?
    - CapsLock + Shift + letter?
    - Something else?

2. On the controlled side, Caps Lock appears to be applied only temporarily while handling an input event, then restored afterward. Also, lockModes CapsLock is only added for letter keys. Given that, how do key-up or non-character events benefit from preserving the cached Caps Lock state?

3. I tested long-pressing keys on an iPad and did not observe KeyRepeatEvent, so the Caps Lock repeat-toggle risk seems theoretical in this environment. Have you observed repeated events in your setup?

4. If every normal letter key-down provides a reliable character, the previous per-event inference seems simpler and sufficient for the main path. What real case requires carrying Caps Lock state across events?

I am not saying the tracker is necessarily wrong.

Still, I would prefer to see a concrete event sequence or reproduction that demonstrates why the added state is needed and why the simpler per-event inference is insufficient.

PR #14671, opened 2026-04-02
fix(ipc): harden local IPC authorization and portable-service bootstrap flow

fufesou:
### IPC auth consolidation + tighter entry points
- Moved IPC security logic into dedicated modules:
  - `src/ipc/auth.rs`: peer identity, session, and executable path checks
  - `src/ipc/fs.rs`: secure handling of IPC parent directories and pid files
- Added stricter checks for service-scoped IPC on Linux/macOS:
  - peer UID must be root or the actual user running the process
  - peer executable path must match the current binary
- The protected `_service` channel (Linux/macOS) now only accepts `SyncConfig` messages — everything else gets rejected

### Unix IPC filesystem hardening
- Reworked how IPC directories and pid files are handled:
  - operations now work with file descriptors and refuse to follow symlinks
  - reduces TOCTOU and path race risks
  - checks owner/permissions and attempts to repair if something's off
- Stale socket/pid cleanup now works against the opened parent inode instead of just path strings
- IPC socket mode `0o0666`. The listener accepts local connections, but authorization happens at accept time before message handling. _service is also limited to SyncConfig. The code documents this near the listener setup and allowlist.

### Windows portable-service hardening
- Added one-time token handshake for portable-service IPC:
  - client sends `AuthToken`, server responds with `AuthResult`
- Shared memory names are now randomized per run; args propagate correctly across elevation and run-as-system paths
- Applied explicit ACLs to shared memory directories and files
- Added trust checks for the logon helper — validates both path and content hash

### Use `SetEntriesInAclW` and `SetNamedSecurityInfoW` instead of `icacls`

  - Protected DACL (no inheritance) + full ACL rebuild
   This requires `PROTECTED_DACL_SECURITY_INFORMATION` and rebuilding the ACL from scratch, not appending to an existing one.
    The current implementation does this in `src/platform/windows/acl.rs:451` and `478`. `icacls` cannot guarantee the same semantics in a stable, programmatic way.

  - Blocking reparse points per node during recursion
  Each target node must be checked, and links or reparse points rejected before ACLs are applied to unintended objects. This is handled in `src/platform/windows/acl.rs:145` and `357`.
  `icacls` in recursive mode has no equivalent per‑node safety check.

  - Structured error codes (Win32) for security decisions
  API‑level error codes are needed for deterministic handling and audit logging. The implementation returns `win32_error` directly (`src/platform/windows/acl.rs:66` and `125`).
  `icacls` primarily outputs text, which is not stable enough for security‑critical paths.

  - Automated verification at the same level of detail
  Unit tests read back the DACL to verify policy enforcement, including recursive grants, reparse point rejection, type checking, and more (`src/platform/windows/acl.rs:742` and `807`).
  An external command pipeline based on `icacls` cannot achieve the same level of repeatable, structured assertions.

## Security impact
- Harder for unauthorized local processes to talk to sensitive IPC channels
- Portable-service startup goes from "reachable" to "reachable + identifiable + one-time token gated"
- Raises the bar for local symlink and path-race attacks around IPC artifacts

## Tests

- Work on the controlled side (Win7, Win11, Linux, macOS, Android, AppImage, flatpak)
  - [x] cm buttons
  - [x] text chat
- [x] Installed version change password, switch between logon sessions (Win7, Win11, Linux, macOS)
- Win11 portable service, flutter/sciter
  - [x] Local elevation
  - [x] Remote request
  - [x] Remote username/password request
- IPC connection request rejection
  - [x] Rejected if user session is incorrect
  - [x] Rejected if executable path does not match

## Summary by CodeRabbit

* **New Features**
  * Portable-service: per-launch shared-memory names, one-time token handshake, startup watchdog, and improved client/server handshake flows.
  * IPC: service-scoped connect APIs and UID-scoped socket support for user-targeted connections.

* **Security**
  * Stronger accept-time IPC authorization and executable verification; tightened Windows ACLs and hardened socket/PID/shared-memory handling.

* **Reliability**
  * Conditional cleanup of stale IPC artifacts, improved reconnect/retry and startup handling, active-user caching for faster session lookups.


fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

rustdesk:
@codex review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

rustdesk:
# Review

The IPC hardening leaves root-to-user connection-manager paths unmigrated for key Linux/macOS service scenarios, causing the service to connect to the wrong per-UID socket directory. These are blocking regressions for installed remote-control flows.

## Findings

### [P1] Resolve the CM uid outside headless mode

`/Users/hzhou/rustdesk/src/server/connection.rs:4914-4919`

On Linux when the service process is root and the desktop is not headless, `user` stays `None`, but `run_as_user(..., None, ...)` still launches `--cm` as the active desktop user. With the paired hbb_common per-UID IPC paths, deriving `cm_uid` only from `user` means the retry loop falls back to `ipc::connect("_cm")` as uid 0 and never reaches `/tmp/{app}-{active_uid}/ipc_cm`, so normal installed-service sessions fail to attach to the connection manager.

### [P1] Avoid UID-scoping macOS IPC before caller migration

`/Users/hzhou/rustdesk/libs/hbb_common/src/config.rs:849-853`

Including macOS in the new per-euid IPC parent path breaks existing root-to-user IPC flows: RustDesk's macOS service still launches `--cm` with `launchctl asuser <active uid>`, while its reconnects call `ipc::connect("_cm")` as root. After this change the service looks in `/tmp/{app}-0` and the CM listens in `/tmp/{app}-{active_uid}`, so installed macOS service sessions cannot connect to CM unless the macOS callers get a uid-targeted connect path too.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4315078529

> [P1] Resolve the CM uid outside headless mode

Incorrect.

`--server` is run as the user process in non-headless mode.
`--cm` is also run as the user process.

> [P1] Avoid UID-scoping macOS IPC before caller migration

Incorrect.

`--cm` is started by `--server`, not the "macOS service".

Both processes run in the user scope.

rustdesk:
• The portable-service bootstrap path can reject valid SYSTEM children on common partial identity failures, and preexisting shmem parent directories bypass the intended ACL/reparse validation. These are
  regressions in a security-critical startup path.

  Full review comments:

  - [P1] Treat partial identity failures as unavailable — /Users/hzhou/rustdesk/src/ipc/auth.rs:653-654
    On Windows portable-service startup, ProcessIdToSessionId can succeed while is_process_running_as_system returns None (for example when OpenProcessToken on the SYSTEM child is denied). Because the
    fallback only runs when the session id is also None, this path rejects the legitimate SYSTEM child before the ACL + one-time-token handshake, so portable service startup can fail on hardened/non-elevated
    hosts.
  - [P2] Validate preexisting shmem parent directories — /Users/hzhou/rustdesk/src/server/portable_service.rs:329-331
    When the portable-service shmem directory already exists and the controller is not SYSTEM (normal before elevation), this condition skips set_path_permission_for_portable_service_shmem_dir, which is also
    the only reparse/type/ACL validation for the parent. A local user can pre-create or leave ProgramData\...\portable_service_shmem with broad ACLs/a junction, causing the token-bearing shmem flink to be
    created under an untrusted parent where another account can observe/delete it before the SYSTEM process opens it.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4317946965

> [P1] Treat partial identity failures as unavailable

A better check is used now. And the comments are added.

> [P2] Validate preexisting shmem parent directories

Added.

21pages:
Tested Windows portable/installed, Linux installed/AppImage/Flatpak, macOS portable/installed, and Android.

  Key observations:
  - Linux installed switch-user behavior is better in this PR: each user has its own `--cm` process, so two `--cm` processes are observed. In release, `--cm` remains in the original process.
  - macOS installed:
    - Logout + switch user, reboot + login from lock screen, and lock screen without switching user work.
    - Lock screen switch A -> B (both pr and nightly not work):
      After switching to user B, there are two `--server`, and the screen randomly connects to either User A or User B. The mouse can click normally on the desktop. Connection may fail sometimes.

rustdesk:
• The patch still has regressions in the Windows portable-service startup hardening: unsafe preexisting shmem parents can be accepted, and in-progress elevated startups can be mistaken for stale and
  restarted. These can break or weaken a security-critical bootstrap path.

  Full review comments:

  - [P2] Validate existing shmem parent ACLs — /Users/hzhou/rustdesk/src/server/portable_service.rs:336-336
    When portable_service_shmem already exists and the controller is not SYSTEM, this path only checks that the directory is not a reparse point; it accepts whatever DACL was already there. If another local
    account pre-creates this directory under the user-accessible app folder with broad inherited rights, the randomized shared-memory flink is created under a parent that other accounts can delete or replace
    during SYSTEM bootstrap, causing the child to open a missing/untrusted shmem object. Existing parents should have the expected protected DACL or fail closed before use.
  - [P2] Track elevated portable-service bootstrap stages — /Users/hzhou/rustdesk/src/server/portable_service.rs:850-851
    This stale-start check only looks for a process whose first argument is --portable-service, but the bootstrap quickly hands off to --elevate and then --run-as-system while STARTING is still true. A
    second elevation/start request during that window can reset STARTING, clear the shared memory/token, and launch another bootstrap, causing the original SYSTEM child to fail its handshake or multiple
    starts to race. Please also recognize the propagated shmem/elevation stages before treating STARTING as stale.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4394474753

These look more like additional hardening opportunities than practical regressions in the normal elevation flow.

1. For the preexisting `portable_service_shmem` parent directory case:

- The normal path creates the parent directory and immediately applies the portable-service shmem directory ACL.
- The shmem flink itself is also hardened right after creation. If applying the file ACL fails, the newly created shmem handle is dropped, cleanup is attempted, and startup fails.
- The SYSTEM stage also goes through `SharedMemory::flink()` before opening the existing shmem. Since that process is SYSTEM, it reapplies the parent directory ACL before `open_existing()` continues.

So the parent directory is not left with broad permissions for the whole bootstrap lifetime.

The remaining gap is limited to a preexisting parent directory with unexpected ACLs, before the SYSTEM stage gets a chance to repair it. In that narrow window, another local user may be able to interfere with the flink path and cause startup/handshake failure.

That looks like a local DoS or bootstrap reliability issue, not a stable privilege-boundary bypass. The token handshake and the per-flink ACL both reduce the practical impact.

2. For the `STARTING` stale check:

- A normal portable elevation is single-shot: the controller creates shmem, writes the one-time IPC token, launches `--portable-service`, and the shmem argument is propagated through `--elevate` and `--run-as-system`.
- The SYSTEM child then opens the same shmem, reads the token, connects back to `_portable_service`, and completes the IPC handshake.
- This path is stable as long as a second start request is not injected while the first bootstrap is still between `STARTING=true` and `RUNNING=true`.

The stale check currently only looks for a process whose first argument is `--portable-service`, so it can miss the propagated `--elevate` / `--run-as-system` stages. That is a real race if duplicate elevation requests are sent during the bootstrap window. However, in the intended product flow, elevation is normally triggered once: either by the controlled side requesting portable elevation, or by the controller sending a direct/logon elevation request. The UI/protocol path does not normally issue repeated start requests during the same bootstrap.

The practical impact of the race is also limited: a duplicate request can clear the old runtime shmem/token and cause the first SYSTEM child to fail its handshake, possibly resulting in failed startup, retry, or repeated UAC. It does not appear to create a successful unauthorized elevation path.

-----------------------------------

Given that, both items are valid hardening observations, but they do not look like blocking P2 regressions for the normal portable-service elevation path.

They can be addressed as follow-up robustness improvements if the goal is to make the bootstrap path tolerate malicious local interference or duplicate scripted elevation requests.

rustdesk:
# Code Review: PR #14671 — fix(ipc): harden local IPC authorization and portable-service bootstrap flow

**Author:** fufesou • **Branch:** `fix/harden-ipc` • **Size:** ~4,300 LOC added across 13 files
**URL:** https://github.com/rustdesk/rustdesk/pull/14671

## Overview

A wide-ranging defense-in-depth pass on local IPC. Three themes:

1. **Accept-time authorization** for service-scoped IPC channels: SO_PEERCRED/`getpeereid` on Unix, `GetNamedPipeClientProcessId` + token introspection on Windows, plus an executable-path/SHA-256 cross-check against the current binary.
2. **Filesystem hardening** for the Linux/macOS IPC parent dir and PID file via fd-bound `O_NOFOLLOW`/`fstat`/`fchmod`/`unlinkat` sequences (TOCTOU-resistant).
3. **Portable-service bootstrap** on Windows: per-launch randomized shared-memory name, one-time 64-hex-char token handshake, programmatic ACL hardening (`SetEntriesInAclW` + `SetNamedSecurityInfoW`) replacing `icacls` shell-out, reparse-point rejection, and a startup watchdog.

The code is well-commented, defensive, and includes ~30 new unit tests covering the policy decisions and Windows ACL invariants. Overall the change is a clear net win for local-attack-surface posture.

---

## Strengths

- **Right primitives.** Replacing path-based PID/socket cleanup with `unlinkat`-via-`O_NOFOLLOW`-opened parent FD genuinely closes a TOCTOU class. The `test_scrub_preexisting_ipc_parent_entries_should_bind_to_opened_inode_not_path` test demonstrates the property well.
- **Programmatic ACL.** `acl.rs` correctly uses `PROTECTED_DACL_SECURITY_INFORMATION` to break inheritance and rebuild from scratch, with structured Win32 error codes — all goals stated in the PR description that `icacls` cannot deliver.
- **Token gating is one-shot.** `consume_runtime_ipc_token_if_match` clears both the in-process slot and the shmem-resident copy on success; `restore_runtime_ipc_token_after_failed_handshake` carefully checks generation before restoring — replay and stale-handshake races are handled.
- **CSPRNG-backed token.** `hbb_common::rand::random::<u64>()` resolves to a thread-local ChaCha CSPRNG; 64-byte tokens are sufficient.
- **Throttled rejection logs.** `UnauthorizedIpcLogThrottle` (5s window with suppressed-count tail) avoids log flooding while still surfacing the rejected-peer fields for forensics.
- **Fail-closed on critical bootstrap paths.** `_portable_service` listener creation returns an error if SDDL construction fails (`src/ipc.rs:268`); `elevate_or_run_as_system` exits if the shmem arg is malformed.

---

## Issues / suggestions

### 1. Linux active-uid cache may return stale identity to authorization (`src/platform/linux.rs:2643-2674`)

`get_active_userid()` and `get_active_username()` now consult a cache that's refreshed only inside `start_os_service`'s loop (`src/platform/linux.rs:2636`). `active_uid_strict()` in `src/ipc/auth.rs:781` calls `get_active_userid()` and is the basis for `is_allowed_service_peer_uid` decisions.

Risk: between user sessions switching and the next loop iteration refreshing the cache, the previous user's UID can still authorize on protected service-scoped IPC. The window is small (one loop tick) but it's a real authorization regression vs. fresh lookup. Recommend either:
- Bypass the cache for authorization-path callers (route them to `get_values_of_seat0` directly), or
- Drive cache invalidation off seat0 changes rather than the loop's polling cadence.

### 2. Windows `_cm` and other non-service pipes silently lose `Everyone` access (`src/ipc/auth.rs:634`)

`should_allow_everyone_create_on_windows` now returns true only for `""` and service postfixes. Before this PR, `SecurityAttributes::allow_everyone_create()` was unconditional. After: `_cm`, `_file_transfer`, `_terminal_*` get `SecurityAttributes::empty()`, which produces a default DACL based on the creator's token.

This is mostly a hardening, but it's a behavior change with cross-session implications. The PR test plan covers Win7/Win11 installed/portable, so primary flows are validated, but the change deserves an explicit mention in release notes — if a downstream packager runs the cm process under a different user than the connecting process, they'll see new connection failures with no obvious cause.

### 3. Dead/confused comment in `authorize_windows_portable_service_ipc_connection` (`src/ipc/auth.rs:1287`)

```rust
// Don't use `peer_pid.is_some() && peer_is_system.is_none() && peer_is_system.is_none();` here.
let identity_unavailable = peer_pid.is_some() && peer_is_system.is_none();
```

The "don't use" comment shows a *triple* conjunction that's tautologically equal to the actual line below it. Either it's a leftover from refactoring or it's trying to communicate a subtle distinction that isn't visible. Drop it or rewrite to explain what was *actually* rejected.

### 4. `set_path_permission` recursion has a small TOCTOU gap (`src/platform/windows/acl.rs:3230-3279`)

The recursive walk does `fs::symlink_metadata` to reject reparse points per node (good), but the subsequent `apply_grant_sid_allow_ace_to_path` calls `GetNamedSecurityInfoW`/`SetNamedSecurityInfoW` on the *path*, which follow reparse points. An attacker with write access to the helper directory could swap in a junction between the metadata check and the ACL apply.

Mitigation in practice: the helper directory itself is ACL'd to current-user + SYSTEM, so an unprivileged attacker can't perform the swap. Still, prefer opening with `CreateFileW(FILE_FLAG_OPEN_REPARSE_POINT)` and using the handle-based `SetSecurityInfo` to fully close the gap.

### 5. `test_ensure_secure_ipc_parent_dir_creates_parent_with_expected_mode` is umask-dependent (`src/ipc/fs.rs:2373-2401`)

The test asserts `mode == 0o0700` and `should_scrub == false` after `mkdir(0o0700)`. With a restrictive umask (e.g. `0o0700`), the kernel will produce `0o0000`, the function will then `fchmod` to repair, return `should_scrub == true`, and the test fails. CI generally runs umask 022 so this is unlikely to bite, but it's a brittle assertion — either set umask explicitly in the test or assert that the final mode equals expected (which it will after the repair).

### 6. `start_portable_service` lock acquisition in `start_result` closure (`src/server/portable_service.rs:4895-5096`)

The closure does `clear_runtime_shmem_state()` (acquires IPC_RUNTIME_TOKEN → SHMEM → SHMEM_RUNTIME_NAME), then re-acquires `SHMEM` and writes to `SHMEM_RUNTIME_NAME` while holding it, then later acquires `IPC_RUNTIME_TOKEN`. The orderings are consistent (IPC_RUNTIME_TOKEN < SHMEM < SHMEM_RUNTIME_NAME), but you're juggling three globals with manual `drop()` interleaving across ~150 lines. Worth extracting a `RuntimeShmemHandle { token, shmem, name }` struct guarded by a single mutex; the resulting code would be smaller and impossible to misorder.

### 7. Forced-exit watchdog 3s budget (`src/server/portable_service.rs:4619-4639`)

`run_exit_check` force-exits 3s after `EXIT` flips, gated by `FORCE_EXIT_ARMED`. If a worker is doing legitimate teardown (e.g., a slow shmem flink unlink retry, or video pipeline shutdown), 3s can be tight. The comment correctly labels this a last resort, but consider making the timeout configurable or aligning it with the 20×200ms = 4s flink retry budget downstream — currently the watchdog can fire *before* the flink retry loop completes.

### 8. `connect_to_user_session` is now Windows-only (`src/ipc.rs:473-479`)

Previously it lacked a cfg gate; now it's `#[cfg(windows)]`. Make sure no non-Windows caller still references it (a quick `rg connect_to_user_session` would confirm) — the diff doesn't show a use site update, so either there were none or the build will complain.

### 9. Hash recompute on every accept (`src/ipc/auth.rs:1011-1043`)

`portable_service_helper_is_trusted` SHA-256s both peer and current binaries on every connection. For the portable service helper this is a one-shot at startup, so it's fine, but if this code path ever fans out to higher-rate channels, cache the current binary's hash in a `OnceLock`.

### 10. Minor

- The `// TODO:` blocks in `start_ipc` (`src/server/connection.rs:4090, 4103`) about `_rx_desktop_ready` hot-loop and `id -u --` hardening are reasonable to defer, but please file follow-up issues so they don't fade.
- `parity-tokio-ipc` bumped from `0.7.3-5` → `0.7.3-6` and `hbb_common` to a new commit — both opaque to this review; reviewers downstream should verify they don't carry unrelated changes.

---

## Test coverage

Strong. The new tests exercise:
- Peer-UID policy matrix (linux/macos and windows session-scoped + portable-service variants)
- Symlink rejection on parent dir and PID file
- FD-bound scrub correctness under path swap (the actual TOCTOU property)
- ACL policy assertions against the live DACL (system/admin/current-user/auth-users grants and Everyone/Users denials, plus `SE_DACL_PROTECTED`)
- Reparse-point rejection on each entry point
- Helper trust requires content match (SHA-256), not just path match
- Frame-length bounds in shared memory

What's missing:
- No test exercises the `identity_unavailable` fall-through in `authorize_windows_portable_service_ipc_connection`
- No tests around the lock-ordering / state-machine of `start_portable_service` (legitimately hard to test, but a regression here would be subtle)
- macOS service-IPC `gui+service` binary allowance has no unit test for the dual-binary case

---

## Security assessment

Net positive. The change closes the local-symlink/path-race attack class against IPC artifacts on Linux/macOS, materially raises the bar for unauthorized local processes on Windows portable-service bootstrap (random shmem name + ACL + one-time token + executable identity), and replaces shell-invoked `icacls` with a programmatic, error-propagating ACL pipeline that the unit tests can actually verify.

The remaining residual risks (#1, #4) are narrow and well-mitigated by the surrounding ACL/token defenses; #2 is a deliberate hardening with downstream-compat implications worth flagging in the changelog.

## Recommendation

**Approve with minor changes** — addressing #1 (active-uid cache freshness on the auth path), #3 (the dead comment), and ideally #5 (umask-robust test) before merge. The rest are good follow-ups.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4395497333

1. Linux active-uid cache may return stale identity to authorization (src/platform/linux.rs:2643-2674)
3. Dead/confused comment in authorize_windows_portable_service_ipc_connection (src/ipc/auth.rs:1287)
5. test_ensure_secure_ipc_parent_dir_creates_parent_with_expected_mode is umask-dependent (src/ipc/fs.rs:2373-2401)

Addressed in 46e2fde

rustdesk:
# Code Review — PR #14671: `fix(ipc): harden local IPC authorization and portable-service bootstrap flow`

**Author:** fufesou · **Status:** Open · **Scope:** ~4,300 LOC, 13 files

## Overview

A defense-in-depth hardening of RustDesk's local IPC across Linux, macOS, and Windows. The PR introduces two new modules — `src/ipc/auth.rs` (peer authorization policy) and `src/ipc/fs.rs` (TOCTOU-resistant filesystem ops) — and rewires `src/ipc.rs` to gate every connection at accept-time. On Windows it also redesigns portable-service bootstrap with an SDDL-restricted named pipe + one-time token handshake. Includes ~30 unit tests.

## What's Good

- **Layered policy.** Peer UID, PID-derived exe path match, Windows session ID, SYSTEM check, file-mode/owner enforcement, and SDDL DACLs combine well — no single check is load-bearing.
- **TOCTOU-aware filesystem ops.** `src/ipc/fs.rs` opens parent dirs with `O_NOFOLLOW|O_DIRECTORY|O_CLOEXEC` and operates on the FD via `fstat/fchmod/fchown/unlinkat`. The path-swap test (`test_scrub_preexisting_ipc_parent_entries_should_bind_to_opened_inode_not_path`) actually proves the property, not just the happy path.
- **Fail-closed defaults.** SDDL build failure on `_portable_service` returns `Err`, socket `set_permissions` failure removes the socket and returns. Missing `peer_uid` (`is_some_and`) rejects.
- **Throttled rejection logging** with a suppressed-count summary — good DoS hygiene given the `0o0666` service socket is world-connectable by design.
- **Comments document *why*.** The "uinput vs `_service` allowlist scope" note in `start()` and the portable-service identity-unavailable rationale are exactly the kind of policy commentary that prevents future drift.

## Issues to Address

### High

1. **Token RNG provenance is unclear (`src/ipc.rs`, `generate_one_time_ipc_token`).**
   Uses `hbb_common::rand::random::<u64>()`. The portable-service privilege boundary depends on this token being unguessable. Confirm the underlying RNG is a CSPRNG (`OsRng` / `ThreadRng`); if it routes to a non-cryptographic generator, replace with `getrandom::getrandom` over 32 bytes. Worth a comment asserting "CSPRNG: <which one>" so future refactors don't regress it.

2. **Token comparison is likely not constant-time.**
   `portable_service_ipc_handshake_as_server` calls `validate_token(&token)`. If the implementation uses `==` on `String`, it's vulnerable to a (local) timing side channel. Use `subtle::ConstantTimeEq` or a manual byte-wise compare with no early-exit. The handshake design implies attacker-as-local-user is in scope, so this matters.

3. **`identity_unavailable` short-circuit allows the connection (`auth.rs`, `authorize_windows_portable_service_ipc_connection`).**
   When `peer_pid.is_some() && peer_is_system.is_none()`, the function logs and returns `true` regardless of whether `ensure_peer_executable_matches_current_by_pid_opt` succeeded. The exe-mismatch case still proceeds — the warning and the success path differ only in log text. Consider returning `false` when the exe verification fails *and* identity is unavailable; that's the strictest you can get without breaking the documented bootstrap edge case.

### Medium

4. **`Ok(None)` doesn't break on Windows / non-service Unix postfixes (`src/ipc.rs`, `start()`).**
   New code only breaks on `Ok(None)` for the `_service` postfix on Linux/macOS. On Windows and other postfixes, a closed stream falls through and the loop calls `next().await` again. This is also true of the prior `_ => {}`, so it's pre-existing — but since you're already restructuring this branch, fix it for all platforms (`Ok(None) => break;` outside the cfg-gated block).

5. **`#[inline]` overuse.** Many large functions (`ensure_secure_ipc_parent_dir`, `write_pid_file`, `portable_service_ipc_handshake_as_server`, `authorize_*`) carry `#[inline]`. These are not hot paths and the attribute risks code bloat without measurable benefit. Drop `#[inline]` on anything over ~5 LOC unless profiling shows a win.

6. **`portable_service_helper_is_trusted` hashes both binaries on every connection.** SHA-256 over two PE files per connection adds disk I/O. Throttling on the listener side mitigates abuse, but consider caching `current_exe`'s hash in a `OnceLock` since it can't change for the process lifetime.

7. **Exe-path normalization on Windows is shallow.** `executable_paths_match` lowercases and strips `\\?\`, but doesn't account for `\??\` NT-namespace paths or 8.3 short names. `fs::canonicalize` upstream usually normalizes these, but if a peer process is launched via an 8.3 path on a volume with short names enabled, canonicalize may still hand back the long form and you're fine — worth a comment asserting that assumption.

### Low / Style

- **`peer_uid` reads from `inner.get_ref()` without locking.** `parity-tokio-ipc` exposes the raw fd; concurrent close from another task could race the `getsockopt`. Risk is low (the fd is owned by `Framed` and won't be closed until `inner` drops), but a sentence in a comment confirming the lifetime would help future refactors.
- **`active_uid_strict` on Linux logs `warn!` on parse failure.** Given a misconfigured seat could fire this on every connection, fold it into the throttled-log path or downgrade to `debug!` — the rejection itself is already logged.
- **Repeated `cfg(any(target_os = "linux", target_os = "macos", target_os = "windows"))`.** Consider `#[cfg(unix_or_windows)]` via a Cargo feature or a tiny `#[cfg(...)] mod` indirection — the current attributes are noisy.
- **`write_pid_file` byte construction** is `to_string()` then `as_bytes()` — fine, but `format!("{}\n", pid).into_bytes()` (or similar) avoids one allocation step. Minor.
- **`scrub_preexisting_ipc_parent_entries`** only removes `ipc{postfix}` and `ipc{postfix}.pid`. If a future change introduces another sibling artifact (e.g. a lock file), this will silently leave it behind. Worth a comment listing the canonical artifact set.

## Test Coverage

Strong on policy and TOCTOU primitives:
- UID matrix (`test_service_peer_uid_policy`)
- Windows session policy (`test_windows_server_peer_policy`, `test_windows_portable_service_peer_policy`)
- Path normalization (`test_executable_paths_match_windows_normalization`)
- Symlink rejection in `write_pid_file` and `ensure_secure_ipc_parent_dir`
- Inode-bound scrub vs. path-swap attack
- Helper trust requires content match, not just path match

Gaps worth filling:
- No test exercises the `portable_service_ipc_handshake_as_server` reject path (wrong token, missing token, type mismatch).
- No integration test asserts that connecting from a non-active UID is rejected end-to-end on Linux (only the policy function is unit-tested).
- The `Ok(None)` loop-exit fix (if applied) needs a test that drops the stream and asserts the task terminates.

## Risks

- **Behavioral regressions on edge installs.** Service IPC sockets switching from `0o0777` to `0o0666` and pid files from `0o0777` to `0o0600` is correct, but any third-party tooling that read these files as a non-owner will break. Worth calling out in the changelog.
- **AppImage / flatpak / portable installs** have historically had quirky parent-dir ownership. The `fchown` adoption path under root mitigates this, but distros where the service runs as a non-root user with a foreign-owned `/tmp/rustdesk-*` parent will get a hard `PermissionDenied`. Confirmed in test matrix per PR description, but worth re-verifying.
- **`hbb_common` submodule bump.** Reviewers should look at that change separately — security-sensitive code shouldn't piggyback on an opaque submodule pointer move.

## Verdict

The architecture is right and the test coverage is unusually thorough for an IPC hardening change. Address #1 (RNG provenance) and #2 (constant-time token compare) before merge — those are the only items I'd consider blocking. #3, #4 are worth a follow-up commit on this branch. The rest are polish.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4402857571

1. Token RNG provenance

Confirmed. Comment is added in 4365b1b.

`hbb_common` re-exports `rand 0.8`, and `rand::random()` uses `ThreadRng`, which is seeded from `OsRng`. Added a comment at the token generation site to make the CSPRNG dependency explicit.

2. Token comparison

Fixed in 4365b1b.

Replaced the direct string comparison with a fixed-length byte-wise comparison without early exit. I kept it dependency-free instead of adding `subtle` just for this check.

3. `identity_unavailable`

Kept this behavior intentionally.

This path handles hosts where SYSTEM identity lookup is unavailable during portable-service bootstrap. Executable verification is still logged, but the final boundary remains the restricted pipe ACL plus the one-time token handshake. Failing closed here can break valid bootstrap flows.

4. `Ok(None)` handling

Kept the service-only handling.

`Ok(None)` here means a malformed or unexpected frame, not a closed stream. A closed/reset stream is returned as `Err("reset by the peer")`. Master also ignores `Ok(None)`, so changing all postfixes to break would be a behavior change. `_service` still fails closed.

5. `#[inline]`

Leaving this out of this security fix.

Removing them is a cleanup-only change and does not affect the IPC authorization behavior.

6. Helper hash caching

Leaving this uncached for now.

7. Windows exe-path normalization

Added a comment for this.

Callers canonicalize paths before matching; the local normalization only handles remaining Win32 spelling differences such as slash direction, `\\?\`, and case.

rustdesk:
# Code Review: rustdesk/rustdesk#14671 — IPC Hardening

## Overview
Layered local-IPC hardening across Linux/macOS/Windows: dedicated `src/ipc/auth.rs` (peer UID/PID/session/exe-hash checks) and `src/ipc/fs.rs` (FD-bound, `O_NOFOLLOW`-based parent dir + PID file ops). Adds a one-time 64-char token handshake for the Windows portable-service bootstrap, an explicit SDDL DACL replacing shell `icacls`, and a `SyncConfig`-only message allowlist on the world-mode `_service` socket. ~4.3k LOC, ~30 unit tests, 27 commits, depends on a patched `parity-tokio-ipc` (0.7.3-5 → 0.7.3-6, fork commit) plus an `hbb_common` submodule bump.

The threat model and chosen primitives are right. Most of the comments below are about a few load-bearing details where the implementation should be tightened before merge.

## Blocking / High-confidence issues

- **Token RNG provenance must be cryptographic.** `generate_one_time_ipc_token` uses `hbb_common::rand::random::<u64>()`. The 64-hex-char (256-bit) format is fine, but only if the source is a CSPRNG. If `hbb_common::rand` re-exports `rand::random` backed by `ThreadRng` that's acceptable; if it's seeded from a non-CSPRNG (or worse, a deterministic seed), the privilege boundary collapses. **Replace with `getrandom::getrandom(&mut [u8; 32])` directly** — removes the indirection and makes the guarantee local to this function. (Also, `debug_assert_eq!` on the length invariant is a release no-op; either `const _: () = assert!(...)` or just write a single `getrandom` call.)

- **Constant-time token comparison.** `validate_token: F: FnMut(&str) -> bool` delegates the comparison to the caller, which almost certainly does `==` on `&str`. Over a local pipe with 3s budget the timing channel is narrow but not zero, and constant-time is essentially free. Either change the signature so the server takes the *expected* token and does the compare with `subtle::ConstantTimeEq`/byte-wise no-early-exit internally, or document at the call sites that callers must use a constant-time primitive. Right now this is silently caller's problem.

- **Identity-unavailable bootstrap fallback in `authorize_windows_portable_service_ipc_connection`.** Per the open review thread on `src/ipc/auth.rs` around line 1287, the path returns `true` when peer identity is unobtainable even though exe verification failed. Bootstrap convenience is not a reason to fail open on a SYSTEM-only channel. Make it fail-closed: if exe match fails, return `false` regardless of identity availability. The token handshake is the right compensating control for the genuine "we don't know who you are yet" case.

## Should-fix

- **Linux peer-exe verification race.** macOS uses inode comparison; Windows uses `QueryFullProcessImageNameW` + SHA-256. On Linux the typical implementation reads `/proc/<pid>/exe`, which is a TOCTOU between getting the peer PID and reading the symlink (peer can `execve` in between). Either `open("/proc/<pid>/exe", O_RDONLY)` early and hash via the FD, or pin the PID's exe inode at peer-cred time. Worth confirming the current Linux path does one of these — I couldn't see it in the excerpt.

- **Socket mode 0o0666 is load-bearing.** Any local user can `connect()` to `_service` and `_portable_service`. That's intentional and gated by the per-connection authorization, but it means any bug in `is_allowed_service_peer_uid`, `executable_paths_match`, or the throttle/reject path becomes a local-privilege issue. Worth (a) a top-of-file comment in `auth.rs` stating "0o0666 is intentional; security depends entirely on the gates below" and (b) a fuzz/property test that the gate is reached *before* any side-effecting parsing of message payloads. The current `SyncConfig`-only allowlist is good defense-in-depth here.

- **Hash on every connection.** `portable_service_helper_is_trusted` re-hashes both peer and current exe per accept. Cache the current-exe hash in a `OnceLock<[u8; 32]>` keyed by `(path, mtime, len)` (or just path if you're confident the binary doesn't get replaced live). Per-peer hash still needed.

- **Cross-platform `Ok(None)` handling in `start()`.** Reviewer flagged that the closed-stream break is gated on `_service` postfix + Linux/macOS only. Either every postfix on every platform should break on closed stream (likely the safe behavior) or there's a reason it differs that's worth a comment. Right now this looks like an accidental asymmetry.

- **`terminal_count_candidate_uids` includes UID 0 when running as root.** Counting "terminal sessions" against root seems off — root doesn't typically have a console session in the sense this is measuring. If the goal is "all UIDs that might own a session," resolving from `who`/`utmp`/logind would be more correct than `[active_uid, 0]`.

## Nits

- **SDDL is built via `format!`.** `D:P(A;;GA;;;SY)(A;;GA;;;{user_sid})` is fine *if* `current_process_user_sid_string()` always returns a syntactically valid SID. Add a `debug_assert!` (or proper validation) that the returned string matches `^S-1-…$` so a regression in the SID resolver can't silently produce a permissive descriptor when `from_sddl` happens to accept a malformed input.

- **`#[inline]` overuse.** `ensure_secure_ipc_parent_dir`, `portable_service_ipc_handshake_as_server`, etc. are not hot paths. Drop the `#[inline]` from anything not single-call-site-trivial; let LLVM decide.

- **`io::Error::new(io::ErrorKind::Other, …)` repeated.** Rust 1.74+ has `io::Error::other(msg)`. Mechanical, but reduces noise.

- **`debug_assert_eq!(IPC_TOKEN_LEN % IPC_TOKEN_U64_HEX_LEN, 0)` is a release no-op.** Make it `const _: () = assert!(IPC_TOKEN_LEN % IPC_TOKEN_U64_HEX_LEN == 0);` to fail at compile time.

- **`hbb_common` submodule + patched `parity-tokio-ipc` fork bump.** Two transitive code changes ride along with this PR. Reviewer can't see the fork diff from the PR view; consider including the fork commit's notes in the PR body.

## Tests

The added unit tests cover the right invariants: symlink rejection on PID files, `O_NOFOLLOW` parent open, scrub idempotence/ownership, peer-UID policy, helper-trust SHA. Gaps worth filling:

- An end-to-end "unauthorized peer" test on the Unix `_service` socket: open as a non-allowed UID and assert (a) connection drops, (b) no `Data::SyncConfig` is processed, (c) the throttle counter increments rather than producing per-connection log spam.
- A handshake timeout test: server should reject after `PORTABLE_SERVICE_IPC_HANDSHAKE_TIMEOUT_MS` even when client hangs, and not leak the listener slot.
- A test that verifies `validate_token` is reached *only* after UID/exe gates pass (so a slow handshake from a disallowed peer can't tie up a slot).

## Summary

The architecture is correct: O_NOFOLLOW + FD-bound ops, explicit DACL, exe-hash trust, message allowlist, throttled rejection. The three things that should not merge as-is are: (1) confirm/replace the token RNG with `getrandom`, (2) make token comparison constant-time at the IPC layer rather than at the call site, and (3) close the `identity_unavailable` fail-open path. Everything else is polish.

rustdesk:
• The UID-scoped IPC path change is not fully propagated to the non-headless root service path, which can break CM IPC startup on normal Linux desktop installs.

  Review comment:

  - [P2] Resolve the active-user CM socket for root launches — /Users/hzhou/rustdesk/src/server/connection.rs:5030-5035
    When the Linux service is running as root in a normal non-headless desktop session, run_as_user(args, None, ...) still starts --cm as the active user, but this PR moves non-service IPC sockets to UID-
    scoped paths. Leaving cm_uid as None makes the follow-up reconnect use root's _cm path instead of the spawned user's path, so the service can fail to attach to the connection manager whenever it was not
    already running.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4404425263

Invalid.

`--server` The server process always runs as the user process, not the root process.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4403751257

1. Token RNG provenance

Fixed in 5872441.

Token generation now uses `OsRng` directly and propagates RNG failure. The token is still 32 random bytes encoded as 64 lowercase hex chars.

2. Constant-time token comparison

Fixed in 5872441.

The fixed-length byte-wise comparison now lives in `ipc.rs`, and the portable-service validator uses it. The handshake API also documents that validators must use a constant-time check.

3. `identity_unavailable` fallback

Kept as-is.

This fallback covers hosts where SYSTEM identity lookup is unavailable during bootstrap. The connection still has to pass the restricted pipe ACL and the one-time token handshake. Failing closed here can break valid bootstrap flows.

4. Linux peer-exe verification race

Valid hardening point.

The current Linux check is already behind peer UID / active-user validation, and installed binaries are not normally user-writable. FD/inode pinning would be a stronger defense-in-depth improvement, but it is not a regression introduced here.

5. Socket mode `0o0666`

Intentional.

The listener accepts local connections, but authorization happens at accept time before message handling. `_service` is also limited to `SyncConfig`. The code documents this near the listener setup and allowlist.

6. Hash on every connection

Not changing this now.

This path runs during portable-service bootstrap, not on a hot data path. Caching adds invalidation assumptions for little benefit here.

7. `Ok(None)` handling

Kept service-only.

`Ok(None)` here means a malformed frame, not a closed stream. Closed/reset streams return `Err("reset by the peer")`. Added a comment to make that explicit.

8. `terminal_count_candidate_uids` includes UID 0

Kept.

UID 0 is a fallback IPC owner when the caller is root. Replacing this with logind/utmp would be a larger behavior change.

9. SDDL SID validation

Added a `debug_assert!` for the expected `S-1-*` SID format before building the SDDL string.

10. `#[inline]` cleanup

Done for the larger non-hot functions touched by this PR. Small trivial helpers were left alone.

11. `io::Error::other`

Leaving this as-is. It is a mechanical cleanup and not part of the IPC hardening behavior.

12. Token length invariant

Fixed. Replaced the release no-op `debug_assert_eq!` with a compile-time assertion.

rustdesk:
Architecturally solid and a clear security improvement. The constant-time comparison and CSPRNG handling correctly address prior review feedback. Before merge, I'd want resolved:                           
  
  1. Strict numeric UID parsing on Linux (no empty-string fallback).                                                                                                                                           
  2. macOS bundle peers: signature/team-id check, not just filename match.
  3. RNG failure path explicitly fail-closed and documented.                                                                                                                                                   
  4. CHANGELOG note for the 0o0666 pipe permission change.                                                                                                                                                     
  5. Add a concurrent-stress test for the parent-FD scrub path.                                                                                                                                                
                                                                                                                                                                                                               
  Items 1–3 are blocking from a security standpoint; 4–5 can be follow-ups.

rustdesk:
> [#14671 (comment)](https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4404425263)
> 
> Invalid.
> 
> `--server` The server process always runs as the user process, not the root process.

Codex：  A narrower follow-up may still be worth checking for the root fallback path (`desktop.username == "root"` / login Wayland / headless paths using `start_server(None)`), but the normal non-headless service
  path is not broken by this.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4405391466

1. Linux UID parsing

This should not be a blocking item.

The security-sensitive paths already fail closed on invalid or empty UID.

`active_uid_strict()` returns `None` on parse failure, uinput rejects when the active UID cannot be parsed, and the headless CM path parses the UID before using `connect_for_uid()`. 

2. macOS bundle peers

Team-id verification would be a stronger check, but it should not block this PR.

The `_service` peer is already limited to root or the active user, the executable check is constrained to the same bundle directory and GUI/service binary names, and the channel only accepts `SyncConfig`. This can be a follow-up hardening item.

3. RNG failure

Already fail-closed.

`OsRng` failure is propagated, and token generation happens before shared memory/runtime state is created.

4. CHANGELOG note

There is no repo changelog file in this branch.

PR note is updated.

5. Parent-FD scrub stress test

A stress test should not block this.

The important behavior is that scrub is bound to the opened parent fd, not the path name, and there is already a deterministic path-swap test for that.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4405422141

Checked the Linux service flow again.

For normal desktop sessions, `--server` is started through `run_as_user(Some(desktop.uid, desktop.username))`, so `start_ipc()` runs as the desktop user and the default `_cm` path is already the user-scoped path.

The root fallback cases do not look like the reported regression:

- `desktop.username == "root"`: `--server` and CM both run on the root side, so the `_cm` IPC path matches.
- login Wayland/headless: this is not the normal CM UI path. The headless CM path already resolves the target user and uses `connect_for_uid()`.

So the normal non-headless CM path does not need a change for this item.

rustdesk:
• The patch adds useful IPC hardening, but the portable-service authorization has a fail-open identity-unavailable path and the startup de-duplication misses the actual run-as-system process. These can
  undermine the new security and reliability guarantees.

  Full review comments:

  - [P1] Reject identity-unavailable portable-service peers — /Users/hzhou/rustdesk/src/ipc/auth.rs:719-719
    When peer SYSTEM identity cannot be resolved, this branch still accepts the connection and proceeds to the token handshake even if executable verification failed. The _portable_service pipe and shared-
    memory ACLs both grant the current user access, and the token is stored in that shared memory, so a same-user non-SYSTEM peer in an environment where token/process querying fails can pass the ACL+token
    gates without proving it is the SYSTEM portable service.
  - [P2] Check the run-as-system portable service process — /Users/hzhou/rustdesk/src/server/portable_service.rs:850-851
    During normal portable-service startup, --portable-service is only the first bootstrap argument; the elevated/SYSTEM process that actually runs run_portable_service() is relaunched with --run-as-system
    plus the shmem argument. If a second start request arrives while STARTING is true but before RUNNING is set, this check can falsely conclude there is no in-flight service, clear the active shmem/token,
    and launch a competing instance.

rustdesk:
Recommendation

  Approve with changes. The architecture is right and the implementation is meticulous. Before merging:

  1. Confirm (or include in this PR) that the hbb_common bump removes the 0o0777 parent-dir creation in Config::ipc_path(). Otherwise the parent-dir hardening has a bypass.
  2. Add a comment in is_allowed_service_peer_uid explaining the "root always allowed + executable match enforced separately" policy.
  3. Either unify on get_active_userid_fresh() for all Linux authz call sites, or document why the cache is acceptable for _service authz.
  4. Validate that Ok(None) on _service doesn't tear down legit connections under the bumped parity-tokio-ipc.

  Nothing else looks like a blocker. Strong work overall — this is the kind of PR that's easy to get wrong in subtle ways and the author has clearly reasoned about the boundaries.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4412126929

> [P1] Reject identity-unavailable portable-service peers — /Users/hzhou/rustdesk/src/ipc/auth.rs:719-719

Incorrect.

As the comments say:

This fallback is for hosts where SYSTEM identity lookup is not reliable during bootstrap.

In that path, executable verification is only telemetry; it is not the final auth gate. The connection still has to pass the restricted pipe ACL and the one-time token handshake.

Failing closed on executable introspection failure can break valid SYSTEM bootstrap cases.

> [P2] Check the run-as-system portable service process

Keeping this as-is.

This was already discussed in https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4395254512.

The stale check can miss the intermediate `--elevate` / `--run-as-system` phases, but only if a second start is injected while `STARTING=true` and `RUNNING=false`.

The expected product flow is single-shot. The likely impact is failed/retried startup or repeated UAC, not an unauthorized elevation path. Keeping this as follow-up robustness, not a blocker.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4412138735

Comments are added in 9fd3d30.

1. hbb_common parent-dir creation

Confirmed.

The bumped `hbb_common` no longer creates or chmods Linux/macOS IPC parent dirs from `Config::ipc_path()`. Unix parent hardening is done at listener setup.

2. `is_allowed_service_peer_uid` comment

Added a comment.

Root is allowed only at the UID gate; service-scoped peers still go through executable matching before being accepted.

3. active UID cache vs fresh lookup

Added a comment.

`_service` uses the cached active UID for stable config sync with the service loop.

Uinput has a higher-risk input path and uses a fresh active UID lookup in its own authorizer.

4. `Ok(None)` on `_service`

This was already clarified in https://github.com/rustdesk/rustdesk/pull/14671#issuecomment-4403473809. `Ok(None)` here means a full frame failed to deserialize into `Data`; peer close/reset returns `Err("reset by the peer")`. Breaking `_service` on malformed frames is intentional fail-closed behavior.

LucasFarinha92:
1.4.6 installed + EXE 1.4.6 = OK
Service updated to 1.4.7 = problem appears
EXE 1.4.6 or 1.4.7 presents the same problem when the service is 1.4.7
The ID is displayed correctly
The password remains permanently in "Generating..."
The tray GUI shows the password normally
The problem only occurs when opening a second instance (custom EXE)

PR #14681, opened 2026-04-03
Add macOS option-command swap toggle for non-macOS remote sessions

poneding:
#14670 

This adds a new session-level Swap option-command key toggle for macOS clients when connecting to non-macOS peers.
It is separate from the existing Swap control-command key option, and the two toggles are mutually exclusive.

I use RustDesk almost every day, and this is a feature I really want in actual use.

When working from macOS to Windows/Linux machines, swapping Control and Command helps in some cases, but it still does not cover the Option/
  Command habit on macOS. For my workflow, having an Option/Command swap toggle makes remote keyboard use much more natural.

  ## Behavior

  - show Swap option-command key only when:
      - local client is macOS
      - remote peer is not macOS
  - keep it session-level, same as the existing keyboard swap option
  - make it mutually exclusive with Swap control-command key

  ## Implementation notes

  - no new submodule dependency
  - no hbb_common config field added for this
  - the new toggle uses the existing session option storage path
  - keyboard and mouse modifier remapping are both handled

  I reviewed the code carefully and built/tested it locally myself.

  I also verified the toggle interaction logic locally, including mutual exclusion with the existing Swap control-command key option.

## Summary by CodeRabbit

* **New Features**
  * Added a new keyboard toggle option to swap Option and Command keys on macOS systems, available when connecting to a Mac from a non-Mac platform.
  * Added Chinese (Simplified and Traditional) translations for the new toggle.

### DIFF
--- a/flutter/lib/common/widgets/toolbar.dart
+++ b/flutter/lib/common/widgets/toolbar.dart
@@ -853,6 +853,22 @@ List<TToggleMenu> toolbarKeyboardToggles(FFI ffi) {
         child: Text(translate('Swap control-command key'))));
   }
 
+  if (ffiModel.keyboard && isMacOS && pi.platform != kPeerPlatformMacOS) {
+    final option = 'allow_swap_option_command_key';
+    final value =
+        bind.sessionGetToggleOptionSync(sessionId: sessionId, arg: option);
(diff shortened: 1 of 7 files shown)


rustdesk:
@codex review

PR #14682, opened 2026-04-03
add brute-force protection for one-time password

21pages:
# Summary

  Harden brute-force protection for the one-time password by tracking consecutive failed login attempts against the current password, rotating it on the 10th failure, and resetting the counter only after a successful one-time-password authentication.

  # Tested

  - Verified the one-time password does not rotate after 9 consecutive failed login attempts
  - Verified the one-time password rotates on the 10th consecutive failed login attempt
  - Verified a successful one-time-password authentication resets the failure counter
  - Verified a successful permanent-password authentication does not reset the one-time-password failure counter
  - Verified rotating the one-time password resets the tracked state

# Related

- https://www.teamviewer.com/en/global/support/knowledge-base/teamviewer-classic/security/security-features/brute-force-protection

## Summary by CodeRabbit

* **Bug Fixes**
  * Temporary passwords now auto-rotate after 10 consecutive failed validations for the same temporary password.
  * Failure tracking is scoped to the current temporary password and resets when the temporary password changes or after any successful authentication.
  * A warning is logged when rotation occurs, including the failure count and connection IP.
  * Brute-force protection is now integrated into the login flow.

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -2032,8 +2032,10 @@ impl Connection {
             minute: i32,
             failures: i32,
         }
-        static TEMPORARY_PASSWORD_FAILURES: std::sync::LazyLock<Mutex<State>> =
-            std::sync::LazyLock::new(|| Mutex::new(State::default()));
+        lazy_static::lazy_static! {
+            static ref TEMPORARY_PASSWORD_FAILURES: Mutex<State> =
+                Mutex::new(State::default());
+        }
 
         if !password::temporary_enabled() {
             return;
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -2020,9 +2020,11 @@ impl Connection {
         self.validate_password_plain(storage)
     }
 
-    // This is coarse brute-force protection for the temporary password.
-    // We track consecutive full login failures against the current temporary password budget
-    // across all sources. Only a temporary-password success clears this state.
+    // This is coarse brute-force protection for the current temporary password value.
+    // We only care whether the active temporary password itself was presented correctly,
+    // not whether later authorization steps succeed. A successful temporary-password
+    // match clears this state immediately, and the counter also resets whenever the
+    // temporary password changes or is rotated.
     fn check_update_temporary_password(&self, temporary_password_success: bool) {
         const MAX_CONSECUTIVE_FAILURES: i32 = 10;
         #[derive(Default)]
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -2041,13 +2041,12 @@ impl Connection {
             return;
         }
 
+        let minute = (get_time() / 60_000) as i32;
+        let mut state = TEMPORARY_PASSWORD_FAILURES.lock().unwrap();
         let current_password = password::temporary_password();
         if current_password.is_empty() {
             return;
         }
-
-        let minute = (get_time() / 60_000) as i32;
-        let mut state = TEMPORARY_PASSWORD_FAILURES.lock().unwrap();
         if state.password != current_password {
             state.password = current_password;
             state.minute = minute;
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -1993,11 +1993,6 @@ impl Connection {
         constant_time_eq(&hasher2.finalize()[..], &self.lr.password[..])
     }
 
-    #[inline]
-    fn validate_one_password(&self, password: &str) -> bool {
-        self.validate_password_plain(password)
-    }
-
     fn validate_password_plain(&self, password: &str) -> bool {
         if password.is_empty() {
             return false;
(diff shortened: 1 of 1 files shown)


rustdesk:
@codex review

rustdesk:
@codex review

rustdesk:
claude code review

  Issues & Concerns                                                                                                                                                                                          
                   
  1. Mutex held during password rotation (medium severity)
                                                                                                                                                                                                               
  password::update_temporary_password() is called while TEMPORARY_PASSWORD_FAILURES mutex is held. If that function triggers UI updates or acquires other locks, this creates deadlock risk. The fix is        
  straightforward — capture the needed data, drop the lock, rotate, then re-acquire:                                                                                                                           
                                                                                                                                                                                                               
  if state.failures >= MAX_CONSECUTIVE_FAILURES {                                                                                                                                                            
      let failures = state.failures;                                                                                                                                                                           
      drop(state);
      password::update_temporary_password();                                                                                                                                                                   
      let mut state = TEMPORARY_PASSWORD_FAILURES.lock().unwrap();                                                                                                                                           
      let new_password = password::temporary_password();                                                                                                                                                       
      log::warn!("... failures={}, ip={}", failures, self.ip);
      state.password = new_password;                                                                                                                                                                           
      state.failures = 0;                                                                                                                                                                                      
      return;
  }                                                                                                                                                                                                            
                                                                                                                                                                                                             
  This does introduce a TOCTOU window, but since double-rotation is benign, it's safer than holding the lock.                                                                                                  
  
  2. lazy_static + Mutex inside a method body                                                                                                                                                                  
                                                                                                                                                                                                             
  Defining State struct and lazy_static! block inside a method is unconventional. It compiles fine, but it hides global state in a surprising location. Consider moving the State struct and the static ref to 
  module level or at least to impl Connection level for visibility.                                                                                                                                          
                                                                                                                                                                                                               
  3. Storing the password in plaintext in State                                                                                                                                                              

  The State struct holds password: String — the current temporary password in cleartext in memory. This is used only for change-detection (comparing against password::temporary_password()). A hash or an     
  opaque generation counter / password ID would avoid keeping an extra copy. The risk is low since the password is already in memory elsewhere, but it's worth noting.
                                                                                                                                                                                                               
  4. Counter is per-process, not per-peer                                                                                                                                                                      
  
  The failure counter is global (single Mutex<State>) — any failed connection from any IP increments it. An attacker could trigger rotation by making 10 attempts, but so could 10 different legitimate users  
  mistyping. For a remote desktop tool this is likely acceptable (few concurrent users), but worth documenting explicitly.                                                                                   
                                                                                                                                                                                                               
  5. Race between success check and failure increment                                                                                                                                                          
  
  The success path (validate_password → check_update_temporary_password(true)) and failure path (update_failure → check_update_temporary_password(false)) are called at different points in the login flow. If 
  two connections authenticate concurrently — one succeeding, one failing — the ordering is non-deterministic. A success could reset the counter just before a 10th failure, or vice versa. This seems       
  acceptable for a coarse protection, but worth acknowledging.                                                                                                                                                 
                                                                                                                                                                                                             
  6. .unwrap() on mutex lock

  let mut state = TEMPORARY_PASSWORD_FAILURES.lock().unwrap();                                                                                                                                                 
  
  If any code path panics while holding this lock, the mutex becomes poisoned and all future connections will panic here. Consider using .lock().unwrap_or_else(|e| e.into_inner()) to recover from poisoning  
  gracefully, or at minimum document why panicking is acceptable.                                                                                                                                            
                                                                                                                                                                                                               
  ---                                                                                                                                                                                                        
  Minor Observations
                    
  - Removal of validate_one_password: Clean and correct. It was a trivial wrapper adding no value.
  - Log message: Good — includes failure count and IP. Consider also logging the peer ID for correlation with other logs.                                                                                      
  - MAX_CONSECUTIVE_FAILURES = 10: Reasonable default. Not configurable, which keeps things simple — fine for now.                                                                                             
                                                                                                                                                                                                               
  ---                                                                                                                                                                                                          
  Verdict                                                                                                                                                                                                      
                                                                                                                                                                                                             
  The security intent is sound and the implementation is mostly clean. The main actionable item is releasing the mutex before calling password::update_temporary_password() to avoid potential deadlocks. The
  other points are lower priority but worth considering. No test coverage changes in this PR — the author mentions manual testing of the scenarios, which is reasonable for this kind of stateful server logic.

21pages:
The main issue is that `password::update_temporary_password()` here is the server-side `hbb_common::password_security` function, not the UI/IPC path. In practice it just writes a new value into `TEMPORARY_PASSWORD` (`Arc<RwLock<String>>`) and then reads it back. I don’t see a reverse lock ordering involving `TEMPORARY_PASSWORD_FAILURES`, so the deadlock scenario looks speculative. Releasing the mutex before rotation would also reintroduce the race window.

  The other points mostly look like design notes rather than correctness issues:
  - the counter is intentionally global/coarse rather than per-peer;
  - success/failure ordering is not strictly linear under concurrency, but that seems acceptable for a coarse brute-force protection mechanism;
  - `lazy_static!` inside the method is unconventional, but not incorrect;
  - storing another plaintext copy in `State` is low risk because the temporary password is already kept in memory elsewhere;
  - handling mutex poisoning more gracefully could be nice, but that is a broader style issue in this file, not something specific to this change.

rustdesk:
• The new brute-force counter is wired to password validation rather than actual authentication outcomes. It resets too early on partial/failed logins and misses the recent-session success path, so the
  advertised consecutive-failure behavior is not preserved.

  Full review comments:

  - [P2] Defer clearing the one-time-password counter until auth succeeds — /Users/hzhou/web-rustdesk/src/server/connection.rs:2083-2083
    When 2FA is enabled, or when the login is rejected later because err_msg is non-empty, this line clears the new failure budget as soon as the temporary password matches. In those flows the client has not
    completed a successful one-time-password authentication, so it can break the "10 consecutive failures" streak without ever becoming authorized.
  - [P2] Reset the one-time-password counter on recent-session logins — /Users/hzhou/web-rustdesk/src/server/connection.rs:2454-2456
    On reconnects that are accepted by is_recent_session(false), the request never reaches this tracker, because both the success reset and the failure increment live only around validate_password(). A
    successful recent-session login therefore does not break the "consecutive failures" streak, so after a transient reconnect the temporary password can rotate even though a valid one-time-password login
    happened in between.

21pages:
The intended scope of this counter is narrower than full authentication outcome tracking. It is only meant to track consecutive wrong attempts against the currently active temporary password itself. We intentionally do not tie it to later authorization steps such as 2FA completion or other post-password rejection paths. Once the temporary password matches, that brute-force streak is considered broken.

  For the recent-session path, the counter is keyed to the current temporary password value, and it is reset whenever that temporary password changes or rotates. In the normal reconnect flow, an authorized session close rotates the temporary password, so a later recent-session login is not expected to carry forward the previous password's failure streak. I updated the code comment to make this behavior explicit.

tremoar:
What happens if a brute-force attack doesn't stop making impossible match the password for the legitimate user?

Maybe should be considered a temporary password per each whitelisted ID or past connections with a separated counter, this way the legitimate user can login with the public temporary password or with its assigned temporary password.

PR #14695, opened 2026-04-06
flutter: shift after one shot IME capitalization

amirHdev:
## Summary
Fix soft keyboard Shift in Android getting stuck after typing single uppercase letter in remote sessions.
Rustdesk can keep its local mobile Shift state enabled after Gboard uses Shift for single uppercase character so the next characters are still sent as shifted.
Fixes #14673

## Summary by CodeRabbit

* **Bug Fixes**
  * Detects and clears stale mobile Shift states during key input, preventing a stuck Shift modifier from affecting typing across input paths.

* **Tests**
  * Added unit tests covering stale-Shift detection and release across mobile vs non-mobile and modifier vs non-modifier scenarios.

* **Chores**
  * Enabled Flutter test tooling in dev dependencies.

### DIFF
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 import 'dart:math';
+import 'package:flutter/foundation.dart';
 import 'dart:ui' as ui;
 
 import 'package:desktop_multi_window/desktop_multi_window.dart';
@@ -718,6 +719,13 @@ class InputModel {
           hasTrackedShiftKeyDown: toReleaseRawKeys.lastLShiftKeyEvent != null ||
               toReleaseRawKeys.lastRShiftKeyEvent != null,
         )) {
+      if (kDebugMode) {
+        debugPrint(
+          'input: releasing stale mobile Shift before replaying tracked raw '
+          'key-up (logicalKey=${e.logicalKey.keyLabel}, '
+          'actualShiftPressed=${e.isShiftPressed}, cachedShiftPressed=$shift)',
+        );
+      }
       _releaseTrackedRawShiftKeyEventIfNeeded();
     }
 
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -621,6 +621,8 @@ class InputModel {
     }
   }
 
+  // Safe: this only re-dispatches synthesized Shift key-up events.
+  // The key-up path clears the tracked Shift state so this does not loop.
   void _releaseTrackedShiftKeyEventIfNeeded() {
     final leftShift = toReleaseKeys.lastLShiftKeyEvent;
     final rightShift = toReleaseKeys.lastRShiftKeyEvent;
@@ -632,6 +634,8 @@ class InputModel {
     }
   }
 
+  // Safe: this only re-dispatches synthesized Shift key-up events.
+  // The raw key-up path clears the tracked Shift state so this does not loop.
   void _releaseTrackedRawShiftKeyEventIfNeeded() {
     final leftShift = toReleaseRawKeys.lastLShiftKeyEvent;
     final rightShift = toReleaseRawKeys.lastRShiftKeyEvent;
@@ -710,6 +714,8 @@ class InputModel {
       legacyKeyboardModeRaw(e);
     }
 
+    // On some mobile soft-keyboard paths, Flutter may leave cached Shift state
+    // set even though the current raw key event is not shifted anymore.
     if (e is RawKeyDownEvent &&
         shouldReleaseStaleMobileShift(
           isMobile: isMobile,
@@ -759,6 +765,8 @@ class InputModel {
       iosCapsLock = _getIosCapsFromCharacter(e);
     }
 
+    // Update cached modifier state before sending the event. The stale mobile
+    // Shift release check below relies on this cached state.
     if (e is KeyUpEvent) {
       handleKeyUpEventModifiers(e);
     } else if (e is KeyDownEvent) {
@@ -796,6 +804,9 @@ class InputModel {
         }
       }
     }
+
+    // On some mobile soft-keyboard paths, Flutter may leave cached Shift state
+    // set even though the current key event is not shifted anymore.
     if (e is KeyDownEvent &&
         shouldReleaseStaleMobileShift(
           isMobile: isMobile,
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -722,7 +722,6 @@ class InputModel {
           cachedShiftPressed: shift,
           actualShiftPressed: e.isShiftPressed,
           logicalKey: e.logicalKey,
-          character: e.character,
           hasTrackedShiftKeyDown: toReleaseRawKeys.lastLShiftKeyEvent != null ||
               toReleaseRawKeys.lastRShiftKeyEvent != null,
         )) {
@@ -813,10 +812,8 @@ class InputModel {
           cachedShiftPressed: shift,
           actualShiftPressed: HardwareKeyboard.instance.isShiftPressed,
           logicalKey: e.logicalKey,
-          character: e.character,
-          hasTrackedShiftKeyDown:
-              toReleaseKeys.lastLShiftKeyEvent != null ||
-                  toReleaseKeys.lastRShiftKeyEvent != null,
+          hasTrackedShiftKeyDown: toReleaseKeys.lastLShiftKeyEvent != null ||
+              toReleaseKeys.lastRShiftKeyEvent != null,
         )) {
       _releaseTrackedShiftKeyEventIfNeeded();
     }
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -15,6 +15,7 @@ import 'package:get/get.dart';
 import '../../models/model.dart';
 import '../../models/platform_model.dart';
 import '../../models/state_model.dart';
+import 'input_modifier_utils.dart';
 import 'relative_mouse_model.dart';
 import '../common.dart';
 import '../consts.dart';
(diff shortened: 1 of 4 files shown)


rustdesk:
@codex review

Zacorich:
This PR was also fixed by https://github.com/rustdesk/rustdesk/pull/14709 where all modifier keys are handled and reset correctly.

TheLonelyDevil9:
THANK you for this, it was driving me nuts. Hope 1.4.7 releases soon @rustdesk

rustdesk:
# Claude Code Review: PR #14695 — "flutter: shift after one shot IME capitalization"

## Overview

This PR fixes a bug (#14673) where Android soft keyboards (Gboard) leave Shift in a "stuck" state after one-shot capitalization during remote sessions. It adds a pure utility function `shouldReleaseStaleMobileShift()` and integrates it into the two key event handlers (`handleKeyEvent` and `handleRawKeyEvent`) to synthesize Shift key-up events when a stale state is detected.

### Changed Files

- `flutter/lib/models/input_model.dart` — Integrated stale Shift detection into key event handlers
- `flutter/lib/models/input_modifier_utils.dart` — New utility function to detect stale mobile Shift states
- `flutter/test/input_modifier_utils_test.dart` — 9 unit tests covering all branches
- `flutter/pubspec.yaml` — Enabled `flutter_test` in dev dependencies

## Positives

- **Clean separation of concerns**: The detection logic is extracted into a standalone, pure function in `input_modifier_utils.dart` — easy to test and reason about.
- **Good test coverage**: 9 unit tests covering all branches of the utility function (non-mobile, empty/null character, modifier keys, cached vs actual shift, tracked guard).
- **Defensive guards**: Multiple early-return conditions prevent false positives (non-mobile, actual shift still pressed, modifier key events, null/empty characters, no tracked key-down).

## Issues & Suggestions

### 1. Missing `shift = false` after releasing tracked shift events (Medium severity)

`_releaseTrackedShiftKeyEventIfNeeded()` dispatches synthesized key-up events through `handleKeyEvent`/`handleRawKeyEvent`. These handlers do reset `shift = false` on key-up events for shift keys (lines 605-610, 665-667), so this should be handled — but only if the synthesized events actually flow through those code paths. The PR should verify (or add a comment confirming) that the dispatched events properly reset the `shift` boolean and clear the tracked shift references, to avoid an infinite re-trigger loop where `shouldReleaseStaleMobileShift` keeps returning `true`.

### 2. Potential re-entrancy in `handleKeyEvent` / `handleRawKeyEvent`

The new code calls `handleKeyEvent(leftShift)` from within what appears to be the same handler's call chain. If `handleKeyEvent` has side effects or guards (like the `isViewOnly` early return at line 688), the synthesized shift-up event might be silently swallowed or cause unexpected state. Consider:
- Adding a comment explaining why re-entrancy is safe here
- Or dispatching the release asynchronously / via a flag that's checked before the main event processing

### 3. Where exactly is the hook inserted in `handleKeyEvent`?

The `shouldReleaseStaleMobileShift` check is inserted in the key-down path, but its exact position relative to `handleKeyDownEventModifiers(e)` (line 723) matters:
- If it runs **before** `handleKeyDownEventModifiers`, then `shift` still reflects the previous (stale) state — correct.
- If it runs **after**, then `shift` may have already been updated by the current event, making the check unreliable.

The PR should ensure (and document) that the check happens **before** modifier state updates for the current event.

### 4. `pubspec.yaml` change enables `flutter_test` globally

Uncommenting `flutter_test` in `dev_dependencies` is the right thing, but confirm this was intentionally commented out (perhaps to work around a build issue). If there was a reason it was disabled, this change could break CI for other targets.

### 5. Test gap: no integration-level test

The unit tests cover the pure function well, but there's no test verifying the actual event dispatch flow — that calling `_releaseTrackedShiftKeyEventIfNeeded()` properly resets `shift` and clears `lastLShiftKeyEvent`/`lastRShiftKeyEvent`. A widget test or mock-based test for the `InputModel` would increase confidence that the end-to-end flow works.

### 6. Minor: both left and right shift are always released

In `_releaseTrackedShiftKeyEventIfNeeded`, both `lastLShiftKeyEvent` and `lastRShiftKeyEvent` are released if non-null. On mobile soft keyboards, typically only one shift variant is tracked. This is harmless but could be tightened — or at minimum deserves a comment explaining why both are checked.

## Security

No concerns — this is purely input event handling with no external I/O or data persistence.

## Summary

| Area | Assessment |
|------|-----------|
| **Correctness** | Core logic is sound; re-entrancy and ordering need verification |
| **Code quality** | Clean, well-structured utility function with good separation |
| **Test coverage** | Pure function fully covered; integration-level testing missing |
| **Risk** | Low — mobile-only code path with multiple defensive guards |

## Recommendation

**Approve with suggestions** — the approach is well-structured and low-risk. Before merging, the author should:

1. Confirm/document that the stale-shift check runs before `handleKeyDownEventModifiers()`
2. Add a brief comment explaining why the recursive `handleKeyEvent` call is safe
3. Consider an integration test for the full dispatch cycle

amirHdev:
@rustdesk Thanks
I added clarifying comments around the stale mobile Shift flow including the ordering assumptions.

rustdesk:
@amirHdev Your code is very good. Could you simply explain 1, 2, 3 of the claude code review seperately?

fufesou:
@amirHdev Hi, thanks a lot for the PR.

I can't reproduce this issue on my physical phone or the Emulator. GBoard was tested on both devices.

Please tell me what I was missing.

I will :
1. Try Pixel or Pixel 9a again.
2. Read the code.


https://github.com/user-attachments/assets/8d2a7038-da98-485a-a196-13dc5c90b7ee


Could you please also upload videos before and after the fix?

Please make sure the controlled side is 1.4.5.

https://github.com/rustdesk/rustdesk/issues/14673#issuecomment-4272970716

amirHdev:
> @amirHdev Your code is very good. Could you simply explain 1, 2, 3 of the claude code review seperately?

Thank you for the detailed review.
1. missing shift = false after release
the synthesized key up event hits handleKeyEvent lines 605 to 610 which sets shift = false. no infinite loop

2. Re entrancy safety
I added comment: // Safe: release event only synthesizes key-up which does not re-trigger stale shift check

3. Hook position in handleKeyEvent
check runs before handleKeyDownEventModifiers(e). added inline comment too

4. pubspec.yaml change
It was commented out accidentally in a previous PR. flutter_test is required for the new test file. 

5. Integration test gap
Added a widget test in input_model_test.dart that simulates the full event flow

6. Both shifts released
I added comment too: // Mobile keyboards track only one shift but checking both is harmless and keeps logic simple

All 6 points addressed. PR updated with comments and integration test. 
Thanks again!

amirHdev:
@fufesou Thanks.
Honestly I'm not 100% sure it's because of the emulator but that's my best guess. I've seen a few differences between emulators and real devices when it comes to keyboard behavior. things like input lag or how Gboard handles Shift and timing sensitivity, etc.

On my physical Android 16 device I can reproduce this every single time without fail. same steps same controlled side version 1.4.5 it just happens.

Maybe the emulator handles Shift state differently? or maybe there's a slight delay that changes the timing? I'm not entirely sure

If you have a real Android device handy could you try it there? that would help a lot. otherwise I've added unit tests and widget tests that cover the logic so at least we have automated coverage or manually test on my Device again too .

Let me know what you think.

fufesou:
> On my physical Android 16 device I can reproduce this every single time without fail. same steps same controlled side version 1.4.5 it just happens.

Hi @amirHdev, Thanks for the feedback.

Yes, you are right. Your test steps indicate this is a "Shift" issue, and you have fixed it.

BTW, do you mean you can produce it every time, both 1.4.6 (current master logic) and 1.4.5 on the controlled/remote side? Or only 1.4.5 on the controlled side?

Sorry, but I have a physical Android, but I still can't reproduce it.

fufesou:
@TheLonelyDevil9 Hi, do you still experience this issue when using 1.4.6 on the controlled(Windows) side?

rustdesk:
• The patch fixes the printable-character path but leaves stale Shift active for subsequent non-text keys, which can still send unintended shifted input to the remote side.

  Review comment:

  - [P2] Release stale Shift for non-text keys as well — /Users/hzhou/rustdesk/flutter/lib/models/input_modifier_utils.dart:21-21
    When a stale one-shot mobile Shift exists and the next key is non-text, such as Enter, Backspace, Tab, or an arrow key, character is typically null or empty, so this returns false and no synthetic Shift-
    up is sent before the key is forwarded. The remote side still has the earlier Shift down, so those keys are delivered as Shift+Enter/Shift+Backspace/etc. and the modifier remains stuck until a printable
    character is typed.

======================== update ============================

Addressed in d90a841

rustdesk:
> P2

@fufesou

fufesou:
## Tests

- [x] Does `pubspec.yaml` need to be changed?
- [x] Physical keyboard not affected. (1.4.5 & 1.4.6 , the controlled side , Windows)

fufesou (approved):
Not reproduced.
But this PR does not introduce regression.

rustdesk:
Issues
                                                                                                                                                                                                               
  1. Ordering bug in handleRawKeyEvent — release fires after the offending key is already sent                                                                                                                 
  
  In handleKeyEvent (the modern path, primary on Android), the new block is placed before the legacy/map-mode dispatch (if (isMobileAndMapMode || isDesktopAndMapMode) { ... }), so a synthesized Shift-up is  
  sent to the peer first, shift becomes false, and only then does the current key (d) flow through with the corrected modifier state. ✓
                                                                                                                                                                                                               
  In handleRawKeyEvent, the new block is placed after mapKeyboardModeRaw(e, iosCapsLock) / legacyKeyboardModeRaw(e) at input_model.dart:758. That means the original key has already been dispatched to the    
  peer with the stale shift = true cached state — so the very first post-one-shot key is still delivered uppercase, and only subsequent keys are corrected.
                                                                                                                                                                                                               
  Suggested fix: move the new if (e is RawKeyDownEvent && shouldReleaseStaleMobileShift(...)) block to before the if ((isDesktop || isWebDesktop) && keyboardMode == kKeyMapMode) { ... } else {               
  legacyKeyboardModeRaw(e); } dispatch, mirroring the placement in handleKeyEvent.
                                                                                                                                                                                                               
  In practice the raw path may rarely fire on Android Gboard (soft-keyboard input typically lands on KeyEvent, not RawKeyEvent), so the user-visible impact is small — but the asymmetry will bite if the raw  
  path is exercised, and the comment in both spots claims identical intent.


======================== update ============================

Addressed in b330a4b

fufesou:
https://github.com/rustdesk/rustdesk/pull/14695#issuecomment-4318550736

> Ordering bug in handleRawKeyEvent — release fires after the offending key is already sent

Seems reasonable and is fixed in b330a4b.

@amirHdev, could you please help to confirm?

amirHdev:
@fufesou Thanks
confirmed. I pulled the latest changes and releasing Shift before dispatching the raw key event fixes the ordering issue

fufesou:
@coderabbitai review

amirHdev:
@fufesou What you think about nitpick ? good to push into the code?

fufesou:
@amirHdev Which nitpick, this one https://github.com/rustdesk/rustdesk/pull/14695#pullrequestreview-4175821279 ?

Yes, it's better to add documentation.

Also, please add `debugPrint()` before `_releaseTrackedRawShiftKeyEventIfNeeded()` and include its output in the docs.

amirHdev:
@fufesou It fails in the `Install vcpkg dependencies` step before the Flutter build starts. but same workflow passed for armv7 and x86_64-linux-android

So this seems isolated to the Android arm64 `vcpkg` dependency bootstrap path likely in the Android overlayports
Could you familiar with Android vcpkg take a look?

rustdesk:
> same workflow passed for armv7 and x86_64-linux-android

No worries, the CI can fail sometimes because of network problem in the runner.

PR #14705, opened 2026-04-08
fix: bootstrap install LaunchAgent in user session

### LINKED ISSUE #10709: MacOS Custom Client with "hide-tray=Y" Parameter that Makes Client Run as LaunchDaemon Prevents Keyboard Input on Remote Sessions
### Bug Description

When building the RustDesk client with the hide-tray=Y config parameter, the client installs as a LaunchDaemon instead of a LaunchAgent. As a result, when accessing the machine remotely (from Win/Mac to Mac ARM64/Intel), the mouse functions correctly but the keyboard does not register any input. I have already granted all the necessary macOS permissions (Screen Recording, Accessibility, and Input Monitoring).

### How to Reproduce

	1.	Build the RustDesk client using the hide-tray=Y parameter.
	2.	Install the client, which will run as a LaunchDaemon.
	3.	Connect remotely to the machine.
	4.	Attempt to use the keyboard — it fails to work, while the mouse operates normally.

### Expected Behavior

Keyboard input should work correctly during remote sessions. Alternatively, there should be an option to hide the tray icon while still running as a LaunchAgent, so that the application remains in the user context and maintains full input functionality.

MacOS 15.3 (Apple Silicon) -->  MacOS 12.7.6 (Intel)

1.3.7 -> 1.3.2

### Screenshots

Actual Behavior:
Running RustDesk as a LaunchDaemon (via the hide-tray=Y parameter) causes it to operate outside the user session, which prevents it from receiving keyboard events, even though all relevant permissions are in place.

### Additional Context

It appears that the issue stems from the inherent limitations of LaunchDaemons on macOS, which do not have access to user session events like keyboard inputs. A potential solution might involve either adjusting how the hide-tray=Y parameter works or offering a separate configuration to hide the tray icon while still running as a LaunchAgent. Pin protect the tray icon would also work, the main idea is to prevent a bunch of corporate Karens to quit the client app or stop the service.

Any assistance or suggestions to resolve this issue would be greatly appreciated. Thanks!

amirHdev:
Fixes #10709 

## Summary by CodeRabbit

* **Tests**
  * Added macOS installation script tests to validate service bootstrap/restart behavior and user-specific installation handling.

* **Bug Fixes**
  * Improved macOS installer: more robust user-domain service enablement with conditional agent loading, tolerant handling of missing preference files, and clearer logging for service load outcomes.

### DIFF
--- a/src/platform/macos.rs
+++ b/src/platform/macos.rs
@@ -319,6 +319,7 @@ mod tests {
             .and_then(|file| file.contents_utf8())
             .map(correct_app_name)
             .expect("install.scpt should be embedded");
+        let app_name = crate::get_app_name();
 
         assert!(
             install.contains("launchctl bootstrap gui/$uid")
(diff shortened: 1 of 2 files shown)


rustdesk:
@codex review

rustdesk:
@codex review

amirHdev:
> @codex review

What you think ?

PR #14836, opened 2026-04-17
fix(client): serialize X11 keyboard grab and debounce focus feedback

sergiuszm:
## Summary

- Fixes keyboard input getting stuck on the wrong session (or stopping entirely) when switching between two fullscreen RustDesk sessions on separate monitors on Linux/X11
- Breaks the XGrabKeyboard focus-change feedback loop that blocks keyboard input for seconds during workspace switches

## Problem

When two RustDesk sessions run fullscreen on separate monitors on Linux/X11 (tested with MATE/Marco), keyboard input gets stuck on the wrong session or stops working entirely after clicking between monitors. The same workflow from a Windows client has no issues.

**Root cause 1 -- Race condition:** Each Flutter isolate (one per DesktopMultiWindow session window) calls `change_grab_status(Run/Wait)` concurrently via FFI. `SyncReturn` only serializes within a single Dart isolate, so two Rust threads race on `KEYBOARD_HOOKED` and the rdev grab channel. A stale `Wait` from session A can clobber session B's freshly acquired grab.

**Root cause 2 -- X11 focus feedback loop:** `XGrabKeyboard` shifts focus away from the Flutter window, triggering `PointerExit`, which releases the grab, restoring focus, triggering `PointerEnter`, which re-grabs -- cycling at ~10 Hz. This is the "rdev issues" referenced in `remote_page.dart` that led to focus-based grab handling being disabled on Linux.

## Fix

1. **Serialized grab transitions** via `GRAB_STATE: Mutex<GrabOwnerState>` that tracks which session (by `lc.session_id`) currently owns the grab
2. **Stale Wait filtering** -- ignores `Wait` from sessions that don't own the grab, so a late `PointerExit` from session A cannot release session B's grab
3. **Debounce (300 ms)** -- suppresses `Wait` events that arrive shortly after a `Run` for the same session, breaking the X11 focus feedback loop
4. **Debounce refresh** -- idempotent `Run` calls (enterView while already owner) refresh the debounce timer, keeping the grab stable during normal use

`src/keyboard.rs`, `src/ui_session_interface.rs`, and `src/flutter_ffi.rs` are changed. Client-side only -- remote machines need no changes. Windows/macOS behavior is preserved (the debounce is `#[cfg(target_os = "linux")]` only).

## Environment

- RustDesk version: 1.4.6
- Client OS: Linux Mint with MATE (X11)
- Remote OS: Linux Mint with MATE
- Setup: Two separate RustDesk sessions, each fullscreen on its own monitor

## Test plan

- [ ] Two fullscreen sessions on separate monitors -- keyboard follows mouse click between them
- [ ] Workspace switch does not cause multi-second keyboard lockout
- [ ] F12/hotkeys are captured by the remote session, not the local OS
- [ ] Single session: keyboard works normally after entering/leaving the remote image
- [ ] Windows/macOS clients are unaffected (no behavioral change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale or conflicting keyboard grabs between sessions and added a debounced release on Linux to avoid rapid toggles and spurious releases.
  * Ensured deferred releases no longer release a different session’s grab.

* **Improvements**
  * Grab transitions are session-aware and idempotent, improving input ownership and handoff reliability.
  * More robust ordering of release/handoff operations to reduce unexpected remote key behavior.

### DIFF
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -146,7 +146,7 @@ pub mod client {
     }
 
     pub fn start_grab_loop() {
-        let mut lock = IS_GRAB_STARTED.lock().unwrap_or_else(|e| e.into_inner());
+        let mut lock = IS_GRAB_STARTED.lock().unwrap();
         if *lock {
             return;
         }
@@ -167,7 +167,7 @@ pub mod client {
         let mut run_grab_after_unlock = None;
         #[cfg(target_os = "linux")]
         let mut disable_after_unlock = false;
-        let mut gs = GRAB_STATE.lock().unwrap_or_else(|e| e.into_inner());
+        let mut gs = GRAB_STATE.lock().unwrap();
         match state {
             GrabState::Ready => {}
             GrabState::Run => {
@@ -248,8 +248,7 @@ pub mod client {
                             std::thread::spawn(move || {
                                 std::thread::sleep(std::time::Duration::from_millis(remaining));
                                 let release_keys = {
-                                    let mut gs =
-                                        GRAB_STATE.lock().unwrap_or_else(|e| e.into_inner());
+                                    let mut gs = GRAB_STATE.lock().unwrap();
                                     // Release only if no new Run has refreshed the grab since.
                                     if gs.owner == Some(session_id) && gs.last_grab == snapshot {
                                         let to_release = take_remote_keys();
@@ -731,7 +730,7 @@ pub fn is_long_press(event: &Event) -> bool {
 }
 
 fn take_remote_keys() -> HashMap<Key, Event> {
-    let mut to_release = TO_RELEASE.lock().unwrap_or_else(|e| e.into_inner());
+    let mut to_release = TO_RELEASE.lock().unwrap();
     std::mem::take(&mut *to_release)
 }
 
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -84,11 +84,16 @@ pub mod client {
 
     /// Tracks grab ownership and serializes transitions across threads.
     ///
-    /// On Linux/X11, `XGrabKeyboard` can cause a focus-change feedback loop:
-    /// grab -> focus shifts -> PointerExit -> ungrab -> focus returns ->
-    /// PointerEnter -> grab -> ... at ~10 Hz. `last_grab` lets us debounce
-    /// spurious `Wait` events that arrive shortly after a `Run` for the same
-    /// session - these are X11 focus feedback, not real user actions.
+    /// Multiple Flutter isolates (one per session window) call
+    /// `change_grab_status(Run/Wait)` concurrently. Without serialization a
+    /// stale `Wait` from session A can clobber session B's freshly acquired
+    /// grab on any desktop OS.
+    ///
+    /// Windows and macOS are less susceptible in practice because the Flutter
+    /// side triggers `enterView` only after a mouse click inside the window,
+    /// but we cannot rely on that. On Linux/X11, `XGrabKeyboard` can also
+    /// cause a focus-change feedback loop (~10 Hz), so `last_grab` debounces
+    /// spurious `Wait` events that arrive shortly after a `Run`.
     #[derive(Default)]
     struct GrabOwnerState {
         owner: Option<u64>,
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -82,8 +82,31 @@ lazy_static::lazy_static! {
 pub mod client {
     use super::*;
 
+    /// Tracks grab ownership and serializes transitions across threads.
+    ///
+    /// On Linux/X11, `XGrabKeyboard` can cause a focus-change feedback loop:
+    /// grab -> focus shifts -> PointerExit -> ungrab -> focus returns ->
+    /// PointerEnter -> grab -> ... at ~10 Hz. `last_grab` lets us debounce
+    /// spurious `Wait` events that arrive shortly after a `Run` for the same
+    /// session - these are X11 focus feedback, not real user actions.
+    struct GrabOwnerState {
+        owner: Option<u64>,
+        last_grab: Option<std::time::Instant>,
+    }
+
+    impl Default for GrabOwnerState {
+        fn default() -> Self {
+            Self { owner: None, last_grab: None }
+        }
+    }
+
+    /// How long after a grab acquisition we suppress Wait from the same session.
+    /// Must exceed one full X11 feedback cycle (~100 ms: 50 ms enable + 50 ms disable).
+    const GRAB_DEBOUNCE_MS: u128 = 300;
+
     lazy_static::lazy_static! {
         static ref IS_GRAB_STARTED: Arc<Mutex<bool>> = Arc::new(Mutex::new(false));
+        static ref GRAB_STATE: Arc<Mutex<GrabOwnerState>> = Arc::new(Mutex::new(GrabOwnerState::default()));
     }
 
     pub fn start_grab_loop() {
@@ -96,33 +119,98 @@ pub mod client {
     }
 
     #[cfg(not(any(target_os = "android", target_os = "ios")))]
-    pub fn change_grab_status(state: GrabState, keyboard_mode: &str) {
+    pub fn change_grab_status(state: GrabState, keyboard_mode: &str, session_id: u64) {
         #[cfg(feature = "flutter")]
         if !IS_RDEV_ENABLED.load(Ordering::SeqCst) {
             return;
         }
+        // Serialize transitions so a stale `Wait` from a previous owner cannot
+        // clobber a fresh `Run` from a different session window.
+        let mut gs = GRAB_STATE.lock().unwrap();
         match state {
             GrabState::Ready => {}
             GrabState::Run => {
                 #[cfg(windows)]
                 update_grab_get_key_name(keyboard_mode);
+
+                // Idempotent: if this session already owns the grab, just
+                // refresh the debounce timer (proves the session is still
+                // actively focused) and skip the actual grab call.
+                if gs.owner == Some(session_id) {
+                    gs.last_grab = Some(std::time::Instant::now());
+                    log::debug!("[grab] Run(0x{:x}): already owner, refresh debounce", session_id);
+                    return;
+                }
+
+                log::info!(
+                    "[grab] Run(0x{:x}): prev_owner={}, mode={}",
+                    session_id,
+                    gs.owner.map_or("none".to_string(), |id| format!("0x{:x}", id)),
+                    keyboard_mode,
+                );
+
                 #[cfg(any(target_os = "windows", target_os = "macos", target_os = "linux"))]
-                KEYBOARD_HOOKED.swap(true, Ordering::SeqCst);
+                KEYBOARD_HOOKED.store(true, Ordering::SeqCst);
 
                 #[cfg(target_os = "linux")]
-                rdev::enable_grab();
+                {
+                    // On handoff, explicitly release any prior owner's X11 grab
+                    // before taking our own. This keeps the rdev control thread
+                    // and the X server in a consistent state even if the prior
+                    // owner never sent `Wait` (e.g. disconnected or raced).
+                    if gs.owner.is_some() {
+                        log::info!("[grab] handoff: disable_grab before re-grab");
+                        rdev::disable_grab();
+                    }
+                    rdev::enable_grab();
+                }
+                gs.owner = Some(session_id);
+                gs.last_grab = Some(std::time::Instant::now());
             }
             GrabState::Wait => {
+                // Drop stale `Wait` events that do not correspond to the
+                // current grab owner. This prevents a late PointerExit from
+                // session A from releasing session B's freshly acquired grab.
+                if gs.owner != Some(session_id) {
+                    log::debug!(
+                        "[grab] Wait(0x{:x}): ignored, owner={}",
+                        session_id,
+                        gs.owner.map_or("none".to_string(), |id| format!("0x{:x}", id)),
+                    );
+                    return;
+                }
+
+                // Debounce: on Linux/X11, XGrabKeyboard causes a focus-change
+                // feedback loop (grab -> PointerExit -> ungrab -> PointerEnter ->
+                // grab -> ...). Suppress Wait if the grab was acquired recently
+                // by this same session -- it is X11 feedback, not a real leave.
+                #[cfg(target_os = "linux")]
+                if let Some(t) = gs.last_grab {
+                    let elapsed = t.elapsed().as_millis();
+                    if elapsed < GRAB_DEBOUNCE_MS {
+                        log::debug!(
+                            "[grab] Wait(0x{:x}): debounced ({}ms < {}ms)",
+                            session_id, elapsed, GRAB_DEBOUNCE_MS,
+                        );
+                        return;
+                    }
+                }
+
+                log::info!("[grab] Wait(0x{:x}): releasing grab", session_id);
+
                 #[cfg(windows)]
                 rdev::set_get_key_unicode(false);
 
                 release_remote_keys(keyboard_mode);
 
                 #[cfg(any(target_os = "windows", target_os = "macos", target_os = "linux"))]
-                KEYBOARD_HOOKED.swap(false, Ordering::SeqCst);
+                KEYBOARD_HOOKED.store(false, Ordering::SeqCst);
 
                 #[cfg(target_os = "linux")]
                 rdev::disable_grab();
+
+                gs.owner = None;
+                gs.last_grab = None;
             }
             GrabState::Exit => {}
         }
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -211,9 +211,9 @@ pub mod client {
                             std::thread::spawn(move || {
                                 std::thread::sleep(std::time::Duration::from_millis(remaining));
                                 let mut gs = GRAB_STATE.lock().unwrap_or_else(|e| e.into_inner());
-                                gs.deferred_pending = false;
                                 // Release only if no new Run has refreshed the grab since.
                                 if gs.owner == Some(session_id) && gs.last_grab == snapshot {
+                                    gs.deferred_pending = false;
                                     log::info!("[grab] Wait(0x{:x}): deferred release", session_id);
                                     release_remote_keys(&mode);
                                     KEYBOARD_HOOKED.store(false, Ordering::SeqCst);
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -105,6 +105,7 @@ pub mod client {
 
     /// How long after a grab acquisition we suppress Wait from the same session.
     /// Must exceed one full X11 feedback cycle (~100 ms: 50 ms enable + 50 ms disable).
+    #[cfg(target_os = "linux")]
     const GRAB_DEBOUNCE_MS: u128 = 300;
 
     lazy_static::lazy_static! {
--- a/src/keyboard.rs
+++ b/src/keyboard.rs
@@ -108,7 +108,7 @@ pub mod client {
     }
 
     pub fn start_grab_loop() {
-        let mut lock = IS_GRAB_STARTED.lock().unwrap();
+        let mut lock = IS_GRAB_STARTED.lock().unwrap_or_else(|e| e.into_inner());
         if *lock {
             return;
         }
@@ -124,7 +124,7 @@ pub mod client {
         }
         // Serialize transitions so a stale `Wait` from a previous owner cannot
         // clobber a fresh `Run` from a different session window.
-        let mut gs = GRAB_STATE.lock().unwrap();
+        let mut gs = GRAB_STATE.lock().unwrap_or_else(|e| e.into_inner());
         match state {
             GrabState::Ready => {}
             GrabState::Run => {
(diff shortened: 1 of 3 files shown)


rustdesk:
# Review Findings (Codex)

- [P1] Don't discard the previous owner's key releases on session handoff — `/Users/hzhou/rustdesk/src/keyboard.rs:165-171`
  When `session A` still has keys held and `session B` acquires the grab first, this early return drops A's `Wait` without calling `release_remote_keys()`. Because pressed keys are tracked globally in `TO_RELEASE`, the eventual local key-up will be sent to whichever session is current next, so A can be left with a stuck modifier/key after a monitor switch. That is exactly the multi-session handoff path this patch changes.

- [P2] Debounced `Wait` can leave the keyboard grab active after a real leave — `/Users/hzhou/rustdesk/src/keyboard.rs:180-188`
  If the pointer genuinely leaves the session within `GRAB_DEBOUNCE_MS` of the latest `Run`, this branch returns before `rdev::disable_grab()`, `KEYBOARD_HOOKED.store(false, ...)`, or `gs.owner = None` run. In that case the client keeps intercepting keyboard input even though focus has already moved away, until some later enter/leave cycle happens to clean it up.

sergiuszm:
> # Review Findings (Codex)
> 
>     * [P1] Don't discard the previous owner's key releases on session handoff — `/Users/hzhou/rustdesk/src/keyboard.rs:165-171`
>       When `session A` still has keys held and `session B` acquires the grab first, this early return drops A's `Wait` without calling `release_remote_keys()`. Because pressed keys are tracked globally in `TO_RELEASE`, the eventual local key-up will be sent to whichever session is current next, so A can be left with a stuck modifier/key after a monitor switch. That is exactly the multi-session handoff path this patch changes.

This is a pre-existing issue rather than a regression from this patch. The stale-Wait filter is intentional (without it, A's late PointerExit clobbers B's grab -- the exact bug we're fixing). The deeper problem is that TO_RELEASE and release_remote_keys are global -- they route through cur_session, which has already switched to B by the time A's stale Wait arrives. Even if we called release_remote_keys there, the releases would go to B's peer, not A's.

Fixing this properly needs per-session key-release tracking, which is a larger refactor I'd rather not mix into this PR.

>     * [P2] Debounced `Wait` can leave the keyboard grab active after a real leave — `/Users/hzhou/rustdesk/src/keyboard.rs:180-188`
>       If the pointer genuinely leaves the session within `GRAB_DEBOUNCE_MS` of the latest `Run`, this branch returns before `rdev::disable_grab()`, `KEYBOARD_HOOKED.store(false, ...)`, or `gs.owner = None` run. In that case the client keeps intercepting keyboard input even though focus has already moved away, until some later enter/leave cycle happens to clean it up.

This is handled by the second commit (eb42d879). When a Wait is debounced, we spawn a deferred release thread that sleeps for the remaining debounce window, then re-checks whether the grab state is unchanged (same owner, same last_grab snapshot). If nothing refreshed it, the grab is released. So a genuine leave within the debounce window does get cleaned up.

rustdesk:
Full review comments: (codex)

  - [P1] Key grab ownership by Flutter window ID, not lc.session_id — /Users/hzhou/rustdesk/src/ui_session_interface.rs:873-880
    self.lc.session_id is connection-scoped, not window-scoped: flutter::sessions::insert_session() keeps one Session per (peer_id, conn_type), and insert_peer_session_id() only adds another UI handler to
    that same Session. When two desktop windows belong to the same peer connection, both enter()/leave() calls pass the same id here, so change_grab_status() cannot distinguish the windows, treats the second
    window's Run as already owner, and still lets the first window's later Wait interfere. In that multi-window case the new serialization logic does not actually serialize per window, so the grab race this
    patch is meant to fix still reproduces.
  - [P2] Re-arm deferred release after last_grab is refreshed — /Users/hzhou/rustdesk/src/keyboard.rs:224-229
    This deferred_pending guard is only a boolean, so once one debounced Wait has spawned a timer, every later Wait returns here even if a newer Run has already refreshed gs.last_grab (or ownership has
    changed). The old timer can then only cancel on its snapshot check, which leaves owner, KEYBOARD_HOOKED, and the X11 grab set with no replacement release scheduled. On Linux/X11 this is reachable in the
    normal feedback loop (Wait -> Run refresh -> real leave within the new 300 ms window).

sergiuszm:
Addressed both findings in a069b1d:

**P1 (window-scoped ID):** `session_enter_or_leave()` now derives the grab owner ID from the per-window `SessionID` UUID (`_session_id.as_u128() as u64`) and calls `change_grab_status` directly, bypassing `Session::enter/leave` (which keep their Sciter-compatible signature unchanged). Two windows viewing the same peer now get distinct grab owners.

**P2 (deferred-release re-arming):** `deferred_pending` is reset to `false` in both the idempotent Run path (same owner refreshes `last_grab`) and the owner handoff path (new owner takes over). An in-flight timer from a previous state self-cancels on its stale snapshot check, so no double-release is possible, and the next Wait can always spawn a fresh timer.

Also applied the CodeRabbit nitpick (derive `Default`).

src/keyboard.rs
```diff
@@ -82,8 +82,29 @@ lazy_static::lazy_static! {
...
+    /// Tracks grab ownership and serializes transitions across threads.
+    ///
+    /// On Linux/X11, `XGrabKeyboard` can cause a focus-change feedback loop:
+    /// grab -> focus shifts -> PointerExit -> ungrab -> focus returns ->
+    /// PointerEnter -> grab -> ... at ~10 Hz. `last_grab` lets us debounce
+    /// spurious `Wait` events that arrive shortly after a `Run` for the same
+    /// session - these are X11 focus feedback, not real user actions.
+    #[derive(Default)]
+    struct GrabOwnerState {
+        owner: Option<u64>,
```
afurm, line 99: Line 131: `let mut gs = GRAB_STATE.lock().unwrap();` — calling .unwrap() on a Mutex lock result will panic if the mutex is poisoned. While poisoning is rare (it only occurs when a thread panics while holding the lock), it means any previous panicking thread makes this code unreachable. Prefer: `GRAB_STATE.lock().expect("GRAB_STATE mutex")` with a descriptive message, or handle PoisonError explicitly with `lock().unwrap_or_else(|poisoned| poisoned.into_inner())`.

sergiuszm: Thanks. Good catch! Fixed in: [a815a06](https://github.com/rustdesk/rustdesk/pull/14836/commits/a815a06210b6d467a7f30daa3386d7a1858c5cd5)

src/keyboard.rs
```diff
@@ -82,12 +82,33 @@ lazy_static::lazy_static! {
...
 
+    /// Tracks grab ownership and serializes transitions across threads.
+    ///
+    /// On Linux/X11, `XGrabKeyboard` can cause a focus-change feedback loop:
+    /// grab -> focus shifts -> PointerExit -> ungrab -> focus returns ->
+    /// PointerEnter -> grab -> ... at ~10 Hz. `last_grab` lets us debounce
+    /// spurious `Wait` events that arrive shortly after a `Run` for the same
+    /// session - these are X11 focus feedback, not real user actions.
+    #[derive(Default)]
+    struct GrabOwnerState {
```
fufesou, line 98: Why not decorated by `#[cfg(target_os = "linux")]`

fufesou:
> **Root cause 1 -- Race condition:** Each Flutter isolate (one per DesktopMultiWindow session window) calls `change_grab_status(Run/Wait)` concurrently via FFI. `SyncReturn` only serializes within a single Dart isolate, so two Rust threads race on `KEYBOARD_HOOKED` and the rdev grab channel. A stale `Wait` from session A can clobber session B's freshly acquired grab.

I can reproduce the issue.
No need to use two full screens. Moving fast between two windows can reproduce it.

Your PR solved this problem very well, and the design is excellent. 👍👍👍 

While I'm not entirely sure your PR code is completely error-free, the general logic is clear and useful.

The root cause is correctly identified, but my tests suggest a more likely reason.

### My tests (master branch)

I have added some logs

<details open><summary>Code patch</summary>

```
diff --git a/flutter/lib/common/widgets/remote_input.dart b/flutter/lib/common/widgets/remote_input.dart
index 5871033db..326ecedab 100644
--- a/flutter/lib/common/widgets/remote_input.dart
+++ b/flutter/lib/common/widgets/remote_input.dart
@@ -628,8 +628,14 @@ class RawPointerMouseRegion extends StatelessWidget {
         cursor: inputModel.isViewOnly
             ? MouseCursor.defer
             : (cursor ?? MouseCursor.defer),
-        onEnter: onEnter,
-        onExit: onExit,
+        onEnter: (evt) {
+          debugPrint('================================= MouseRegion onEnter');
+          onEnter?.call(evt);
+        },
+        onExit: (evt) {
+          debugPrint('================================= MouseRegion onExit');
+          onExit?.call(evt);
+        },
         child: child,
       ),
     );
diff --git a/flutter/lib/desktop/pages/remote_page.dart b/flutter/lib/desktop/pages/remote_page.dart
index 29e710bbc..004a01fd6 100644
--- a/flutter/lib/desktop/pages/remote_page.dart
+++ b/flutter/lib/desktop/pages/remote_page.dart
@@ -487,6 +487,7 @@ class _RemotePageState extends State<RemotePage>
   }
 
   void enterView(PointerEnterEvent evt) {
+    debugPrint('================================= enterView');
     _ffi.canvasModel.rearmEdgeScroll();
 
     _cursorOverImage.value = true;
@@ -509,6 +510,7 @@ class _RemotePageState extends State<RemotePage>
   }
 
   void leaveView(PointerExitEvent evt) {
+    debugPrint('================================= leaveView');
     _ffi.canvasModel.disableEdgeScroll();
 
     if (_ffi.ffiModel.keyboard) {
diff --git a/src/flutter_ffi.rs b/src/flutter_ffi.rs
index e29133687..620c60f17 100644
--- a/src/flutter_ffi.rs
+++ b/src/flutter_ffi.rs
@@ -612,6 +612,7 @@ pub fn session_handle_flutter_raw_key_event(
 // As rust is multi-thread, it is possible that enter() is called before leave().
 // This will cause the keyboard input to take no effect.
 pub fn session_enter_or_leave(_session_id: SessionID, _enter: bool) -> SyncReturn<()> {
+    log::info!("=========================== session: {}, enter: {}", _session_id, _enter);
     if let Some(session) = sessions::get_session_by_session_id(&_session_id) {
         let keyboard_mode = session.get_keyboard_mode();
```

</details>

Sometimes I can get the following outputs, and then I can't input anymore.

When moving the mouse from Window A to Window B. `onEnter()` may be triggered first, then `onExit()`.

<details open><summary>The outputs</summary>

```
flutter: ================================= MouseRegion onEnter
flutter: ================================= enterView
flutter: ================================= MouseRegion onExit
flutter: ================================= leaveView
[INFO  librustdesk::flutter_ffi] =========================== session: c2cac68e-b824-4354-819c-ddf980b573ef, enter: true
[INFO  librustdesk::flutter_ffi] =========================== session: 77fd6367-3e63-456a-aa89-7273cf5a69af, enter: false
```

</details>


### I agree with your design

It is hard to make sure `onExit()` is called before `onEnter()`.

`owner` is a very good design.

A more comprehensive fix could touch both Flutter and Rust, but that would add complexity. Your fix looks good enough.

fufesou:
> **Root cause 2 -- X11 focus feedback loop:** `XGrabKeyboard` shifts focus away from the Flutter window, triggering `PointerExit`, which releases the grab, restoring focus, triggering `PointerEnter`, which re-grabs -- cycling at ~10 Hz. This is the "rdev issues" referenced in `remote_page.dart` that led to focus-based grab handling being disabled on Linux.

I can't reproduce it. 

I have added the logs in `session_enter_or_leave()` in `src/flutter_ffi.rs`.

But it seems reasonable.
Could you please upload a video before and after the fix?

fufesou:
I have also experienced another issue.

1. Start RustDesk
2. Connect to a client with 2 displays
3. Click the 2nd display to open another window

When I enter or leave Window 2, Flutter does not emit `onEnter()` or `onExit()`.
Then I can't click or input any keys.

But if I close Window 2 and open it again, the callback can be emitted correctly.

It's another issue, but it will be great if you can reproduce it and solve it in a **new PR**.

Ubuntu 24.04.3 TLS, GNOME, x11


https://github.com/user-attachments/assets/b817c76f-0508-44cd-ade7-3c9f1d685ed4

sergiuszm:
> > **Root cause 2 -- X11 focus feedback loop:** `XGrabKeyboard` shifts focus away from the Flutter window, triggering `PointerExit`, which releases the grab, restoring focus, triggering `PointerEnter`, which re-grabs -- cycling at ~10 Hz. This is the "rdev issues" referenced in `remote_page.dart` that led to focus-based grab handling being disabled on Linux.
> 
> I can't reproduce it.
> 
> I have added the logs in `session_enter_or_leave()` in `src/flutter_ffi.rs`.
> 
> But it seems reasonable. Could you please upload a video before and after the fix?

This happens randomly - I don't have a reliable way to reproduce this either. I can only provide a debug log from the time it happened.
[debug.log](https://github.com/user-attachments/files/26870956/debug.log)

fufesou (changes requested):

src/keyboard.rs
sergiuszm: The ownership serialization (`GrabOwnerState`) fixes a cross-platform race condition: multiple Flutter isolates calling `change_grab_status(Run/Wait)` concurrently can clobber each other's state on any desktop OS, not just Linux. The doc comment highlights the X11 feedback loop as the most visible symptom, but the stale-`Wait` filtering and owner tracking are needed on Windows and macOS too.

Gating the struct to Linux would re-expose the race on other platforms. I could trim the doc comment to emphasize the general problem and mention X11 as a specific case - would that help?

sergiuszm:
> I have also experienced another issue.
> 
>     1. Start RustDesk
> 
>     2. Connect to a client with 2 displays
> 
>     3. Click the 2nd display to open another window
> 
> 
> When I enter or leave Window 2, Flutter does not emit `onEnter()` or `onExit()`. Then I can't click or input any keys.
> 
> But if I close Window 2 and open it again, the callback can be emitted correctly.
> 
> It's another issue, but it will be great if you can reproduce it and solve it in a **new PR**.
> 
> Ubuntu 24.04.3 TLS, GNOME, x11
> flutter-linux-no-events.mp4

I am unable to reproduce this issue. Linux Mint 22.3, Mate, X11 - both client and host. I tried multiple times.

https://github.com/user-attachments/assets/a59b0659-b7b1-40e5-a42e-b807ddfecac3

src/keyboard.rs
fufesou: Thanks.

> not just Linux

My mistake.

> I could trim the doc comment to emphasize the general problem and mention X11 as a specific case - would that help?

It should be better.

fufesou:
I've reviewed the code; no other logic needs to change.

> > > **Root cause 2 -- X11 focus feedback loop:** `XGrabKeyboard` shifts focus away from the Flutter window, triggering `PointerExit`, which releases the grab, restoring focus, triggering `PointerEnter`, which re-grabs -- cycling at ~10 Hz. This is the "rdev issues" referenced in `remote_page.dart` that led to focus-based grab handling being disabled on Linux.
> 
> This happens randomly - I don't have a reliable way to reproduce this either. I can only provide a debug log from the time it happened. [debug.log](https://github.com/user-attachments/files/26870956/debug.log)

I still haven't been able to reproduce it. How frequently does this occur?

I'm just asking because the fix is harmless.

It should fix the issue you mentioned.

> Gating the struct to Linux would re-expose the race on other platforms. I could trim the doc comment to emphasize the general problem and mention X11 as a specific case - would that help?

When changing the comments, you can also mention that Windows and macOS do not have this issue because the Flutter side includes special logic. Then the "Enter" event is triggered only after a mouse click inside the window.

However, we can't rely on it.

The struct is harmless. It's better to keep it on Windows and macOS.

sergiuszm:
> I've reviewed the code; no other logic needs to change.
> 
> > > > **Root cause 2 -- X11 focus feedback loop:** `XGrabKeyboard` shifts focus away from the Flutter window, triggering `PointerExit`, which releases the grab, restoring focus, triggering `PointerEnter`, which re-grabs -- cycling at ~10 Hz. This is the "rdev issues" referenced in `remote_page.dart` that led to focus-based grab handling being disabled on Linux.
> > 
> > 
> > This happens randomly - I don't have a reliable way to reproduce this either. I can only provide a debug log from the time it happened. [debug.log](https://github.com/user-attachments/files/26870956/debug.log)
> 
> I still haven't been able to reproduce it. How frequently does this occur?
> 
> I'm just asking because the fix is harmless.
> 
> It should fix the issue you mentioned.
> 
> > Gating the struct to Linux would re-expose the race on other platforms. I could trim the doc comment to emphasize the general problem and mention X11 as a specific case - would that help?
> 
> When changing the comments, you can also mention that Windows and macOS do not have this issue because the Flutter side includes special logic. Then the "Enter" event is triggered only after a mouse click inside the window.
> 
> However, we can't rely on it.
> 
> The struct is harmless. It's better to keep it on Windows and macOS.

I've updated the doc comment to clarify the cross-platform rationale. The debounce (last_grab) remains specifically for the Linux/X11 feedback loop.

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

src/keyboard.rs
```diff
@@ -96,33 +122,136 @@ pub mod client {
...
+                                    release_remote_keys(&mode);
+                                    KEYBOARD_HOOKED.store(false, Ordering::SeqCst);
+                                    rdev::disable_grab();
+                                    gs.owner = None;
+                                    gs.last_grab = None;
+                                } else {
+                                    log::debug!(
+                                        "[grab] Wait(0x{:x}): deferred release cancelled (grab refreshed)",
+                                        session_id,
+                                    );
```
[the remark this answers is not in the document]
fufesou: This comment seems to be a good catch on lock contention, but unlocking before release/disable alone is unsafe due to race conditions.

src/ui_session_interface.rs
```diff
@@ -870,12 +870,14 @@ impl<T: InvokeUiSession> Session<T> {
 
     #[cfg(not(any(target_os = "android", target_os = "ios")))]
     pub fn enter(&self, keyboard_mode: String) {
-        keyboard::client::change_grab_status(GrabState::Run, &keyboard_mode);
+        let session_id = self.lc.read().unwrap().session_id as u128;
+        keyboard::client::change_grab_status(GrabState::Run, &keyboard_mode, session_id);
```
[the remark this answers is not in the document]
sergiuszm: `Session::enter/leave` is the non-Flutter (sciter) UI path, which doesn't support multi-window sessions. Only the Flutter path (`session_enter_or_leave` in `flutter_ffi.rs`) can have multiple windows viewing the same peer, and that path already uses the full per-window UUID. No collision risk here.

src/keyboard.rs
```diff
@@ -96,33 +122,137 @@ pub mod client {
...
                 update_grab_get_key_name(keyboard_mode);
+
+                // Idempotent: if this session already owns the grab, just
+                // refresh the debounce timer (proves the session is still
+                // actively focused) and skip the actual grab call.
+                if gs.owner == Some(session_id) {
+                    gs.last_grab = Some(std::time::Instant::now());
+                    // Reset so the next Wait can spawn a fresh deferred-release
+                    // timer with an up-to-date snapshot of last_grab.
+                    gs.deferred_pending = false;
```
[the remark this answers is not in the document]
sergiuszm: This is intentional. When an idempotent `Run` refreshes the debounce timer, the next `Wait` needs to spawn a fresh deferred-release thread with the updated `last_grab` snapshot. The old timer self-cancels via snapshot mismatch (`gs.last_grab != snapshot`), so there's never more than one timer doing actual work. Without the reset, a genuine leave after a debounce refresh would be silently dropped because `deferred_pending` would still be `true` from the stale timer.

fufesou:
Hi @sergiuszm, please address https://github.com/rustdesk/rustdesk/pull/14836#pullrequestreview-4144461265 and the last comment in https://github.com/rustdesk/rustdesk/pull/14836#pullrequestreview-4151461510

"Resolve conversation" if they are solved or there is no need to change.

rustdesk:
• The grab ownership handoff can suppress the previous session's key release, leaving remote keys stuck in a common session-switching scenario with held modifiers. This is a behavioral regression in the area
  the patch is trying to fix.

  Review comment:

  - [P2] Release previous keys before overwriting grab owner — /Users/hzhou/rustdesk/src/keyboard.rs:172-172
    When handing off while the previous session still has keys in TO_RELEASE (for example holding Ctrl/Alt in session A and clicking session B, which on X11 can land inside the 300 ms debounce window), this
    owner overwrite makes A's deferred Wait cancel and any later Wait(A) is ignored as stale. Since release_remote_keys routes through the current session, the physical key-up after the switch goes to B
    instead of A, leaving A with a stuck modifier/key; the previous owner's keys need to be released or tracked per owner before invalidating its pending release.

sergiuszm:
Good catch on the scenario. However, this is a pre-existing limitation of `TO_RELEASE` being global and `release_remote_keys` routing through `get_cur_session()`, not a regression introduced by this PR.

Simply calling `release_remote_keys` before overwriting the owner doesn't help here: on the `enter` path, `set_cur_session_id_` switches the current session to B *before* `change_grab_status(Run, B)` is called (`flutter_ffi.rs:622-627`). So by the time we reach the handoff in `Run`, the current session is already B, and any release would go to B's remote instead of A's.

A proper fix requires per-session `TO_RELEASE` tracking (keyed by owner) so releases route through the correct session. That's a larger change to pre-existing architecture and would be better handled in a separate PR.

On master today, the same stuck modifier scenario exists (and is worse) because there's no ownership tracking at all - a stale `Wait` from A can fully clobber B's grab, losing keyboard input entirely rather than just a stuck modifier.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14836#issuecomment-4318077805

> The grab ownership handoff can suppress the previous session's key release, leaving remote keys stuck in a common session-switching scenario with held modifiers. 

There is no need to change.

Users don't hold modifier keys while moving the mouse between windows.

The issue only happens when pointing to the same controlled side -- releasing the modifiers is fine in that scenario.

Moreover, the debounce window is only 300 ms.

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou (approved):

PR #14848, opened 2026-04-19
fix(linux): enable mouse side buttons in remote sessions

sergiuszm:
## Summary

- Mouse back/forward buttons (X11 buttons 8/9) were silently dropped in Linux remote sessions because Flutter's Linux embedder never delivers these button events to Dart code
- Intercept buttons 8/9 at the GDK level via `button-press-event`/`button-release-event` handlers on all windows (main + sub-windows) and forward them through a dedicated platform channel (`org.rustdesk.rustdesk/side_buttons`) to the active session's InputModel
- Add `forward` to the `MouseButtons` enum (only `back` was defined)
- Add defensive `XSetPointerMapping` call during enigo init to extend the X11 core pointer button map on servers with minimal X configurations

## Details

**Client side (C)** — `flutter/linux/my_application.cc`:
- GDK signal handler filters for buttons 8/9, ignores double/triple-click synthetic events
- Per-window `FlMethodChannel` with proper GObject lifecycle (`g_object_set_data_full`)
- Sub-windows handled via unified `on_subwindow_created` callback (also chains existing Wayland shortcuts inhibition)

**Client side (Dart)** — `flutter/lib/models/input_model.dart`:
- Separate `_sideButtonDownModel` tracking ensures button-up is routed to the model that received button-down, even if the pointer has left the remote view
- `disposeSideButtonTracking()` clears stale references on session close

**Server side (Rust)** — `libs/enigo/src/linux/xdo.rs`:
- `ensure_x11_button_map()` extends core pointer to 9+ buttons if needed (defensive, most X servers already have sufficient buttons)

## Test plan
- [x] Tested on X11: side buttons (back/forward) register correctly in remote session
- [ ] Wayland: not tested (no Wayland test environment), but side button code is X11-only and Wayland shortcuts inhibition path is preserved
- [ ] Multi-window: sub-window callback handles side button setup for remote session windows

## Summary by CodeRabbit

* **New Features**
  * Linux side-mouse-button support extended to both back and forward buttons, forwarded to the active remote view.

* **Bug Fixes**
  * Better tracking and cleanup of side-button routing when views/sessions close; unmatched release events are dropped to prevent stray actions.

* **Chores**
  * Added X11 pointer-button diagnostics to detect and log Linux mouse compatibility issues.

### DIFF
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -339,6 +339,10 @@ class InputModel {
   static final Map<MouseButtons, InputModel> _sideButtonDownModels = {};
   static bool _sideButtonChannelInitialized = false;
 
+  /// Each Flutter engine (main window + sub-windows from desktop_multi_window)
+  /// runs its own Dart isolate with its own statics. Called from initEnv()
+  /// which runs per-engine, so each isolate registers its own handler tied
+  /// to its own set of InputModels.
   static void initSideButtonChannel() {
     if (!Platform.isLinux) return;
     if (_sideButtonChannelInitialized) return;
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -387,7 +387,10 @@ class InputModel {
         .toList();
     for (final mb in held) {
       _sideButtonDownModels.remove(mb);
-      unawaited(sendMouse('up', mb));
+      // Best-effort release; session may already be tearing down.
+      unawaited(sendMouse('up', mb).catchError((Object e) {
+        debugPrint('[InputModel] failed to release side button $mb: $e');
+      }));
     }
   }
 
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -488,6 +488,7 @@ class InputModel {
   bool get isRelativeMouseModeSupported => _relativeMouse.isSupported;
 
   InputModel(this.parent) {
+    initSideButtonChannel();
     sessionId = parent.target!.sessionId;
     _relativeMouse = RelativeMouseModel(
       sessionId: sessionId,
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -359,16 +359,20 @@ class InputModel {
         if (type == 'down') {
           final model = _activeSideButtonModel;
           if (model != null &&
-              !(model.isViewOnly && !model.showMyCursor)) {
+              !(model.isViewOnly && !model.showMyCursor) &&
+              model.keyboardPerm &&
+              !model.isViewCamera) {
             _sideButtonDownModels[mb] = model;
-            await model.sendMouse(type, mb);
+            await model._sendMouseUnchecked(type, mb);
           }
         } else {
           // Only route 'up' when we recorded the matching 'down';
           // dropping avoids sending unpaired 'up' to an unrelated session.
+          // Use _sendMouseUnchecked to bypass permission checks so the
+          // release always goes through even if permissions changed.
           final model = _sideButtonDownModels.remove(mb);
           if (model != null) {
-            await model.sendMouse(type, mb);
+            await model._sendMouseUnchecked(type, mb);
           }
         }
       }
@@ -388,7 +392,7 @@ class InputModel {
     for (final mb in held) {
       _sideButtonDownModels.remove(mb);
       // Best-effort release; session may already be tearing down.
-      unawaited(sendMouse('up', mb).catchError((Object e) {
+      unawaited(_sendMouseUnchecked('up', mb).catchError((Object e) {
         debugPrint('[InputModel] failed to release side button $mb: $e');
       }));
     }
@@ -1033,13 +1037,20 @@ class InputModel {
     return evt;
   }
 
+  /// Send mouse event unconditionally (no permission checks).
+  /// Used for side button releases that must go through even if permissions
+  /// changed after the matching down was sent.
+  Future<void> _sendMouseUnchecked(String type, MouseButtons button) async {
+    await bind.sessionSendMouse(
+        sessionId: sessionId,
+        msg: json.encode(modify({'type': type, 'buttons': button.value})));
+  }
+
   /// Send mouse press event.
   Future<void> sendMouse(String type, MouseButtons button) async {
     if (!keyboardPerm) return;
     if (isViewCamera) return;
-    await bind.sessionSendMouse(
-        sessionId: sessionId,
-        msg: json.encode(modify({'type': type, 'buttons': button.value})));
+    await _sendMouseUnchecked(type, button);
   }
 
   void enterOrLeave(bool enter) {
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -333,9 +333,10 @@ class InputModel {
   // Flutter's Linux embedder drops X11 button 8/9 events, so we capture them
   // natively via GDK and forward through the platform channel.
   static InputModel? _activeSideButtonModel;
-  // Tracks the model that received a side button down event, so the matching
-  // up event is routed there even if the pointer has left the remote view.
-  static InputModel? _sideButtonDownModel;
+  // Tracks per-button which model received a side button down event, so the
+  // matching up event is routed there even if the pointer has left the view
+  // or a different button was pressed in between.
+  static final Map<MouseButtons, InputModel> _sideButtonDownModels = {};
   static bool _sideButtonChannelInitialized = false;
 
   static void initSideButtonChannel() {
@@ -354,14 +355,14 @@ class InputModel {
         if (type == 'down') {
           final model = _activeSideButtonModel;
           if (model != null) {
-            _sideButtonDownModel = model;
+            _sideButtonDownModels[mb] = model;
             await model.sendMouse(type, mb);
           }
         } else {
           // Route 'up' to the model that received the matching 'down',
           // even if the pointer has since left the remote view.
-          final model = _sideButtonDownModel ?? _activeSideButtonModel;
-          _sideButtonDownModel = null;
+          final model =
+              _sideButtonDownModels.remove(mb) ?? _activeSideButtonModel;
           if (model != null) {
             await model.sendMouse(type, mb);
           }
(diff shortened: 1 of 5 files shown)


rustdesk:
# Code Review: PR #14848 — fix(linux): enable mouse side buttons in remote sessions

**PR:** https://github.com/rustdesk/rustdesk/pull/14848
**Author:** sergiuszm
**Base:** `rustdesk:master` @ `ac124c068`
**Head:** `sergiuszm:fix/client-mouse-side-buttons` @ `8bea4b699`
**Stats:** +227 / -6, 5 files changed, 3 commits

## Overview

Fixes Linux back/forward mouse buttons (X11 buttons 8/9) being dropped by Flutter's Linux embedder. Strategy:

1. **Client (C)** — GDK `button-press-event` / `button-release-event` handlers on main + sub-windows forward buttons 8/9 through a dedicated `org.rustdesk.rustdesk/side_buttons` platform channel.
2. **Client (Dart)** — adds `MouseButtons.forward` enum variant, a static routing table keyed by button to match down→up across view-leave and multi-session scenarios, plus disposal cleanup on session close.
3. **Server (Rust)** — `ensure_x11_button_map()` defensively extends the X11 core pointer button map to ≥9 buttons so `XTestFakeButtonEvent` for 8/9 is not silently dropped.

The author has already iterated through two CodeRabbit review passes (struct-offset unsafety, per-button tracking, view-only gating). End-to-end wiring is correct: `MOUSE_BUTTON_BACK` / `MOUSE_BUTTON_FORWARD` already exist server-side in `src/server/input_service.rs:1139-1144,1157-1162`.

## Strengths

- **Correct GObject lifecycle** in `side_buttons_init_for_window` via `g_object_set_data_full`; signal handlers release cleanly with the window.
- **Double-init guard** on the window keeps signal `user_data` from dangling.
- **Thread-safety** — GDK callbacks and `fl_method_channel_invoke_method` stay on the GTK main thread; no cross-thread hazards.
- **Safer Rust** — uses an independent `XOpenDisplay` rather than relying on `xdo_t`'s private struct layout. `MappingBusy` retry loop is appropriate.
- **Down-owner-by-button** map correctly handles overlapping back/forward presses across sessions (fixed from single-slot design in first revision).
- **Good comments** on the non-obvious bits (why the channel exists, why we drop unpaired `up`, why we skip `GDK_2BUTTON_PRESS`).

## Issues & Suggestions

### 1. `on_subwindow_created` — check X11-only build paths (`flutter/linux/my_application.cc`)

Previous code registered the sub-window callback only when both `GDK_WINDOWING_WAYLAND` **and** `HAS_KEYBOARD_SHORTCUTS_INHIBIT` were defined. The PR now registers `on_subwindow_created` unconditionally. Inside, the Wayland-specific call is still guarded, so the body itself is safe — but confirm that calling `desktop_multi_window_plugin_set_window_created_callback` on X11-only builds (where it was previously never called) does not regress behavior.

### 2. Unpaired `up` window is bounded but not zero (`flutter/lib/models/input_model.dart:348-368`)

If the user toggles view-only *between* down and up, the down is filtered but a later up can arrive unpaired. Current code correctly drops the orphan up because `_sideButtonDownModels[mb]` was never set. The inverse (user toggles *out* of view-only between down and up) would leave a button held on the peer. `disposeSideButtonTracking` sends compensating `up` on session close, so the stuck-button window is bounded to the session lifetime. Acceptable, but worth noting in a comment.

### 3. `XSetPointerMapping` has a global side effect (`libs/enigo/src/linux/xdo.rs`)

This mutates the **global** X11 pointer map, not a per-client view. On a typical desktop with ≥9 buttons already mapped it is a no-op, but on minimal/headless X servers it extends the map for the lifetime of the X server, affecting every other X client. Risk is low (only appends identity entries for buttons beyond the current count), but:

- Add a comment making the global scope explicit.
- Consider running lazily on the first back/forward event rather than at every `EnigoXdo::default()` construction — currently it re-runs per enigo context.

### 4. `ensure_x11_button_map` can block init for up to 500 ms

On `MappingBusy` the function sleeps 5 × 100 ms. Fine for one-shot init, but surprising if `EnigoXdo` is ever constructed on a latency-sensitive path. No change required; worth flagging.

### 5. `initSideButtonChannel` per-isolate assumption is implicit

Each Flutter engine (main + sub-window engines from `desktop_multi_window_plugin`) runs its own Dart isolate with its own copy of `_activeSideButtonModel` and `_sideButtonChannelInitialized`. `initEnv` runs per-engine, so this works correctly — but the design is implicit. A one-line comment in `initSideButtonChannel` explaining that each engine registers its own handler, tied to its own set of InputModels, would help future maintainers.

### 6. Cast defensiveness (`input_model.dart` channel handler)

`args['button'] as String` and `args['type'] as String` will throw on malformed payloads. The sender is trusted native code, so this is fine in practice; `as String?` + null-check would be more defensive.

### 7. Test coverage

No automated tests — acceptable given the X11/GDK-integration nature, but the author flagged Wayland as "not tested." GDK normalizes X11 and Wayland pointer events, so the path likely works on Wayland too, but a smoke test would de-risk the claim. The PR description's "Multi-window: sub-window callback handles side button setup" is also unchecked.

## Security

- No new untrusted inputs; method-channel payload is constructed entirely in-process by the C++ handler.
- `XSetPointerMapping` requires the invoking user to own the X session — no new privilege boundary.

## Performance

- GDK handler returns `FALSE` for non-8/9 buttons — no hot-path cost on regular clicks.
- `fl_method_channel_invoke_method` with null callback is fire-and-forget; no blocking.
- Server-side X11 map extension runs once per `EnigoXdo` construction (see §3-4 for notes).

## Verdict

Solid, well-scoped fix with thoughtful responses to earlier review feedback. The main item worth addressing before merge is **§3** (comment on global side effect of `XSetPointerMapping`, and consider running it lazily). **§1** and **§5** are worth a line of clarification each. Everything else is nit-level. No blocking correctness issues.

fufesou:
- Wayland: not tested (no Wayland test environment), but side button code is X11-only and Wayland shortcuts inhibition path is preserved
- Multi-window: sub-window callback handles side button setup for remote session windows

Both work perfectly in my tests.

Ubuntu GNOME Wayland, X11
Archlinux KDE Wayland, GNOME Wayland

sergiuszm:
My last commit ([3583378](https://github.com/rustdesk/rustdesk/pull/14848/commits/35833789dbd639ad42c960966bb2b721a83365f3)) addresses §1, §3 and §5.

fufesou:
@coderabbitai review

flutter/linux/my_application.cc
```diff
@@ -29,6 +29,78 @@ void try_set_transparent(GtkWindow* window, GdkScreen* screen, FlView* view);
...
+  g_signal_connect(window, "button-release-event",
+    G_CALLBACK(on_side_button_event), channel);
+}
+
+static void on_subwindow_created(FlPluginRegistry* registry) {
+#if defined(GDK_WINDOWING_WAYLAND) && defined(HAS_KEYBOARD_SHORTCUTS_INHIBIT)
+  wayland_shortcuts_inhibit_init_for_subwindow(registry);
+#endif
+  // Set up side button forwarding for sub-windows.
+  FlView* view = FL_VIEW(registry);
```
fufesou, line 96: Check `NULL`?

flutter/linux/my_application.cc
```diff
@@ -29,6 +29,78 @@ void try_set_transparent(GtkWindow* window, GdkScreen* screen, FlView* view);
...
+  GtkWidget* toplevel = gtk_widget_get_toplevel(GTK_WIDGET(view));
+  if (toplevel != NULL && GTK_IS_WINDOW(toplevel)) {
+    FlMethodChannel* channel = side_buttons_create_channel(fl_view_get_engine(view));
```
fufesou, line 99: Check `NULL`?

flutter/linux/my_application.cc
sergiuszm: Done

flutter/linux/my_application.cc
sergiuszm: Done

fufesou:
@sergiuszm Hi, please also address https://github.com/rustdesk/rustdesk/pull/14848#pullrequestreview-4144466514 , and close the conversations after addressing

flutter/linux/my_application.cc
```diff
@@ -29,6 +29,80 @@ void try_set_transparent(GtkWindow* window, GdkScreen* screen, FlView* view);
...
+static const char* kSideButtonChannelName = "org.rustdesk.rustdesk/side_buttons";
+
+static gboolean on_side_button_event(GtkWidget* widget, GdkEventButton* event, gpointer user_data) {
+  if (event->button != 8 && event->button != 9) {
+    return FALSE;
+  }
+  // Ignore GDK_2BUTTON_PRESS / GDK_3BUTTON_PRESS (double/triple-click synthetic
+  // events) - only handle real press and release.
+  if (event->type != GDK_BUTTON_PRESS && event->type != GDK_BUTTON_RELEASE) {
+    return FALSE;
```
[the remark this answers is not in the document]
sergiuszm: GDK's multi-click sequence delivers GDK_BUTTON_PRESS first for each physical press, then GDK_2BUTTON_PRESS/GDK_3BUTTON_PRESS as additional synthetic events. So the 2nd/3rd presses are already captured by the regular GDK_BUTTON_PRESS handler - the 2BUTTON_PRESS is a duplicate that should be ignored, not a replacement. Dropping it is the correct behavior.

flutter/lib/models/input_model.dart
```diff
@@ -327,6 +329,80 @@ class ToReleaseKeys {
...
+              model.keyboardPerm &&
+              !model.isViewCamera) {
+            _sideButtonDownModels[mb] = model;
+            // Fire-and-forget to avoid blocking the platform channel handler.
+            unawaited(model._sendMouseUnchecked(type, mb).catchError((Object e) {
+              debugPrint('[InputModel] failed to send side button $type for $mb: $e');
+            }));
+          }
+        } else {
+          // Only route 'up' when we recorded the matching 'down';
```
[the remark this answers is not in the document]
sergiuszm: Both sides of the channel are in-process native code we control - the C handler only ever sends "back"/"forward" and "down"/"up". Adding validation for values that can't appear would be dead code. If the native side is ever extended with new values, the Dart handler would need updating regardless.

fufesou:
Hi @sergiuszm, please address https://github.com/rustdesk/rustdesk/pull/14848#pullrequestreview-4144466514 and the last comment in https://github.com/rustdesk/rustdesk/pull/14848#pullrequestreview-4151742152

"Resolve conversation" if they are solved or there is no need to change.

PR #14850, opened 2026-04-20
Fix default Android API version mismatch between vcpkg and rest of build (for working on android 6)

basilgello:
I had to invoke RustDesk on an ancient device running Android 6 and discovered the app could not start because of missing symbol `__write_chk` in `librustdesk.so`.

Debugging revealed that while we still use Android 21 as minimal SDK version, vcpkg maintainers have silently bumped default Android NDK API level to 28 in the default Android triplets across the years.

The fix is to keep Android triplets as overlay triplets in RustDesk source tree along with the overlay ports we already have.

I also refactored `build_android_deps.sh` as a separate commit to get rid of warnings by `shellcheck` and `shfmt` tools.

Built and tested on Android 6, 8, 9, 16.

## Summary by CodeRabbit

* **New Features**
  * Added Android build configurations for ARM, ARM64, x86 and x86_64 to support multiple ABIs.

* **Chores**
  * Improved build tooling: stricter environment validation, safer tool invocation, clearer usage/error messages, and more robust artifact handling.


basilgello:
@rustdesk @21pages

rustdesk:
Good job, but the change is huge, though some are renaming. I can not take this risk.

rustdesk:
> vcpkg maintainers have silently bumped default Android NDK API level to 28 in the default Android triplets across the years.

Really? interesting, rustdesk works quite well on old android 7. we fixed several crash bug related to android 7 last year.

basilgello:
Are you confused by many renames or API downgrade?

rustdesk:
YES

basilgello:
> Really? interesting, rustdesk works quite well on old android 7. we fixed several crash bug related to android 7 last year.

For the "slient" API bump on vcpkg side: on March 20 2025 the https://github.com/microsoft/vcpkg/pull/44424 was merged.

And it took some time to propagate changes to a "stable" baseline and then some more time for RustDesk to pick the new baseline.

So if the change was in April - May, then it could still use 25 as default API which is exactly Android 7.1: https://github.com/microsoft/vcpkg/pull/44424/files#diff-0e2a5cfbf5997f799a8d4404126cf9e1fef0732cdc58c2b4a866f5704d42c54e

basilgello:
> YES

OK, I can get rid of renames :)

To do that I can just introduce `res/vcpkg-triplets` directory so `res/vcpkg` still is used for ports.

Or we can patch triplets in `vcpkg` during build but its ugly :)

basilgello:
@rustdesk Okay, I have got rid of renames cluttering the PR.

If this PR ever gets merged, you can do the renames yourself:

```shell
mkdir res/vcpkg-ports
mv res/vcpkg/* res/vcpkg-ports
mv res/vcpkg-ports res/vcpkg/ports
mv res/vcpkg-triplets res/vcpkg/triplets
```

and edit `vcpkg.json` to replace `res/vcpkg` to `res/vcpkg/ports` and `res/vcpkg-triplets` to `res/vcpkg/triplets` to achieve the same result as these renames.

basilgello:
@rustdesk Did I address your concern or there is something more?

rustdesk:
## Issues / suggestions (claude code)

1. **Unprofessional error message.** `flutter/build_android_deps.sh` (new line ~75 in the patched build): `echo 'WTF: Where did installed artifact directory go?'`. Replace with something like `echo "ERROR: expected ${VCPKG_ROOT}/installed/arm-neon-android after vcpkg install, but it is missing" >&2`. Same line: in this branch the previous behavior was a silent skip, but now you `return 1`. That's a behavior change — if `vcpkg install` ever produces the artifact under a different name in a future vcpkg release, all armv7 CI builds break. Worth being intentional about.

2. **Lost CI debug output.** The old script had `head -n 100 "${VCPKG_ROOT}/buildtrees/ffmpeg/build-$VCPKG_TARGET-rel-out.log" || true` which surfaced the first 100 lines of the FFmpeg build log on success — useful for diagnosing CI flakes. The PR drops it. Non-blocking, but consider keeping it (or the `|| true` variant) since it's free insurance.

3. **API level duplication.** `VCPKG_CMAKE_SYSTEM_VERSION 21` is hardcoded in 4 triplet files. If/when min API gets bumped (and the Cargo/Gradle side can change independently), it's easy for these to drift again — exactly the failure mode this PR is fixing. A short comment in each file (e.g. `# Must match minSdkVersion in flutter/android/app/build.gradle`) would prevent the next regression.

4. **`pushd`/`popd` unchecked.** `pushd "${SCRIPTDIR}/.."` followed by `popd` doesn't check return values, and the script doesn't `set -euo pipefail`. Since the surrounding refactor is shellcheck-driven, adding `set -eu` at the top is a small additional win — though it could surface other latent issues, so reasonable to defer.

5. **Usage string only mentions one ABI.** `Usage: build_android_deps.sh <ANDROID-ABI>` is unchanged but doesn't list valid values. Minor: `Usage: build_android_deps.sh <arm64-v8a|armeabi-v7a|x86_64|x86>`.

6. **Mixed quoting style.** Some strings switched to single quotes (`'ERROR: ...'`, `'*** [Start] ...'`), others stayed double-quoted. Internally consistent enough, but worth noting for a follow-up cleanup.

rustdesk:
# Review (codex)

The Android API triplets are added correctly, but the refactored ARM rename path is no longer idempotent when reusing an existing vcpkg installation, which can break normal incremental builds.

## Findings

- [P2] Keep armeabi vcpkg reruns idempotent - `flutter/build_android_deps.sh:81`
  When `armeabi-v7a` deps are already built and the same `VCPKG_ROOT` is reused, the previous run has renamed `installed/arm-neon-android` to `installed/arm-android` while vcpkg's status database still marks the `arm-neon-android` packages installed. A subsequent `vcpkg install` can therefore no-op and leave no `arm-neon-android` directory, so this new `else` aborts even though the usable `arm-android` tree exists, breaking incremental/cached ARM builds.

basilgello:
@rustdesk Thanks for review! Or I should thank Anthropic and MS? ;)

Pushed the slighlty more polished version of `build_android_deps.sh` but it is worth to note the following:

1. The action to rename `arm-neon-android` to `arm-android` vcpkg directory inside `build_android_deps.sh` was a hack in 2023 and is still a hack in 2026.
   Proper fix is to edit the following places across crates so that `arm` architectecture on `android` is mapped to `arm-neon-android` not to `arm-android` (just like it is done for `x64-windows-static`):

   https://github.com/rustdesk/rustdesk/blob/master/build.rs#L61
   https://github.com/rustdesk/rustdesk/blob/master/libs/scrap/build.rs#L47
   https://github.com/rustdesk-org/magnum-opus/blob/master/build.rs#L32
   https://github.com/rustdesk-org/hwcodec/blob/master/build.rs#L135

   If these crates are modified and versions bumped appropriately, the part https://github.com/rustdesk/rustdesk/blob/master/flutter/build_android_deps.sh#L74 can be removed.

   Unfortunately I have no access to the crates, so I need you or @21pages / @fufesou to do it in one go.

2. Hardcoding 21 as Android NDK API level also should be refactored into `flutter-build.yml`.
   After this PR gets merged and vcpkg-triplets folder renames are applied, I will do a follow-up PR not to clutter this one.

What do you think?

rustdesk:
Thank you very much. Too busy lately, I will check later.

rustdesk:
# Code Review: PR #14850 — Fix default Android API version mismatch

## Overview

The PR fixes a regression where `librustdesk.so` failed to load on Android 6 (API 23) due to a missing `__write_chk` symbol. Root cause: vcpkg silently bumped the default Android NDK API level from 21 to 28 in its built-in triplets. The fix introduces local overlay triplets under `res/vcpkg-triplets/` that pin `VCPKG_CMAKE_SYSTEM_VERSION 21`, registers them in `vcpkg.json`, and refactors `flutter/build_android_deps.sh` to be shellcheck/shfmt-clean.

## What's good

- **Core fix is correct.** Pinning `VCPKG_CMAKE_SYSTEM_VERSION 21` in overlay triplets is the right lever — it isolates RustDesk from upstream vcpkg default drift and matches the documented minSdk.
- **Tight scope.** Only 7 files change (the diff is 215 lines); the earlier "51 file" rename churn appears to have been dropped.
- **Cleaner dispatch.** Moving the `ANDROID_ABI` switch above the build (instead of inside a `build()` function that was only ever called once) removes dead variables (`HOST_TAG`, `TOOLCHAIN`, `ABI`).
- **NEON symlink is now ABI-gated.** Wrapping the `arm-neon-android → arm-android` step in `if [ "${ANDROID_ABI}" = 'armeabi-v7a' ]` avoids running the check on every architecture.
- **`mv` → `ln -sf`** lets the script be re-run without losing the canonical install dir.

## Issues & suggestions

### Correctness

- **`ln -sf` doesn't handle an existing directory at the target** (`flutter/build_android_deps.sh:130`). If `arm-android` is a real directory from a previous `mv`-era build, `ln -sf foo arm-android` creates `arm-android/arm-neon-android` (a symlink *inside* the dir) instead of replacing it. Use `ln -snf` — the `-n` flag treats the existing destination dir as a file and replaces it.
- **Grammar / clarity in the error heredoc** (`flutter/build_android_deps.sh:137`): "'vcpkg install' seem to complete" → "seems to complete". Also the heredoc body prints to **stdout**; an `ERROR:` message belongs on stderr. Add `1>&2` to the `cat`.
- **Non-idiomatic heredoc.** `cat 0<<.a … .a` works but is jarring — the `0<<` is a redundant redirect from FD 0, and `.a` is an unusual delimiter. Prefer `cat >&2 <<EOF … EOF`.
- **Loss of FFmpeg debug log dump.** The removed `head -n 100 .../build-$VCPKG_TARGET-rel-out.log || true` is a noticeable hit to CI diagnosability — when FFmpeg fails on CI, that snippet is what people look at first. Worth keeping (under a `|| true` so a missing log doesn't fail the script).

### Robustness

- **No `set -euo pipefail`.** With the script's heavier use of `pushd/popd` and command chaining, the lack of `-e` means a failed `vcpkg install` followed by `popd` and the symlink step will still exit 0. Adding `set -euo pipefail` at the top would also subsume the manual empty-variable checks via `-u`.
- **`ANDROID_NDK_HOME` is checked but never used** in the new script body (the toolchain path was removed along with `TOOLCHAIN`). vcpkg's Android toolchain file consumes it via the environment, so the check is still meaningful — but a one-line comment noting "consumed by vcpkg's android toolchain" would prevent a future cleanup PR from removing the check.

### Consistency / style

- **Tab indentation in the .sh file.** The rest of the repo's shell scripts mix styles; shfmt's tab default is fine, but worth confirming with the maintainer (`@rustdesk` already flagged churn concerns).
- **All four triplets are near-duplicates.** Not blocking, but if a future change ever needs to touch all four (e.g., bumping to API 23 in a year), a small shared `.cmake` `include()` would help. Keep as-is for this PR — the duplication is shallow.

### Risk

- **API 21 vs Android 7 (API 24).** `@rustdesk` flagged that existing Android 7 work needs care. This PR explicitly *lowers* the floor back to 21, which restores Android 6 — but if any vcpkg dep (e.g., a newer openssl/curl rev pulled later) now needs ≥ 23 symbols, you'll see runtime link errors on API 21. Worth a CI smoke-load test on an API-21 emulator.
- **Behavior change with symlink vs `mv`.** Anything downstream that consumed `arm-neon-android/` directly used to find it empty after the build; it will now resolve through the symlink. Likely fine and intentional, but flag for reviewer attention.

### Test coverage

- No tests added or changed. Acceptable for a build-system fix, but the PR description's mention of manual validation on Android 6/8/9/16 should make it into the commit message or CI matrix if possible.

## Recommendation

**Approve with minor changes.** The fix itself (overlay triplets pinning API 21) is correct and well-scoped. The script refactor is a net win but ships a few small regressions worth fixing before merge: `ln -snf` instead of `ln -sf`, restore the FFmpeg log head, add `set -euo pipefail`, and clean up the heredoc (stderr + standard delimiter + grammar).

CatzTimes:
I can confirm this issue on a real old Android device.

Device:
- Vivo V3
- Snapdragon 616
- Android 5.1 (API 22)

RustDesk 1.4.3 and newer crash immediately on startup with:

java.lang.UnsatisfiedLinkError:
dlopen failed: cannot locate symbol "__write_chk"
referenced by "librustdesk.so"

RustDesk 1.4.1 still works correctly on the same device.

So this PR seems to match the exact issue affecting Android 5.x devices too.

I previously reported the same crash behavior here:
https://github.com/rustdesk/rustdesk/issues/15025

rustdesk:
@CatzTimes please test https://github.com/rustdesk/rustdesk/releases/tag/nightly

CatzTimes:
> @CatzTimes please test https://github.com/rustdesk/rustdesk/releases/tag/nightly

https://github.com/rustdesk/rustdesk/releases/download/nightly/rustdesk-1.4.9-armv7-signed.apk
[0.log](https://github.com/user-attachments/files/29890094/0.log)
There are no changes in the error log.

rustdesk:
> There are no changes in the error log.

@basilgello could you also confirm? It has been merged, but there is still `__write_chk ` problem in nightly build

PR #14855, opened 2026-04-20
Preserve leading and trailing spaces in OS passwords

ali-joghataei:
This PR fixes an issue where OS passwords were being trimmed before being saved or submitted.

Some operating system passwords may intentionally include whitespace at the beginning or end. In the current flow, the UI was calling trim() on the OS password in both Flutter and Sciter paths, which caused the stored/sent value to differ from what the user actually entered.

This change removes trimming only for the os-password field in the affected login and save flows, while keeping the existing behavior unchanged for other fields such as usernames and the main RustDesk password.

**Updated areas**
- Flutter OS password save flow
- Flutter session login flow
- Sciter OS password dialog flow
- Sciter session login flow

**Result**
- OS passwords are now stored and submitted exactly as entered
- Leading and trailing spaces are preserved
- Existing trimming behavior for unrelated fields remains unchanged

**Notes**
- No backend config change was required
- The issue was caused in the UI layer before persistence/submission

## Summary by CodeRabbit

* **Bug Fixes**
  * OS password fields now preserve exactly what users enter (including leading/trailing spaces) across prompts, dialogs, and login attempts.
  * Fixed an issue where the OS password could be incorrectly taken from the username input; passwords are now read from the correct password input.

### DIFF
--- a/flutter/lib/common/widgets/dialog.dart
+++ b/flutter/lib/common/widgets/dialog.dart
@@ -902,7 +902,7 @@ _connectDialog(
         }
       }
       final osUsername = osUsernameController?.text.trim() ?? '';
-      final osPassword = osPasswordController?.text.trim() ?? '';
+      final osPassword = osPasswordController?.text ?? '';
       final password = passwordController?.text.trim() ?? '';
       if (passwordController != null && password.isEmpty) return;
       if (rememberAccount) {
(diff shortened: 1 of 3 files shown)


rustdesk:
NO, we will not accept this PR. I would suggest you add a checkbox like "Preserve leading and trailing spaces" for this rather than changing default behavior.

PR #14914, opened 2026-04-26
fix(android): reduce Connected & Screen messages when host screen is off

### LINKED ISSUE #11479: RustDesk iOS shows "Successful - Connected, waiting for image..." when connecting to Android target with the screen off and dozing
### Bug Description

When I connect to an Android target using RustDesk iOS app, a message shows "Successful - Connected, waiting for image..." when connecting to Android target with the screen off (it has a screen lock, but it still has a screen lock timeout so it's not locked just a black screen and dozing).

### How to Reproduce

1. Make sure your Android target phone screen lock is enabled with pattern swipe.
2. Make sure your Android target phone screen lock time out is at least couple of minutes or more. 
3. Turn off the target screen (it will still not lock because of the timeout you configured in step 2). Wait for it to go into dozing state.
4. Check to make sure the screen is dozing by running the following command from adb: adb shell dumpsys power | grep -E "mWakefulness", it should return mWakefulness=Dozing.
5. Connect to the target from RustDesk iOS client app using any versions between 1.2.2 to 1.3.9 (last version).
### Actual Behavior

A message shows "Successful - Connected, waiting for image..." - You can't do anything, if you try to tap it nothing happens.

### Expected Behavior

Being able to see the phone screen and control it.

### Workaround

Turn on the Android target screen and then try to connect again.
Or use RustDesk iOS client app version 1.1.10 where this issue isn't happening (note that it has a different issue with a black screen showing when you first connect that you need to tap before seeing the screen: https://github.com/rustdesk/rustdesk/issues/7311 that I reported that I was hoping that it will get addressed in the newer versions but it looks like the problem become even worst).

iOS 14 -> Android 11

1.2.2-1.3.9 -> 1.1.10-1

### Screenshots

Image

### Additional Context

The same issue doesn't happen with TeamViewer under the same conditions - So if the screen is off and dozing, you can still wake it up when you connect. RustDesk iOS client app version 1.1.10 was able to wake up from dozing, but since version 1.2.2 it no longer is able to allow you to wake up the screen. The issue also happens if you try to connect to it from RustDesk Desktop client version 1.1.9.

amirHdev:
## Summary
The Android controlled-side wake flow for incoming screen share sessions, when an authorized session starts and the Android host is non interactive, it wakes the device proactively. where clients can get stuck on `Connected, waiting for image...` when the Android host screen is off.

Wake behavior now runs automatically on authorized incoming Android screen share sessions, `HOME` may change device state differently too.

Closes #11479

## Summary by CodeRabbit

* **New Features**
  * Activate idle devices automatically with coordinated wake management and timeouts.
  * Automatic screen refresh with one retry and a “cannot wake screen” notification when limits are exceeded.
  * Throttled wake attempts to avoid repeated wake spam.
  * Keyguard-aware behavior to avoid unwanted UI actions when the device is locked.
  * Optional automatic navigation to the home screen when accessibility input is open and cancellation of pending wake/refresh when the service stops.

### DIFF
--- a/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
+++ b/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
@@ -91,7 +91,8 @@ class InputService : AccessibilityService() {
 
     private val volumeController: VolumeController by lazy { VolumeController(applicationContext.getSystemService(AUDIO_SERVICE) as AudioManager) }
 
-    fun wakeUpDevice() {
+    // Just HOME. Caller handles the actual wake lock.
+    internal fun goHome() {
         performGlobalAction(GLOBAL_ACTION_HOME)
     }
 
--- a/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
+++ b/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
@@ -91,6 +91,10 @@ class InputService : AccessibilityService() {
 
     private val volumeController: VolumeController by lazy { VolumeController(applicationContext.getSystemService(AUDIO_SERVICE) as AudioManager) }
 
+    fun wakeUpDevice() {
+        performGlobalAction(GLOBAL_ACTION_HOME)
+    }
+
     @RequiresApi(Build.VERSION_CODES.N)
     fun onMouseInput(mask: Int, _x: Int, _y: Int) {
         val x = max(0, _x)
--- a/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/MainService.kt
+++ b/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/MainService.kt
@@ -62,6 +62,8 @@ const val VIDEO_KEY_FRAME_RATE = 30
 
 class MainService : Service() {
 
+    private val wakeRetryHandler = Handler(Looper.getMainLooper())
+
     @Keep
     @RequiresApi(Build.VERSION_CODES.N)
     fun rustPointerInput(kind: Int, mask: Int, x: Int, y: Int) {
@@ -128,6 +130,7 @@ class MainService : Service() {
                     }
                     if (authorized) {
                         if (!isFileTransfer && !isStart) {
+                            wakeAndRefreshIncomingScreenIfNeeded("authorized_connection")
                             startCapture()
                         }
                         onClientAuthorizedNotification(id, type, username, peerId)
@@ -228,6 +231,41 @@ class MainService : Service() {
     private lateinit var notificationChannel: String
     private lateinit var notificationBuilder: NotificationCompat.Builder
 
+    private fun wakeScreen(reason: String) {
+        if (wakeLock.isHeld) {
+            Log.d(logTag, "WakeLock release before wake, reason:$reason")
+            wakeLock.release()
+        }
+        Log.d(logTag, "Wake screen, reason:$reason")
+        wakeLock.acquire(5000)
+    }
+
+    private fun wakeAndRefreshIncomingScreenIfNeeded(reason: String) {
+        if (powerManager.isInteractive) {
+            return
+        }
+        wakeRetryHandler.removeCallbacksAndMessages(null)
+        wakeScreen(reason)
+        wakeRetryHandler.postDelayed({
+            if (powerManager.isInteractive) {
+                Log.d(logTag, "Skip delayed wake refresh, device already interactive, reason:$reason")
+                return@postDelayed
+            }
+            // HOME works better here than faking a tap.
+            InputService.ctx?.wakeUpDevice()
+            FFI.refreshScreen()
+        }, 500)
+        wakeRetryHandler.postDelayed({
+            if (powerManager.isInteractive) {
+                Log.d(logTag, "Skip retry wake refresh, device already interactive, reason:$reason")
+                return@postDelayed
+            }
+            wakeScreen("$reason-retry")
+            InputService.ctx?.wakeUpDevice()
+            FFI.refreshScreen()
+        }, 1200)
+    }
+
     override fun onCreate() {
         super.onCreate()
         Log.d(logTag,"MainService onCreate, sdk int:${Build.VERSION.SDK_INT} reuseVirtualDisplay:$reuseVirtualDisplay")
@@ -249,6 +287,7 @@ class MainService : Service() {
     }
 
     override fun onDestroy() {
+        wakeRetryHandler.removeCallbacksAndMessages(null)
         checkMediaPermission()
         stopService(Intent(this, FloatingWindowService::class.java))
         super.onDestroy()
--- a/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/MainService.kt
+++ b/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/MainService.kt
@@ -62,19 +62,32 @@ const val VIDEO_KEY_FRAME_RATE = 30
 
 class MainService : Service() {
 
+    companion object {
+        private const val WAKELOCK_TIMEOUT_MS = 5000L
+        private const val WAKE_POLL_INTERVAL_MS = 100L
+        private const val WAKE_MAX_WAIT_MS = 2000L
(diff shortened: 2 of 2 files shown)


rustdesk:
# Code Review: PR #14914

**Title:** fix(android): reduce Connected & Screen messages when host screen is off
**Closes:** #11479

## Overview

When an Android device is the host (controlled side) and its screen is off / non-interactive, remote clients get stuck on "Connected, waiting for image..." because `MediaProjection` does not emit frames while the display is in doze-suspend. This PR adds a wake-and-refresh sequence in the `add_connection -> authorized` path, before `startCapture()`.

### Key changes

- **`InputService.kt`** — adds `wakeUpDevice()` which calls `performGlobalAction(GLOBAL_ACTION_HOME)`.
- **`MainService.kt`** — adds `wakeRetryHandler`, `wakeScreen()`, and `wakeAndRefreshIncomingScreenIfNeeded()`. Reuses the existing `wakeLock(ACQUIRE_CAUSES_WAKEUP | SCREEN_BRIGHT_WAKE_LOCK)`. Sequence: t=0 acquire wake lock → t=500ms HOME + `FFI.refreshScreen()` → t=1200ms re-acquire + retry. Cleanup is done in `onDestroy`.

---

## How do other tools "wake" Android?

| Tool | Approach | Notes |
|---|---|---|
| **scrcpy** | Injects `KEYCODE_WAKEUP` at session start, `KEYCODE_POWER` / `KEYCODE_SLEEP` on exit | Goes through ADB / `InputManager.injectInputEvent`, requires system uid or adb shell |
| **TeamViewer Host / AnyDesk** | OEM add-ons (signed with vendor keys) call system-only APIs like `PowerManager.goToSleep`. The non-OEM build holds `SCREEN_BRIGHT_WAKE_LOCK` continuously to keep the screen *from sleeping*, rather than waking it after the fact | True "wake" needs OEM privilege |
| **AirDroid / Vysor** | Bring up a transparent Activity with `FLAG_KEEP_SCREEN_ON` + `setShowWhenLocked` to keep the screen alive during the session | Also sidesteps the "already off, must wake" problem |
| **KDE Connect / Plasma** | `PowerManager.WakeLock(FULL_WAKE_LOCK or ACQUIRE_CAUSES_WAKEUP)` — same family as this PR, deprecated since API 17 | Likewise cannot unlock; only turns the panel on |

**Conclusion:** without a system-signed APK, there is no legitimate way to wake an Android device from doze *and* unlock it. Third-party apps have two realistic options: (a) keep a wake lock held continuously so the screen never sleeps, or (b) use `ACQUIRE_CAUSES_WAKEUP` to bring the panel up to the lock screen. This PR picks (b).

---

## Is the approach reasonable?

**Reasonable in principle, but flawed in detail.**

### What's right
- The sequencing is correct: take the wake lock to power up the panel → wait for SurfaceFlinger to come back → trigger `HOME` to nudge WindowManager into a redraw → push a key frame via `FFI.refreshScreen()`. That is the standard remediation for `MediaProjection` going silent after doze-suspend.
- Cleaning up the handler in `onDestroy` is correct.

### Issues

1. `SCREEN_BRIGHT_WAKE_LOCK` and `ACQUIRE_CAUSES_WAKEUP` have been **deprecated since API 17**. Many ROMs (Xiaomi MIUI, Huawei EMUI, Meizu Flyme) silently ignore them, so hit rate in the China/OEM market will be poor.
2. `performGlobalAction(GLOBAL_ACTION_HOME)` is a **no-op while the device is non-interactive**. The PR uses a fixed 500 ms delay to wait for `isInteractive`, but wake-up latency varies wildly across devices (Pixel ~300 ms, older devices 1 s+). 1200 ms with only one retry is a rigid heuristic. A `BroadcastReceiver` for `Intent.ACTION_SCREEN_ON`, or a short polling loop on `powerManager.isInteractive`, would be far more robust.
3. `wakeScreen()` is a misleading name — it only releases/acquires a wake lock; the actual wake/HOME/refresh logic lives in `wakeAndRefreshIncomingScreenIfNeeded`.
4. **No fallback / no signal on failure.** If after 1200 ms the device is still not interactive (accessibility disabled, lock screen, OEM blocking), there is no further retry, no log warning, and no notification to the remote — the user experience degrades back to the original "waiting for image" with no diagnostic.
5. `onDestroy` clears the handler but does not call `wakeLock.release()`. The instance is shared with `rustPointerInput`'s wake path; if a 5-second wake lock is still held during destruction it will delay the system entering doze.
6. Magic numbers `500`, `1200`, `5000` should be named constants.

---

## Security implications

These are the risks introduced by the change, ordered by severity.

### 1. Lock-screen content disclosure (medium — most important)
`ACQUIRE_CAUSES_WAKEUP` only **turns the panel on**; it does **not** unlock the device. What lights up is the lock screen, which typically displays:

- Notification previews (SMS 2FA codes, banking / IM message snippets, email subjects)
- Media controls, calendar reminders
- AOD / Always-On info

Immediately afterward, `FFI.refreshScreen()` forces a fresh frame to the remote. The result: **any "authorized" remote session can wake the device without the owner's knowledge and capture the lock-screen contents — including 2FA codes and message previews.**

The previous behavior ("screen is off → remote only sees black") was effectively an implicit security barrier. This PR removes it. Mitigations:

- Check `KeyguardManager.isKeyguardLocked()`. When locked, either skip `refreshScreen()` entirely or substitute a placeholder frame.
- Surface a UI / notification telling the owner that a remote session has woken the screen and is capturing the lock-screen view.

### 2. Remote nuisance / coercion — wake DoS (low–medium)
"Authorized" includes any *always-allow* client. An attacker who was approved once can repeatedly initiate connections to:

- Flash the screen at night for harassment / coercion.
- Hold wake locks back-to-back, drain the battery, defeat doze, and shorten device life.
- There is **no throttling**: every authorized add_connection runs `acquire(5000)` and triggers `HOME`.

Mitigation: rate-limit (e.g. at most one wake per N seconds), or only wake on the first connection in a session per peer.

### 3. Side effects of HOME (low)
`performGlobalAction(GLOBAL_ACTION_HOME)` will:

- **Dismiss sensitive foreground UI** the owner might be using — banking app, KeyChain, an authenticator app entering a code, etc. If a wake fires at the wrong moment, the owner is yanked back to the launcher and any in-progress input is lost or misrouted.
- Interrupt system dialogs (permission prompts, biometric prompts).
- Race with the owner's own input.

Not a vulnerability per se, but coupling a global UI action to "remote authorized" is a privilege-amplification surface.

### 4. Increased exposure of `InputService.ctx?.wakeUpDevice()` (informational)
`wakeUpDevice()` is now a publicly reachable method on a static-held `InputService` reference, with no caller authentication. In-process risk is low today, but if any future Binder/IPC surface ever exposes `InputService`, this becomes an LPE pivot. Recommend `internal` visibility plus a caller-stack assertion ("MainService only").

### 5. Frame-leak race (theoretical)
`MediaProjection` behavior during doze-suspend is OEM-dependent. In edge cases, `refreshScreen()` may pick up the **last frame buffered before sleep** (e.g. a banking screen, payment page) and ship it to the remote. Android usually clears the buffer, but some OEMs do not. Same root cause as #1; same mitigations apply.

---

## Recommendations (priority order)

1. **Check `KeyguardManager.isKeyguardLocked`** inside `wakeAndRefreshIncomingScreenIfNeeded`. If locked, either skip wake entirely, or wake without pushing a frame, and notify the owner via UI. This is the most important fix.
2. **Throttle**: cap to N wakes per session, with a cooldown of ≥10 s.
3. Replace fixed 500/1200/5000 ms timings with named constants and prefer an `ACTION_SCREEN_ON` `BroadcastReceiver` over a static delay.
4. Add `if (wakeLock.isHeld) wakeLock.release()` in `onDestroy`.
5. Add a fallback when `performGlobalAction(GLOBAL_ACTION_HOME)` fails (accessibility disabled): log a warning and notify the remote, instead of silently leaving the client on "waiting for image".
6. Make `InputService.wakeUpDevice` `internal` (or guard it with an access check).

**Bottom line:** the direction is correct and consistent with industry option (b); the implementation, however, attaches a side-effecting global action to "authorized" with no throttle and **opens a lock-screen capture path**. The lock-screen / Keyguard guard must be in place before this is merged.

amirHdev:
@rustdesk Can you sanity check the lockscreen and notification preview concern with this approach?

amirHdev:
@rustdesk what you think ?

PR #14918, opened 2026-04-27
let accept window popup after windows session chosen

rustdesk:
fix: https://www.reddit.com/r/rustdesk/comments/1svrk2c/terminal_servers_rust_window_pops_up_on_all/

Test cases
- [ ] multiple sessions
- [ ] one session case
- [ ] auto reconnect
- [ ] non-windows
- [ ] If introduce security (auth) problem.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Windows login behavior to ensure the permission window displays only in the confirmed session when users select a specific Windows session during login.

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -315,6 +315,15 @@ pub struct Connection {
     #[cfg(not(any(target_os = "android", target_os = "ios")))]
     terminal_user_token: Option<TerminalUserToken>,
     terminal_generic_service: Option<Box<GenericService>>,
+    // When the host has multiple sessions and the client must pick one,
+    // defer notifying the CM (which would pop the permission window in the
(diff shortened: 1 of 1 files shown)


21pages:
*  `SelectedSid` is only handled after authorization. For click-approval mode, authorization depends on showing the accept window first. If accpet window is not shown before auth, click approval cannot work.

- [ ] `try_start_cm_ipc()` also needs to be deferred in this scenario. If the  accept windowprocess starts but does not receive any client/Login data, Flutter accept windowtreats `_clients` as empty and closes itself after about 6 seconds.

- [ ] `wait_for_windows_session_id_confirm` is only enabled in `handle_windows_specific_session()`, so branches that call `try_start_cm()` before that still send `Data::Login` immediately and can still show the accept window in the wrong session.

PR #14975, opened 2026-05-06
fix: iOS auto-reconnect stall when prior session was on a non-primary display

michaelotis:
### Summary
iOS RustDesk → Windows host with multiple monitors hangs at "Connected, Waiting for image" after auto-reconnecting from a network drop, but only when the prior session was viewing a non-primary display. Reconnects from primary always succeed.

### Root cause
On reconnect the iOS client does **not** re-issue its display selection (no `CaptureDisplays` / `SwitchDisplay` arrives). The host correctly defaults `display_idx` to `PRIMARY_DISPLAY_IDX` in `Connection::start()` and brings up only `monitor0`. The iOS UI is still in "display N" mode from the previous session and silently waits for / drops mismatched frames.

Verified against real service logs from the failing case: host emits `new video service: monitor0` and the matching snapshot, but `monitor1` is never created during the entire auto-reconnected session lifetime.

### Fix
Persist the most recent active display set per remote peer id in a new `LAST_VIEW_BY_PEER` map (in-memory, same pattern as the adjacent `LOGIN_FAILURES` / `SESSIONS` statics). On `handle_login_request_without_validation`, look up the stored set for the peer's `my_id` — if it differs from primary-only, validate against currently-present displays (drops stale entries from unplugged monitors), pre-set `self.display_idx` so the upcoming `PeerInfo` carries the restored index, and stash the validated set in `pending_display_set`. After `try_sub_monitor_services()` finishes, `apply_pending_display_set()` runs `self.capture_displays(&[], &[], &set)` to restore the capturer.

Behaviour for the common case (primary-only / fresh peer / anonymous peer with empty `my_id`) is unchanged.

### Verification
Reproduced and verified on Windows 11 host (`flutter,hwcodec,vram` build) with a 2-monitor setup against an iOS RustDesk client through self-hosted hbbs 1.1.15.

**Failing run (before patch):** `monitor0` created on reconnect; no `monitor1` ever created; iOS stalls indefinitely.

**Passing run (this patch):**
```
L36   #1381 Connection opened
L84   Persisted view [1] for peer 1130614991
L111  #1381 Connection closed: Timeout            <- airplane mode 30s
L144  #1382 Connection opened                     <- auto-reconnect
L145  Will restore previously viewed displays [1] for peer 1130614991
L155  Restoring previously viewed display set [1] after services subscribed
L162  new video service: monitor1                 <- without any iOS message
```
iOS resumes display 2 video; no stall.

### Notes
- Persistence is in-memory only (matches existing per-peer state pattern; lost on host restart, which is acceptable since the iOS client also reissues its session view on a fresh connection).
- Single-file diff: `src/server/connection.rs`.

## Summary by CodeRabbit

* **New Features**
  * Per-peer display preferences are remembered and automatically restored on reconnect and login, improving session continuity.
  * Last-view is persisted when users switch displays so previous displays are retained across connections.
  * Restored display sets are validated and clamped to available displays to avoid invalid selections.

* **Bug Fixes**
  * Display switch requests received before video/services are ready are deferred and applied once services initialize, preventing lost switches.

### DIFF
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -76,6 +76,10 @@ lazy_static::lazy_static! {
     static ref SWITCH_SIDES_UUID: Arc::<Mutex<HashMap<String, (Instant, uuid::Uuid)>>> = Default::default();
     static ref WAKELOCK_SENDER: Arc::<Mutex<std::sync::mpsc::Sender<(usize, usize)>>> = Arc::new(Mutex::new(start_wakelock_thread()));
     static ref WAKELOCK_KEEP_AWAKE_OPTION: Arc::<Mutex<Option<bool>>> = Default::default();
+    // Last-active display set per remote peer id. Restored on next login from
+    // the same peer so that auto-reconnect resumes on the previously selected
+    // display (iOS does not re-issue its display selection on auto-reconnect).
+    static ref LAST_VIEW_BY_PEER: Arc::<Mutex<HashMap<String, Vec<usize>>>> = Default::default();
 }
 
 fn constant_time_eq(a: &[u8], b: &[u8]) -> bool {
@@ -292,6 +296,14 @@ pub struct Connection {
     file_remove_log_control: FileRemoveLogControl,
     last_supported_encoding: Option<SupportedEncoding>,
     services_subed: bool,
+    // Buffered SwitchDisplay (from early arrival) and pending display set to
+    // restore after services are subscribed. Together these fix the iOS
+    // auto-reconnect "waiting for image" hang when the prior session was on
+    // a non-primary display: iOS does not re-issue its display selection on
+    // auto-reconnect, so the host must restore the previously-active display
+    // set itself based on remembered per-peer state.
+    pending_display_switch: Option<usize>,
+    pending_display_set: Option<Vec<usize>>,
     delayed_read_dir: Option<(String, bool)>,
     #[cfg(target_os = "macos")]
     retina: Retina,
@@ -482,6 +494,8 @@ impl Connection {
             file_remove_log_control: FileRemoveLogControl::new(id),
             last_supported_encoding: None,
             services_subed: false,
+            pending_display_switch: None,
+            pending_display_set: None,
             delayed_read_dir: None,
             #[cfg(target_os = "macos")]
             retina: Retina::default(),
@@ -1733,6 +1747,14 @@ impl Connection {
                         self.retina.set_displays(&displays);
                     }
                     pi.displays = displays;
+                    // Clamp against the just-built `pi.displays` in case
+                    // self.display_idx was pre-set in handle_login_request_without_validation
+                    // from a snapshot that has since become stale (a monitor
+                    // unplugged in the brief window between login validation
+                    // and PeerInfo emit).
+                    if self.display_idx >= pi.displays.len() {
+                        self.display_idx = *display_service::PRIMARY_DISPLAY_IDX;
+                    }
                     pi.current_display = self.display_idx as _;
                     #[cfg(not(any(target_os = "android", target_os = "ios")))]
                     {
@@ -1790,6 +1812,8 @@ impl Connection {
         } else if sub_service {
             if !wait_session_id_confirm {
                 self.try_sub_monitor_services();
+                self.apply_pending_display_switch().await;
+                self.apply_pending_display_set().await;
             }
         }
         true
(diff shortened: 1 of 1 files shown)


michaelotis:
Thanks @coderabbitai — both findings are valid, addressed in the latest push (commit a638848).

### Changes since the initial commit

1. **Display enumeration failure now skips restore** instead of trusting stored indices. On `try_get_displays()` error we emit a `log::warn!` and use an empty Vec, so a transient enumeration failure can't push an out-of-range `current_display` into `PeerInfo`.
2. **Explicit `SwitchDisplay` clears `pending_display_set`** in the deferred path. A real client-side display selection now wins over login-time auto-restore — previously a pending restore could revert the user's newer choice once `apply_pending_display_set()` ran.
3. **Camera-mode peers no longer write to `LAST_VIEW_BY_PEER`** (in both `capture_displays` and `switch_display_to`). Persisting only when `video_source == VideoSource::Monitor` prevents camera-mode display indices from leaking into the monitor restore path on a future session.

### Known limitations (deliberately deferred)

- **`LAST_VIEW_BY_PEER` is unbounded.** Same lifetime semantics as `LOGIN_FAILURES`/`SESSIONS` — happy to add LRU/cap eviction if maintainers prefer.
- **Multi-display restore has a 1-frame `current_display` desync.** The pre-set of `self.display_idx` only fires when `valid.len() == 1`; for multi-display sets `PeerInfo` ships with primary as `current_display` and the actual set is applied right after via `apply_pending_display_set`. Single-display is the dominant case for iOS and works cleanly.
- **Stored indices are positional, not display-identity-bound.** If a user remaps which physical monitor is primary in Windows between sessions, the stored index could point to a different physical display. Not a regression — the existing display-selection protocol has the same property.
- **`wait_session_id_confirm` path:** if the host is in multi-user-session-confirm mode and the client never sends `SelectedSid`, the pending state is consumed by the eventual `try_sub_monitor_services` call from the SelectedSid handler. If that handler never runs (connection drops), the state is dropped — no worse than the pre-patch behaviour, which had no restore at all.

Verified end-to-end on Windows 11 (`flutter,hwcodec,vram` build) → iOS RustDesk via self-hosted hbbs 1.1.15: prior session on display 2 → 30s airplane-mode timeout → auto-reconnect cleanly resumes display 2.

michaelotis:
@coderabbitai good catch — fixed in `f44a967`.

`apply_pending_display_set` now re-validates the stored set against `display_service::try_get_displays()` immediately before calling `capture_displays`, instead of trusting the validation that ran at login time. This closes the window where a monitor can be unplugged or remapped between login and apply (notably during a `wait_session_id_confirm` / `SelectedSid` delay).

Also extracted the validation logic into a `Self::validate_display_set` helper so login-time and apply-time use identical rules and we don't drift.

michaelotis:
@coderabbitai applied in `915bf72`.

`pi.current_display = self.display_idx` is now preceded by a clamp against `pi.displays.len()` (with fallback to `*display_service::PRIMARY_DISPLAY_IDX`), so even if a monitor was unplugged in the brief window between login-time validation and PeerInfo emit, the client receives a safe, in-range `current_display` rather than a stale snapshot value.

michaelotis:
@coderabbitai thanks — agreed it's a worthwhile follow-up. As I noted up-thread under "Known limitations", the unbounded growth of `LAST_VIEW_BY_PEER` matches the lifetime semantics of the adjacent `LOGIN_FAILURES`/`SESSIONS` statics, so I'd rather defer the eviction policy (size cap vs. time-based vs. LRU) and any new metrics surface to a separate PR once a maintainer decides which knob fits the project's broader preferences. Happy to bolt it on here if a maintainer would prefer it inline.

michaelotis:
@Copilot thanks — addressing both points (commit `89f6f6f`):

### Wayland safety (line 3848) — fixed

Switched `validate_display_set()` from `display_service::try_get_displays()` to `display_service::get_sync_displays()`. `get_sync_displays()` reads from the cached `SYNC_DISPLAYS` so we don't reintroduce the `Display::all()` blocking/crash risk that `check_displays_changed()` deliberately avoids on Wayland.

If `SYNC_DISPLAYS` is empty (e.g., a login racing `update_get_sync_displays_on_login`), the helper now returns an empty `Vec`, which the calling `if !valid.is_empty()` guard already handles by conservatively skipping the restore — strictly safer than the previous code path which would have called `Display::all()` and on Wayland could have hung.

### Unbounded `LAST_VIEW_BY_PEER` growth (line 82) — same call as the prior CodeRabbit nitpick

This is the same concern @coderabbitai flagged earlier; my reasoning is unchanged ([up-thread](https://github.com/rustdesk/rustdesk/pull/14975#issuecomment-4392693335)): the lifetime semantics match `LOGIN_FAILURES` / `SESSIONS` / `SWITCH_SIDES_UUID`, none of which have eviction either, so adding it asymmetrically here would be inconsistent. The `(Instant, Vec<usize>)` + TTL pattern you suggested is a nice fit if maintainers want eviction across the board — happy to either bolt it on here or open a follow-up issue, but I'd rather not invent a policy unilaterally.

PR #14981, opened 2026-05-07
fix(mobile): resolve Chinese IME input buffering on iOS

nickhoo555:
## Summary

After selecting a Chinese character candidate from the pinyin IME on iOS, the committed character is not sent to the remote until the next keystroke. For example, typing 你好 (characters sent correctly on macOS/Android) results in only the first character appearing on the remote, with the second delayed until further input.

## Root Cause

Flutter upstream bug [#174159](https://github.com/flutter/flutter/issues/174159): on iOS, `TextEditingController.value.composing` does not reset to `(-1,-1)` after the IME commits text. The stale composing range still covers the committed CJK characters, so the code that blocks input forwarding while composing (to avoid sending pinyin letters) incorrectly continues to block the committed characters.

## Fix

Detect whether the text inside the composing range contains Latin letters (a-z). Real pinyin/romaji composing always produces Latin letters. If the composing range covers only non-Latin characters (CJK), it is the stale Flutter bug case — treat it as committed text and send immediately.

## Tested

iPad mini (iOS 26.4) with Chinese Pinyin IME. Typing "你好吗" now sends each character immediately upon candidate selection.

## Test plan

- [ ] iOS Chinese pinyin input sends characters immediately after candidate selection
- [ ] iOS English input unaffected
- [ ] macOS/Android input unaffected (composing range resets correctly on these platforms)

## Bug Fixes
* Fixed iOS keyboard input behavior for CJK characters by improving detection between text being composed versus already committed to ensure proper input handling.
* Optimized the soft keyboard's text processing for more reliable and accurate character entry on iOS devices.

### DIFF
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -157,6 +157,12 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
     }
   }
 
+  // Tracks the composing range from the previous onChanged call.
+  // Comparing against the previous composing range lets us detect when
+  // characters transition from composing → committed, even when the iOS
+  // IME hasn't fully cleared the composing range yet.
+  TextRange _lastComposing = TextRange.empty;
+
   // For client side
   // When swithing from other app to this app, try to sync clipboard.
   void trySyncClipboard() {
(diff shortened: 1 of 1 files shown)


rustdesk:
Please upload video before and after the fix.

rustdesk:
We will address this ourself.

PR #14985, opened 2026-05-08
Harden os password （terminal windows and headless linux) anti brute force

fufesou:
Windows terminal sessions could previously submit OS username/password attempts without a dedicated rate or failure policy. The new policy tracks OS credential failures separately and applies a backoff window after repeated failures.

## Changes

- Add a dedicated OS-credential failure policy with exponential backoff for terminal OS login attempts.
- Add a Windows terminal OS-login concurrency gate so only one credential check runs at a time.
- Move terminal OS credential authorization into the logon response path, and start CM IPC only after terminal OS authorization succeeds.
- Add Linux headless OS-auth checks before desktop startup, and record password failures for failed headless X-session login attempts.
- Validate the target device before running Linux headless OS-auth work, so wrong-target requests do not consume failure budget or trigger PAM/X-session work.
- Split Linux X-session startup errors into auth and environment failures, and avoid logging usernames or PAM details for auth failures.

## Desc

1. For terminal OS-login, CM IPC startup is deferred until the OS credential check and terminal service-user validation have completed. This avoids starting the connection-manager IPC path for failed terminal OS-login attempts. Terminal sessions without a supplied OS username keep the existing current-session user flow.

2. Blocked OS-credential attempts use the existing alarm audit path for backoff and concurrency events.

3. Terminal OS-login failures intentionally share the existing per-source login failure bucket with regular password login. This keeps the lockout behavior conservative across authentication entry points, while the OS-credential backoff adds an additional global throttle for OS password attempts.

4. Linux headless startup now checks the existing login-failure state before trying to start a desktop session with supplied OS credentials. Password-related X-session failures are mapped to the normal password error path, while environment/startup failures keep the desktop-session error path.

5. Linux headless OS-auth work now runs only after the request has been validated for the current device. This avoids touching PAM or X-session startup for wrong-target requests.

6. Linux authentication failures use a generic log detail instead of including the username or PAM error text, so credential-failure metadata is not exposed in logs.

7. The OS-credential backoff is intentionally not keyed by client IP. A client may reconnect through different source IPs, so an IP-only policy would be easy to bypass.

8. The backoff state is also not cleared by a single successful OS-credential login. This is intentional: after too many failed attempts, the penalty should stay in place for the idle reset window. To clear it immediately, restart the RustDesk service.

9. Some rare failure messages reuse existing English text, such as `"Please try 1 minute later"`. These paths are not expected during normal use, and adding new translated strings for them is not necessary for this change.

## Tests

- [x] Normal connect
- Terminal connect
  - [x] password
  - [x] accept
- Terminal (Run as administrator)
  - [x] connect
  - [x] wrong username/password 

## Summary by CodeRabbit

* **New Features**
  * Login throttling with exponential backoff and scoped failure tracking for OS-credential logins.
  * Enhanced terminal OS login authorization with concurrency gating and conditional startup sequencing.
  * Earlier rejection of invalid terminal usernames and improved handling of authorization flows.

* **Bug Fixes**
  * Clearer, context-sensitive login messages distinguishing authentication vs environment/startup failures.
  * More consistent handling and reporting of desktop start and OS-credential failures across platforms.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rustdesk/rustdesk/pull/14985)

### DIFF
--- a/src/platform/linux_desktop_manager.rs
+++ b/src/platform/linux_desktop_manager.rs
@@ -29,6 +29,19 @@ lazy_static::lazy_static! {
     static ref DESKTOP_MANAGER: Arc<Mutex<Option<DesktopManager>>> = Arc::new(Mutex::new(None));
 }
 
+const DESKTOP_LOGIN_MIN_FAILURE_DELAY: Duration = Duration::from_secs(1);
+
+#[derive(Debug)]
+struct DesktopLoginAuthFailed;
+
+impl std::fmt::Display for DesktopLoginAuthFailed {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("desktop login auth failed")
+    }
+}
+
+impl std::error::Error for DesktopLoginAuthFailed {}
+
 #[derive(Debug)]
 struct DesktopManager {
     seat0_username: String,
@@ -94,11 +107,24 @@ fn detect_headless() -> Option<&'static str> {
     None
 }
 
-pub fn try_start_desktop(_username: &str, _passsword: &str) -> String {
+pub fn try_start_desktop(_username: &str, _password: &str) -> String {
+    try_start_desktop_with_auth_result(_username, _password).message
+}
+
+pub(crate) struct DesktopStartResult {
+    pub message: String,
+    pub auth_failed: bool,
+}
+
+pub(crate) fn try_start_desktop_with_auth_result(
+    _username: &str,
+    _password: &str,
+) -> DesktopStartResult {
     debug_assert!(crate::is_server());
-    if _username.is_empty() {
+    let normalized_username = _username.trim();
+    if normalized_username.is_empty() {
         let username = get_username();
-        if username.is_empty() {
+        let message = if username.is_empty() {
             if let Some(msg) = detect_headless() {
                 msg
             } else {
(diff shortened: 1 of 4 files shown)


fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

rustdesk:
• The patch introduces regressions in terminal current-session authorization and Linux PAM error classification that can cause hangs/task leaks or incorrect lockouts in specific but realistic environments.

  Full review comments:

  - [P2] Validate current-session terminal logins before CM IPC — /Users/hzhou/rustdesk/src/server/connection.rs:2445-2446
    On Windows terminal requests that omit os_login.username, this branch now starts CM IPC before fill_terminal_user_token() runs; start_ipc waits while the host is at prelogin, and the approval/no-password
    branches below only call prepare_terminal_login_for_authorization() for OS-login attempts. In that current-session prelogin case the old code immediately returned “No active console user…”, but the new
    flow can leave the client/pending IPC task waiting until someone logs in, so run terminal preparation for the session-user path before starting/prompting CM.
  - [P2] Preserve PAM system errors as environment failures — /Users/hzhou/rustdesk/src/platform/linux_desktop_manager.rs:334-337
    When PAM authentication fails because of a system/config/account error rather than bad credentials, this branch still converts every authenticate() error to Auth; try_start_desktop() then reports
    LOGIN_MSG_PASSWORD_WRONG, and the connection code records it as a login failure. A PAM outage or misconfiguration can therefore consume the headless login failure budget and lock out clients instead of
    reporting an environment/startup failure; only map real credential failures to Auth and keep other PAM errors as Env with sanitized details.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14985#issuecomment-4414269875

1. [P2] Validate current-session terminal logins before CM IPC

Fixed in aab7c6c.

The control (local) side shows: "No active console user logged on, please connect and log on first.".

However, an external connection manager process (the "accept window") also runs in the background.

It would be better to avoid running this process.

2. [P2] Preserve PAM system errors as environment failures

Keep as is.

Rare case and no secure issue.

rustdesk:
• The patch leaves a failure-accounting bypass in a core security-hardening path: terminal OS credential failures do not accumulate in the intended shared per-IP bucket when the RustDesk password is valid.

  Review comment:

  - [P2] Defer clearing failures until OS auth succeeds — /Users/hzhou/rustdesk/src/server/connection.rs:2583-2583
    In the Windows terminal OS-login path with a valid RustDesk password but invalid os_login credentials, this clears the default per-IP failure bucket before send_logon_response_and_keep_alive() runs
    prepare_terminal_login_for_authorization(). The later OS credential failure then re-adds only a fresh single entry, so repeating with the same valid RustDesk password prevents OS-login failures from
    accumulating in the shared per-IP lockout bucket; defer this clear until terminal OS authorization has succeeded.

fufesou:
https://github.com/rustdesk/rustdesk/pull/14985#issuecomment-4414467971

1. [P2] Defer clearing failures until OS auth succeeds — /Users/hzhou/rustdesk/src/server/connection.rs:2583-2583

Keep as is.

RustDesk password success clears RustDesk password failures; the separate OS credential policy tracks OS credential failures and is intentionally not cleared by RustDesk password success.

rustdesk:
But it does not address the newer line 3698 issue. In the PR patch, prepare_terminal_login_for_authorization() calls:

  update_failure_with_scope(failure, true, 0, FailureScope::TerminalOsLogin)

  after OS credential success. The remove = true path in update_failure_with_scope deletes the shared LOGIN_FAILURES entry without checking the scope. That means OS credential success can clear RustDesk
  password failure state before RustDesk auth or approval is complete.

  So the distinction is:

  - 2583: RustDesk password success clears RustDesk password failures. The linked explanation can justify this.
  - 3698: OS credential success also clears RustDesk password failures. The linked explanation does not justify this, and it conflicts with the stated separation.

  The likely fix is: TerminalOsLogin success should not remove the default LOGIN_FAILURES bucket; only completed RustDesk authentication should clear it.

fufesou:
> https://github.com/rustdesk/rustdesk/pull/14985#issuecomment-4414826662

Done. Terminal OS login is now fully separated from LOGIN_FAILURES[0] and only uses the OS credential policy.

21pages:
Tested:

  * RustDesk password:
    * Per source IP, more than 6 wrong attempts within the same minute are blocked with the 1-minute retry message.
    * A successful RustDesk password login before the hard limit clears the failure bucket for that IP.
    * Failure buckets are separated by source IP.

  * RustDesk password hard limit:
    * After more than 30 accumulated wrong attempts from the same source IP, further login attempts are rejected before password verification.
    * Before reaching this hard limit, a successful RustDesk password login clears the accumulated failures.
    * After the hard limit is reached, entering the correct RustDesk password no longer clears the failure state because password verification is not reached.

  * Windows Terminal Run as administrator:
    * The first two wrong OS credential attempts do not trigger backoff.
    * Starting from the third wrong OS credential attempt, backoff is applied and the backoff duration increases with subsequent failures.
    * A successful OS credential login does not clear the OS credential failure count or backoff state.
    * The OS credential failure count is reset only after the idle reset window, currently 2 hours without new OS credential failures.

21pages (approved):

PR #15004, opened 2026-05-10
Allow macOS monitor switching in privacy mode

wysuperfly:
## Summary
- allow desktop clients to show monitor controls while macOS privacy mode is active
- skip the legacy Display 1 requirement for macOS privacy mode
- keep the existing privacy-mode display-switch restriction for other peer platforms

## Rationale
macOS privacy mode blacks out all online displays via CoreGraphics gamma tables and fails the privacy-mode enable path if not all displays can be protected. Switching the remote display therefore does not weaken the local privacy protection for macOS peers.

## Testing
- git diff --check
- not run: dart/flutter is not available in this local environment

## Summary by CodeRabbit

* **Bug Fixes**
  * Display switching now functions during privacy mode on macOS, removing the previous restriction that required switching to Display 1 first.
  * Monitor-selection menu visibility improved for multi-display setups when privacy mode is active on macOS.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rustdesk/rustdesk/pull/15004)

### DIFF
--- a/flutter/lib/common/widgets/toolbar.dart
+++ b/flutter/lib/common/widgets/toolbar.dart
@@ -16,6 +16,12 @@ import 'package:get/get.dart';
 
 bool isEditOsPassword = false;
 
+// macOS privacy mode blacks out all online displays, so switching the remote
+// display does not weaken the local privacy protection.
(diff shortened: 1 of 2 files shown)


luochen28:
macOS is fully supported. When will it support Windows?

PR #15057, opened 2026-05-16
fix: show hide cm in settings and fixed window not being hidden on new connection

BaniGuru:
Enabled settings for **Hide Connection management window** in Security tab

image

Disabled when not using password

image

On connection, no connection management window shows, making the user control remote screen without displaying this window so that the remote user does not get distracted,
really required when doing presentation

Applied fix in the connection area to wait for setting load and hide cm window initially when this settings enabled. There was bug before which was causing this window to always display due to setting wasn't loaded with await before creating screen and has been fixed with this change.

Also, the checkbox in settings wasn't refreshing after user change this settings and has been fixed using setHideCm added function.

**Reason for change**: Sometime connection drop and reconnects making this window shows in middle and distract the presentation. Also sometime remote user wants control in middle of presentation and don't want to distract users with this window displayed while presentation is happening.

## Release Notes

* **Bug Fixes**
  * Fixed connection manager window hiding state synchronization between user settings and application behavior
  * Connection manager window now correctly loads hiding preference from persisted settings during startup
  * Improved consistency of hiding behavior when password verification or specific approval modes are enabled
  * Enhanced real-time synchronization when users toggle the hiding setting

[Review Change Stack](https://app.coderabbit.ai/change-stack/rustdesk/rustdesk/pull/15057?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/flutter/lib/desktop/pages/desktop_setting_page.dart
+++ b/flutter/lib/desktop/pages/desktop_setting_page.dart
@@ -1227,8 +1227,8 @@ class _SafetyState extends State<_Safety> with AutomaticKeepAliveClientMixin {
             if (usePassword && !isChangePermanentPasswordDisabled())
               _SubButton('Set permanent password', setPasswordDialog,
                   permEnabled && !locked),
-            // if (usePassword)
-            //   hide_cm(!locked).marginOnly(left: _kContentHSubMargin - 6),
+            if (usePassword)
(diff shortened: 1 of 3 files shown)


rustdesk:
We have stated our reason about why, won't state again.

PR #15061, opened 2026-05-16
fix(android): clean up stale disconnected session in session_add (fixes #15060)

### LINKED ISSUE #15060: Android client: tap-connect hangs at "Connecting..." indefinitely after extended background; force-stop fixes
### Bug Description

On the Android client, after the app has been open in the background for an extended period (hours-to-days, typically combined with the OS deep-sleeping the app via Doze and/or network switches between WiFi and cellular), tapping a peer ID to connect results in a "Connecting..." spinner that never resolves and no error message ever appears. **Force-stopping the app from Android Settings → Apps → RustDesk → Force Stop and reopening makes the very next connect attempt succeed immediately.**

The wedge is purely in-process — the device's network is healthy throughout (the same phone can reach an HTTPS URL on the rendezvous server via the system browser during the failed-connect window). The Rust core never opens a socket for the connect attempt, so the broker sees **zero packets** from the phone's WAN IP during the stuck-state.

Mechanism (verified by code reading; tracker search for terms like "android background spinner", "android doze", "stuck connecting", "force stop" returned no matching reports):

1. On mobile, `sessionId` is a **constant per-app-lifetime UUID** (`_constSessionId` at `flutter/lib/models/model.dart:57`), unlike desktop which generates a fresh UUID per session.
2. When a prior connection wedges silently — typically because Android Doze pauses the io_loop long enough that no error path fires and Flutter never calls `sessionClose` — the stale session remains in the global `SESSIONS` map.
3. On the next user tap, `FFI.start` (`model.dart:3677`) calls `bind.sessionAddSync`. Rust's `session_add` bails with `"same session id is found"` (`src/flutter.rs:1331` in 1.4.6).
4. **The Flutter side ignores this error** — the return is bound to a `// ignore: unused_local_variable` at `model.dart:3723`.
5. `bind.sessionStart` is then called regardless. `session_start_` finds the stale handler with its old `event_stream` still attached, computes `is_connected = h.event_stream.is_some() == true`, and the short-circuit at `flutter.rs:1411` skips spawning a new `io_loop`. The function returns `Ok(())`.
6. Flutter shows "Connecting..." and waits for events that never arrive. No timeout fires because no `Client::start` ever ran, so `CONNECT_TIMEOUT` (18s) never gets to apply.

Foreground network switches do **not** trigger this — while the app is awake, the existing TCP/UDP error paths fire on stream errors, `io_loop` calls `on_establish_connection_error`, the Flutter UI dismisses the connection page, and `RemotePage.dispose() → gFFI.close() → bind.sessionClose() → remove_session_by_session_id` cleans up the stale session before the next attempt.

### How to Reproduce

1. Configure the Android client with a known-working ID/rendezvous server. Connect to a peer successfully once.
2. Leave the app open in the background. Switch the phone between WiFi and cellular at least once. Let the device sleep for several hours.
3. Re-open the app. Tap connect on the same peer ID (or a different one).
4. → "Connecting..." spinner forever, no error.
5. Force-stop the app, reopen, try the same connect → succeeds immediately.

Hard to reproduce on-demand: requires hours-to-days of real-phone backgrounding; emulators do not reliably reproduce. Easier **synthetic** repro: kill the rendezvous broker mid-session, wait for io_loop's 30-second receive watchdog to fire (state transitions to `Disconnected`), then try to reconnect to the same peer.

### Expected Behavior

Tapping connect should either succeed (using fresh sockets bound to the current default network) or surface an error within a reasonable timeout. Process force-stop should not be required to recover.

Android (controller) -> any (irrelevant — bug is purely on controller side)

1.4.6 (verified by code reading against main as of 2026-05-16)

### Screenshots

Not applicable — UI shows the standard "Connecting..." spinner indefinitely, with no error to capture. Logcat from inside the stuck state would be the diagnostic of interest; reporter has not yet captured this because force-stopping (the only known recovery) ended the diagnostic window before logs were pulled. Happy to provide a logcat dump on next reproduction if maintainers want it.

### Additional Context

A PR with a minimum-diff fix follows: detect the stale-disconnected case in `session_add` and clean up the prior session before bailing. The fix piggybacks on the existing `ConnectionRoundState` machinery — when `state == Disconnected`, the prior io_loop has provably exited (it calls `set_disconnected(round)` on exit at `src/client/io_loop.rs:341`) and the session can be safely removed without racing.

Tangentially related: comment at `src/rendezvous_mediator.rs:264` (added 2024-02-17 in `964c2ed2b5`) reads "temporarily disable exponential backoff for android before we add wakeup trigger to force connect in android" — the wakeup-trigger work it anticipates would also help the heartbeat path recover faster from this class of state, but the load-bearing fix for the user-visible bug is the session-state cleanup in `session_add`.

mwaddip:
Fixes #15060.

- Mobile reuses one constant `sessionId` per app lifetime (`_constSessionId` at `flutter/lib/models/model.dart:57`), unlike desktop which generates a fresh UUID per session.
- When a prior connection wedges silently after Android Doze suspends the io_loop long enough that no error path fires and `sessionClose` is never called, the entry stays in the `SESSIONS` map.
- The next user tap calls `sessionAddSync`, which bails with `"same session id is found"`. The Flutter side **ignores** this error (`// ignore: unused_local_variable` at `flutter/lib/models/model.dart:3723`) and calls `sessionStart` regardless.
- `session_start_` then finds the stale handler with its old `event_stream` still attached, the `is_connected` short-circuit at `src/flutter.rs:1411` prevents a new `io_loop` from spawning, and the user is stuck on `"Connecting..."` with no error — no `Client::start` ever runs, so `CONNECT_TIMEOUT` never fires.
- Fix: detect the stale case in `session_add` via the existing `ConnectionRoundState::Disconnected` state (the io_loop sets it on exit at `src/client/io_loop.rs:341`) and clean up the prior session before bailing. Safe — when state is `Disconnected`, the prior `io_loop` has provably exited.
- Foreground network switches are unaffected because the existing error paths fire while the app is awake, which leads to `RemotePage.dispose() → gFFI.close() → bind.sessionClose() → remove_session_by_session_id` cleaning up the session before the next attempt.

Diff is small: +18 / −3 across two files. Adds a `pub fn is_disconnected(&self) -> bool` accessor on `ConnectionRoundState` to expose the existing state machine outside its module.

## Tests

- [x] `cargo check --features flutter` on Linux x86_64 — passes
- [x] `cargo ndk -t arm64-v8a check --features flutter` (cross-compile for Android target) — passes
- [ ] On-device repro on a real Android phone with a self-hosted broker. Original reporter (filer of #15060) is set up for this but the bug is non-deterministic (hours-to-days of backgrounding). Synthetic repro path: kill broker mid-session, wait for io_loop's 30-second receive watchdog to fire (state → `Disconnected`), then reconnect to the same peer. Will follow up here once verified.

Open to any naming / placement adjustments — happy to iterate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session management to properly handle stale disconnected sessions, allowing new connections to proceed instead of being rejected when reconnecting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rustdesk/rustdesk/pull/15061?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/flutter.rs
+++ b/src/flutter.rs
@@ -1324,11 +1324,22 @@ pub fn session_add(
 
     // to-do: check the same id session.
     if let Some(session) = sessions::get_session_by_session_id(&session_id) {
-        if session.lc.read().unwrap().conn_type != conn_type {
+        // Mobile reuses one constant session_id for the entire app lifetime
+        // (_constSessionId in flutter/lib/models/model.dart). When a prior
+        // connection wedges silently — typically after Android Doze pauses the
+        // io_loop long enough that no error path fires and session_close is
+        // never called — the entry stays in SESSIONS. Detect that here and
+        // clean it up so the new session_add can succeed instead of bailing,
+        // which would leave the user stuck on a "Connecting..." spinner with
+        // no error surfaced (the Flutter side ignores sessionAddSync's return).
+        if session.connection_round_state.lock().unwrap().is_disconnected() {
+            sessions::remove_session_by_session_id(&session_id);
+        } else if session.lc.read().unwrap().conn_type != conn_type {
             bail!("same session id is found with different conn type?");
+        } else {
+            // The same session is added before?
+            bail!("same session id is found");
         }
-        // The same session is added before?
-        bail!("same session id is found");
     }
 
     LocalConfig::set_remote_id(&id);
--- a/src/ui_session_interface.rs
+++ b/src/ui_session_interface.rs
@@ -128,6 +128,10 @@ impl ConnectionRoundState {
             true
         }
     }
+
+    pub fn is_disconnected(&self) -> bool {
+        matches!(self.state, ConnectionState::Disconnected)
+    }
 }
 
 impl Default for ConnectionRoundState {


21pages:
@mwaddip Thanks for the detailed write-up.

I still have a few questions.

  1. On mobile, my current understanding is:
  - `sessionAddSync()` is only called when opening a new remote session page

https://github.com/rustdesk/rustdesk/blob/377547fa1128823d6c5a4ea17b1310640264713b/flutter/lib/mobile/pages/remote_page.dart#L94

https://github.com/rustdesk/rustdesk/blob/377547fa1128823d6c5a4ea17b1310640264713b/flutter/lib/models/model.dart#L3724

  - `sessionClose()` is called when that page is disposed

https://github.com/rustdesk/rustdesk/blob/377547fa1128823d6c5a4ea17b1310640264713b/flutter/lib/mobile/pages/remote_page.dart#L138

https://github.com/rustdesk/rustdesk/blob/377547fa1128823d6c5a4ea17b1310640264713b/flutter/lib/models/model.dart#L3937

When the remote page is closed, the old session should already be removed. So if the user has already left the remote page and then starts a new connection, I would not expect a stale session to still be there. 

  ```rust
  if session
      .connection_round_state
      .lock()
      .unwrap()
      .is_disconnected()
  {
      sessions::remove_session_by_session_id(&session_id);
  }
```

Could you add a temporary log right before `sessions::remove_session_by_session_id(&session_id)` and share the log output? I want to check whether this code path is really reached when you hit this issue on Android.

  For Android logs, an easy way is:

  - on the `Share screen` tab, enable `Transfer file`
  - restart the app, so permission for writing logs to external storage takes effect
  - then collect the logs from the path documented here: https://github.com/rustdesk/rustdesk/wiki/FAQ#access-logs

  2. Could you describe your "kill hbbs/hbbr and reconnect" reproduce steps in more detail? I haven’t been able to reproduce it locally. One important detail is that auto-reconnect does not call sessionAddSync(), so I want to understand the exact sequence in your test that leads back into session_add().
  3. In `io_loop.rs`, when `set_disconnected()` is called, the later cleanup work has not finished yet. That cleanup also calls `has_sessions_running()`. Because of that, calling `remove_session_by_session_id()` immediately after seeing `is_disconnected()` may cause cleanup problems or races with the old session’s shutdown path. So I think it would be better to first confirm the two points above. If this is indeed caused by a stale session not being removed, then we can look at the safest way to clean it up.

**Comment by mwaddip** [comment]
Hey, thanks for your answer. On encountering the actual bug live last night it seems the issue is indeed deeper. Pressing Cancel is supposed to remove the session but the issue of not being able to connect remains. 

Building a debug build so I can dive into it proper.

**Comment by mwaddip** [comment]
Got a debug build with entry-point logs on the affected phone. Wedge captured live today:

```
[wedge-debug] session_add ENTRY id=... session_id=4416cc77-...
[wedge-debug] session_start_ ENTRY id=... session_id=4416cc77-...
```

Then silence — no `rendezvous server: ...`, no `Client::start`, no io_loop activity. So Rust IS called on every tap; `session_start_` is short-circuiting at the `!is_connected && is_first_ui_session` check before spawning io_loop. RemotePage.dispose → sessionClose doesn't run before the new connect (process kept alive across background; SESSIONS still holds the stale entry).

Will refactor to address your race concern (avoid removing during io_loop's post-`set_disconnected` cleanup window) and add a log on each branch to pin down session_add bail vs is_connected vs is_first_ui_session.

mwaddip:
Closing in favor of #15143.

Live diagnostics — a native thread dump of a wedged client plus self-hosted hbbs/hbbr logs — disproved the premise here: on the wedge, `io_loop` is alive and the session is still connected and streaming, not disconnected and not failing to spawn. The actual cause is that `sessionClose` can be skipped when the app is backgrounded while `RemotePage.dispose()` is running, leaking the session; the reconnect then re-attaches to it (mobile reuses a constant `sessionId`) and stays on "Connecting…". #15143 fixes that directly. Full write-up in #15060.

PR #15062, opened 2026-05-16
feat: add JIS keyboard support for macOS in RustDesk (AI)

### LINKED ISSUE #15003: JIS keyboard is not well supported
### Bug Description

The key mappings from Windows to macOS are not well supported when using a JIS keyboard.

1. The output is the `ANSI` keyboard layout.
2. Japanese can't be typed in Codex, but it can be typed in Claude Code.

### How to Reproduce

1. Windows -> macOS
2. Use the "Map mode"
3. Input "Shift" + "2", the output is `@`. But it should be `"` when using a `JIS` keyboard.
4. Input Japanese in Claude code and Codex. 

### Expected Behavior

1. Only `ISO` and non-`ISO` layouts are available now—`JIS` should also be supported.
2. It should work fine in Codex.

Windows -> osx

1.4.6 -> 1.4.6

### Screenshots

No.

### Additional Context

_No response_

amirHdev:
This fixes the issue where Shift+2 produced @ instead of " enables Japanese input in Codex and preserves full ANSI/ISO layout support as a fallback

Fixes #15003

## Summary by CodeRabbit

* **New Features**
  * Added JIS (Japanese) keyboard support on macOS with fallback to standard mapping.
  * Added a new config option: "enable-perm-change-in-accept-window".
  * Exposed keyboard layout types (Unknown/ANSI/ISO/JIS) and keyboard-type-aware input handling for macOS.

* **UI**
  * Keyboard layout chooser now shows explicit layout options (ANSI/ISO/JIS/Unknown), uses a responsive picker, and increased dialog height.

* **Bug Fixes**
  * Improved detection and handling of macOS keyboard input sources for more consistent mapping.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rustdesk/rustdesk/pull/15062?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6920,7 +6920,7 @@ dependencies = [
 [[package]]
 name = "rdev"
 version = "0.5.0-2"
-source = "git+https://github.com/rustdesk-org/rdev#f9b60b1dd0f3300a1b797d7a74c116683cd232c8"
+source = "git+https://github.com/fufesou/rdev?rev=154c76278d4ab27a3dce5739a3d4af1a9e42f0e4#154c76278d4ab27a3dce5739a3d4af1a9e42f0e4"
 dependencies = [
  "cocoa 0.24.1",
  "core-foundation 0.9.4",
@@ -6929,6 +6929,7 @@ dependencies = [
  "dispatch",
  "enum-map",
  "epoll",
+ "foreign-types 0.3.2",
  "inotify",
  "lazy_static",
  "libc",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,6 +104,9 @@ portable-pty = { git = "https://github.com/rustdesk-org/wezterm", branch = "rust
 system_shutdown = "4.0"
 qrcode-generator = "4.1"
 
+[patch."https://github.com/rustdesk-org/rdev"]
+rdev = { git = "https://github.com/fufesou/rdev", rev = "154c76278d4ab27a3dce5739a3d4af1a9e42f0e4" }
+
 [target.'cfg(target_os = "windows")'.dependencies]
 winapi = { version = "0.3", features = [
     "winuser",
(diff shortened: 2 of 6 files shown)


fufesou:
@amirHdev Please confirm you can reproduce the issue and that this PR fixes it.

I doubt it will, because "Map mode" doesn't work with `enigo`.

https://github.com/rustdesk/rustdesk/issues/15003#issuecomment-4468820582

amirHdev:
@fufesou Yes,
I addressed the map mode concern by moving the macOS map mode simulation path to rdev’s keyboard type aware VirtualInput ( as visual testing for now and for validation together ) . now selected macOS keyboard type through KeyEvent and applies ANSI , ISO, JIS or Unknown when simulating raw macOS key events

Neustradamus:
@amirHdev: Thanks for your work!

fufesou:
The changes in https://github.com/fufesou/rdev/pull/2 (**AI implementation**) seemed too significant back then, but it actually just adds keyboard type support.

amirHdev:
@rustdesk  Could you clarify what (AI) in the PR title is meant to indicate?
Does it refer to the linked fufesou/rdev#2 change or to this PR itself? If there’s a specific concern about the implementation, I’m happy to address it directly

fufesou:
> @rustdesk Could you clarify what (AI) in the PR title is meant to indicate? Does it refer to the linked [fufesou/rdev#2](https://github.com/fufesou/rdev/pull/2) change or to this PR itself? If there’s a specific concern about the implementation, I’m happy to address it directly

https://github.com/rustdesk/rustdesk/pull/15062#issuecomment-4525077618

https://github.com/fufesou/rdev/pull/2 is too complicated.

amirHdev:
Thank you that makes sense @fufesou
To be sure, is the AI problem primarily related to `fufesou/rdev#2`, instead of this PR?
It is used in this PR since macOS JIS keyboard emulation relies on keyboard-aware key simulation. If the primary problem is the dependency I can make the dependency smaller or even split this PR to be only for RustDesk
Thanks

fufesou:
> If the primary problem is the dependency
Yes

> I can make the dependency smaller

You can do it.

Splitting this PR to be only for RustDesk may not be a good idea.

PR #15072, opened 2026-05-19
Fix iOS CJK soft keyboard input

CodeEagle:
## Summary

Fixes iOS soft keyboard handling for CJK IMEs in remote sessions.

The previous iOS text diff logic treated any active composing range as text that should not be sent. Some iOS IMEs keep a composing range briefly after the user confirms a candidate, so committed Chinese text could be held until another key, such as space, caused another text change.

## Changes

- Extracted iOS soft keyboard diffing into `ios_soft_keyboard_input.dart`.
- Tracks pending composing text so pinyin, stroke, bopomofo, Japanese kana, and Korean jamo/hangul composition are not sent prematurely.
- Sends committed CJK text immediately via `sessionInputString`, including Japanese kanji/kana and Korean hangul once composition is committed.
- Uses a text controller listener on iOS so commits where only the composing range changes are still observed.
- Added focused tests for Chinese, Japanese, Korean, ASCII, and backspace behavior.

## Validation

```bash
cd flutter
fvm flutter test --no-pub test/ios_soft_keyboard_input_test.dart test/input_modifier_utils_test.dart
```

All tests passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved iOS soft keyboard input handling for better support of input composition in languages such as Chinese, Japanese, and Korean.

* **Tests**
  * Added comprehensive test suite for iOS soft keyboard input diffing logic.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rustdesk/rustdesk/pull/15072?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -22,6 +22,7 @@ import '../../models/input_model.dart';
 import '../../models/model.dart';
 import '../../models/platform_model.dart';
 import '../../utils/image.dart';
+import '../utils/ios_soft_keyboard_input.dart';
 import '../widgets/dialog.dart';
 import '../widgets/custom_scale_widget.dart';
 
@@ -66,6 +67,7 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
   Orientation? _currentOrientation;
   final _uniqueKey = UniqueKey();
   Timer? _iosKeyboardWorkaroundTimer;
+  String? _iosComposingValue;
 
   final _blockableOverlayState = BlockableOverlayState();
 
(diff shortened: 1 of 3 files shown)


fufesou:
`1` can't be sent to the controlled/remote side. It also deletes the previous characters.

fufesou:
I can reproduce the Bopomofo issue on iOS with the Traditional Chinese Zhuyin keyboard.

Steps: type `ㄋ`, then `ㄧ`, then the third-tone mark `ˇ`, but do not select the candidate yet. The remote side already receives input before `你` is committed. The final committed character can still become `你`, but the uncommitted composing text should not be sent to the remote side at all.

Possible fix: treat Bopomofo tone marks like `ˇ`, `ˊ`, `ˋ`, and `˙` as part of Bopomofo composition, and keep holding the text while the composing range is still valid. A test with `ㄋㄧˇ` in an active composing range should cover this case.

====================== update ===========================

 I also saw a similar issue with the Japanese keyboard. While composing mixed ASCII/kana text, the remote side receives the local composing text before I select a candidate. After selecting the candidate, the remote text changes to the selected result, but the uncommitted composing text should not be sent in the first place.

This may not be the same path as the Bopomofo issue, since ASCII -> Japanese kana is already handled in  `_isComposingTransition()`. It may be worth checking whether the composing range is valid for those intermediate updates, and whether `previousComposingValue` is preserved across them.

rustdesk:
Close since no response.

PR #15143, opened 2026-05-29
fix(android): close session on dispose to prevent reconnect wedge

### LINKED ISSUE #15060: Android client: tap-connect hangs at "Connecting..." indefinitely after extended background; force-stop fixes

mwaddip:
## Problem

Fixes #15060.

On the Android/Flutter client, closing a session and immediately reconnecting to the same peer (with an app background/foreground in between) leaves the client stuck on "Connecting…" indefinitely — while the original connection is still alive and streaming.

Root cause: `RemotePage.dispose()` only reaches `bind.sessionClose` at the tail of `gFFI.close()`, behind several awaits (the `enable_soft_keyboard` platform call, `setCanvasConfig`, `imageModel.update`). If the app is backgrounded while the page is disposing, dispose can be suspended before reaching it, so the Rust session is never torn down — io_loop, the relay socket, and the session handler leak. Because mobile reuses a constant `sessionId`, the next reconnect re-attaches to the leaked session in `session_start_` (it correctly skips spawning a second io_loop) and never re-emits the connection state, so the UI stays on "Connecting…" while the orphaned io_loop keeps streaming.

## Fix

Dispatch `sessionClose` at the start of `dispose()`, so teardown happens synchronously on route pop, before any backgrounding can interrupt it. The `sessionClose` in `gFFI.close()` becomes a no-op once the session is already removed. One line + comment, mobile-only.

## Verification

Reproduced and fixed against a self-hosted hbbs/hbbr (used as ground truth for client-side vs server-side):

- **Before:** a single relay session that never closes across close+reconnect; a native thread dump of the wedged client shows `io_loop` alive and the relay still streaming video underneath the "Connecting…" dialog.
- **After:** close → server logs `Relay … closed` → reconnect → a fresh relay session. Confirmed end-to-end at the UI, at the broker (relay closed + new request), and at the socket (fresh established session).

## Note

This supersedes #15061, which targeted a "stale disconnected session blocks io_loop spawn" premise. Live diagnostics disproved that: io_loop spawns fine and the session is connected (not disconnected) — the actual defect is the teardown being skippable when the app is backgrounded mid-dispose.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remote session disconnection handling to ensure sessions close promptly, preventing connections from getting stuck in "Connecting…" state and reducing session leakage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rustdesk/rustdesk/pull/15143?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -126,6 +126,14 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
   @override
   Future<void> dispose() async {
     WidgetsBinding.instance.removeObserver(this);
+    // Close the session up-front. `gFFI.close()` below only calls `sessionClose`
+    // after several awaits (canvas save, image update, the `enable_soft_keyboard`
+    // platform call), so if the app is backgrounded while this page is disposing,
+    // dispose can be suspended before reaching it and the connection is never torn
+    // down. The reconnect then re-attaches to the leaked session and is stuck on
+    // "Connecting...". Dispatching it here makes teardown happen synchronously on
+    // pop; the `sessionClose` in `gFFI.close()` becomes a no-op once removed.
+    unawaited(bind.sessionClose(sessionId: sessionId));
     // https://github.com/flutter/flutter/issues/64935
     super.dispose();
     gFFI.dialogManager.hideMobileActionsOverlay(store: false);


21pages:
I agree that an unclosed session could explain the behavior. If the previous session is not properly closed, the controlling side will not start a new `Client::start` process. The client will remain stuck on "Connecting", while the accept window on the controlled side still shows the connection as active.

However, could you provide logs demonstrating the issue before and after this fix? The proposed root cause is that `dispose()` may be interrupted before it completes, leaving the session uncleared. We would need some evidence (logs, traces, or a reproducible scenario) to confirm that this is actually happening and that the fix addresses the underlying issue.

`sessionClose` is now being called twice. On Android, there is currently no session cleanup logic that depends on the session after the earlier sessionClose call, so this does not cause any issues at the moment. However, additional cleanup logic that requires access to the session may be added in the future, which could make this behavior problematic.

rustdesk:
• The change can race with mobile's reused session ID and the existing later cleanup path, allowing a delayed dispose to close a newly created session. This affects the reconnect scenario the patch is
  intended to fix.

  Review comment:

  - [P2] Prevent late cleanup from closing a fresh session — `/Users/hzhou/rustdesk/flutter/lib/mobile/pages/remote_page.dart:136-136`
    On mobile the sessionId is reused, and this fire-and-forget close runs while the same dispose() still later awaits and calls `gFFI.close()`, which calls bind.sessionClose again. If the user reconnects
    after this early close removes the old Rust session but before the delayed cleanup resumes (for example in the background/foreground flow this PR targets), the later close can target the newly added
    session with the same ID and tear down the fresh connection.

mwaddip:
Yes. The natural trigger is a timing race — the app being backgrounded while `RemotePage.dispose()` is mid-flight, before it reaches `sessionClose` — so for a deterministic, repeatable capture I used a debug build with extra logging in `session_add` / `session_start_` / `session_close`, and reproduced the leak by short-circuiting the teardown (equivalent to dispose being suspended before `sessionClose` runs). It's also visible independently on my self-hosted relay (hbbr).

### Before — leak: close doesn't tear the session down, so the reconnect re-attaches to the still-live session

Client:
```
session_start_ branch check: is_connected=false ... → SPAWNING io_loop   # connect
                                                                         # close (back → OK): no session_close ran — session leaked
session_start_ branch check: is_connected=true  ... → SKIP spawn         # reconnect re-attaches → UI stuck on "Connecting..."
```
Relay (hbbr) — the first connect's relay session just stays open across close+reconnect (it only closes ~23 min later, when I reinstalled):
```
12:06:00  New relay request 4efeb247… :36959 got paired     # connect
          (close + reconnect happen here — no relay events at all)
12:29:32  Relay of :36959 closed                            # only when the process was replaced
```

### After — fix: close tears the session down, reconnect is fresh

Client:
```
session_start_ branch check: is_connected=false ... → SPAWNING io_loop   # connect
session_close found and removing session_id=1b740119…                    # close → teardown runs up-front
session_start_ branch check: is_connected=false ... → SPAWNING io_loop   # reconnect = fresh
```
Relay (hbbr):
```
11:53:45  New relay request b84ac48b… :37685 got paired     # connect
11:53:58  Relay of :37685 closed                            # close → torn down
11:54:05  New relay request bdd5fa01… :33463 got paired     # reconnect = fresh session
```

So the leak signature is: close produces no `session_close`, the relay never closes, and the reconnect re-attaches (`is_connected=true → SKIP`). The fix makes `sessionClose` run synchronously at the start of `dispose()`, so teardown can't be skipped and the reconnect is a clean fresh connect.

For reference, this matches a real (non-forced) wedge I caught earlier: a native thread dump of the live wedged client showed `io_loop` still alive and the relay still streaming video (~2.3 Mbps, Recv‑Q 0) underneath the "Connecting..." dialog, and the broker showed that relay session had never closed across the user's close+reconnect. Happy to share those raw captures too.

mwaddip:
> • The change can race with mobile's reused session ID and the existing later cleanup path, allowing a delayed dispose to close a newly created session. This affects the reconnect scenario the patch is intended to fix.
> 
> Review comment:
> 
>     * [P2] Prevent late cleanup from closing a fresh session — `/Users/hzhou/rustdesk/flutter/lib/mobile/pages/remote_page.dart:136-136`
>       On mobile the sessionId is reused, and this fire-and-forget close runs while the same dispose() still later awaits and calls `gFFI.close()`, which calls bind.sessionClose again. If the user reconnects
>       after this early close removes the old Rust session but before the delayed cleanup resumes (for example in the background/foreground flow this PR targets), the later close can target the newly added
>       session with the same ID and tear down the fresh connection.

Tested this, the race can be reproduced although in reality it's a matter of milliseconds. The real issue is deeper and would require a refactor of the mobile architecture: each connection gets its own FFI/sessionId like desktop already does (desktop is FFI(null) → fresh Uuid(); mobile hard-codes _constSessionId)

The current fix is technically a workaround that resolves the main issue at the cost of a potential race that is unlikely to ever hit in reality. Addressing the race would become a workaround over a workaround.

PR #14933, opened 2026-04-30
Keyboard shortcuts

rustdesk:
feat(shortcuts): user-configurable keyboard shortcuts for session actions              

<img width="1196" height="432" alt="7525e96d-398f-4a37-95b9-c04446a01bf4" src="https://github.com/user-attachments/assets/f48798e9-60ce-4bea-a0b8-d800964f9eab" />

<img width="1388" height="1534" alt="66fe3207-85ef-46eb-9672-89ac5a20b832" src="https://github.com/user-attachments/assets/4ea40e25-ff95-45cb-bac1-ba6cb8e48008" />
                                                                                                                      
                                                                                                                                                                                                               
  Adds a keyboard shortcut feature (Rust matcher + Dart UI + cross-language                                                                                                                                    
  parity tests) that lets users bind combinations like Ctrl+Alt+Shift+P to                                                                                                                                     
  session actions. Bindings are stored in LocalConfig under                                                                                                                                                    
  `keyboard-shortcuts`; the matcher gates dispatch on `enabled` and                                                                                                                                            
  `pass_through` flags so flipping the master switch off is a hard stop.                                                                                                                                       
                                                                                                                                                                                                               
  Wire-up summary:                                                                                                                                                                                             
  - src/keyboard/shortcuts.rs: matcher, default bindings, parity test against                                                                                                                                  
    flutter/test/fixtures/default_keyboard_shortcuts.json                                                                                                                                                      
  - src/keyboard.rs: shortcut intercept in process_event{,_with_session},                                                                                                                                      
    feature-gated to `flutter`; runs before key swapping so users bind to                                                                                                                                      
    physical keys                                                                                                                                                                                              
  - src/flutter_ffi.rs: main_reload_keyboard_shortcuts +                                                                                                                                                       
    main_get_default_keyboard_shortcuts; reload_from_config seeded in main_init                                                                                                                                
  - flutter/lib/common/widgets/keyboard_shortcuts/: shared config page body,                                                                                                                                   
    recording dialog, shortcut display formatter, action group registry                                                                                                                                        
  - flutter/lib/desktop/pages/desktop_keyboard_shortcuts_page.dart and                                                                                                                                         
    flutter/lib/mobile/pages/mobile_keyboard_shortcuts_page.dart: platform                                                                                                                                     
    shells around the shared body                                                                                                                                                                              
  - flutter/lib/models/shortcut_model.dart: per-session ShortcutModel +                                                                                                                                        
[...]

PR #15196, opened 2026-06-05
Add Windows privacy screen mode

QuentinCrane:
## Summary

- Add an optional Windows privacy screen implementation (`privacy_mode_impl_privacy_screen`) shown as Mode 3.
- Start a small helper process in the active desktop session that creates a black topmost overlay for every physical monitor.
- Mark overlay windows with `WDA_EXCLUDEFROMCAPTURE` so the controlled side is black locally while the remote side can still see and operate the desktop.
- Reuse the existing Windows privacy-mode input hook and allow monitor switching while Mode 3 is active.

## Research

- UU Remote distinguishes display-off/virtual-screen behavior from a privacy screen / black-screen protection flow: https://uuyc163.com/security/
- Microsoft documents `WDA_EXCLUDEFROMCAPTURE` as available starting with Windows 10 version 2004, so this mode is only advertised behind the same support gate used by the existing exclude-from-capture implementation: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowdisplayaffinity
- Refreshed open PR search before submitting; the visible open matches were unrelated security-hardening / keyboard-shortcut PRs (#14674, #14933), not a multi-monitor privacy-screen implementation.

## Tests

- `rustfmt --edition 2021 --config skip_children=true --check src\privacy_mode.rs src\core_main.rs src\privacy_mode\win_privacy_screen.rs`
- `cargo check --lib`
  - Exits 0 on Windows after configuring Rust, MSVC Build Tools, libclang, and the minimal vcpkg static libs needed by the build scripts.
  - Reports existing warnings in unrelated files.
- `C:\flutter\bin\dart.bat analyze lib\common\widgets\toolbar.dart lib\desktop\widgets\remote_toolbar.dart lib\consts.dart`
  - Exits 0.
  - Reports existing info-level deprecation / style hints in these large files.

## Manual Test Plan

1. Windows 10 version 2004+ controlled side with at least two physical monitors, connected from Android or another client.
2. Select privacy Mode 3 and turn on privacy mode.
3. Verify every local physical monitor is black while the remote client still sees the real desktop.
4. Verify remote input still works and local physical keyboard/mouse input is blocked.
5. Verify local `Ctrl+P` exits privacy mode and restores input.
6. Switch remote displays while Mode 3 is active and confirm local monitors remain covered.
7. Change display topology while Mode 3 is active and confirm the overlay refreshes to the new monitor layout.
8. Turn privacy mode off / end the session and confirm all overlays are removed.

## Follow-up

This PR intentionally keeps the first implementation to a black privacy screen. Custom privacy-screen images can be added in a follow-up PR with dedicated UI/options and image validation.


fufesou:
Hi @QuentinCrane , thanks for the contribution.

The structure generally fits the existing `PrivacyMode` design, and the split between the Flutter option, the helper process, and the Windows implementation is easy to follow.

However, the PR needs to settle the protection model and a few correctness details.

1. The main problem is coverage. The overlay is created as a normal topmost window, so `WS_EX_TOPMOST` and `HWND_TOPMOST` only keep it above standard desktop windows. That does not cover shell UI like Start, taskbar flyouts, or above-lock surfaces. If you want Mode 3 to match Mode 1, look at `src/privacy_mode/win_topmost_window.rs` and the `RustDeskTempTopMostWindow` `WindowInjection.dll` path: it uses `CreateWindowInBand` in a high band, which is the real reason the overlay stays above shell surfaces. If Mode 3 is intentionally injection-free, that limitation should be stated plainly in the PR.

2. The same applies to secure desktop, UAC, and the lock screen. A normal interactive desktop window disappears there, so the local screen can become visible during those transitions. Please confirm whether that is expected behavior and document it, or change the implementation if above-lock coverage is required.

3. The overlay refresh path is also too aggressive. `WM_SETTINGCHANGE` currently destroys and recreates every overlay, which can cause flicker and a brief uncovered gap for reasons that have nothing to do with display changes. I would keep destructive rebuilds tied to real display events only.

4. Two smaller correctness points still matter: `GetMessageW` needs to treat `-1` as an error return, and the helper should be checked on a mixed-DPI setup so its monitor rectangles do not leave uncovered strips.

The rest is minor.

Issue #15220, opened 2026-06-08
View Only mode does not fully disable clipboard synchronization (local→remote clipboard still works)


littlestejan:
### Bug Description

When entering View Only mode, the "Disable clipboard" option becomes checked and locked in the UI, indicating that clipboard synchronization should be disabled.

However, clipboard synchronization is only disabled in the remote -> local direction. Clipboard content copied on the controlling side is still automatically synchronized to the remote side.

This results in behavior that is inconsistent with both the UI state and user expectations.

### How to Reproduce

1. Establish a remote connection.
2. Ensure "Disable clipboard" is NOT enabled.
3. Enter View Only mode.
4. Observe that the "Disable clipboard" option becomes checked and disabled (greyed out).
5. Copy new text on the controlling side.
6. Observe that the copied text is still synchronized to the remote side.
7. Copy text on the remote side.
8. Observe that the copied text is NOT synchronized back to the controlling side.

### Expected Behavior

When View Only mode forces "Disable clipboard" to be enabled:

* Local (controller) -> Remote clipboard synchronization should be blocked.
* Remote -> Local clipboard synchronization should be blocked.

Alternatively, the UI should not indicate that clipboard synchronization is disabled if clipboard data can still be transmitted in one direction.

### Operating system(s) on local (controlling) side and remote (controlled) side

Windows 11 -> Windows 11

### RustDesk Version(s) on local (controlling) side and remote (controlled) side

1.4.7 -> 1.4.7

### Screenshots

https://drive.google.com/file/d/1NGoFjkzPjDZHBA6pPWl72E3ruR5z0X7S/view?usp=drive_link

### Additional Context

After reviewing the source code, the issue appears to be caused by inconsistent clipboard state handling between View Only mode and clipboard synchronization logic.

Observations:

* View Only mode sends disable_clipboard=true to the remote side.
* The UI also forces the "Disable clipboard" checkbox into a checked and locked state.
* However, the local client's runtime clipboard configuration remains unchanged.
* The function responsible for determining whether local clipboard updates should be sent (is_text_clipboard_required()) does not consider view_only status.

As a result:

* Remote -> Local clipboard synchronization is disabled.
* Local -> Remote clipboard synchronization continues to function.

Additionally, the remote side appears to accept incoming clipboard messages even when disable_clipboard is active, because clipboard message handling checks self.clipboard but does not verify self.disable_clipboard.

This results in View Only mode only partially disabling clipboard synchronization.

PR #15224, opened 2026-06-08
Fix clipboard synchronization not fully disabled in View Only mode

littlestejan:
## Summary

Fix clipboard synchronization behavior in View Only mode. #15220 

Previously, when entering View Only mode, the UI forced the "Disable clipboard" option to appear enabled, but clipboard synchronization was only disabled in the remote -> local direction. Clipboard updates from the controlling side could still be sent to the remote side.

This PR makes View Only mode consistently disable clipboard synchronization in both directions and correctly restores clipboard functionality when View Only mode is disabled.

## Root Cause

The issue was caused by multiple independent clipboard permission checks that did not consistently account for the `view_only` state.

### Sending side

`is_text_clipboard_required()` only checked:

* `disable_clipboard`

but did not consider:

* `view_only`

As a result, the controlling side continued sending clipboard updates while in View Only mode.

### Receiving side

The controlled side accepted incoming clipboard messages using:

```rust
if self.clipboard
```

which ignored the runtime `disable_clipboard` state.

Therefore, clipboard messages could still be applied even when clipboard synchronization was intended to be disabled.

### State refresh

When View Only mode was toggled, clipboard listener state was not refreshed.

This caused clipboard synchronization from the controlling side to remain inactive after leaving View Only mode until the clipboard option was manually toggled.

### Server-side validation

Replace:

```rust
if self.clipboard
```

with:

```rust
if self.clipboard_enabled()
```

for both:

* `Clipboard`
* `MultiClipboards`

message handling.

### Client-side clipboard permission checks

Update clipboard requirement checks to respect:

```rust
!view_only.v
```

for:

* text clipboard
* file clipboard

### Clipboard state refresh

Refresh clipboard listener state when:

```text
view-only
```

is toggled, ensuring clipboard synchronization is restored immediately after leaving View Only mode.

### Defensive receive-side checks

Prevent clipboard updates from being applied locally while View Only mode is active.

## Tested

Environment:

* Local: Linux
* Remote: Windows

Verified:

### Normal mode

* Local -> Remote clipboard sync works
* Remote -> Local clipboard sync works

### View Only mode

* Local -> Remote clipboard sync is blocked
* Remote -> Local clipboard sync is blocked

### Leaving View Only mode

* Clipboard synchronization resumes automatically in both directions

### Disable Clipboard option

* Clipboard synchronization is blocked in both directions when enabled
* Clipboard synchronization resumes normally when disabled

## Notes

File clipboard related permission checks were updated for consistency with text clipboard behavior. These paths were not directly tested due to unavailable file clipboard support in the current Linux test environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clipboard operations are now blocked in view-only sessions so remote updates won't overwrite local view-only clients.
  * Permission checks for clipboard and file copy/paste were unified to prevent unexpected clipboard activity.
  * UI now updates reliably when clipboard-related options (disable clipboard / view-only) change, reflecting actual permission state.

### DIFF
--- a/src/client/io_loop.rs
+++ b/src/client/io_loop.rs
@@ -1426,7 +1426,8 @@ impl<T: InvokeUiSession> Remote<T> {
                     self.handler.set_cursor_position(cp);
                 }
                 Some(message::Union::Clipboard(cb)) => {
-                    if !self.handler.lc.read().unwrap().disable_clipboard.v {
+                    let lc = self.handler.lc.read().unwrap();
+                    if !lc.disable_clipboard.v && !lc.view_only.v {
                         #[cfg(not(any(target_os = "android", target_os = "ios")))]
                         update_clipboard(vec![cb], ClipboardSide::Client);
                         #[cfg(target_os = "ios")]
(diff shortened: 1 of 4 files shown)


littlestejan:
@coderabbitai please re-review latest changes

PR #15236, opened 2026-06-10
Linux/Wayland: forward system shortcuts (Alt+Tab, Super) to the remote peer

VitalyOstanin:
## Problem

On Wayland, system shortcuts (Alt+Tab, Super, ...) pressed inside a remote
session window are handled by the local compositor instead of being
forwarded to the remote peer. On X11 they were forwarded as before, so this
is a regression for users whose desktop now runs a Wayland session.

Refs #5117, #4237.

## Solution

Use the `keyboard-shortcuts-inhibit-unstable-v1` protocol. While a remote
session window has keyboard focus, the client asks the compositor (via
`zwp_keyboard_shortcuts_inhibit_manager_v1`) to stop handling its own
shortcuts, so they reach the session window and are forwarded to the remote
peer. The inhibitor is released on focus leave; the compositor can also
release it natively (e.g. Super+Esc on GNOME), after which the next focus
enter re-creates it.

No new user-facing settings: capture follows window focus.

## Implementation notes

- New Linux-only module `src/platform/wayland_keyboard_inhibit.rs`.
- Wraps GTK's existing Wayland connection (`gdk_wayland_display_get_wl_display`)
  on a dedicated `wayland-client` event queue, leaving GTK's own queue
  untouched, and binds the inhibit manager from the registry.
- All GTK/GDK and Wayland work is marshalled to the GTK main thread via
  `glib::MainContext::invoke`; the public entry points may be called from a
  Flutter FFI worker thread.
- Every failure path -- X11, a compositor without the protocol, a null
  surface/seat, or any Wayland error -- is a no-op, so existing behavior is
  preserved.
- A singleton inhibitor plus surface-id comparison avoids the fatal
  `already_inhibited` protocol error on the shared GTK connection.

## Scope

Client-only. The server and the remote side are untouched.

## Testing

Built the Flutter Linux client (release) and verified on GNOME Wayland:
- Alt+Tab switches windows on the remote host.
- Super opens the remote Start menu.
- Super+Esc releases the inhibitor (shortcuts become local again).
No-op confirmed on X11.

## Summary by CodeRabbit

* **New Features**
  * Linux (Wayland): app now inhibits compositor/system shortcuts (e.g., Alt+Tab, Super) when a session window is active to avoid interference; gracefully no-ops on unsupported Wayland/GTK setups.
* **Bug Fixes**
  * Improved enter/leave handling to avoid stale-race inhibition and ensure reliable creation/release of the Wayland shortcut inhibitor when focus changes.


fufesou:
@VitalyOstanin Hi, thanks for the PR. However, I think this PR is a duplicate of #14302.

I have tested 1.4.7, and the "forward system shortcuts (Alt+Tab, Super) to the remote peer" feature works fine.

Could you please let me know if it does not work on your machines?

VitalyOstanin:
@fufesou Thanks for pointing this out, and apologies for the duplicate effort — I wasn't aware of #14302 when I started this. You're right: the merge commit is already in current master, and the C++ embedder implementation covers the same case. I installed the official 1.4.7 and confirmed the built-in inhibition works on my Wayland setup too (Alt+Tab and Super are forwarded to the remote peer).

One suggestion: the older issues/discussions around this (e.g. #13013 and related Wayland "system shortcuts" threads) still read as open/unresolved, which is partly why I ended up reimplementing it. It might help to note there that this already shipped in 1.4.7, so others don't go down the same path.

Closing this as a duplicate. Thanks again!

PR #15273, opened 2026-06-14
Fix mouse right click and ESC key exit

hjjhjj1000:
### Problem
When using RustDesk to connect to a Windows device from Android:
1. Right mouse click causes RustDesk to exit instead of sending right click to remote
2. No easy way to disconnect from session

### Solution
1. Modified onLongPress() to always send right click event to remote system
2. Changed ESC button to disconnect from session
3. Added new Send Esc button to send ESC key to remote Windows system

## Summary by CodeRabbit

* **New Features**
  * Added a **“Send Esc”** option to transmit the escape key to the remote system.
* **Bug Fixes / Behavior Changes**
  * Unified long-press behavior across all platforms to consistently perform a right-click.
  * Updated the **Esc** button to close/disconnect the current remote session instead of sending the escape key.
* **Chores**
  * Improved Flutter CI build uploads by enabling artifact uploading with a custom tag.

### DIFF
--- a/.github/workflows/flutter-ci.yml
+++ b/.github/workflows/flutter-ci.yml
@@ -21,4 +21,5 @@ jobs:
   run-ci:
     uses: ./.github/workflows/flutter-build.yml
     with:
-      upload-artifact: false
+      upload-artifact: true
+      upload-tag: "custom-build"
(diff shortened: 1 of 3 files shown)


rustdesk:
> Fix mouse right click

Thank you very much. You can provide a PR for this only (if you can find the root reason, actually we can not reproduce this issue). We won't accept the other changes.

Issue #15291, opened 2026-06-15
Windows crash when copying Excel clipboard data in remote session (FileGroupDescriptorW/FileContents with file_transfer_enabled=false)


Romy997:
### Bug Description

RustDesk crashes during a Windows-to-Windows remote session when copying data in Excel on the remote/controlled side.

The crash appears related to clipboard handling. Right before the crash, RustDesk logs show clipboard formats `FileGroupDescriptorW` and `FileContents` being processed even though file copy/paste is disabled for this peer (`file_transfer_enabled: false`).

The crash is reproducible several times in the same session. It has appeared as both:

- `librustdesk.dll` access violation: `0xc0000005`
- `ntdll.dll` heap corruption: `0xc0000374`

A Windows error dialog also reported that `RustDesk.exe` tried to read memory at `0xFFFFFFFFFFFFFFFF`.

### How to Reproduce

1. Start a Windows-to-Windows RustDesk remote desktop session.
2. Keep clipboard sync enabled.
3. Keep file copy/paste disabled for the peer.
4. Open Microsoft Excel on the remote/controlled Windows machine.
5. Copy a range of cells/data in Excel.
6. RustDesk on the controlling side crashes shortly after the clipboard update.

Observed RustDesk log lines immediately before one crash:

```text
[2026-06-15 22:23:00.991322 +08:00] DEBUG [libs\clipboard\src\platform\windows.rs:1117] client_format_list called, client id: 0, format_list: [(49415, "FileGroupDescriptorW"), (49369, "FileContents")]
[2026-06-15 22:23:00.991376 +08:00] DEBUG [src\client\io_loop.rs:383] Process clipboard message from system, stop: false, is_stopping_allowed: true, view_only: true, server_file_transfer_enabled: true, file_transfer_enabled: false
```

### Expected Behavior

RustDesk should not crash when Excel places rich/virtual file clipboard formats on the clipboard.

If file copy/paste is disabled, RustDesk should ignore or safely reject `FileGroupDescriptorW` / `FileContents` clipboard payloads instead of processing them into a crash path.

### Operating system(s) on local (controlling) side and remote (controlled) side

Windows 11 24H2, build 10.0.26100 -> Windows

### RustDesk Version(s) on local (controlling) side and remote (controlled) side

1.4.7.65 -> 1.4.7

### Screenshots

Windows error dialog text:

```text
RustDesk.exe - Application Error

[...]

PR #15300, opened 2026-06-16
try to fix https://github.com/rustdesk/rustdesk/issues/15291

rustdesk:
Root cause (verified)

  The crash signatures — heap corruption 0xc0000374 and an access violation reading 0xFFFFFFFFFFFFFFFF — are the classic fingerprint of free() being called on an uninitialized pointer. In
  libs/clipboard/src/windows/wf_cliprdr.c the format-mapping table does exactly that:
  
  - map_ensure_capacity() grows clipboard->format_mappings with realloc() — which leaves the new slots uninitialized (garbage).
  - clear_format_map() then frees map->name for every slot up to map_capacity — including those never-written, garbage slots → heap corruption / wild free.

  The format count fed to wf_cliprdr_server_format_list() comes straight from the peer's FormatList (Rust server_format_list forwards the count verbatim, it is not capped), so this is reachable from
  clipboard data the peer controls. It's the same hardening class the maintainers already addressed in #15142 (wf_cliprdr_file_group_descriptor_size_valid + the invariant test), which is why I extended that
  exact test harness.
  
  This was the convergent finding of a 50-agent adversarial audit (flagged by 6 findings across 5 independent angles, rated real by ~10 verifiers) and matches my own full read of the module.

  What I changed (git diff --stat: 2 files, +105/−4)

  1. map_ensure_capacity() — ZeroMemory() the realloc-grown region so freed-as-pointers slots are always NULL; return BOOL and propagate realloc failure.
  2. wf_cliprdr_server_format_list() — ensure capacity before writing slot i and bail on failure, so a write can never land past the allocation (removes the old write-then-grow fragility and the
  realloc-failure OOB).
  3. tests/test_invariant_wf_cliprdr.c — two regression tests: grown slots are zero-initialized, and clear_format_map() is safe after a realloc growth.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a memory corruption issue in clipboard format mapping that could occur during storage expansion.

* **Tests**
  * Added regression tests to validate clipboard format mapping safety and proper cleanup operations.

### DIFF
--- a/libs/clipboard/src/windows/wf_cliprdr.c
+++ b/libs/clipboard/src/windows/wf_cliprdr.c
@@ -1406,10 +1406,10 @@ static UINT32 get_remote_format_id(wfClipboard *clipboard, UINT32 local_format)
 	return local_format;
 }
 
-static void map_ensure_capacity(wfClipboard *clipboard)
+static BOOL map_ensure_capacity(wfClipboard *clipboard)
 {
 	if (!clipboard)
-		return;
+		return FALSE;
 
 	if (clipboard->map_size >= clipboard->map_capacity)
 	{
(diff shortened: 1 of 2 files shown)


fufesou:
@coderabbitai review

PR #15315, opened 2026-06-17
Feature: Make the address book & accessible devices side panels resizable to some degree

StealUrKill:
Replace the fixed-width left panels in the Address book and Accessible devices tabs with a draggable splitter.

The chosen width is clamped to 120-300px and remembered as a local option so it survives restarts.

The divider is invisible but still shows the resize cursor on hover.

Ran dart format on existing modified files to keep consistent.

## Issue
If you have long names in the Address book or Accessible devices tabs they become two lines and formatting looks off. Adding some extra space clears it up.

# Preview

### Before
Screenshot 2026-06-16 192502

### After
Screenshot 2026-06-16 192753

https://github.com/user-attachments/assets/960788e2-4946-4f1a-be00-1cde6d56599f

## Summary by CodeRabbit

**New Features**
* Added resizable landscape side panels for the Address Book tags view and the My Group devices view.
* Users can drag the divider to adjust available space between sections (with RTL-aware behavior).
* Panel widths are saved and restored across visits, including divider-aware sizing for smoother, more consistent resizing.

### DIFF
--- a/flutter/lib/common/widgets/address_book.dart
+++ b/flutter/lib/common/widgets/address_book.dart
@@ -9,6 +9,7 @@ import 'package:flutter_hbb/common/hbbs/hbbs.dart';
 import 'package:flutter_hbb/common/widgets/peer_card.dart';
 import 'package:flutter_hbb/common/widgets/peers_view.dart';
 import 'package:flutter_hbb/consts.dart';
+import 'package:flutter_hbb/common/widgets/resizable_side_panel.dart';
 import 'package:flutter_hbb/desktop/widgets/popup_menu.dart';
 import 'package:flutter_hbb/models/ab_model.dart';
 import 'package:flutter_hbb/models/platform_model.dart';
@@ -36,6 +37,8 @@ class AddressBook extends StatefulWidget {
 
 class _AddressBookState extends State<AddressBook> {
   var menuPos = RelativeRect.fill;
+  final _tagsPanel = ResizablePanelController(
+      optionKey: kOptionAbTagsPanelWidth, defaultWidth: 200, maxWidth: 300);
 
   @override
   Widget build(BuildContext context) => Obx(() {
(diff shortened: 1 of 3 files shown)


flutter/lib/common/widgets/resizable_side_panel.dart
```diff
@@ -0,0 +1,48 @@
...
+    return Listener(
+      onPointerUp: (_) => _persist(),
+      child: DraggableDivider(
+        axis: Axis.vertical,
+        onPointerMove: _onDrag,
+        padding: const EdgeInsets.symmetric(horizontal: 4.0),
+        color: Colors.transparent,
+      ),
+    );
+  }
```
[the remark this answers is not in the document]
StealUrKill: Applied suggestion, rebuilt, and tested locally. Resizing and width persistence still works as expected.

rustdesk:
• The feature works for LTR layouts but introduces inverted resizing behavior in supported RTL locales. This should be fixed before the patch is considered fully correct.

  Review comment:

  - [P2] Account for RTL when applying drag delta — /Users/hzhou/rustdesk/flutter/lib/common/widgets/resizable_side_panel.dart:40-40
    In RTL locales (Arabic/Hebrew/Farsi are supported), these Rows lay the side panel on the right, so moving the divider left should increase the panel width. Passing the raw positive-right delta.dx to
    _onDrag makes the width change in the opposite direction, so resizing feels broken or clamps the wrong way for those users; flip the sign based on Directionality or the panel's visual side.

rustdesk:
Issues / risks worth addressing

  - ⚠️  Perf during active drag (main note): the Obx(() => Container(... width: width.value, child: Column([dropdown, tagHeader, _buildTags(), permission]) )) puts the entire panel subtree inside the 
  width-reactive scope. So every drag frame rebuilds _buildTags() / the device list, not just the width. With many tags/devices this can jank while dragging. Tighten the reactive scope — e.g. keep children
  outside the Obx and only make a SizedBox/AnimatedContainer width reactive. (Steady-state is unaffected.)
  - ⚠️  maxWidth: 300 vs old fixed 200 lets the user drag the AB panel wider than before; on a very narrow main window that could trigger a RenderFlex overflow (cosmetic stripes) where the old 200px didn't.
  Consider clamping max to a fraction of available width, or rely on the window's min size. Low severity (self-inflicted, transient).
  - Minor: DraggableDivider is somewhat redundant here (its drag callback is unused; it's used only for cursor + transparent line) — a MouseRegion would do, but it's harmless. The ~9px hit target is a bit
  thin for grabbing.

StealUrKill:
> Issues / risks worth addressing
> 
> * ⚠️  Perf during active drag (main note): the Obx(() => Container(... width: width.value, child: Column([dropdown, tagHeader, _buildTags(), permission]) )) puts the entire panel subtree inside the
>   width-reactive scope. So every drag frame rebuilds _buildTags() / the device list, not just the width. With many tags/devices this can jank while dragging. Tighten the reactive scope — e.g. keep children
>   outside the Obx and only make a SizedBox/AnimatedContainer width reactive. (Steady-state is unaffected.)
> * ⚠️  maxWidth: 300 vs old fixed 200 lets the user drag the AB panel wider than before; on a very narrow main window that could trigger a RenderFlex overflow (cosmetic stripes) where the old 200px didn't.
>   Consider clamping max to a fraction of available width, or rely on the window's min size. Low severity (self-inflicted, transient).
> * Minor: DraggableDivider is somewhat redundant here (its drag callback is unused; it's used only for cursor + transparent line) — a MouseRegion would do, but it's harmless. The ~9px hit target is a bit
>   thin for grabbing.

I rebased and updated per review. 

- build panel subtree outside the width-reactive Obx
- replace redundant DraggableDivider with MouseRegion and 12px grab strip
- clamp rendered width via LayoutBuilder so a wide panel can't overflow a narrow window
- flip drag delta by Directionality so resizing isn't inverted in RTL

I tested the RTL as the best of my ability and set my local to an RTL locale and changed the locale in RustDesk as well. Not quite sure if that was fixed. Is there anyway for someone to test it?

flutter/lib/common/widgets/address_book.dart
```diff
@@ -76,38 +76,43 @@ class _AddressBookState extends State<AddressBook> {
       });
 
   Widget _buildAddressBookLandscape() {
-    return Row(
```
rustdesk, line 76: Can we cancel this change?

flutter/lib/common/widgets/my_group.dart
```diff
@@ -63,39 +63,42 @@ class _MyGroupState extends State<MyGroup> {
   }
 
   Widget _buildLandscape() {
-    return Row(
```
rustdesk, line 61: Can we cancel this change? I do not hope there will be possible regression introduced.  RTL support is not that important.

StealUrKill: Are you wanting the full commit canceled?

https://github.com/rustdesk/rustdesk/pull/15315/commits/fad445ca16606f1c6618b389d2067e389bbee1c9

rustdesk: I do not know, I just do not want to introduce change to my_group and address_book files. Unnecessary risks.

StealUrKill:
Will come back to this at a later date.

PR #15337, opened 2026-06-19
fix(linux): reap leftover logind session procs on headless teardown

### LINKED ISSUE #15183: [Linux] Orphaned PipeWire sockets cause logind session leak and "Session Already Running" loop
### Bug Description

The 1.4.6 changelog has no mention of pipewire socket cleanup, logind session leaks, or display detection fixes. So this bug is **not fixed** in 1.4.6 — it's worth filing. Here's a ready-to-use bug report:

---

**Title:** `[Linux] Orphaned PipeWire sockets cause logind session leak and "Session Already Running" loop`

**Environment:**
- RustDesk version: 1.3.8 (confirmed present, not fixed in 1.4.6)
- OS: Ubuntu 22.04
- Display server: X11 (GDM)
- Audio: PipeWire

**Bug Description:**

Each RustDesk connection/disconnection creates PipeWire socket files in `/run/user/1000/` (named `pipewire-rob-<PID>`, `pipewire-rob-<PID>-manager`, etc.) that are never cleaned up on session end. Over time these accumulate — in our case 33 orphaned sockets were found.

These orphaned sockets hold logind sessions in a permanent `closing` state. Because logind cannot clean up these sessions, each new login increments the X11 display number (`:2` → `:19` in our case). Since the RustDesk service has a hardcoded `DISPLAY` in its systemd override, it eventually points to a dead display, causing all subsequent connection attempts to fail with "Session Already Running".

**Workaround:**
```bash
sudo find /run/user/1000/ -name "pipewire-rob-*" -delete
sudo systemctl kill -s SIGUSR1 systemd-logind
```
Then update `DISPLAY` in `/etc/systemd/system/rustdesk.service.d/override.conf` to match the new active display from `loginctl show-session <id> | grep Display`.

**Related issues:** #5176, #7703, #10197

### How to Reproduce

1. Install RustDesk on Ubuntu 22.04 with PipeWire
2. Make and disconnect multiple RustDesk sessions over days/weeks
3. Check accumulation: `ls /run/user/1000/pipewire-rob-* | wc -l`
4. Check logind state: `loginctl list-sessions` — many sessions stuck in `closing`
5. Attempt new RustDesk connection → "Session Already Running" dialog

### Expected Behavior

PipeWire sockets created by RustDesk should be cleaned up when a session ends. Logind sessions should not accumulate in `closing` state.

**Actual behavior:** 33 orphaned PipeWire sockets found. 20 logind sessions stuck in `closing`. X11 display number incremented to `:19`. RustDesk service unable to connect to display, causing "Session Already Running" on all subsequent connections.

macOS 14.8.5 > Ubuntu 24.04.4 LTS

1.4.6 > 1.3.8

### Screenshots

n/a

### Additional Context

Versions affected: Confirmed in 1.3.8; socket cleanup code not present in 1.4.6 source; likely affects all Linux versions.

TBX3D:
On headless login the desktop manager opens a PAM session, which makes pam_systemd register a logind session and put the spawned Xorg + window manager and their children (e.g. pipewire) in a "session-<id>.scope" cgroup. Teardown only killed the Xorg and wm pids, so the rest of the session kept running, holding the logind session in "closing" and leaking runtime sockets and X display numbers on every reconnect.

Capture the session scope cgroup from a child pid and, on teardown, kill the remaining processes in it, excluding our own service process which pam moved into the scope too.

Closes #15183

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux X11 desktop shutdown by identifying the correct systemd session scope and terminating all processes within that scope, not just the X server and its direct children.
  * Reduces orphaned/lingering background processes after closing a desktop session.
  * Cleans up stale X11 lock and socket files for the allocated display during teardown.
* **Tests**
  * Added coverage for extracting the session scope from Linux cgroup data and recursively collecting descendant PIDs for cleanup.

### DIFF
--- a/src/platform/linux_desktop_manager.rs
+++ b/src/platform/linux_desktop_manager.rs
@@ -462,10 +462,14 @@ impl DesktopManager {
         let (child_xorg, child_wm) = Self::start_x11(uid, gid, username, display_num, &envs)?;
         is_child_running.store(true, Ordering::SeqCst);
 
+        // capture the logind session scope (from a live child) for teardown, see
+        // reap_session_scope.
+        let scope_dir = Self::session_scope_dir(child_xorg.id());
+
         log::info!("Start xorg and wm done, notify and wait xtop x11");
         allow_err!(tx_res.send("".to_owned()));
 
-        Self::wait_stop_x11(child_xorg, child_wm);
+        Self::wait_stop_x11(child_xorg, child_wm, scope_dir, display_num);
         log::info!("Wait x11 stop done");
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


TBX3D:
addressed all three: the resolver now handles cgroup v1/hybrid (picks the systemd controller mount, verifies cgroup.procs exists) instead of assuming the v2 unified path, truncates to the session-*.scope so a nested child scope isn't matched, and surfaces the procs-read / unexpected signal errors (skipping the benign ESRCH race) instead of swallowing them.

rustdesk:
• The patch improves basic session cleanup but misses processes in descendant cgroups, which leaves the targeted leak unresolved for nested cgroup layouts.

  Review comment:

  - [P2] Reap descendant session cgroups as well — /Users/hzhou/rustdesk/src/platform/linux_desktop_manager.rs:705-705
    When a process in the desktop session is placed in a child cgroup below session-*.scope, this stores only the parent scope's cgroup.procs. cgroup.procs lists only processes directly attached to that
    cgroup, not descendants, so leftover PipeWire/desktop children in nested cgroups survive teardown and can still keep the logind session in closing; walk the scope tree recursively or use cgroup.kill
    where available.

TBX3D:
You're right, cgroup.procs only lists the processes directly in the scope, so a de
sktop that moves pipewire/apps into descendant cgroups was missed. reap_session_sco
pe now walks the whole subtree, recursing into the session scope's child cgroups (w
orks for v1 and v2), with a test covering the descendant case.

I went with the recursive walk rather than cgroup.kill because the service process
is itself in the scope (pam_systemd put it there), so cgroup.kill would take it dow
n too; and the SIGTERM-then-SIGKILL pass lets pipewire unlink its sockets before ex
it, which a straight SIGKILL skips, and those leaked sockets are the other half of
 #15183.

rustdesk:
🔴 Primary concern — reaping the scope may kill other RustDesk processes, not just desktop cruft
  
  reap_session_scope kills every PID in session-<id>.scope except std::process::id(). The premise (which I confirmed in code) is that pam placed the --server process itself into the scope, and everything
  else there is desktop session leftovers. But I traced the spawn paths and that "everything else" is not obviously limited to the desktop:

  - The process running start_xdesktop is the --server process (rendezvous_mediator.rs:139, is_server-gated).
  - Xorg and wm are spawned with a plain Command::spawn().uid().gid() — no setsid/new cgroup (start_x_server L741, start_x_window_manager L774) — so they inherit the --server cgroup = the session scope. ✅
  Correct, and these are exactly what we want to kill.
  - But the connection manager (--cm / --cm-no-ui) is also spawned from the same --server process, via run_as_user → sudo -u <user> … <exe> --cm-no-ui (linux.rs:1100/1141), again a plain Command::spawn()
  with no setsid, and the child is tracked in super::CHILD_PROCESS (connection.rs:5384/5392).

  Whether that sudo child re-homes into a separate cgroup depends entirely on whether sudo's PAM stack runs pam_systemd (i.e. session … pam_systemd.so in /etc/pam.d/sudo / common-session). That is distro-
  and config-dependent — not guaranteed. If it does not re-home, the CM (and any IPC helper spawned the same way) sits in session-<id>.scope and gets SIGKILL-ed on desktop teardown — potentially killing the
  connection manager of a still-active or restarting session (start_x_session calls stop_children() → teardown on every new session).

  This is the assumption the whole PR rests on, and it's only partially in the author's control. Please:

  - Empirically confirm on the target distros that cat /proc/<cm_pid>/cgroup is not under the captured session scope (and same for any other long-lived --server child). If it differs everywhere, the fix is
  safe as written and this should be stated in the PR/commit.
  - If any tracked child can share the scope, harden the exclusion — e.g. skip the PIDs in crate::server::CHILD_PROCESS in addition to std::process::id(), since those are exactly the RustDesk-owned children
  you must not reap. Relying on a single self-PID exclusion is fragile against future code that forks more helpers from --server.

  ---
  🟡 Secondary — does excluding our own PID actually let the session fully release?
  
  After reaping, the scope still contains the --server PID (deliberately skipped). The leak being fixed is "logind session stuck in closing because the scope is non-empty." I confirmed the PAM session is
  released: pam::Client (rustdesk-org fork) defaults close_on_drop = true and open_session() sets has_open_session = true, so when client drops at the end of start_x_session_thread, pam_close_session →
  setcred(Delete_Cred) → pam_end all run — and this happens after wait_stop_x11/reaping returns, which is the right order.

  That said, the scope still has our PID in it at close_session time. Worth a sentence in the PR confirming (from the reproduction in #15183) that logind/systemd actually reaps the now---server-only scope
  and the display/socket leak is gone end-to-end — rather than just that the child processes died. The before/after loginctl list-sessions + ls /tmp/.X11-unix would nail it.

TBX3D:
You're right that the single self-pid exclusion is too fragile. The --cm-no-ui is spawned from the --server through sudo, and whether it re-homes into its own cgroup depends on sudo's PAM stack running pam_systemd, which isn't guaranteed. I hardened it rather than rely on that: reap_session_scope now also skips every pid in crate::server::CHILD_PROCESS, so the connection manager and any future --server-spawned helper is spared even when it shares the scope, and only the desktop leftovers are reaped.

For the end-to-end check I ran the #15183 reproduction over 5 reconnect cycles with one long-lived leader (KillUserProcesses=no). Before the fix the reused session scope accumulates the leftover desktop procs and X sockets and the display number climbs: ls /tmp/.X11-unix shows X40..X44 and the session is stuck closing. After it, the scope holds only the --server, /tmp/.X11-unix is empty, and the closing session has no leaked resources; it releases as soon as the --server exits or restarts. So the socket/display leak is cleared on every teardown, and at most one leader-only session is ever closing, bounded by the single --server.

rustdesk:
🟡 Residual gap in the new sparing — it keys on the wrapper PID, not the worker

  This is the one substantive caveat on the fix. For the headless root path, the CM is spawned by run_as_user as sudo -u <user> … <exe> --cm-no-ui (linux.rs:1100/1141), and the Child pushed to CHILD_PROCESS
  is the sudo process. So c.id() in spared is the sudo PID, not the actual --cm-no-ui PID:

  - If sudo exec-replaces itself (no pty/monitor), the worker keeps sudo's PID → spared correctly. ✅
  - If sudo runs the command under a monitor process (use_pty, the default in recent Debian/Ubuntu/Fedora sudoers), the real --cm-no-ui is a descendant with a different PID → not in spared → still reaped if
  it shares the scope. ⚠️ 

  The gap only bites when both (a) sudo uses a monitor and (b) the CM didn't re-home into its own cgroup. Those may be anti-correlated (a monitor/pty usually implies an active PAM session, which tends to
  re-home), so the real-world exposure is probably small — but it's not guaranteed. Options, in rough order of robustness:

  1. Spare the subtree: when a spared PID is found in the scope, also spare its descendants (walk /proc/<pid>/stat PPIDs, or build a parent→children map from the collected scope PIDs). This makes sparing
  robust to any wrapper-fork. Most correct, modest extra code.
  2. Document + verify: confirm on target distros that cat /proc/<cm_pid>/cgroup is outside the captured scope (i.e. the CM always re-homes), and note that the sparing is a belt-and-suspenders for the
  exec-in-place case only.
  3. Leave as-is — acceptable if you judge the anti-correlation argument sufficient, but worth a comment in the code noting the wrapper-PID limitation so it isn't mistaken for full coverage.

TBX3D:
Good catch. The tracked pid is the sudo wrapper, so with use_pty the real --cm-no-ui worker is a descendant with a different pid and the wrapper-pid check misses it. I went with option 1: reap_session_scope now spares a tracked pid and any of its descendants, walking each candidate's parent chain (stopping at the --server) and skipping it when a CHILD_PROCESS pid is an ancestor. That spares the whole sudo/monitor/worker subtree, while Xorg and the wm, which are children of the --server rather than of a tracked pid, are still reaped.

PR #15348, opened 2026-06-20
fix(linux): reap a crashed headless session's leftovers on next start

TBX3D:
Follow-up to #15337 (now merged), which reaped the headless logind session scope and removed the X lock files on a normal disconnect. This covers the crash path: if the service or its --server dies before teardown runs, that same scope and the /tmp/.X<n> files leak with nothing to reclaim them, the same way #15183 leaked them.

It records the scope and display when the headless session starts and, on the next --server start, reaps only the one scope and display the previous run recorded, never a scan, so unrelated sessions are untouched. The marker is kept in LocalConfig (host-local, not the synced options map). Reuses #15337's reap_session_scope and cleanup_x_display_files.

Verified the mechanism in a cgroup-v2 VM: a SIGKILLed run leaks the marker, the scope procs and the X files; the next start drains exactly the recorded session, and a second live session beside it is left alone. References #15183, already closed by #15337.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery and cleanup of orphaned desktop sessions after unexpected crashes, preventing stale session artifacts and resource leaks.
* **Tests**
  * Added coverage for parsing and validating the orphaned-session recovery marker, including handling of malformed inputs.

### DIFF
--- a/src/platform/linux_desktop_manager.rs
+++ b/src/platform/linux_desktop_manager.rs
@@ -51,6 +51,7 @@ fn check_desktop_manager() {
 pub fn start_xdesktop() {
     debug_assert!(crate::is_server());
     std::thread::spawn(|| {
+        DesktopManager::recover_orphaned_session();
         *DESKTOP_MANAGER.lock().unwrap() = Some(DesktopManager::new());
 
         let interval = time::Duration::from_millis(super::SERVICE_INTERVAL);
@@ -462,9 +463,10 @@ impl DesktopManager {
         let (child_xorg, child_wm) = Self::start_x11(uid, gid, username, display_num, &envs)?;
         is_child_running.store(true, Ordering::SeqCst);
 
-        // capture the logind session scope (from a live child) for teardown, see
-        // reap_session_scope.
+        // capture the logind session scope (from a live child) for teardown and crash
+        // recovery, see reap_session_scope and recover_orphaned_session.
         let scope_dir = Self::session_scope_dir(child_xorg.id());
+        Self::save_orphaned_marker(&scope_dir, display_num);
 
         log::info!("Start xorg and wm done, notify and wait xtop x11");
         allow_err!(tx_res.send("".to_owned()));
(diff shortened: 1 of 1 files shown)


rustdesk:
🔴 Primary concern — reap_session_scope at recovery has no staleness guard, unlike the display cleanup

  The recorded scope_dir (e.g. .../session-3.scope) persists on disk, but a logind session ID is not stable across reboots — the counter lives in /run (tmpfs) and resets, so session-3.scope can later belong
  to a different, live session. reap_session_scope kills whatever currently lives in that cgroup path, with no check that it's stale or ours (at recovery, CHILD_PROCESS is empty and only the new --server pid
  is excluded). So a stale marker pointed at a reused session path would SIGKILL a live, unrelated desktop session as root.

  - Reachability is timing-gated (which lowers it): within a single boot, session IDs are monotonic — no collision, recovery is correct. The hazard is strictly cross-boot, and on a normal reboot the rustdesk
  system service usually runs recovery (clearing the marker) before any interactive login forms a colliding session → harmless no-op. The dangerous window is narrower: the first post-reboot rustdesk start
  happening after a colliding login already exists (e.g. --server restarted late in uptime by systemd Restart= with the pre-reboot marker still present). Real, but not common.
  - Severity: the consequence is severe (root killing a live session), the probability is low → I'd call it a medium latent correctness bug. The asymmetry is telling: the display side is guarded, the scope
  side isn't.
  - Clean fix: store the boot_id (/proc/sys/kernel/random/boot_id) alongside the marker; on recovery, if it differs from the current boot_id, skip the scope reap (a leaked cgroup cannot survive a reboot, so
  there's nothing legitimate to reap cross-boot — you lose nothing) and at most run the already-guarded display cleanup. This eliminates the collision entirely.

TBX3D:
Good catch, fixed. The marker now carries the boot id (/proc/sys/kernel/random/boot_id) and recovery only reaps the scope when it matches the current boot. Cross-boot the reap is skipped: a leaked cgroup can't survive a reboot, so there's nothing legitimate to reap and the stale session-N.scope collision you described can't happen. The X lock cleanup is pid-guarded already, so it still runs either way. If the boot id can't be read it skips the reap to stay safe.

PR #15349, opened 2026-06-20
fix(rpm): require libxdo for X11 input

TBX3D:
enigo loads libxdo at runtime and every mouse and key event becomes a silent no-op when it is missing (libs/enigo/src/linux/xdo.rs). The deb (libxdo3 | libxdo4) and Arch (xdotool) packages depend on it, but the rpm only Recommends it. On RHEL and Rocky libxdo lives in EPEL, so the weak dependency is skipped on a stock install and X11 sessions connect with dead mouse and keyboard control, with nothing but a log warning.

Make libxdo a hard Requires in both rpm specs, matching the deb and Arch packages, so it is pulled in (or reported as missing at install time) instead of silently absent.

### DIFF
--- a/res/rpm-flutter.spec
+++ b/res/rpm-flutter.spec
@@ -5,8 +5,8 @@ Summary:    RPM package
 License:    GPL-3.0
 URL:        https://rustdesk.com
 Vendor:     rustdesk <info@rustdesk.com>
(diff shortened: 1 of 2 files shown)


rustdesk:
I am sorry NO, libxdo has been removed by many new distro, so we have to put it in recommend.

https://github.com/rustdesk/rustdesk/pull/14096

PR #15361, opened 2026-06-20
fix: refuse Magic Keyboard trackpad pan-zoom on iPad (#15209)

### LINKED ISSUE #15209: Scrolling problem iPad + Magic Keyboard
### Bug Description

iPad/Magic Keyboard client on MacOS host. Two fingers scroll on touchpad should move content on MacOS and not the whole canvas —> unusable and problem introduced in last update.

### How to Reproduce

It needs time, after a couple of minutes (2?) of using the Mac, scrolling, moving windows, the problem shows. Restarting client solves for another couple minutes. Flutter?

### Expected Behavior

Window’s content should be scrolled and not screen/canvas. I use 1:1 screen rendering so there is no need to move canvas around.

iPadOS -> MacOS

Current

### Screenshots

Self evident.

### Additional Context

_No response_

jainnam-1993:
Fixes #15209.

## Symptom

On iPad with a Magic Keyboard, the trackpad's two-finger pan/scroll gestures (delivered by Flutter as `PointerPanZoom*` events from an indirect input device) corrupt the remote canvas: pan delta is applied to the local canvas transform even when there's no useful pan room, leaving the session's view warped until restart.

The bug fires from the trackpad alone — no on-screen touch needed — because Flutter's base `ScaleGestureRecognizer.isPointerPanZoomAllowed` returns `true` unconditionally regardless of the recognizer's `supportedDevices`, so the trackpad gesture is claimed and routed into `onTwoFingerScaleUpdate`.

## Fix

Three small changes under `flutter/`, layered on purpose so any single residual pan path is also covered:

1. **Recognizer-level refusal** (`flutter/lib/common/widgets/gestures.dart`): `CustomTouchGestureRecognizer` overrides `isPointerPanZoomAllowed()` to return `false` on iOS/Android. The recognizer no longer competes for trackpad pan-zoom on mobile; the raw `Listener.onPointerPanZoomUpdate` path in `input_model.dart` — which already sends scrolls to the remote — becomes the sole consumer. Desktop trackpad pinch-zoom is preserved (returns `super.isPointerPanZoomAllowed` on macOS/Linux/Windows).

2. **Pannability predicate** (`flutter/lib/models/model.dart`): new `CanvasModel.isPanScrollUseful()` returns `true` only when the remote display, at the current scale, exceeds the canvas viewport on at least one axis — i.e. when local pan would actually reveal more content.

3. **Two-finger scale guard** (`flutter/lib/common/widgets/remote_input.dart`): in `onTwoFingerScaleUpdate`, the desktop path keeps its `isNotTouchBasedDevice()` gate (preserving existing desktop pinch behaviour); the mobile path drops the stateful `lastDeviceKind` guard (now real-touch by construction, since the recognizer no longer claims trackpad on mobile) and only applies pan when `isPanScrollUseful()` returns true. This avoids panning into black bars when remote content already fits the viewport.

## Tests

Two new pure-Dart unit test files; 11/11 green locally with `flutter test`:

- `flutter/test/canvas_overflow_test.dart` — 6 tests covering the overflow predicate (fits exactly, fits inside, overflows X only / Y only / both, zero-size viewport).
- `flutter/test/gestures_test.dart` — 5 tests covering the recognizer override (`isPointerPanZoomAllowed` returns `false` on iOS/Android, `true` on macOS/Linux/Windows).

## Manual verification

Confirmed on a physical iPad with Magic Keyboard:
- Trackpad two-finger scroll over the remote canvas no longer warps it.
- On-screen one-finger pan still works.
- Pinch-to-zoom still works.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gesture handling on iOS and Android to prevent unwanted trackpad pan-zoom events
  * Enhanced remote canvas panning to only occur when the remote content overflows the viewport
  
* **Tests**
  * Added test coverage for platform-specific gesture recognition behavior
  * Added test coverage for remote canvas overflow detection logic

### DIFF
--- a/flutter/lib/common/widgets/remote_input.dart
+++ b/flutter/lib/common/widgets/remote_input.dart
@@ -458,7 +458,13 @@ class _RawTouchGestureDetectorRegionState
   }
 
   onTwoFingerScaleUpdate(ScaleUpdateDetails d) async {
-    if (isNotTouchBasedDevice()) {
+    // On mobile, only operate on real touch — `CustomTouchGestureRecognizer`
+    // refuses to claim trackpad pan-zoom on mobile (gestures.dart), so any
+    // call here on iOS/Android is a real two-finger touch. The historical
+    // `isNotTouchBasedDevice()` guard was stateful (lastDeviceKind) and gave
+    // the wrong answer after the first incidental screen touch, breaking
+    // trackpad scroll until restart. See rustdesk/rustdesk#15209.
+    if ((isDesktop || isWebDesktop) && isNotTouchBasedDevice()) {
       return;
     }
 
@@ -484,11 +490,16 @@ class _RawTouchGestureDetectorRegionState
                     .toJson()));
       }
     } else {
-      // mobile
+      // mobile: real two-finger touch.
       ffi.canvasModel.updateScale(d.scale / _scale, d.focalPoint);
       _scale = d.scale;
-      ffi.canvasModel.panX(d.focalPointDelta.dx);
-      ffi.canvasModel.panY(d.focalPointDelta.dy);
+      // Only pan the local canvas when the remote view actually overflows
+      // the viewport — otherwise the pan moves nothing but black bars and
+      // surprises the user. (#15209)
+      if (ffi.canvasModel.isPanScrollUseful()) {
+        ffi.canvasModel.panX(d.focalPointDelta.dx);
+        ffi.canvasModel.panY(d.focalPointDelta.dy);
+      }
     }
   }
 
--- a/flutter/lib/common/widgets/gestures.dart
+++ b/flutter/lib/common/widgets/gestures.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_hbb/common/widgets/remote_input.dart';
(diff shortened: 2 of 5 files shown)


fufesou:
Hi @jainnam-1993, thanks for the PR.

Are you facing this issue on the master branch after #15288?

I can't reproduce it locally yet, so it's hard for me to verify.

flutter/lib/common/widgets/remote_input.dart
```diff
@@ -458,7 +458,13 @@ class _RawTouchGestureDetectorRegionState
...
 
   onTwoFingerScaleUpdate(ScaleUpdateDetails d) async {
-    if (isNotTouchBasedDevice()) {
+    // On mobile, only operate on real touch — `CustomTouchGestureRecognizer`
+    // refuses to claim trackpad pan-zoom on mobile (gestures.dart), so any
+    // call here on iOS/Android is a real two-finger touch. The historical
+    // `isNotTouchBasedDevice()` guard was stateful (lastDeviceKind) and gave
+    // the wrong answer after the first incidental screen touch, breaking
+    // trackpad scroll until restart. See rustdesk/rustdesk#15209.
+    if ((isDesktop || isWebDesktop) && isNotTouchBasedDevice()) {
```
fufesou, line 467: `(isDesktop || isWebDesktop)` should always be `false` on mobile, then `((isDesktop || isWebDesktop) && isNotTouchBasedDevice())` is always false?

fufesou:
This PR breaks the touch operations on my iPad.

rustdesk:
Close since https://github.com/rustdesk/rustdesk/pull/15288 has fixed the issue, and this pr introduce bigger problems.

PR #15362, opened 2026-06-20
feat: floating window shows live remote screen (foreground service)

WallnerRepo:
## Summary
- FloatingWindowService now runs as foreground service to prevent Android from killing the process when app goes to background
- Floating window displays live remote screen stream (10fps) instead of static icon
- Aspect ratio matches remote screen (16:9, 16:10, etc.)
- Free drag anywhere (no forced side-snapping)
- Size presets via popup: Small 120px / Medium 320px / Large half screen / XL full screen
- Double-tap to minimize to 48px dot, double-tap to restore
- Size persisted in SharedPreferences

## Changes
- `FloatingWindowService.kt`: Foreground service with notification, frame bitmap rendering, aspect ratio resize, size submenu, double-tap mini toggle
- `MainActivity.kt`: New `update_floating_frame` method channel handler
- `remote_page.dart`: Periodic frame push (10fps) with correct aspect ratio
- `AndroidManifest.xml`: Added `FOREGROUND_SERVICE_SPECIAL_USE` permission and `foregroundServiceType`

## Testing
Built and tested on Android arm64. Connection stays alive when activating floating window mode.

## Summary by CodeRabbit

* **New Features**
  * Added Picture-in-Picture (floating window) mode for remote sessions on Android
  * Floating window now supports double-tap to toggle mini mode and drag to reposition
  * Real-time frame updates for floating window display

* **Chores**
  * Updated Android build configuration and Gradle dependencies
  * Improved native library initialization error handling

### DIFF
--- a/flutter/android/app/build.gradle
+++ b/flutter/android/app/build.gradle
@@ -117,7 +117,7 @@ android {
         release {
             // TODO: Add your own signing config for the release build.
             // Signing with the debug keys for now, so `flutter run --release` works.
-            signingConfig signingConfigs.release
+            signingConfig signingConfigs.debug
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules'
         }
     }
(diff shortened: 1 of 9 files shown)


rustdesk:
so WHY?

rustdesk:
Thank you, but I do not think this change is necessary, which introduces unnecessary regression.

PR #14577, opened 2026-03-18
fix: support ext_data_control_v1 Wayland protocol for clipboard sync (AI)

night-hood:
## Summary

- **Fixes clipboard sync on KDE Plasma 6.5+, wlroots 0.18+, and other compositors that dropped `zwlr_data_control_v1`** in favor of the standardized `ext_data_control_v1` protocol
- Updates `wl-clipboard-rs` from 0.9.0 to 0.9.3 (adds `ext_data_control_v1` support for clipboard read/write)
- Uses a [patched clipboard-master](https://github.com/night-hood/clipboard-master/tree/fix/ext-data-control-v1) that supports both `ext_data_control_v1` (preferred) and `zwlr_data_control_v1` (fallback) for clipboard change detection
- Adds a polling fallback on Linux Wayland to catch clipboard changes missed by the event-driven listener
- Adds protobuf-level deduplication to avoid spamming identical clipboard content during polling

## Root Cause

KDE Plasma 6.5 removed `zwlr_data_control_manager_v1` from the compositor registry. Both `clipboard-master` (change detection) and `wl-clipboard-rs` 0.9.0 (clipboard I/O) only supported this old protocol. On affected compositors:

1. `clipboard-master` could not bind `zwlr_data_control_manager_v1`, so `is_primary_selection_supported()` failed, causing a fallback to X11 mode
2. `wl-clipboard-rs` 0.9.0 could not read/write the Wayland clipboard

Result: first copy/paste could work via XWayland, but subsequent clipboard updates were never forwarded to the remote peer.

### clipboard-master fork ([diff](https://github.com/night-hood/clipboard-master/compare/master...fix/ext-data-control-v1))
- `wayland.rs`: Protocol-agnostic enums (`DataControlManager`, `DataControlDevice`) wrapping both `ext_data_control_v1` and `zwlr_data_control_v1`
- `wayland.rs`: Registry dispatch prefers `ext_data_control_v1`, falls back to `zwlr_data_control_v1`
- `wayland.rs`: Full `Dispatch` implementations for all `ext_data_control_v1` types
- `x11.rs`: Custom Wayland protocol probe replacing `is_primary_selection_supported()` (which only checked `zwlr`)
- `Cargo.toml`: `wayland-protocols` bumped to >=0.32.6 with `"staging"` feature

### rustdesk (this PR)
- `Cargo.toml`: Point `clipboard-master` to fork with `ext_data_control_v1` support
[...]

PR #15366, opened 2026-06-21
fix(wayland): clipboard, support ext-data-control

fufesou:
This PR is a duplicate of https://github.com/rustdesk/rustdesk/pull/14577

## Testing

```
 wayland-info 2>/dev/null | grep -E 'zwlr_data_control_manager_v1|ext_data_control_manager_v1'
```

- [x] Windows <-> Linux X11
- [x] Windows <-> KDE Plasma < 6.5
  - Fedora 42 KDE, KDE Plasma 6.3.4 (`zwlr_data_control_manager_v1` only)
  - manjaro 24.0.8 KDE, KDE Plasma 6.3.6 (`zwlr_data_control_manager_v1` only)
  - KUbuntu 22.04, KDE Plasma 5.24.7
- [x] Windows <-> KDE Plasma 6.5+
  - KUbuntu 26.04, KDE Plasma 6.6.4 (`ext_data_control_manager_v1` only)
  - Fedora 44 KDE, KDE Plasma 6.6.4  (`ext_data_control_manager_v1` only)
- [x] Windows <-> Ubuntu 26.04 GNOME

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clipboard listener stability by automatically detecting and cleaning up stale background listener thread handles before registering new subscribers, reducing the risk of lingering completed listeners and enhancing overall reliability.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1324,7 +1324,7 @@ dependencies = [
 [[package]]
 name = "clipboard-master"
 version = "4.0.0-beta.6"
-source = "git+https://github.com/rustdesk-org/clipboard-master#ddc39f00a6211959489ae683aa6ae6eedf03a809"
+source = "git+https://github.com/rustdesk-org/clipboard-master#578fae4805afeb263552a566740991248b18822f"
 dependencies = [
  "objc",
  "objc-foundation",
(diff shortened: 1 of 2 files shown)


fufesou:
@coderabbitai review

rustdesk:
- 🟡 The first cleanup_stale_listener call (before the insert) is redundant. Because the whole function holds CLIPBOARD_LISTENER and a thread can only go finished→stays-finished (never un-finish), any
  thread that's finished before the insert is still finished after it — so the post-insert call alone covers every case (I worked through all three: dead-before, dies-during-insert, still-running). The
  post-insert call is the meaningful one (it catches a thread that dies during the subscribers.insert, then restarts with the new tx already in the shared map). Recommend either dropping the first call or 
  adding a comment explaining why both exist — as written, the duplicated call reads like a copy-paste slip.

PR #15371, opened 2026-06-21
feat(mobile): inline persistent terminal — multiple tabs alongside the live screen

WallnerRepo:
## Summary

Adds an embeddable, multi-tab **persistent terminal** to the mobile remote
session. You can open one or more terminal tabs in a resizable sheet **over the
live screen stream**, all on a single authenticated connection — so you can
watch the remote screen and run shell commands at the same time, like an
SSH + tmux split, without leaving RustDesk.

Desktop already has a terminal; on mobile you previously had to choose between
the screen and the (beta) full‑screen terminal. This brings them together.

> _Screenshots / screen recording: TODO before opening._

## What's included

- **Multiple tabs over one connection.** `TerminalConnectionManager` keeps a
  single FFI per peer (reference‑counted); each tab is a distinct `terminal_id`.
  One live `TerminalView` per tab in an `IndexedStack`, so switching tabs never
  drops keystrokes or breaks repaint.
- **Persistent sessions.** Enables the native `terminal-persistent` option, so
  sessions survive disconnect/idle. On reconnect, the server's surviving session
  ids are offered as tabs to reattach (replaying buffered output). Typing `exit`
  closes the tab and reaps the session.
- **Keyboard.** While the terminal is open, keystrokes go to it (not the remote
  desktop). A Termius‑style accessory bar adds sticky **Ctrl/Alt** modifiers
  that combine with the next key — from the bar or the system keyboard — plus
  Esc/Tab/arrows/Home/End/PgUp/PgDn.
- **UI.** Resizable, maximizable bottom sheet that stays clear of the status bar,
  gesture nav and soft keyboard; One Dark theme; legible monospace.
- **Robustness.** Hardens the shared terminal stack it builds on: use‑after‑
  dispose guards in `TerminalModel`, identity‑checked terminal‑event routing, an
  error‑safe per‑peer connection pool, and guarded padding math.

The desktop split‑view reuses the same `InlineTerminalPanel`.

## Scope & safety

- **Client‑only.** No changes under `src/` or `libs/` — the server and wire
  protocol are untouched (uses only existing FFI/proto).
- 10 files changed, all under `flutter/lib/`. One new file:
  `flutter/lib/desktop/pages/inline_terminal_panel.dart`.
- `flutter analyze` is clean (no new errors/warnings).

## How it works

Each tab is a `TerminalModel` bound to a `terminal_id` over the shared per‑peer
FFI. The panel enables `terminal-persistent`, opens the first tab, and — when the
server reports surviving persistent sessions — restores them as additional tabs.
The native side remaps the lowest surviving session onto the first
`OpenTerminal`, so reconnection transparently resumes where you left off.

## Testing

- Android arm64 (Flutter 3.24.5), Linux peer.
- Verified: open/close/switch multiple tabs; `exit` closes tab + session;
  disconnect → reconnect reattaches surviving sessions; Ctrl/Alt combos
  (Ctrl+C/L/R …); fullscreen + soft‑keyboard layout; closing the last tab closes
  the terminal UI.

## Notes

- A separate floating‑window feature is proposed independently in #15362 and is
  intentionally **not** part of this PR.

## Summary by CodeRabbit

* **New Features**
  - Added an embeddable inline terminal panel for desktop and mobile remote sessions
  - Added inline terminal toggle to the remote toolbar and mobile bottom bar
  - Enabled persistent terminal sessions with multi-tab support and session reattachment

* **Bug Fixes**
  - Hardened terminal teardown to avoid updates after disposal and prevent race-condition unregistration
  - Improved mounted-state guards and safer keyboard-metric/padding calculations
  - Strengthened terminal/connection lifecycle cleanup to prevent stale or orphaned connection state
  - Refined mobile canvas layout to account for configurable bottom overlay obstruction


rustdesk:
Thank you very much.

Close since this introduces too big changes to current code base (desktop part), we can not take this risk.

Such thing can only be done by our core dev.

Issue #14823, opened 2026-04-16
Modal window to overwrite files is displayed for each copied file (reopen #13082)


jt-github:
### Bug Description

The modal window to overwrite files is displayed for each copied file (just as was described in #13082, marked as fixed in October of last year).  Even clicking the "X Cancel" button does not work as expected: it does not cancel the transfer and instead just moves on to the next file, regardless of whether or not "Do this for all conflicts" is checked.

### How to Reproduce

Windows 11
RustDesk 1.4.6

I have been selecting multiple directories where all but one file already exists (only one file is new in each directory).  This may or may not occur when only files are selected and no directories are selected for transfer.
1. Initiate a file transfer with multiple directories, where all files are already present at the destination.
2. When the "Overwrite" dialog box appears that states "This file exists, skip or overwrite this file?" with the file name beneath and then "This file is identical with the peer's one.", check the "Do this for all conflicts" checkbox.
3. Click the "X Cancel", "> Skip", or "/ OK" button.
4. The dialog box will return, regardless of which button was clicked.

### Expected Behavior

1. Clicking "X Cancel" should cancel the entire transfer, including removing all the items in the queue on the sidebar, regardless of whether the "Do this for all conflicts" checkbox is checked.
2. Clicking "> Skip" when the "Do this for all conflicts" checkbox is checked should skip ALL existing files without further prompting regardless of how many directories are in the queue.
3. Clicking "/ OK" when the "Do this for all conflicts" checkbox is checked should overwrite ALL existing files without further prompting regardless of how many directories are in the queue.

### Operating system(s) on local (controlling) side and remote (controlled) side

Windows 11 -> Windows 11

### RustDesk Version(s) on local (controlling) side and remote (controlled) side

1.4.6 -> 1.4.6

### Screenshots

<img width="1039" height="630" alt="Image" src="https://github.com/user-attachments/assets/77eb19ef-354f-4c0d-acd8-aae8b55768c7" />

jt-github:
The only "workaround" is to delete all the local files first and then just download everything again from the remote server, but this is a huge waste of time and bandwidth.

fufesou:
Hi, thanks for the feedback.

But I have tested Win -> Win and Win -> macOS, and I can't reproduce the issue.

Please tell me what I was missing.

Or could you please upload a short video?

https://github.com/user-attachments/assets/953b48e0-322e-43da-9986-bb5f7f57dce8

jt-github:
[...]

PR #15385, opened 2026-06-23
refact: file transfer, do this for all conflicts(tasks)

fufesou:
https://github.com/rustdesk/rustdesk/issues/14823

File transfer, refactor "Do this for all conflicts".

There is a difference between this PR and the issue request: canceled tasks are shown in the right panel.

### Before

https://github.com/user-attachments/assets/b06c01c9-4a0d-491e-a650-7c98532aa75f

https://github.com/user-attachments/assets/8bbf03fe-0387-42b1-aeb0-857a93b4f2e1

### After

https://github.com/user-attachments/assets/808c91cd-2ca3-4977-add1-04131874390f

https://github.com/user-attachments/assets/2d30db4c-fd73-4de6-ae26-49a3f1e30087

## Testing

Flutter, sciter, web

- [x] Windows -> Windows, normal file transfer
- [x] Windows -> Windows, resume. No behavior changes. If conflicts occur, the dialog appears during resume, and users must choose again.

## Release Notes

* **New Features**
  * Transfer conflict batching: Multiple file transfers are now grouped under a shared conflict resolution context, enabling consistent overwrite/skip decisions across all transfers in a batch
  * "Remember for all conflicts" functionality: Users can apply their chosen conflict resolution strategy to all similar conflicts within a transfer batch instead of responding to each individually
  * Improved cancellation status display: Transfer jobs now display cancellation state more prominently in status updates

### DIFF
--- a/flutter/lib/models/file_model.dart
+++ b/flutter/lib/models/file_model.dart
@@ -142,12 +142,22 @@ class FileModel {
   }
 
   Future<void> postOverrideFileConfirm(Map<String, dynamic> evt) async {
+    final id = int.tryParse(evt['id']?.toString() ?? '');
+    if (id == null || !jobController.hasTransferConflictJob(id)) {
+      debugPrint("Ignore stale override confirm event: $evt");
+      return;
+    }
     evtLoop.pushEvent(
         _FileDialogEvent(WeakReference(this), FileDialogType.overwrite, evt));
   }
 
   Future<void> overrideFileConfirm(Map<String, dynamic> evt,
       {bool? overrideConfirm, bool skip = false}) async {
+    final id = int.tryParse(evt['id']?.toString() ?? '') ?? 0;
+    if (id == 0 || !jobController.hasTransferConflictJob(id)) {
+      debugPrint("Ignore override confirm for inactive job: $evt");
+      return;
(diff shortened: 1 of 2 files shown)


rustdesk:
• The batching implementation misses the web multi-file selection path unless a shared batch id is propagated with those events, so the new “do this for all conflicts” behavior can fail for browser uploads.

  Review comment:

  - [P2] Propagate one batch id for web selections — /Users/hzhou/rustdesk/flutter/lib/models/file_model.dart:296-297
    On the web path, onSelectedFiles handles one selected handle per event and this registration only groups jobs when the event already carries a shared batchId; otherwise it allocates a new batch per job.
    Since the current webselectFiles bridge still invokes select_files with only is_folder, selecting multiple local files in the browser can create independent batches, so checking “Do this for all
    conflicts” on the first file will not apply to the rest of that selection.

fufesou:
> https://github.com/rustdesk/rustdesk/pull/15385#issuecomment-4775169781

In my local web source, `selectFiles()` generates one `batchId` per picker invocation and emits it with every `selected_files` event, so the runtime behavior is covered when that web source is used.

I will file a new PR in the web repo.

PR #15400, opened 2026-06-23
Add monitor switch button to mobile remote toolbar

tobiasdegen:
Adds a monitor switch button to the mobile remote toolbar when the remote peer has multiple displays.

The button cycles through available displays using the existing CurrentDisplayState and openMonitorInTheSameTab logic, similar to the desktop monitor switch implementation.

Tested locally on iOS with a multi-monitor remote peer.

## Release Notes

* **New Features**
  * Added a display switcher button to the mobile interface. When multiple displays are connected, users can now quickly cycle through them using the new button in the bottom app bar.

### DIFF
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -550,6 +550,63 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
     );
   }
 
+  Widget buildMobileMonitorSwitchButton() {
+    return Obx(() {
+      final pi = gFFI.ffiModel.pi;
+      final total = pi.displays.length;
+
+      if (total < 2) {
+        return const Offstage();
+      }
+
+      final current = CurrentDisplayState.find(widget.id).value;
+      final inRange = current >= 0 && current < total;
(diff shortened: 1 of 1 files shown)


rustdesk:
Thank you, but i am sorry, mobile toolbar is crowded, i do not wanna see it

tobiasdegen:
Hi @rustdesk,  

Thank you for the review.

I understand the concern that the mobile toolbar is already crowded. However, I think the solution should not necessarily be to hide the monitor switch action itself.

On mobile, switching displays is a core remote-control action. In many multi-monitor sessions it is more important than on desktop, because the user only has a very small screen and needs a fast way to move between remote displays during the session.

Would you be open to a different toolbar prioritization instead?

For example, I can rework the PR so that the monitor switch button stays directly accessible in the toolbar, but another less time-critical action, such as starting chat, is moved into the existing three-dot / more menu.

That way the toolbar does not become more crowded, but multi-monitor switching remains quickly accessible where it is needed most. The button would still only be shown when the remote peer has more than one display, so single-monitor sessions would remain unchanged.

I am happy to update the PR in that direction if this approach is acceptable.

PR #15401, opened 2026-06-23
Replace cmd shell-out in consent process check

### LINKED ISSUE #10387: consent.exe endless loop
### Bug Description

Today I had a connection to a Win10 virtual machine, and I noticed it became very slow as soon as I connected with RustDesk.

Looking with Process Explorer, I see that there is an endless-loop of "taskleiste | findstr consent.exe" is running.

Looking at the source code, video_service.rs seems to be the only place where this consent-check is being called from. https://github.com/rustdesk/rustdesk/blob/master/src/server/video_service.rs#L252 The source code comment says "_Note: This function is extremely expensive, do not call it frequently_." , which is strange because this process is *IS* calling is frequently; it is starting/stopping non-stop on the remote computer.

### How to Reproduce

Start RD Connection on a Win10 Virtual Machine

### Expected Behavior

No CPU Load

Win10

1.3.1

### Screenshots

image

### Additional Context

_No response_

merlinz01:
This pull request refactors the way the code checks if the `consent.exe` process is running on Windows. Instead of invoking a shell command to parse process lists, it now uses the internal `get_pids` function for improved reliability and efficiency.

The continual shelling out to `cmd.exe` with the old code made large amounts of noise when trying to debug process creation over remote connections.

Resolves #10390
Resolves #10387

## Summary by CodeRabbit

* **Refactor**
  * Optimized process detection mechanism for improved system efficiency.

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -3113,11 +3113,7 @@ pub fn get_unicode_from_vk(vk: u32) -> Option<u16> {
 }
 
 pub fn is_process_consent_running() -> ResultType<bool> {
-    let output = std::process::Command::new("cmd")
-        .args(&["/C", "tasklist | findstr consent.exe"])
-        .creation_flags(CREATE_NO_WINDOW)
-        .output()?;
-    Ok(output.status.success() && !output.stdout.is_empty())
+    Ok(!get_pids("consent.exe")?.is_empty())
 }
 
 pub struct WakeLock(u32);


rustdesk:
Thanks for bringing this attention, but get pids is not as good as you think, there might be big regression, we will handle this ourselves.

merlinz01:
OK, I will wait to hear from you then.

merlinz01:
Looks like the original function was introduced in fc1af7b0d9c59790c76dc91dee84c4bae4b61a7e.

PR #15406, opened 2026-06-24
fix(fuse): fuse path broken, since ipc path changed

fufesou:
`/tmp/{APP_NAME}` does not exist since https://github.com/rustdesk/rustdesk/pull/14671

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clipboard sharing reliability by deriving the FUSE mount location from the app’s configured IPC path (instead of a hardcoded temporary directory).
  * The mount point directory is now created with stronger error handling—directory creation failures are surfaced to the caller rather than being ignored.
  * Permission-related issues continue to log a warning without blocking startup.
  * Unmount handling is now more resilient, including reduced redundant unmount attempts after successful recovery.

### DIFF
--- a/libs/clipboard/src/platform/unix/fuse/mod.rs
+++ b/libs/clipboard/src/platform/unix/fuse/mod.rs
@@ -4,7 +4,7 @@ use super::filetype::FileDescription;
 use crate::{ClipboardFile, CliprdrError};
 use cs::FuseServer;
 use fuser::MountOption;
-use hbb_common::{config::APP_NAME, log};
+use hbb_common::{config::Config, log};
 use parking_lot::Mutex;
 use std::{
     path::PathBuf,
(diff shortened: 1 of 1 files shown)


fufesou:
@coderabbitai review

rustdesk:
• The path derivation fix is reasonable, but the new early return can prevent cleanup of stale broken FUSE mounts, causing file clipboard initialization to remain stuck after an unclean shutdown.

  Review comment:

  - [P2] Keep stale FUSE unmount recovery path — /Users/hzhou/rustdesk/libs/clipboard/src/platform/unix/fuse/mod.rs:175-181
    If a previous RustDesk crash leaves this mountpoint in a broken FUSE state, create_dir_all can fail while stat'ing the existing mountpoint (for example with ENOTCONN), and this early ? returns before the
    umount cleanup below has a chance to recover it. In that scenario clipboard file paste stays disabled across restarts even though the old code still attempted the unmount; please still try the unmount
    before aborting on mountpoint creation/metadata errors.

fufesou:
https://github.com/rustdesk/rustdesk/pull/15406#issuecomment-4786897210

addressed in 99e591d

Issue #9149, opened 2024-08-22
Invalid format displayed when changing ID


kidzgy:
### Bug Description

The ID I changed meets the requirements and no matter what I enter it shows an invalid format.
![20240822220650](https://github.com/user-attachments/assets/13071e64-b0d0-4bd7-b330-0392df956be8)

### How to Reproduce

My system is Windows 11 23H2 and I use my own ID server .
After installing the program, the ID will be randomly generated, at this time, I entered my ID server address and KEY in the network, and after connecting successfully, I started to modify my ID name, and found that, no matter how I entered the name, it prompted an invalid format.

### Expected Behavior

I hope the name change works.

### Operating system(s) on local side and remote side

Win 11,ver 10.0.22631.3880

### RustDesk Version(s) on local side and remote side

1.3.0

### Screenshots

Picture above

rustdesk:
#763

PR #15412, opened 2026-06-24
fix: custom id registration with self-hosted servers

waradu:
Change ID was sending `RegisterPk` without the device public key, which self-hosted servers can reject even when the requested ID is valid.

This adds the existing device public key to the Change ID registration request so it matches the normal registration flow.

This should close #9150 (#9149)

Screenshot 2026-06-24 at 20 52 44
Screenshot 2026-06-24 at 20 52 52

For my self-hosted RustDesk setup, I used the RustDesk Server LXC for Proxmox.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of connection-related background registration by consistently reusing the same local public key across repeated attempts.
  * Reduced the chance of sending inconsistent key data during registration, helping connections complete more reliably.

### DIFF
--- a/src/ui_interface.rs
+++ b/src/ui_interface.rs
@@ -1500,12 +1500,13 @@ pub async fn change_id_shared_(id: String, old_id: String) -> &'static str {
 
     let mut futs = Vec::new();
     let err: Arc<Mutex<&str>> = Default::default();
+    let pk = Config::get_key_pair().1;
     for rendezvous_server in rendezvous_servers {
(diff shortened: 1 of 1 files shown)


Waradu:
hey @rustdesk why was this pr closed without an explanation?

rustdesk:
It looks like fixing the problem, but it is wrong. because it creates a new id rather than replacement. Please do not @me any more.

PR #15413, opened 2026-06-24
Tray menu opens on left-click after tray-icon ≥0.17 update (regression from 1.4.6)

### LINKED ISSUE #15215: Tray icon in systray Windows (10) confuses left mouse click with both left and right mouse click since v1.4.7
### Bug Description

Image

"Öffnen" and "Vermittlungsdienst stoppen" usually appears only if the users clicks with the right mouse button on the tray icon of rustdesk. now ince 1.4.7 if you click on the tray icon with the left mouse button the main windows of rustdesk appears (which is normal for left mouse button) but also "Öffnen" and "Vermittlungsdienst stoppen" appears which is uncommon for left mouse button but typical for right mouse button.

### How to Reproduce

left mouse button click on tray icon on windows (10)

### Expected Behavior

only main window of rustdesk would appear. i also miss a way where one left mouse click on the tray icon opens the main window and another click with the same mouse button also hides the main windows, as if the tray icon works as a hide and show button of the main window. only around 10 % of all my programs offer this while running in the background.

Windows 10 but i think 11 too , both sides local and remote side

1.4.7 both

### Screenshots

Image

### Additional Context

_No response_

danielmarschall:
# Tray menu opens on left-click after tray-icon ≥0.17 update (regression from 1.4.6)

## Description

Since RustDesk 1.4.7, left-clicking the tray icon on Windows opens both:

1. the RustDesk main window
2. the tray context menu

at the same time.

In RustDesk 1.4.6, a left-click only opened the RustDesk window, while the tray context menu was only shown on right-click.

## Why this is a bug

The current behavior causes two independent UI actions to be triggered by a single left-click:

- Open application window
- Open tray context menu

This leads to overlapping UI elements and inconsistent behavior.

The current implementation therefore produces the following behavior:

| Action | Result |
|----------|----------|
| Left click | Opens RustDesk window AND opens tray menu |
| Right click | Opens tray menu |

This means the tray menu is displayed for both mouse buttons, which is not how RustDesk behaved in previous releases.

## Regression

RustDesk 1.4.6:

- tray-icon 0.14.3
- Left click → open window
- Right click → open tray menu

RustDesk 1.4.7:

- tray-icon 0.21.3
- Left click → open window + open tray menu
- Right click → open tray menu

The issue appears to have been introduced by the tray-icon upgrade from 0.14.3 to 0.21.3.

## Expected behavior:

| Action | Result |
|----------|----------|
| Left click | Open/show RustDesk |
| Right click | Show tray menu |

## Actual behavior

| Action | Result |
|----------|----------|
| Left click | Open/show RustDesk and show tray menu simultaneously |
| Right click | Show tray menu |

## Relevant code in src/tray.rs

The tray icon is created with:

```rust
            let tray = TrayIconBuilder::new()
                .with_menu(Box::new(tray_menu.clone()))
                .with_tooltip(tooltip(0))
                .with_icon(icon.clone())
                .with_icon_as_template(true) // mac only
                .build();
```

This worked fine with tray-icon <0.17 , but starting with tray-icon 0.17 , a new method `with_menu_on_left_click` was added, which is by default `true`, which is a breaking change!

To get the old behavior again, `.with_menu_on_left_click(false)` must be applied to the code.

## Proposed fix

As seen in the PR, I suggest the following fix:

```rust
            let mut builder = TrayIconBuilder::new()
                .with_menu(Box::new(tray_menu.clone()))
                .with_tooltip(tooltip(0))
                .with_icon(icon.clone());

            #[cfg(target_os = "macos")]
            {
              builder = builder.with_icon_as_template(true);
            }

            #[cfg(target_os = "windows")]
            {
                // Required since tray-icon 0.17
                // Fixes #15215, #15222, #15410
                builder = builder.with_menu_on_left_click(false);
            }

            let tray = builder.build();
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved system tray behavior across platforms.
  * MacOS now applies the tray icon template setting only where appropriate.
  * Windows tray menus no longer open on left-click, matching expected behavior.

### DIFF
--- a/src/tray.rs
+++ b/src/tray.rs
@@ -143,12 +143,21 @@ fn make_tray() -> hbb_common::ResultType<()> {
             }
             // We create the icon once the event loop is actually running
             // to prevent issues like https://github.com/tauri-apps/tray-icon/issues/90
-            let tray = TrayIconBuilder::new()
+            let mut builder = TrayIconBuilder::new()
                 .with_menu(Box::new(tray_menu.clone()))
                 .with_tooltip(tooltip(0))
-                .with_icon(icon.clone())
-                .with_icon_as_template(true) // mac only
-                .build();
+                .with_icon(icon.clone());
+            #[cfg(target_os = "macos")]
+            {
+                builder = builder.with_icon_as_template(true);
+            }
+            #[cfg(target_os = "windows")]
+            {
+                // Required since tray-icon 0.17
+                // Fixes #15215, #15222, #15410
+                builder = builder.with_menu_on_left_click(false);
+            }
+            let tray = builder.build();
             match tray {
                 Ok(tray) => _tray_icon = Arc::new(Mutex::new(Some(tray))),
                 Err(err) => {


sensboston (commented):
One more thing I ran into: if the icon is in the "hidden icons" overflow tray, left-click opens the app fine but the Explorer overflow flyout stays open. It can be closed directly: it's its own window (class NotifyIconOverflowWindow), and WM_SYSCOMMAND/SC_CLOSE closes it right away.
Also noticed trySetWindowForeground() in win32_window.cpp only fires when SET_FOREGROUND_WINDOW=1 is set, and run_me() only sets that for --install/--noinstall, so a freshly opened window from the tray never steals focus on its own either.

Patch on top of your diff attached:
[tray_overflow_fix_for_pr.patch](https://github.com/user-attachments/files/29313537/tray_overflow_fix_for_pr.patch)

rustdesk:
> One more thing I ran into: if the icon is in the "hidden icons" overflow tray

@danielmarschall Could you verify this and test the patch?

danielmarschall:
@rustdesk 

Regarding @sensboston comment: No, I cannot verify it.
The tray-overflow window closes correctly on Windows, once the mouse cursor leaves the tray after the window has opened.
Tested with Windows 11 25H2 build 26200.8655.
It is possible that the observed behavior comes from a Windows bug in a different windows version,
but closing a window handle that belong to the operating system is a dirty hack and does not fix the cause.

Here is a video of the normal program behavior when a left-click is performed. My PR is applied, but the patch by @sensboston is NOT applied:

https://github.com/user-attachments/assets/ae022c7c-aa51-4b12-a32c-11142c8f3cf7

---


However, after playing around with the RustDesk icon, I found a strange bug (happens with both hidden and unhidden icon):

Sometimes the program gets into an erroneous state:
1. Left-click => does nothing (Not good)
2. Then you right-click => menu opens (OK)
3. After that, move the mouse a few pixels => the main window opens (Not good)

It feels like the message loop of the tray icon has "forgot" that it should fire on left-click. And once the mouse cursor hovers over the opened menu, the message loop "remembers" to open the main window.

Currently, I have no clear reproduction. It happens 50% of the time, and 50% of the time it works fine.

Here is a video when the application is in the erroneous state:

https://github.com/user-attachments/assets/52bf787a-ed3e-4d3b-8ff2-9239ef306642

In this video, I am clicking with the left mouse button a few times and nothing happens. Then I am clicking with the right button, the menu opens. And then I move the cursor over the context menu (not clicking) and the window opens.

rustdesk:
Great job. Thanks

PR #15419, opened 2026-06-25
fix(android): sync input service state with Flutter

### LINKED ISSUE #15418: Android input control status is not updated after accessibility service changes
### Bug Description

On the Android controlled device, the Input Control switch may display an
incorrect state after enabling or disabling the RustDesk accessibility service
multiple times.

The accessibility service is enabled in Android system settings, but the Input
Control switch in RustDesk may still appear disabled. In some cases, the switch
may also become unresponsive after toggling the screen capture service.

The Android system settings and `dumpsys accessibility` confirm that the
RustDesk InputService is enabled and bound, while the Flutter UI displays a
different state.

### How to Reproduce

1. Open RustDesk on the Android controlled device.
2. Open the Share Screen page.
3. Tap the Input Control switch.
4. Open Android accessibility settings and enable RustDesk Input.
5. Return to RustDesk and confirm that Input Control is enabled.
6. Tap Input Control again to disable it.
7. Repeat the enable and disable process several times.
8. Optionally enable and disable Screen Capture during the test.
9. The Input Control switch may eventually display disabled while the
   accessibility service is enabled, or stop responding when tapped.

### Expected Behavior

The Input Control switch should always reflect the actual lifecycle state of
the RustDesk accessibility service.

It should become enabled when InputService.onServiceConnected() is called, and
disabled when the service is unbound or destroyed.

Disabling the service should not update the Flutter UI before Android completes
the asynchronous accessibility service shutdown.

Local (controlling side): Android.  Remote (controlled side): Android 13, API 33.

1.4.2 -> master (ff226f6d8)

### Screenshots

A screenshot or screen recording showing the RustDesk Input Control switch and
the Android accessibility settings will be attached.

### Additional Context

The issue is caused by the accessibility service lifecycle state not being
actively synchronized with Flutter.

InputService.onServiceConnected() sets `ctx`, but the original implementation
does not notify Flutter that the service has become active.

When disabling input control, MainActivity calls `disableSelf()`, immediately
clears `InputService.ctx`, and reports the disabled state to Flutter. However,
`disableSelf()` is asynchronous, so this can make the UI state diverge from the
actual Android service lifecycle.

The proposed fix:

- Notifies Flutter from InputService.onServiceConnected().
- Notifies Flutter from InputService.onUnbind().
- Notifies Flutter from InputService.onDestroy().
- Removes the premature manual `InputService.ctx = null` and state notification
  from MainActivity.

The original master branch reproduces the issue. After applying the fix, the
accessibility service was enabled and disabled repeatedly on a physical Android
device without reproducing the problem.

Tested fix commit:
d85c9a3058a91c05c5019c424eeebd96bdbccf9e

yuqianglianshou:
## Summary

- Notify Flutter when the Android accessibility service is connected.
- Notify Flutter when the accessibility service is unbound or destroyed.
- Avoid clearing `InputService.ctx` before Android completes `disableSelf()`.

## Problem

The Input Control switch may not match the actual Android accessibility service
state after enabling and disabling the service multiple times.

`InputService.onServiceConnected()` previously updated `ctx` without notifying
Flutter. Also, `MainActivity` cleared `InputService.ctx` immediately after
calling `disableSelf()`, although disabling an accessibility service is
asynchronous.

This could leave the Flutter UI displaying an outdated state.

## Fix

The accessibility service now reports its actual lifecycle state from:

- `onServiceConnected()`
- `onUnbind()`
- `onDestroy()`

`MainActivity` no longer clears `InputService.ctx` or sends a premature disabled
state after calling `disableSelf()`.

## Testing

Tested on a physical Android 13, API 33, arm64 device.

- Reproduced the issue repeatedly with the original master branch.
- Enabled and disabled Input Control multiple times.
- Enabled and disabled Screen Capture between Input Control operations.
- The issue was no longer reproducible after applying this fix.
- `app:compileDebugKotlin` completed successfully.

Fixes #15418

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input mode status tracking by reliably notifying the app when the input service opens or closes, including during disconnect/unbind scenarios.
  * Refined the “stop input” behavior on newer Android versions to prevent stale input state and better synchronize the UI with the current input lifecycle.

### DIFF
--- a/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/MainActivity.kt
+++ b/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/MainActivity.kt
@@ -200,12 +200,13 @@ class MainActivity : FlutterActivity() {
                 "stop_input" -> {
                     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
                         InputService.ctx?.disableSelf()
+                    } else {
+                        InputService.ctx = null
+                        Companion.flutterMethodChannel?.invokeMethod(
+                            "on_state_changed",
+                            mapOf("name" to "input", "value" to InputService.isOpen.toString())
+                        )
                     }
-                    InputService.ctx = null
-                    Companion.flutterMethodChannel?.invokeMethod(
-                        "on_state_changed",
-                        mapOf("name" to "input", "value" to InputService.isOpen.toString())
-                    )
                     result.success(true)
                 }
                 "cancel_notification" -> {
--- a/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
+++ b/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
@@ -8,6 +8,7 @@ package com.carriez.flutter_hbb
 
 import android.accessibilityservice.AccessibilityService
 import android.accessibilityservice.GestureDescription
+import android.content.Intent
 import android.graphics.Path
 import android.os.Build
 import android.os.Bundle
@@ -66,6 +67,16 @@ class InputService : AccessibilityService() {
         var ctx: InputService? = null
         val isOpen: Boolean
             get() = ctx != null
+    }
+
+    private fun notifyInputState() {
+        val inputState = isOpen.toString()
+        Handler(Looper.getMainLooper()).post {
+            MainActivity.flutterMethodChannel?.invokeMethod(
+                "on_state_changed",
+                mapOf("name" to "input", "value" to inputState)
+            )
+        }
     }
 
     private val logTag = "input service"
@@ -716,6 +727,7 @@ class InputService : AccessibilityService() {
     override fun onServiceConnected() {
         super.onServiceConnected()
         ctx = this
+        notifyInputState()
         val info = AccessibilityServiceInfo()
         if (Build.VERSION.SDK_INT >= 33) {
             info.flags = FLAG_INPUT_METHOD_EDITOR or FLAG_RETRIEVE_INTERACTIVE_WINDOWS
@@ -734,7 +746,15 @@ class InputService : AccessibilityService() {
 
     override fun onDestroy() {
         ctx = null
+        // Keep this fallback even though onUnbind usually notifies first.
+        notifyInputState()
         super.onDestroy()
+    }
+
+    override fun onUnbind(intent: Intent?): Boolean {
+        ctx = null
+        notifyInputState()
+        return super.onUnbind(intent)
     }
 
     override fun onInterrupt() {}


rustdesk:
🟡 Main issue — stop_input becomes a no-op below API 24 (regression)

  minSdkVersion is 22, but disableSelf() is API 24+. The removed MainActivity block ran for all SDKs:
```
  "stop_input" -> {
      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
          InputService.ctx?.disableSelf()
      }
      InputService.ctx = null                         // ← removed
      Companion.flutterMethodChannel?.invokeMethod(…) // ← removed
      result.success(true)
  }
```

  On API 22–23, the new code calls neither disableSelf() (SDK-gated) nor the removed fallback — so tapping "stop input" does nothing: ctx stays set, isOpen stays true, input injection continues, and the UI
  is never updated. The old code "soft-stopped" it (null ctx + notify). This is a correctness/safety regression on a supported config (the user loses the in-app way to revoke remote input), and the API-33
  manual test wouldn't surface it.

  Suggested fix — keep the pre-24 fallback in an else:
```
  "stop_input" -> {
      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
          InputService.ctx?.disableSelf()   // onUnbind/onDestroy notify Flutter
      } else {
          InputService.ctx = null           // no disableSelf() pre-24: soft-stop + notify
          Companion.flutterMethodChannel?.invokeMethod(
              "on_state_changed",
              mapOf("name" to "input", "value" to InputService.isOpen.toString())
          )   
      }   
      result.success(true)
  }   
```
  🔵 Minor / risks

  - Redundant double-notify: disabling fires both onUnbind and onDestroy, each doing ctx = null + notify false. Harmless (idempotent), but worth a one-line comment so a future reader doesn't "dedupe" them
  and accidentally drop the earlier/more-responsive onUnbind signal.
  - Per-call Handler allocation in notifyInputState() — negligible (lifecycle-frequency only); a cached Handler would be marginally cleaner.
  - Early onServiceConnected (service enabled while app backgrounded) → flutterMethodChannel is null and the open-state notification is dropped, but the existing check_service handler reconciles on resume,
  so this is acceptable and pre-existing.

rustdesk:
• The patch improves lifecycle reporting but the posted notification still reads mutable global state after it may have changed. That can make lifecycle events report the wrong state during the enable/
  disable races this PR is meant to fix.

  Review comment:

  - [P2] Capture lifecycle state before posting — /Users/hzhou/rustdesk/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt:76-76
    When lifecycle callbacks run close together, this queued Runnable reads InputService.isOpen at execution time rather than at the onServiceConnected/onUnbind/onDestroy callback that queued it. If ctx
    changes before the main-looper task runs during quick enable/disable cycles, the notification can report the later state and drop the actual transition, so snapshot isOpen.toString() before calling post
    and send that captured value.

yuqianglianshou:
Thanks for catching the API 22–23 regression.

I restored the pre-24 fallback in the `else` branch while keeping
`disableSelf()` lifecycle handling for API 24+.

I also added a comment explaining why the `onDestroy()` notification is kept
as a fallback even though `onUnbind()` usually notifies first.

`app:compileDebugKotlin` completed successfully. The PR has been updated.

yuqianglianshou:
I addressed the lifecycle-state race by capturing `isOpen` before posting to
the main looper.

Each queued notification now reports the state from the lifecycle callback that
created it instead of reading the mutable global state later.

`app:compileDebugKotlin` completed successfully. The PR has been updated.

rustdesk:
@21pages test it

flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/MainActivity.kt
```diff
@@ -200,12 +200,13 @@ class MainActivity : FlutterActivity() {
...
+                        Companion.flutterMethodChannel?.invokeMethod(
+                            "on_state_changed",
+                            mapOf("name" to "input", "value" to InputService.isOpen.toString())
+                        )
                     }
-                    InputService.ctx = null
-                    Companion.flutterMethodChannel?.invokeMethod(
-                        "on_state_changed",
-                        mapOf("name" to "input", "value" to InputService.isOpen.toString())
-                    )
```
rustdesk, line 208: Why move into `else`?

yuqianglianshou: I moved it into the `else` branch because on API 24+ `disableSelf()` is available and the real service lifecycle callbacks (`onUnbind()` / `onDestroy()`) now notify Flutter.

If we still clear `ctx` and notify immediately after calling `disableSelf()`, Flutter may receive a disabled state before the AccessibilityService is actually unbound, which is the race this PR is trying to avoid.

For API 22–23, `disableSelf()` is not available, so the old soft-stop behavior is kept in the `else` branch: clear `ctx` and notify Flutter manually.

21pages:
Using ADB commands to repeatedly enable and disable the InputService in the background:

* Before the change, if the Share Screen page remained open, the Input Control status did not update in real time. It was refreshed after switching to the Share Screen page, resuming the app, etc.
* After the change, the Input Control status updates in real time. No issues or regressions were observed after the change.

PR #15429, opened 2026-06-26
fix(android): preserve Hangul/CJK jamo composition in soft-keyboard input (#7727)

### LINKED ISSUE #7727: Client:Android -> Host:Windows10 Android's Korean keypad typing issue.
### Bug Description

First, let me briefly explain the structure of the Korean keyboard as background knowledge. Unlike English typing, Korean typing involves a combination of consonants and vowels to form a single character (each consonant and vowel can also be recognized as an individual character). For example, when typing the keys "ㅇㅏㄴㄴㅕㅇ", the combined output through the key buffer appears as "안녕".

There seems to be a bug related to this part. When trying to type "안녕", only "ㅇㄴ" is entered. This issue has been tested with the Keypad Apps Gboard and Samsung Keyboard (only on Galaxy Phones).

### How to Reproduce

1. Install recent version of Rustdesk on Android and Windows10.
2. Connect Rustdesk from Android to Windows10.
3. Type any Korean character on Notepad.(EX: On Korean language mode, typing "dkssud" -> "안녕")
4. Only "ㅇㄴ" remain.

### Expected Behavior

"안녕" return.

Android -> Windows10

1.2.3 -> 1.2.3-1

### Screenshots

image

### Additional Context

_No response_

realitsyourman:
## Problem

Fixes #7727. On Android (as the controller) typing Korean to a remote host, Hangul jamo composition breaks:

- `안녕` arrives as `ㅇㄴ`
- `만나` arrives as `ㅁ나`

Only the **first jamo of each syllable** survives; medial vowels and final consonants (받침) are dropped. Reported on Galaxy phones with both Samsung Keyboard and Gboard.

## Root cause

`_handleNonIOSSoftKeyboardInput` in `flutter/lib/mobile/pages/remote_page.dart` diffed the hidden soft-keyboard text field by **string length** and did nothing when the length was unchanged:

```dart
if (newValue.length == oldValue.length) {
  // ?
}
```

Korean/CJK IMEs mutate the composing syllable **in place** (ㅁ → 마 → 만). Each precomposed Hangul syllable (U+AC00–U+D7A3) and isolated compatibility jamo is a single UTF-16 code unit, so an in-place mutation keeps `String.length` identical → the no-op branch runs → the medial/final jamo are silently dropped. Only a length-increasing transition (the first jamo of a new syllable, an append) ever reaches the host, which is exactly the observed `안녕 → ㅇㄴ` / `만나 → ㅁ나`.

The iOS handler `_handleIOSSoftKeyboardInput` does **not** have this bug — it diffs by content (common prefix → backspace the changed tail → retype).

## Fix

Factor the iOS content-diff into a small pure helper and route the Android handler through it. The mechanism is *backspace-the-changed-tail then retype*, which converges to the correct host text regardless of whether Android reports a composing range.

- **New** `flutter/lib/models/soft_keyboard_input_utils.dart` — pure `computeSoftKeyboardEdit(old, new)` plus the clipboard-guard and bracket-auto-pair helpers, so the Android-specific behaviour is preserved.
- `_handleNonIOSSoftKeyboardInput` rewired to it. The **iOS handler is left untouched** (only a one-line breadcrumb comment), so there is no iOS regression surface.
- This also fixes a secondary bug: a multi-unit delete previously emitted only a single `VK_BACK`; it now emits one per removed code unit.

## Tests

`flutter/test/soft_keyboard_input_utils_test.dart` — 18 pure unit tests covering: in-place jamo build-up (ㅁ→마→만, 안녕), 받침 delete, multi-unit delete, clipboard paste, bracket auto-pair, and empty-input safety. All pass (`flutter test test/soft_keyboard_input_utils_test.dart`).

## Notes

- PR #9769 addressed a **different** path — the physical-key `usbHidUsage` handling for Backspace/Enter in `input_model.dart` — and did not touch this soft-keyboard composing path, which is why #7727 stayed open.
- Scope is the Flutter mobile client only; no Rust/native/host-injection changes.

## Verification status (honest)

I have automated unit coverage of the diff logic, but I have **not** yet been able to validate on a physical Galaxy device against a live Windows host. I would appreciate a maintainer or the original reporter confirming on-device behaviour, in particular:

- Samsung Korean QWERTY **and** 천지인/Chunjiin 12-key layouts
- whether `_textController.value.isComposingRangeValid` is actually populated during composition on Android (if so, an optional flicker-reducing composing-hold could be added as a follow-up; this PR deliberately does not depend on it)

Marking as draft until on-device confirmation; happy to adjust per maintainer preference (e.g. routing the iOS handler through the shared helper too, if you'd rather unify them).

### DIFF
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -20,6 +20,7 @@ import '../../common/widgets/dialog.dart';
 import '../../common/widgets/remote_input.dart';
 import '../../models/input_model.dart';
 import '../../models/model.dart';
+import '../../models/soft_keyboard_input_utils.dart';
 import '../../models/platform_model.dart';
 import '../../utils/image.dart';
 import '../widgets/dialog.dart';
@@ -286,6 +287,8 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
     var subNewValue = newValue.substring(j + 1);
     var subOldValue = oldValue.substring(j + 1);
 
+    // ponytail: the Android path factors this content-diff out as computeSoftKeyboardEdit() in
+    // models/soft_keyboard_input_utils.dart; kept inline here to avoid touching working iOS code.
     // get common prefix of subNewValue and subOldValue
     var common = 0;
     for (;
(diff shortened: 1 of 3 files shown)


rustdesk:
Thank you, but I am sorry, we can not accept PR about this core logic, we must handle it ourselves very carefully since big regression issues have been caused before by similar cases, you can continue this PR, our dev willl use it as ref.

BTW, there were several attempts to fix https://github.com/rustdesk/rustdesk/issues/7727, all have big regression.

rustdesk:
@fufesou

realitsyourman:
> @fufesou

Please fix this. It’s a really great app, but I can't use it because of this issue.
As quickly as possible.

PR #15437, opened 2026-06-26
Fix input control state not updating after accessibility service changes

### LINKED ISSUE #15418: Android input control status is not updated after accessibility service changes

abhay8463:
Fixes issue where Android input control status is not updated after accessibility service changes.

- Have InputService notify Flutter on service connect/disconnect via onServiceConnected and onUnbind.
- Remove manual ctx = null and UI update from MainActivity.stop_input to avoid race condition.
- This ensures the UI reflects the actual service state.

Fixes rustdesk/rustdesk#15418

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input status updates: the app now informs Flutter when the Android input service connects and disconnects.
  * Prevented stale connection indicators by sending a consistent “input true/false” state tied to the service lifecycle.
  * Updated “stop input” behavior to only disable the service on Android N+ and rely on lifecycle events to clear state.
  * Enhanced accessibility text/selection handling documentation for maintainability.

### DIFF
--- a/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
+++ b/flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/InputService.kt
@@ -26,6 +26,7 @@ import android.accessibilityservice.AccessibilityServiceInfo
 import android.accessibilityservice.AccessibilityServiceInfo.FLAG_INPUT_METHOD_EDITOR
 import android.accessibilityservice.AccessibilityServiceInfo.FLAG_RETRIEVE_INTERACTIVE_WINDOWS
 import android.view.inputmethod.EditorInfo
+import android.content.Intent
 import androidx.annotation.RequiresApi
 import java.util.*
 import java.lang.Character
@@ -34,6 +35,8 @@ import kotlin.math.max
 import hbb.MessageOuterClass.KeyEvent
 import hbb.MessageOuterClass.KeyboardMode
 import hbb.KeyEventConverter
+import com.carriez.flutter_hbb.MainActivity
+import io.flutter.plugin.common.MethodChannel
 
 // const val BUTTON_UP = 2
 // const val BUTTON_BACK = 0x08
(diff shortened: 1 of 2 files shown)


rustdesk:
Here is another one, https://github.com/rustdesk/rustdesk/pull/15419 

:(

rustdesk:
Thank you, close to favor https://github.com/rustdesk/rustdesk/pull/15419

PR #15441, opened 2026-06-26
Feature: Restore the last viewed monitor on auto reconnect

StealUrKill:
## Summary

Implements the feature requested in [#15352](https://github.com/rustdesk/rustdesk/discussions/15352). Reconnect to the same monitor you were viewing.

### What it does

- Remembers the operator's last manually selected remote monitor (per session).
- After an auto reconnect, it will automatically switch back to that monitor or stay.
- No operation on the first connect and when there was no monitor manually selected.

### Testing

- Switch to a non default monitor then pull the network cable and on auto reconnect it returns to the selected monitor. 
- Verified the restore path is shared by all reconnect types (network drop, manual reconnect).
- It is skipped for the new-window / cached-replay path, and is skipped in "use all my displays" mode.

### Preview

https://private-user-images.githubusercontent.com/35749471/613449086-01f3f5b7-e57d-4030-9046-c71f8e6ddbc0.mp4?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODI0OTQ3MDMsIm5iZiI6MTc4MjQ5NDQwMywicGF0aCI6Ii8zNTc0OTQ3MS82MTM0NDkwODYtMDFmM2Y1YjctZTU3ZC00MDMwLTkwNDYtYzcxZjhlNmRkYmMwLm1wND9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA2MjYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNjI2VDE3MjAwM1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTM3ODlmYWM2MmM5MWFlNzQzYzc4YzE4ZjUxODg2NDgzY2I1MTc2YTgwNzU5YzRhYmJmNTEyNjExMjQ1ODVlMmImWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT12aWRlbyUyRm1wNCJ9.U23MQhOK8Qx0wl0sGdzxfwKpePX1oW9fo62JYEGQwxk

## Summary by CodeRabbit

* **New Features**
  * The app now remembers your last selected monitor/display. When you reconnect or refresh a session, it automatically attempts to restore that display in the same tab.

* **Bug Fixes**
  * Restoration is more reliable: it defers until initial setup is complete, validates the saved selection against available displays, respects “use all my displays” mode, and avoids restoring if the session closes during setup.

### DIFF
--- a/flutter/lib/models/model.dart
+++ b/flutter/lib/models/model.dart
@@ -112,6 +112,7 @@ class CachedPeerData {
 class FfiModel with ChangeNotifier {
   CachedPeerData cachedPeerData = CachedPeerData();
   PeerInfo _pi = PeerInfo();
+  int? lastUserDisplay;
   Rect? _rect;
 
   var _inputBlocked = false;
@@ -1400,6 +1401,18 @@ class FfiModel with ChangeNotifier {
       if (_pi.currentDisplay < _pi.displays.length) {
         // now replaced to _updateCurDisplay
         updateCurDisplay(sessionId);
+      }
+      // After an auto reconnect, return to the monitor the user last chose.
+      final last = lastUserDisplay;
+      if (!isCache &&
+          last != null &&
+          last != _pi.currentDisplay &&
+          bind.sessionGetUseAllMyDisplaysForTheRemoteSession(
+                  sessionId: sessionId) !=
+              'Y' &&
+          (last == kAllDisplayValue ||
+              (last >= 0 && last < _pi.displays.length))) {
+        openMonitorInTheSameTab(last, parent.target!, _pi);
       }
       if (displays.isNotEmpty) {
         _reconnects = 1;
--- a/flutter/lib/models/model.dart
+++ b/flutter/lib/models/model.dart
@@ -114,6 +114,7 @@ class FfiModel with ChangeNotifier {
   PeerInfo _pi = PeerInfo();
   int? lastUserDisplay;
   int? pendingMonitorRestore;
+  Timer? _pendingRestoreTimer;
   Rect? _rect;
 
   var _inputBlocked = false;
@@ -251,7 +252,7 @@ class FfiModel with ChangeNotifier {
   clear() {
     _pi = PeerInfo();
     lastUserDisplay = null;
-    pendingMonitorRestore = null;
+    _cancelPendingMonitorRestore();
     _secure = null;
     _direct = null;
     _inputBlocked = false;
@@ -936,6 +937,7 @@ class FfiModel with ChangeNotifier {
       // frame briefly, then shows the Connecting overlay.
       if (_restartReconnectDelayTimer == null) {
         parent.target?.inputModel.setRelativeMouseMode(false);
+        _cancelPendingMonitorRestore();
         bind.sessionReconnect(sessionId: sessionId, forceRelay: false);
         clearPermissions();
         // Retry once more after the silent window so restart reconnect attempts
@@ -1088,10 +1090,17 @@ class FfiModel with ChangeNotifier {
     }
   }
 
+  void _cancelPendingMonitorRestore() {
+    _pendingRestoreTimer?.cancel();
+    _pendingRestoreTimer = null;
+    pendingMonitorRestore = null;
+  }
+
   void reconnect(OverlayDialogManager dialogManager, SessionID sessionId,
       bool forceRelay) {
     // Disable relative mouse mode before reconnecting to ensure cursor is released.
     parent.target?.inputModel.setRelativeMouseMode(false);
+    _cancelPendingMonitorRestore();
     bind.sessionReconnect(sessionId: sessionId, forceRelay: forceRelay);
     clearPermissions();
     dialogManager.dismissAll();
@@ -1410,7 +1419,7 @@ class FfiModel with ChangeNotifier {
       final last = lastUserDisplay;
       pendingMonitorRestore = (!isCache &&
               last != null &&
-              last != _pi.currentDisplay &&
+              last != currentDisplay &&
               bind.sessionGetUseAllMyDisplaysForTheRemoteSession(
                       sessionId: sessionId) !=
                   'Y' &&
@@ -1418,6 +1427,12 @@ class FfiModel with ChangeNotifier {
                   (last >= 0 && last < _pi.displays.length)))
           ? last
           : null;
+      // Fallback if the first image event never reaches this tab (multi-UI).
+      _pendingRestoreTimer?.cancel();
+      if (pendingMonitorRestore != null) {
+        _pendingRestoreTimer = Timer(const Duration(milliseconds: 1500),
+            () => parent.target?._applyPendingMonitorRestore());
+      }
       if (displays.isNotEmpty) {
         _reconnects = 1;
         _offlineReconnectStartTime = null;
@@ -3936,13 +3951,20 @@ class FFI {
       for (final cb in imageModel.callbacksOnFirstImage) {
         cb(id);
       }
-      // Skip if the session closed while the canvas was being set up.
-      final restore = ffiModel.pendingMonitorRestore;
-      ffiModel.pendingMonitorRestore = null;
-      if (restore != null && !closed) {
-        openMonitorInTheSameTab(restore, this, ffiModel.pi,
-            recordSelection: false);
-      }
+      _applyPendingMonitorRestore();
+    }
+  }
+
+  void _applyPendingMonitorRestore() {
+    final restore = ffiModel.pendingMonitorRestore;
+    ffiModel._cancelPendingMonitorRestore();
+    if (restore == null || closed) return;
+    // The display list may have changed since the restore was queued.
+    final displays = ffiModel.pi.displays;
+    if ((restore == kAllDisplayValue && displays.isNotEmpty) ||
+        (restore >= 0 && restore < displays.length)) {
+      openMonitorInTheSameTab(restore, this, ffiModel.pi,
+          recordSelection: false);
     }
   }
 
--- a/flutter/lib/models/model.dart
+++ b/flutter/lib/models/model.dart
@@ -1096,6 +1096,11 @@ class FfiModel with ChangeNotifier {
     pendingMonitorRestore = null;
   }
 
+  void cancelPendingRestoreTimer() {
+    _pendingRestoreTimer?.cancel();
+    _pendingRestoreTimer = null;
+  }
+
   void reconnect(OverlayDialogManager dialogManager, SessionID sessionId,
       bool forceRelay) {
     // Disable relative mouse mode before reconnecting to ensure cursor is released.
@@ -3943,15 +3948,19 @@ class FFI {
     }
     if (ffiModel.waitForFirstImage.value == true) {
       ffiModel.waitForFirstImage.value = false;
+      ffiModel.cancelPendingRestoreTimer();
       ffiModel.resetRestartReconnectState();
       dialogManager.dismissAll();
-      await canvasModel.updateViewStyle();
-      await canvasModel.updateScrollStyle();
-      await canvasModel.initializeEdgeScrollEdgeThickness();
-      for (final cb in imageModel.callbacksOnFirstImage) {
-        cb(id);
+      try {
+        await canvasModel.updateViewStyle();
+        await canvasModel.updateScrollStyle();
+        await canvasModel.initializeEdgeScrollEdgeThickness();
+        for (final cb in imageModel.callbacksOnFirstImage) {
+          cb(id);
+        }
+      } finally {
+        _applyPendingMonitorRestore();
       }
-      _applyPendingMonitorRestore();
     }
   }
 
@@ -3964,7 +3973,7 @@ class FFI {
     if ((restore == kAllDisplayValue && displays.isNotEmpty) ||
         (restore >= 0 && restore < displays.length)) {
       openMonitorInTheSameTab(restore, this, ffiModel.pi,
-          recordSelection: false);
+          recordSelection: false, updateCursorPos: false);
     }
   }
 
--- a/flutter/lib/models/model.dart
+++ b/flutter/lib/models/model.dart
@@ -1414,7 +1414,7 @@ class FfiModel with ChangeNotifier {
               bind.sessionGetUseAllMyDisplaysForTheRemoteSession(
                       sessionId: sessionId) !=
                   'Y' &&
-              (last == kAllDisplayValue ||
+              ((last == kAllDisplayValue && _pi.displays.isNotEmpty) ||
                   (last >= 0 && last < _pi.displays.length)))
           ? last
           : null;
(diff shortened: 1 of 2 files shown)


StealUrKill:
## Update

Small bug found while adding review proposal fixes. Will either push it all in a new commit or put it on hold until the bug can be found and fixed.....

StealUrKill:
@coderabbitai full review

StealUrKill:
@coderabbitai full review

rustdesk:
• The common single-session path can work, but restore state is vulnerable to ordering races and fails in supported multi-session scenarios. A newer manual selection can also be overridden by a stale queued
  restore.

  Full review comments:

  - [P2] Cancel queued restore after a newer manual selection — /Users/hzhou/rustdesk/flutter/lib/common.dart:3355-3355
    If the user selects another monitor after reconnect queues pendingMonitorRestore but before first-image setup completes, only lastUserDisplay changes. The delayed restore then consumes the older target
    and overrides the user's newer choice; clear or update the pending restore when recording a manual selection.

  - [P2] Compare against the reconnect event's display — /Users/hzhou/rustdesk/flutter/lib/models/model.dart:1413-1413
    When the same peer has multiple control sessions, the preceding logic intentionally leaves _pi.currentDisplay unchanged. Since a manual selection made it equal to last, this comparison suppresses
    restoration even though the reconnect event reports the remote primary display; use the parsed currentDisplay to determine whether switching is required.

  - [P2] Queue restore state before frame handling can race — /Users/hzhou/rustdesk/flutter/lib/models/model.dart:1411-1411
    When kOptionTouchMode is unset, handlePeerInfo awaits sessionGetOption before reaching this assignment, while its caller does not await handlePeerInfo. A fast reconnect can therefore deliver the one-shot
    texture event first, causing onEvent2UIRgba to run before the restore is queued and leaving no later notification to apply it.

  - [P2] Do not wait for a frame filtered from this session — /Users/hzhou/rustdesk/flutter/lib/models/model.dart:3940-3943
    For a non-primary tab in a multi-UI session, native rendering routes frames only to handlers already registered for the incoming display, while a fresh remote connection starts on its primary display.
    This tab may therefore never enter onEvent2UIRgba to issue the switch; restoration needs a trigger that does not require a frame routed to the target tab.

StealUrKill:
> • The common single-session path can work, but restore state is vulnerable to ordering races and fails in supported multi-session scenarios. A newer manual selection can also be overridden by a stale queued restore.
> 
> Full review comments:
> 
> * [P2] Cancel queued restore after a newer manual selection — /Users/hzhou/rustdesk/flutter/lib/common.dart:3355-3355
>   If the user selects another monitor after reconnect queues pendingMonitorRestore but before first-image setup completes, only lastUserDisplay changes. The delayed restore then consumes the older target
>   and overrides the user's newer choice; clear or update the pending restore when recording a manual selection.
> * [P2] Compare against the reconnect event's display — /Users/hzhou/rustdesk/flutter/lib/models/model.dart:1413-1413
>   When the same peer has multiple control sessions, the preceding logic intentionally leaves _pi.currentDisplay unchanged. Since a manual selection made it equal to last, this comparison suppresses
>   restoration even though the reconnect event reports the remote primary display; use the parsed currentDisplay to determine whether switching is required.
> * [P2] Queue restore state before frame handling can race — /Users/hzhou/rustdesk/flutter/lib/models/model.dart:1411-1411
>   When kOptionTouchMode is unset, handlePeerInfo awaits sessionGetOption before reaching this assignment, while its caller does not await handlePeerInfo. A fast reconnect can therefore deliver the one-shot
>   texture event first, causing onEvent2UIRgba to run before the restore is queued and leaving no later notification to apply it.
> * [P2] Do not wait for a frame filtered from this session — /Users/hzhou/rustdesk/flutter/lib/models/model.dart:3940-3943
>   For a non-primary tab in a multi-UI session, native rendering routes frames only to handlers already registered for the incoming display, while a fresh remote connection starts on its primary display.
>   This tab may therefore never enter onEvent2UIRgba to issue the switch; restoration needs a trigger that does not require a frame routed to the target tab.

I added the queue restore state before frame handling on its own commit because It physically relocates the touch-mode resolution block inside handlePeerInfo. I did this so it was easier to see why.

flutter/lib/models/model.dart
```diff
@@ -1410,14 +1410,22 @@ class FfiModel with ChangeNotifier {
...
                   'Y' &&
               ((last == kAllDisplayValue && _pi.displays.isNotEmpty) ||
                   (last >= 0 && last < _pi.displays.length)))
           ? last
           : null;
+      // In a multi-UI session the first image event reaches only the tab that
+      // owns the reconnected display, so a pinned tab never hits onEvent2UIRgba.
+      if (pendingMonitorRestore != null &&
+          connType != null &&
+          bind.peerGetSessionsCount(id: peerId, connType: connType.index) > 1) {
```
StealUrKill: Rebased and added reviewers finding into existing commit. Did not want any more compiler errors from not rebasing.

flutter/lib/models/model.dart
```diff
@@ -1359,24 +1363,6 @@ class FfiModel with ChangeNotifier {
     }
 
     final connType = parent.target?.connType;
-    if (isPeerAndroid) {
```
rustdesk, line 1362: You do not have to move this part of code.

StealUrKill: It was moved from this review

[P2] Queue restore state before frame handling can race — /Users/hzhou/rustdesk/flutter/lib/models/model.dart:1411-1411
When kOptionTouchMode is unset, handlePeerInfo awaits sessionGetOption before reaching this assignment, while its caller does not await handlePeerInfo. A fast reconnect can therefore deliver the one-shot
texture event first, causing onEvent2UIRgba to run before the restore is queued and leaving no later notification to apply it.

I will see where i can fix this issue without moving.

StealUrKill: > You do not have to move this part of code.

Ok I removed the relocation and went back to the drawing board. Removed that commit, then found 2 more issues and folded into the last commit. Rebased and tested again.

rustdesk:
• The fallback timer can race the normal first-image initialization path and restore the monitor before the canvas is ready.

  Review comment:

  - [P2] Cancel the fallback once first-image setup starts — /Users/hzhou/rustdesk/flutter/lib/models/model.dart:1433-1434
    When the first image arrives within 1.5 seconds but the awaited canvas initialization exceeds the deadline, this timer invokes _applyPendingMonitorRestore() while onEvent2UIRgba() is still setting up the
    canvas. This switches displays too early, reintroducing the different-monitor-size offset the deferral is intended to prevent; cancel the fallback when first-image handling starts or skip it while setup
    is in progress.

Codex recommend a single-owner state flow:

  - When the first image arrives, synchronously mark canvas initialization as started and cancel the fallback timer before the first await.
  - Restore the pending monitor only after canvas initialization completes.
  - Let the timer restore only while still waiting for the first image.
  - Keep _applyPendingMonitorRestore() idempotent.

```
  void onFirstImage() async {
    _monitorRestoreTimer?.cancel();
    _monitorRestoreTimer = null;
    _isInitializingCanvas = true;

    await _initializeCanvas();

    _isInitializingCanvas = false;
    _applyPendingMonitorRestore();
  }

  void onRestoreTimeout() {
    if (!_isInitializingCanvas) {
      _applyPendingMonitorRestore();
    }
  }
```
  A small enum such as waitingForFirstImage, initializingCanvas, and restored would be cleaner than multiple booleans if the PR already has several related flags.

rustdesk:
# Codex suggestion

 Based on the described behavior, I would remove the fallback timer entirely.

  The timer is logically inconsistent: elapsed time does not mean the canvas is ready. The PR only needs one pending value and one safe execution point:
```
  int? _pendingMonitorRestore;

  Future<void> onEvent2UIRgba() async {
    if (!ffiModel.waitForFirstImage.value) return;

    ffiModel.waitForFirstImage.value = false;

    await canvasModel.updateViewStyle();
    await canvasModel.updateScrollStyle();
    await canvasModel.initializeEdgeScrollEdgeThickness();

    final monitor = _pendingMonitorRestore;
    _pendingMonitorRestore = null;

    if (monitor != null) {
      await _restoreMonitor(monitor);
    }
  }
```
  The overall flow becomes:

  1. Save the desired monitor as pending.
  2. Receive the first image.
  3. Initialize the canvas.
  4. Restore the monitor exactly once.

  If no first image arrives, there is no usable canvas anyway, so restoring after an arbitrary 1.5 seconds provides little value and violates the reason for deferring restoration. This removes the timer,
  cancellation logic, race, and additional in-progress state.

StealUrKill:
> # Codex suggestion
> Based on the described behavior, I would remove the fallback timer entirely.
> 
> The timer is logically inconsistent: elapsed time does not mean the canvas is ready. The PR only needs one pending value and one safe execution point:
> 
> ```
>   int? _pendingMonitorRestore;
> 
>   Future<void> onEvent2UIRgba() async {
>     if (!ffiModel.waitForFirstImage.value) return;
> 
>     ffiModel.waitForFirstImage.value = false;
> 
>     await canvasModel.updateViewStyle();
>     await canvasModel.updateScrollStyle();
>     await canvasModel.initializeEdgeScrollEdgeThickness();
> 
>     final monitor = _pendingMonitorRestore;
>     _pendingMonitorRestore = null;
> 
>     if (monitor != null) {
>       await _restoreMonitor(monitor);
>     }
>   }
> ```
> 
> The overall flow becomes:
> 
> 1. Save the desired monitor as pending.
> 2. Receive the first image.
> 3. Initialize the canvas.
> 4. Restore the monitor exactly once.
> 
> If no first image arrives, there is no usable canvas anyway, so restoring after an arbitrary 1.5 seconds provides little value and violates the reason for deferring restoration. This removes the timer, cancellation logic, race, and additional in-progress state.

### Reasoning for timer

The timer was implemented due to previous review recommendations.

> [P2] Queue restore state before frame handling can race — /Users/hzhou/rustdesk/flutter/lib/models/model.dart:1411-1411
> When kOptionTouchMode is unset, handlePeerInfo awaits sessionGetOption before reaching this assignment, while its caller does not await handlePeerInfo. A fast reconnect can therefore deliver the one-shot
> texture event first, causing onEvent2UIRgba to run before the restore is queued and leaving no later notification to apply it.
> 
> [P2] Do not wait for a frame filtered from this session — /Users/hzhou/rustdesk/flutter/lib/models/model.dart:3940-3943
> For a non-primary tab in a multi-UI session, native rendering routes frames only to handlers already registered for the incoming display, while a fresh remote connection starts on its primary display.
> This tab may therefore never enter onEvent2UIRgba to issue the switch; restoration needs a trigger that does not require a frame routed to the target tab.

Peer-info events are sent to every UI window in flutter.rs:572-587, but image frames are only sent to windows registered for that display in flutter.rs:1227 and flutter.rs:1261.

Because of that, after reconnect, a secondary monitor window can receive peer-info and queue a restore, but receive no frame from the primary display. If no frame reaches the window, `onEvent2UIRgba` never runs, so the normal first-frame restore cannot happen.

The timer is the fallback that calls `sessionSwitchDisplay()` -> `session_switch_display` (flutter.rs:2178). The key step is the peer-side request `s.switch_display()` (flutter.rs:2189), which tells the peer to start capturing the wanted display. Those frames then route to this window (already registered for it) and rendering resumes.

This problem is specific to multi-UI mode. In single-UI mode, the display filter is skipped, so a first frame can still reach the window.

The timer was added to address the earlier frame-delivery and ordering issues. The cancellation fix (not yet pushed, but working) was then added in response to a [later review](https://github.com/rustdesk/rustdesk/pull/15441#issuecomment-4863976890) that found the timer could race the normal first-image initialization path.

**1. If a frame arrives normally, the timer is cancelled before it can interfere.
2. If no frame arrives, the timer performs the restore.
3. With no first frame yet, there is no initialized canvas state that could cause the previous offset problem.**

I believe the timer should remain. Removing it would break the multi-UI reconnect case where peer-info arrives but no matching frame reaches that window.

If the concern is specifically the use of a timer, I am also open to replacing it with an event-driven approach.

The session already has the state to determine when its display isn't part of the active capture set, so the restore could be applied immediately in that case. Because this would only re-subscribe the session to its intended display, there would be no canvas state that needs to be deferred.

The separate ordering issue could then be addressed by queuing the restore before the `sessionGetOption` await.

That approach would remove the timer entirely. I am happy to make that change if an event-driven implementation is preferred. If not I can push the cancellation fix if the timer is ok to stay?

rustdesk:
Let's keep timer, though it looks weird, but it introduces minimum code change.

But here is potential regression (Codex):

  If the first frame arrives and cancels the fallback timer, but canvas initialization subsequently fails or hangs, _applyPendingMonitorRestore() will never run. Because the timer has already been cancelled,
  monitor restoration has no remaining fallback path.

  The practical impact is limited because the canvas is already unusable in this scenario. Only _pendingRestoreTimer should be cancelled; calling _cancelPendingMonitorRestore() would also clear
  pendingMonitorRestore and break restoration entirely.

StealUrKill:
> Let's keep timer, though it looks weird, but it introduces minimum code change.
> 
> But here is potential regression (Codex):
> 
> If the first frame arrives and cancels the fallback timer, but canvas initialization subsequently fails or hangs, _applyPendingMonitorRestore() will never run. Because the timer has already been cancelled, monitor restoration has no remaining fallback path.
> 
> The practical impact is limited because the canvas is already unusable in this scenario. Only _pendingRestoreTimer should be cancelled; calling _cancelPendingMonitorRestore() would also clear pendingMonitorRestore and break restoration entirely.

I added the changes and tested strenuously. The feature worked every time I changed the conditions. I put an Orchestration panel together (per your reviews and suggestions from comments on PRs and discussions). 

This was their answer on the review.

### Orchestration
**GPT‑5.5 (Hermes/codex): LEAVE — a hang would be a deeper FFI failure, not a UI edge; timeout is a magic-number watchdog that doesn't cancel the underlying work; if ever real, fix the bind layer, not the UI.
Claude subagent (repo access): LEAVE — verified against the Rust source that session_get_* are synchronous in-memory RwLock reads with no I/O, so they can't hang (only a global deadlock could, = dead session); a poisoned lock throws, which try/finally already covers.
Kimi (Hermes): LEAVE — hang unlikely and session would be dead anyway; proposed deferring the timer cancel as an alternative (which I rejected — it reintroduces the mid-init race).
GLM‑5.2 (Hermes): LEAVE — local config reads, not I/O; .timeout() doesn't cancel and risks firing on slow-but-not-hung init; raised the defer-cancel idea but dismissed it as unjustified.**

PR #15448, opened 2026-06-27
fix: auto-close terminal tab/window when shell exits

alonginwind:
## Problem

When the user types `exit` in the remote terminal, the terminal tab/window does not close.

### Server (`src/server/terminal_service.rs`)

- Add `child.try_wait()` as a fallback terminal-exit detection mechanism.
  - In addition to checking `reader_thread.is_finished()`, periodically poll the child process with `try_wait()`.
  - This allows Linux to detect shell exit even when the PTY reader thread remains blocked.

#### `terminal_model.dart`

- Add an `onClosed` callback.
- Invoke the callback immediately when the terminal session ends.

#### Desktop (`terminal_page.dart`)

- Set `onClosed` to call `tabController.closeBy(tabKey)`.
- Automatically close the terminal tab.
- If it is the last remaining tab, the terminal window closes automatically.

#### Mobile (`terminal_page.dart`)

- Set `onClosed` to call `closeConnection()`.
- Automatically navigate back to the previous screen when the terminal session ends.

### Linux → Linux, Windows → Linux

- Type `exit` in the remote terminal.
- The terminal tab closes immediately.
- If it is the last tab, the terminal window also closes.

### Android → Linux

- Type `exit` in the remote terminal.
- The terminal page is closed automatically.
- The client navigates back to the previous screen.

https://github.com/user-attachments/assets/750fab9d-ac10-444f-b25a-97b2c984f4f0

**Note:** Windows remote hosts do not close the terminal automatically after `exit`. This issue is not addressed by this change.

### DIFF
--- a/src/server/terminal_service.rs
+++ b/src/server/terminal_service.rs
@@ -1857,14 +1857,32 @@ impl TerminalServiceProxy {
         // Process each session with its own lock
         for (terminal_id, session_arc) in sessions {
             if let Ok(mut session) = session_arc.try_lock() {
-                // Check if reader thread is still alive and we haven't sent closed message yet
+                // Check if the session has ended (reader thread finished or child exited).
+                // On Linux, the PTY reader thread may not return EOF when the shell exits
+                // (the cloned master fd keeps the read side open), so we also poll the child
+                // process via try_wait() as a fallback detection mechanism.
                 let mut should_send_closed = false;
                 if !session.closed_message_sent {
                     if let Some(thread) = &session.reader_thread {
                         if thread.is_finished() {
                             should_send_closed = true;
-                            session.closed_message_sent = true;
                         }
+                    }
+                    if !should_send_closed {
+                        if let Some(child) = &mut session.child {
+                            match child.try_wait() {
+                                Ok(Some(_)) => {
+                                    should_send_closed = true;
+                                }
+                                Ok(None) => {} // still running
+                                Err(e) => {
+                                    log::warn!("Terminal {} child wait error: {}", terminal_id, e);
+                                }
+                            }
+                        }
+                    }
+                    if should_send_closed {
+                        session.closed_message_sent = true;
                     }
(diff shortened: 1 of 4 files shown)


src/server/terminal_helper.rs
```diff
@@ -1028,13 +1028,22 @@ pub fn run_terminal_helper(args: &[String]) -> Result<()> {
 
     exiting.store(true, Ordering::SeqCst);
 
-    // Wait for threads
-    let _ = input_thread.join();
-    let _ = output_thread.join();
-
```
rustdesk, line 1034: Why?

src/server/terminal_service.rs
```diff
@@ -2004,30 +2045,18 @@ impl TerminalServiceProxy {
             for terminal_id in closed_terminals {
                 let mut exit_code = 0;
 
-                if !self.is_persistent {
```
rustdesk, line 2007: Why remove this?

alonginwind: Once child.try_wait() marks a terminal closed here, the persistent cleanup path at Lines 2057-2068 only clears session.child; the service.sessions entry remains. handle_open() at Lines 1096-1165 reconnects any existing entry as if the shell were still alive, so this Linux fallback can now reopen a dead terminal instead of creating a fresh one. Coderabbitai suggests either remove the exited session from service.sessions or make handle_open() recreate sessions whose child has already exited.

src/server/terminal_helper.rs
alonginwind: These two lines cause the Windows controlled-side terminal to hang on exit. The input_thread blocks on pipe read and join() never returns after the shell process exits, so the cleanup path gets stuck.

src/server/terminal_service.rs
rustdesk: Such thing can not trust AI

Issue #565, opened 2022-05-20
Unable to build for android 


Ahsan445:
**Describe the bug you encountered:**

I am trying to build for android using official docs but i am getting this error every time.

I have configured vcpkg many times. reinstalled rust etc, I can't figure out what is the issue.

<pre><font color="#8AE234"><b>ahsan@ahsan-VirtualBox</b></font>:<font color="#729FCF"><b>~</b></font>$ cd Desktop/rusk/rustdesk/
<font color="#8AE234"><b>ahsan@ahsan-VirtualBox</b></font>:<font color="#729FCF"><b>~/Desktop/rusk/rustdesk</b></font>$ cd ..
<font color="#8AE234"><b>ahsan@ahsan-VirtualBox</b></font>:<font color="#729FCF"><b>~/Desktop/rusk</b></font>$ cd ..
<font color="#8AE234"><b>ahsan@ahsan-VirtualBox</b></font>:<font color="#729FCF"><b>~/Desktop</b></font>$ ls
command  <font color="#729FCF"><b>rusk</b></font>  <font color="#729FCF"><b>vcpkg</b></font>
<font color="#8AE234"><b>ahsan@ahsan-VirtualBox</b></font>:<font color="#729FCF"><b>~/Desktop</b></font>$ cd rusk/
<font color="#8AE234"><b>ahsan@ahsan-VirtualBox</b></font>:<font color="#729FCF"><b>~/Desktop/rusk</b></font>$ ls
<font color="#729FCF"><b>android-ndk-r22b</b></font>                   <font color="#EF2929"><b>dep.tar.gz</b></font>  <font color="#729FCF"><b>rustdesk</b></font>
<font color="#EF2929"><b>android-ndk-r22b-linux-x86_64.zip</b></font>  <font color="#729FCF"><b>openssl</b></font>     <font color="#729FCF"><b>vcpkg</b></font>
<font color="#8AE234"><b>ahsan@ahsan-VirtualBox</b></font>:<font color="#729FCF"><b>~/Desktop/rusk</b></font>$ cd rustdesk/
<font color="#8AE234"><b>ahsan@ahsan-VirtualBox</b></font>:<font color="#729FCF"><b>~/Desktop/rusk/rustdesk</b></font>$ ls
<font color="#AD7FA8"><b>128x128@2x.png</b></font>      <font color="#8AE234"><b>entrypoint</b></font>         README-DE.md  README-PL.md
[...]

PR #15469, opened 2026-06-30
Fix/session scope permission audit

fufesou:
## Summary

  - Enforce message scope after a session is authorized.
  - Filter out messages when limited sessions receive out-of-scope messages.
  - Scope login options by connection type so that limited sessions cannot apply unrelated options.

## Notes

 1. Out-of-scope options are filtered before applying in `update_scoped_login_options()`.

2. After housekeeping and legacy no-op compatibility filtering, scoped sessions are limited to their intended protocol surface:
    - `AuthConnType::Remote`: all messages are allowed
    - `AuthConnType::FileTransfer`: file-transfer messages only
    - `AuthConnType::PortForward`: no session protocol messages are allowed
    - `AuthConnType::ViewCamera`: camera/viewing messages and view-camera-scoped options only
    - `AuthConnType::Terminal`: terminal messages and terminal-scoped options only

```rust
        let allowed = match conn_type {
            AuthConnType::Remote => true,
            AuthConnType::FileTransfer => Self::is_file_transfer_scoped_message(msg),
            AuthConnType::PortForward => false,
            AuthConnType::ViewCamera => Self::is_view_camera_scoped_message(msg),
            AuthConnType::Terminal => Self::is_terminal_scoped_message(msg),
        };
```

3. Unexpected authorized messages now trigger a `SessionScopeViolation` alarm audit.

4. ~~New message types are easy to debug: the functions perform a whitelist check, and any unsupported message triggers a disconnection~~.

## ~~Audit logs~~

a

## Compatibility note

Legacy clients may broadcast no-op messages to multiple open sessions:

  - FileTransfer and Terminal tolerate text clipboard and render-broadcast no-op messages, but do not apply them.
  - PortForward tolerates only render-broadcast no-op messages, but does not apply them

## Testing

- [x] Default connections: file transfer, camera view, terminal, RDP, TCP tunneling all work.
- [x] Out-of-scope messages trigger disconnection and are logged in audit logs.

## Summary

- **New Features**
  - Added stricter session-scope enforcement for already-authenticated connections, including session-scope violation auditing with configurable alarm and optional immediate disconnect.

- **Bug Fixes**
  - “Take screenshot” now shows only for desktop default connections; screenshot handling is correctly scoped to the active video source and display.
  - Improved view-camera behavior: suppresses virtual display/privacy, resolution changes, and related auto-login steps.
  - Clipboard sync is now restricted to supported remote/default sessions.
  - Texture/D3D render options now apply only to default or view-camera sessions.

- **Tests**
  - Added message allow/deny and login-option scoping/violation coverage.

### DIFF
--- a/flutter/lib/common/widgets/toolbar.dart
+++ b/flutter/lib/common/widgets/toolbar.dart
@@ -606,7 +606,7 @@ List<TTextMenu> toolbarControls(BuildContext context, String id, FFI ffi) {
   // to-do:
   // 1. Web desktop
   // 2. Mobile, copy the image to the clipboard
-  if (isDefaultConn && isDesktop) {
+  if ((isDefaultConn || ffi.connType == ConnType.viewCamera) && isDesktop) {
     final isScreenshotSupported = bind.sessionGetCommonSync(
         sessionId: sessionId, key: 'is_screenshot_supported', param: '');
     if ('true' == isScreenshotSupported) {
--- a/flutter/lib/common/widgets/toolbar.dart
+++ b/flutter/lib/common/widgets/toolbar.dart
@@ -606,7 +606,7 @@ List<TTextMenu> toolbarControls(BuildContext context, String id, FFI ffi) {
   // to-do:
   // 1. Web desktop
   // 2. Mobile, copy the image to the clipboard
-  if (isDesktop) {
+  if (isDefaultConn && isDesktop) {
     final isScreenshotSupported = bind.sessionGetCommonSync(
         sessionId: sessionId, key: 'is_screenshot_supported', param: '');
     if ('true' == isScreenshotSupported) {
--- a/src/client/io_loop.rs
+++ b/src/client/io_loop.rs
@@ -1090,6 +1090,9 @@ impl<T: InvokeUiSession> Remote<T> {
     }
 
     async fn send_toggle_virtual_display_msg(&self, peer: &mut Stream) {
+        if self.handler.is_view_camera() {
+            return;
+        }
         if !self.peer_info.is_support_virtual_display() {
             return;
         }
--- a/src/client/io_loop.rs
+++ b/src/client/io_loop.rs
@@ -1111,6 +1111,9 @@ impl<T: InvokeUiSession> Remote<T> {
     }
 
     async fn send_toggle_privacy_mode_msg(&self, peer: &mut Stream) {
+        if self.handler.is_view_camera() {
+            return;
+        }
         let lc = self.handler.lc.read().unwrap();
         if lc.version >= hbb_common::get_version_number("1.2.4")
             && lc.get_toggle_option("privacy-mode")
--- a/src/flutter.rs
+++ b/src/flutter.rs
@@ -1437,7 +1437,7 @@ fn try_send_close_event(event_stream: &Option<StreamSink<EventToUI>>) {
 pub fn update_text_clipboard_required() {
     let is_required = sessions::get_sessions()
         .iter()
-        .any(|s| s.is_text_clipboard_required());
+        .any(|s| s.is_default() && s.is_text_clipboard_required());
     #[cfg(target_os = "android")]
     let _ = scrap::android::ffi::call_clipboard_manager_enable_client_clipboard(is_required);
     Client::set_is_text_clipboard_required(is_required);
@@ -1447,13 +1447,16 @@ pub fn update_text_clipboard_required() {
 pub fn update_file_clipboard_required() {
     let is_required = sessions::get_sessions()
         .iter()
-        .any(|s| s.is_file_clipboard_required());
+        .any(|s| s.is_default() && s.is_file_clipboard_required());
     Client::set_is_file_clipboard_required(is_required);
 }
 
 #[cfg(not(target_os = "ios"))]
 pub fn send_clipboard_msg(msg: Message, _is_file: bool) {
     for s in sessions::get_sessions() {
+        if !s.is_default() {
+            continue;
+        }
         #[cfg(feature = "unix-file-copy-paste")]
         if _is_file {
             if crate::is_support_file_copy_paste_num(s.lc.read().unwrap().version)
--- a/src/flutter_ffi.rs
+++ b/src/flutter_ffi.rs
@@ -1227,6 +1227,9 @@ pub fn main_set_local_option(key: String, value: String) {
     if is_texture_render_key {
         let session_event = [("v", &value)];
         for session in sessions::get_sessions() {
+            if !(session.is_default() || session.is_view_camera()) {
+                continue;
+            }
             session.push_event("use_texture_render", &session_event, &[]);
             session.use_texture_render_changed();
             session.ui_handler.update_use_texture_render();
@@ -1234,6 +1237,9 @@ pub fn main_set_local_option(key: String, value: String) {
     }
     if is_d3d_render_key {
         for session in sessions::get_sessions() {
+            if !(session.is_default() || session.is_view_camera()) {
+                continue;
+            }
             session.update_supported_decodings();
         }
     }
--- a/src/flutter_ffi.rs
+++ b/src/flutter_ffi.rs
@@ -1224,10 +1224,12 @@ pub fn main_set_local_option(key: String, value: String) {
     let is_texture_render_key = key.eq(config::keys::OPTION_TEXTURE_RENDER);
     let is_d3d_render_key = key.eq(config::keys::OPTION_ALLOW_D3D_RENDER);
     set_local_option(key, value.clone());
+    let is_render_target =
+        |session: &crate::flutter::FlutterSession| session.is_default() || session.is_view_camera();
     if is_texture_render_key {
         let session_event = [("v", &value)];
         for session in sessions::get_sessions() {
-            if !(session.is_default() || session.is_view_camera()) {
+            if !is_render_target(&session) {
                 continue;
             }
             session.push_event("use_texture_render", &session_event, &[]);
@@ -1237,7 +1239,7 @@ pub fn main_set_local_option(key: String, value: String) {
     }
     if is_d3d_render_key {
         for session in sessions::get_sessions() {
-            if !(session.is_default() || session.is_view_camera()) {
+            if !is_render_target(&session) {
                 continue;
             }
             session.update_supported_decodings();
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -3454,9 +3454,7 @@ impl Connection {
                         self.update_auto_disconnect_timer();
                     }
                     Some(misc::Union::Option(o)) => {
-                        if self.should_handle_render_broadcast_message()
-                            || !Self::is_supported_decoding_only_option(&o)
-                        {
+                        if self.should_update_option_message(&o) {
                             self.update_options(&o).await;
                         }
                     }
@@ -5262,6 +5260,15 @@ impl Connection {
             self.authed_conn_type(),
             Some(AuthConnType::Remote | AuthConnType::ViewCamera)
         )
+    }
+
+    fn should_update_option_message(&self, option: &OptionMessage) -> bool {
+        match self.authed_conn_type() {
+            Some(AuthConnType::Remote) => true,
+            Some(AuthConnType::ViewCamera) => Self::is_view_camera_scoped_option(option),
+            Some(AuthConnType::Terminal) => Self::is_terminal_scoped_option(option),
+            Some(AuthConnType::FileTransfer | AuthConnType::PortForward) | None => false,
+        }
     }
 
     fn authed_conn_type(&self) -> Option<AuthConnType> {
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -5261,7 +5261,7 @@ impl Connection {
             return;
         };
         let Some(conn_type) = self.authed_conn_type() else {
-            // Unrachable, but just in case, we drop the options if the connection type is unknown.
+            // Unreachable, but just in case, we drop the options if the connection type is unknown.
             return;
         };
         let (scoped, violation) = Self::scoped_login_option(conn_type, &option);
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -3436,10 +3436,14 @@ impl Connection {
                     }
                     #[cfg(windows)]
                     Some(misc::Union::ToggleVirtualDisplay(t)) => {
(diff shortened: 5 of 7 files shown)


fufesou:
@coderabbitai review

fufesou:
@codex review

fufesou:
@codex review

fufesou:
@codex review

fufesou:
@coderabbitai review

fufesou:
@codex review

fufesou:
@codex review

fufesou:
@coderabbitai review

rustdesk:
• The clearest regression risks are:

  - Protocol compatibility: Any unlisted message immediately disconnects FileTransfer, Terminal, or ViewCamera sessions. A newer, older, plugin, or customized client may therefore be disconnected for sending
    a harmless message the whitelist does not know.

  - New message types require manual updates: Adding a new Message or Misc variant without updating these scope functions will break limited sessions at runtime. No exhaustive test currently detects this
    omission.

  - Runtime option changes can disconnect sessions: A ViewCamera or Terminal client sending any non-whitelisted OptionMessage field is disconnected. Known automatic senders were fixed, but future UI/
    background senders could reintroduce this.

  - Login-option behavior changes: FileTransfer and PortForward login options are now discarded entirely. ViewCamera and Terminal options are filtered. Custom clients relying on previously accepted options
    will behave differently.


  - Incomplete audit information: Most rejected messages are recorded only as "message" or "misc", making production regressions difficult to diagnose.
  - Over-broad nested whitelists: All current and future FileAction, FileResponse, and TerminalAction subtypes are automatically allowed. A future subtype could unintentionally gain access without a scope
    review.

  - Insufficient tests: Tests sample a few message types but do not cover every top-level and nested variant, actual disconnection, alarm submission, or client/server version combinations.

  The strongest concrete risk is the first one: the change replaces tolerant handling with fail-closed disconnection, creating a protocol-version compatibility hazard.

fufesou:
@codex review

rustdesk:
1. Previously released legacy clients (v1.4.x) still send TogglePrivacyMode / ToggleVirtualDisplay / ChangeResolution on the first frame of a view-camera session. The client-side protection introduced in this PR exists only in new clients; every connection from an old client will still trigger these messages.

2. Even the new client in this PR will trigger it: ui_session_interface.rs:1817 only blocks try_change_init_resolution, but the auto-OS-login path two lines later (should_auto_login → input_os_password) still sends MouseEvent / KeyEvent during a view-camera session. This is an internal inconsistency within the PR and should be fixed.

3. When the Flutter client switches between texture and D3D rendering, it broadcasts Misc::Option{supported_decoding} and RefreshVideo to all open sessions (including file transfer and terminal sessions). These are rejected by the new whitelist, and closing the session can terminate an ongoing file transfer.

4. Repeated LoginRequest messages arriving after authorization (due to a race condition between password submission and the CM Accept click) were previously explicitly tolerated, but are now treated as violations.

fufesou:
@codex review

fufesou:
@coderabbitai review

fufesou:
@codex review

fufesou:
@coderabbitai review

fufesou:
@codex review

rustdesk:
Here are the false-positive findings written in English, ready to paste as PR comments:

---
1. Already-shipped view-camera clients trigger scope violations on every connect — src/server/connection.rs (is_view_camera_scoped_misc)

Pre-PR clients (e.g. 1.4.x) unconditionally send Misc::ToggleVirtualDisplay / Misc::TogglePrivacyMode on the first video frame, and Misc::ChangeResolution via try_change_init_resolution, when the peer config has privacy-mode / virtual-display / open-in-original-resolution persisted from prior remote sessions to the same peer. None of these are in the view-camera allowlist, and the compensating client-side guards added in this PR only exist in clients built with it. Every view-camera connection from an existing client then hits the violation path: false SessionScopeViolation alarms with allow-scope-violation-alarm, or instant disconnect with allow-scope-violation-close.

2. Even the patched client still sends out-of-scope messages in view-camera sessions — src/ui_session_interface.rs:1817

The PR gates try_change_init_resolution behind !self.is_view_camera(), but two lines below, the auto-OS-login path still runs: should_auto_login() → input_os_password() sends several MouseEvent messages plus a KeyEvent carrying the OS password sequence. With lock-after-session-end, auto-login, and a stored os-password in the shared peer config (all persisted from earlier remote sessions), a fully legitimate view-camera session gets flagged — closed immediately after connect with the close option, or false alarms otherwise. This one is internal to the PR itself and should be fixed before merge.

3. Local render-option toggles are broadcast to file-transfer/terminal sessions and rejected — src/server/connection.rs (is_file_transfer_scoped_misc / terminal allowlist)

When the user flips "use texture rendering" (or D3D render) in desktop settings, main_set_local_option (src/flutter_ffi.rs) iterates sessions::get_sessions() — which includes file-transfer and terminal sessions — and each session sends Misc::Option{supported_decoding} plus Misc::RefreshVideo over the wire. Both are rejected by the file-transfer/terminal allowlists. With allow-scope-violation-close enabled, toggling a purely local render setting aborts an in-flight file transfer; with the alarm option, it posts false SessionScopeViolation audits.

4. A duplicate LoginRequest after authorization is now a scope violation — src/server/connection.rs:2504

Previously a LoginRequest arriving after authorized == true was explicitly tolerated (if self.authorized { return true; }). The new scope check runs before the LoginRequest branch, and LoginRequest is not in any non-Remote allowlist. If the host user clicks Accept in the connection manager at the same moment the remote user submits the password, the second LoginRequest lands after authorization and is classified as a violation — closing a just-accepted legitimate session with the close option, or posting a false alarm.

---
Common thread worth stating in the PR discussion: all four are invisible under default config (violations only warn-log and drop the message), but they make the allow-scope-violation-alarm / allow-scope-violation-close options unsafe to enable in real deployments until the allowlists cover what legitimate clients actually send.

fufesou:
@codex review

fufesou:
@coderabbitai review

rustdesk:
# Review: PR #15469 — Fix/session scope permission audit

https://github.com/rustdesk/rustdesk/pull/15469 (head: `fufesou/rustdesk@e857d0c`, base: `master`, +696/−28 across 7 files; depends on hbb_common PR #565)

This PR adds post-authorization message-scope enforcement to the controlled side: file-transfer, terminal, view-camera, and port-forward sessions now pass every inbound message through per-connection-type whitelists (`src/server/connection.rs`, +664 lines), with out-of-scope messages dropped, logged, and optionally alarmed (`allow-scope-violation-alarm`) or disconnected (`allow-scope-violation-close`, both opt-in). It also scopes login options by connection type, re-keys screenshots by `(VideoSource, display_idx)`, and adds client-side gates (toolbar, `io_loop`, `flutter_ffi`). The enforcement direction is sound and closes real holes (e.g. pre-PR, a file-transfer session could write the host clipboard and a camera session could inject keystrokes) — but the whitelists miss several message flows that legitimate, already-shipped clients routinely send, which turns normal usage into "violations."

All line references are to the PR head revision.

## Findings (most severe first)

1. **`src/server/connection.rs:5426` — Clipboard/MultiClipboards broadcasts from current clients are classified as scope violations in file-transfer and terminal sessions.** Flutter clients with a remote session open broadcast every clipboard copy to *all* sessions (`src/flutter.rs:1455`), including FT/terminal ones — the server never sends `Permission::Keyboard=false` for those types, so `is_text_clipboard_required()` stays true, and the PR leaves `send_clipboard_msg` untouched so even updated clients still broadcast. The routine "remote session + Transfer file/Terminal window" workflow now produces a warn log per copy by default, an unbounded audit-HTTP flood with alarm on, and kills the FT/terminal session mid-use with close on. Pre-PR this traffic was silently accepted. Fix direction: add clipboard messages to the FT/terminal broadcast-compat exemption (drop-without-flag, like `RefreshVideo`) and/or send the Permission messages at logon as the view-camera branch does (line 1950).

2. **`src/server/connection.rs:5452` — chat with terminal sessions is silently broken at default settings.** The desktop CM side panel renders an editable chat page for terminal clients and auto-opens on receive; the controlling client receives and replies via `misc::ChatMessage` over the terminal connection — all of which worked pre-PR. `is_terminal_scoped_misc` allows only `Option`, so the user's reply is now dropped with a warn (plus alarm/close if enabled). The view-camera whitelist includes `ChatMessage` (line 5481); terminal needs it too.

3. **`src/server/connection.rs:5463` — released 1.4.1–1.4.8 clients send auto-login input on view-camera connect, now flagged (or fatal).** For peers with `os-password` + `auto-login` + `lock-after-session-end` configured, released clients run `input_os_password` on `PeerInfo` in camera sessions, emitting ~5 MouseEvent + 2 KeyEvent messages. The PR fixes this client-side (`ui_session_interface.rs` gates it behind `!is_view_camera()`), implicitly acknowledging the flow, but old clients still send it: dropping the input is correct hardening, yet with `allow-scope-violation-close=Y` the legitimate camera session dies on connect, and with alarm on it files false `SessionScopeViolation` audits. These inputs deserve the same "tolerated no-op" carve-out the PR gives `TogglePrivacyMode`/`ChangeResolution`.

4. **`src/server/connection.rs:5289` — no rate limiting on violation handling: one audit HTTP task per out-of-scope message.** `handle_authorized_scope_violation` does two `Config::get_bool_option` reads, a warn log, and — with alarm on, close off — `post_alarm_audit` `tokio::spawn`s an HTTP POST *per message* (line 1511). A malicious authorized FT peer looping KeyEvents (or just finding 1's benign clipboard traffic) generates unbounded task creation and a self-inflicted flood of the audit server while the connection stays open. A per-connection (or per-message-family) once-latch plus cached config reads would cap it.

5. **`src/server/connection.rs:3457` — the runtime option path validates with the scoped filter but applies the original unfiltered message, and the four field enumerations fail open on future fields.** `should_update_option_message` uses `scoped_view_camera_option`/`scoped_terminal_login_option` only for their boolean verdict, then `update_options(&o)` applies the full message; the carefully built scoped copy is used only at login. The parallel hand-enumerated field lists (5395, 5532, 5547, plus `option_has_any_field`) cover every current `OptionMessage` field, but a future proto field passes all of them unflagged and is then applied verbatim in scoped sessions — the security allow-list silently rots into a stale deny-list with no compile-time backstop. Applying the scoped copy on both paths, and expressing the checks as "clear allowed fields on a clone, compare to default" (protobuf derives `PartialEq`), would make it fail closed and collapse ~40 triplicated lines.

6. **`src/server/connection.rs:5306` — the "login options outside scope" warning fires on essentially every view-camera login from stock clients.** `get_option_message` (`src/client.rs:2589`, untouched by this PR) has no view-camera branch and sends `enable_file_transfer=Yes` whenever the default-on `enable-file-copy-paste` option is set and view-only is off — so `option_has_non_view_camera_login_field` flags virtually every camera login. Filtering is harmless (camera sessions ignore the field) but the warn is permanently noisy, burying the real violations the log is meant to surface.

7. **`src/server/connection.rs:5507` — unknown `ImageQuality` enum values make `scoped_view_camera_option` bail with an *empty* option, discarding in-scope fields.** The `let-else` early-return drops `supported_decoding`/`custom_fps`/`disable_audio` at login (pre-PR, an unknown value skipped that one field and applied the rest), and mid-session the same message becomes a hard scope violation. Latent today (the enum has been frozen for years), and fail-closed is defensible for a security filter — but the return is also asymmetric: `Some(empty)` here vs `None` in the terminal path (5354), and field-level ignoring would match pre-PR behavior for benign QoS messages.

8. **`src/server/connection.rs:5377` — pre-PR clients can trip a port-forward scope violation during the login handshake.** Old clients broadcast texture-render/d3d option changes to all sessions including port-forward tunnels, and `port_forward.rs` forwards `Data::Message` to the wire until `PeerInfo` arrives; if one lands in the ~1-RTT window where the server is authorized but still in `on_message` (waiting on an in-flight TestDelay echo, connection.rs:939), `PortForward => false` fires the violation and close=Y tears down the fresh tunnel. Narrow race, but the fix is cheap: extend the render-broadcast compat exemption to `PortForward`.

## Minor cleanups (below the cut)

- `connection.rs:5396`, `5551` — `option.supported_decoding.clone().take().is_some()` deep-clones a protobuf message for a presence check on the per-message hot path; `is_some()` on the `MessageField` suffices (AGENTS.md: "Avoid unnecessary `.clone()`").
- `connection.rs:1524` — the new alarm payload uses `"peer_id"` and omits `"name"`, while every other `post_alarm_audit` caller sends `"ip"`/`"id"`/`"name"`; audit-server ingestion of this alarm type will show empty peer fields.
- The new handler guards use `!self.view_camera` (3442, 3447, 3544, 3550, 4135) while the gate keys off `authed_conn_type()` — two sources of truth for the same policy, and the view-camera compat messages are admitted by the gate but neutralized per-handler, so every future privileged handler must remember its own guard.

## Candidates examined and cleared

The authorized-`LoginRequest` early-return (2505) is fine — no released client re-sends a login post-authorization, and freezing `self.lr` prevents an authorized peer rewriting the identity used in audit records. The screenshot display-index collision is *fixed* by this PR's `(VideoSource, display_idx)` re-keying, not introduced (old servers remain affected regardless of client). Note also that both enforcement options are opt-in (`allow-*` defaults to off), so out of the box the PR only filters and warn-logs — the PR description's "trigger a SessionScopeViolation alarm audit" applies only once an admin sets `allow-scope-violation-alarm=Y`.

---

*Review method: 8 independent finder passes (line-by-line, removed-behavior, cross-file tracing, reuse, simplification, efficiency, altitude, conventions) over the PR diff and head files, followed by one adversarial verifier per surviving candidate. 37 raw candidates → deduped → 8 confirmed findings; refuted candidates listed above.*

fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@codex review

fufesou:
@codex review

fufesou:
@coderabbitai review

fufesou:
@codex review

fufesou:
@codex review

fufesou:
@codex review

fufesou:
@coderabbitai review

21pages:
* Tested older client versions as the controller with different connection types. Try selecting menu items whenever possible. Test versions including 1.2.3, 1.3.1, 1.3.5, 1.3.9, and 1.4.0 through 1.4.8.
* Tested alarm log reporting and connection disconnection when the option is enabled.
* Analyzed the Message field with codex.

21pages (approved):

PR #15493, opened 2026-07-02
fix(clipboard): make CLIPRDR format-map growth checked

fufesou:
This is not a typical fix—it's hardening.

The Windows CLIPRDR format-list handler relies on `map_ensure_capacity()` while processing peer-provided formats.

The previous helper only attempted growth: if realloc() failed, it returned silently, and the caller continued processing. A later iteration could then index past the allocated format_mappings array.

Make format-map growth a checked operation. The handler now validates the peer-provided format count, ensures the mapping array is large enough before writing entries, and aborts processing if growth fails. Newly allocated slots are zeroed, so existing cleanup can safely run after partial processing.

Also, bind remote format names before measuring/converting them. The chosen limits follow Windows clipboard/atom constraints:
  - registered clipboard format IDs use 0xC000..0xFFFF
  - string atom names are limited to 255 bytes

## Testing

- [x] Windows -> Windows, files & dirs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remote clipboard format handling for Windows sessions, adding bounded validation for format counts and format-name sizes.
  * Added stricter checks for malformed or oversized clipboard format lists and names, including safer UTF-8 to UTF-16 conversion and empty-name rejection.
  * Reduced clipboard-related failures by failing fast on invalid input and consistently resetting internal state on errors.

### DIFF
--- a/libs/clipboard/src/windows/wf_cliprdr.c
+++ b/libs/clipboard/src/windows/wf_cliprdr.c
@@ -41,6 +41,12 @@
 
 /* Maximum number of clipboard streams accepted from a remote peer (integer overflow / DoS guard) */
 #define WF_CLIPRDR_MAX_STREAMS 16384
+/* Registered clipboard formats use IDs 0xC000 through 0xFFFF.
+ * https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerclipboardformatw */
+#define WF_CLIPRDR_MAX_FORMATS 0x4000u
+/* Registered format names are string atoms, limited to 255 bytes.
+ * https://learn.microsoft.com/en-us/windows/win32/dataxchg/about-atom-tables */
+#define WF_CLIPRDR_MAX_FORMAT_NAME_BYTES 255u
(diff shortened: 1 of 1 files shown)


fufesou:
@codex review

rustdesk:
• The safety changes are generally sound, but the new byte-based name limit rejects clipboard format names that the Windows wide-character API accepts.

  Review comment:

  - [P2] Count decoded characters instead of UTF-8 bytes — /Users/hzhou/rustdesk/libs/clipboard/src/windows/wf_cliprdr.c:49-49
    When a valid format name contains multibyte UTF-8 characters, this byte cap rejects it even if its UTF-16 name is within RegisterClipboardFormatW's 255-character limit; for example, 128 CJK characters
    occupy 384 UTF-8 bytes. Validate the converted WCHAR count instead, otherwise such names cause the entire remote format list to fail.

fufesou:
@codex review

fufesou:
@coderabbitai review

PR #15514, opened 2026-07-04
fix: non-E2EE show dialog

fufesou:
Show the dialog before continuing when an E2EE connection cannot be established.

## Preview

https://github.com/user-attachments/assets/b9f02f84-2384-43f8-9ef1-2e4c084f4310

image

image

## Testing

Flutter, sciter, web, mobile

- [x] Normal connections
- [x] E2EE connections. Input password, remember password, click

## Summary by CodeRabbit

* **New Features**
  * Added an insecure-connection prompt with **Continue** or **Disconnect**, including the ability to remember your confirmation for the session.
  * Added an E2EE verification-unavailable warning with an explicit **Continue** option.
* **Bug Fixes**
  * Improved insecure-connection flow so the correct dialogs and actions reliably proceed or terminate.
  * Ensured unsecured connections are blocked when the confirmation is denied, and made disconnect timing behavior more consistent (including web).
* **Localization**
  * Added the new **Continue** and E2EE warning strings across supported languages.


fufesou:
@codex review

fufesou:
@coderabbitai review

rustdesk:
• The web implementation still proceeds without confirmation, and native sessions can hang when the peer disappears during the prompt.

  Full review comments:

  - [P1] Gate insecure web connections before continuing — /Users/hzhou/rustdesk/flutter/lib/web/bridge.dart:1919-1922
    When Flutter web connects and secure() returns false, the transport in flutter/web/js/src/connection.ts emits connection_ready and continues without entering the Rust confirmation flow. Moreover,
    globals.js has no setByName('common', ...) handler, making this method a no-op. Add equivalent web-side prompting and connection gating.

  - [P2] Observe peer closure while awaiting confirmation — /Users/hzhou/rustdesk/src/client/io_loop.rs:394-394
    If the peer disconnects or the network drops while this prompt is open, only the UI receiver is polled; the peer stream and normal timeout loop are inactive. The session therefore remains connected with
    the dialog open until the user responds. Also wait for peer closure or another liveness signal here.

fufesou:
https://github.com/rustdesk/rustdesk/pull/15514#issuecomment-4881022583

> [P2] Observe peer closure while awaiting confirmation — /Users/hzhou/rustdesk/src/client/io_loop.rs:394-394

This isn't worth changing here. During the confirmation dialog, we shouldn't call `peer.next()` in a loop without buffering – we'd risk dropping handshake messages. If the peer disconnects while the dialog is open, it just leaves a stale dialog; it doesn't bypass confirmation. I'd keep this PR focused on the E2EE gating.

21pages:
Tested Flutter desktop, web, and Sciter as the controller.

 * Port forward is not handled

21pages (approved):

fufesou:
@codex review

fufesou:
@codex review

rustdesk:
• The confirmation flow can time out active transports, permanently hangs command-line port forwarding, and fails to terminate GUI port-forward listeners after rejection.

  Full review comments:

  - [P1] Provide confirmation for CLI port forwarding — /Users/hzhou/rustdesk/src/port_forward.rs:124-124
    When --port-forward reaches an unsecured ID-based connection, this awaits ContinueInsecureConnection or Close, but cli::Session::msgbox only logs this message type and sends neither event. The command
    therefore hangs permanently after the first local client connects instead of prompting or failing.

  - [P2] Keep the transport alive while awaiting confirmation — /Users/hzhou/rustdesk/src/client/io_loop.rs:192-192
    While this await is pending, the peer stream is not read and TestDelay traffic is not answered. If the warning remains open for roughly the server's 30-second timeout, the server closes the stream, so
    selecting Continue subsequently fails with a reset; this also affects the port-forward path using the same helper.

  - [P2] Propagate port-forward rejection to the listener — /Users/hzhou/rustdesk/src/port_forward.rs:124-125
    When a GUI user selects Disconnect, the helper consumes Data::Close and this converts it into a generic error. listen then reports a connection error and resumes its outer loop because the close event is
    gone, leaving the session's local listener alive after the tab or window was closed.

fufesou:
@codex review

rustdesk:
- [P1] Gate insecure Flutter-web connections — /Users/hzhou/rustdesk/flutter/lib/web/bridge.dart:1919-1921
    When E2EE negotiation fails in Flutter web, connection.ts emits connection_ready and immediately continues without showing this dialog. Moreover, window.setByName has no common handler, so this added
    response path is ignored; web sessions therefore still proceed insecurely without confirmation.


  - [P2] Keep the peer serviced while awaiting confirmation — /Users/hzhou/rustdesk/src/client/io_loop.rs:192-192
    If the dialog remains open for roughly 30 seconds, this await prevents polling peer.next() and replying to TestDelay; the remote server's SEC30 inactivity check then closes the connection. Selecting
    Continue afterward consequently resumes onto an already-reset stream, including in the analogous port-forward flow.

  - [P2] Cache insecure approval for the port-forward listener — /Users/hzhou/rustdesk/src/port_forward.rs:128-129
    Each accepted local TCP socket calls this confirmation independently, and a Continue result is not retained. Multi-connection clients such as browsers, database tools, or reconnecting RDP clients will
    therefore be blocked by the warning repeatedly within the same port-forward session.

  - [P2] Let Enter activate the focused Continue button — /Users/hzhou/rustdesk/src/ui/msgbox.tis:196-199
    For an insecure-connection dialog, the keydown handler routes Enter through submit(), but this branch always calls cancel(). Thus pressing Enter while the Continue button is keyboard-focused disconnects
    the session instead of activating that button.

fufesou:
https://github.com/rustdesk/rustdesk/pull/15514#issuecomment-4890488112

> [P1] Gate insecure Flutter-web connections — /Users/hzhou/rustdesk/flutter/lib/web/bridge.dart:1919-1921

The web side is not updated.

> [P2] Keep the peer serviced while awaiting confirmation — /Users/hzhou/rustdesk/src/client/io_loop.rs:192-192

Not worth.

https://github.com/rustdesk/rustdesk/pull/15514#issuecomment-4881082028

> [P2] Cache insecure approval for the port-forward listener — /Users/hzhou/rustdesk/src/port_forward.rs:128-129

Not worth. This dialog is a rare case.

> [P2] Let Enter activate the focused Continue button — /Users/hzhou/rustdesk/src/ui/msgbox.tis:196-199

Intended.

21pages:
Tested remote desktop and port forwarding with Flutter Desktop, Web, and Sciter acting as the controller.

21pages (approved):

OmlineEditor:
I keep seeing this message about an insecure connection. I tried opening a ticket, but the developer gives generic answers and doesn't help me solve the problem. I checked all the keys, and they're the same on the server and in my program. What's the problem? What should I do to make this insecure connection message go away?

My: https://github.com/rustdesk/rustdesk/discussions/15578
Another person: https://github.com/rustdesk/rustdesk/discussions/15544

PR #15530, opened 2026-07-07
fix: ci: macos: allow signed but not notarized dmg

foresee-io:
Community developers may use self-signed Apple Developer certificates without paying the developer annual fee.

The previous CI process assumed that users carrying `MACOS_P12_BASE64` would also carry `MACOS_NOTARIZE_JSON` . However, this is not feasible for them. Therefore, this Pull Request strengthens the validation in this regard, thereby:

Allowing these developers to use CI to build self-signed packages.

## Summary by CodeRabbit

* **New Features**
  * macOS builds now conditionally enable notarization when the required notarization secret is present.
* **Bug Fixes**
  * Improved the macOS signing workflow so notarization steps run only when all necessary credentials are available.
  * Added stricter validation for the macOS codesign identity (alphanumeric-only), failing early on invalid values.
  * Prevented notarization from running unconditionally during the signed DMG creation step.

### DIFF
--- a/.github/workflows/flutter-build.yml
+++ b/.github/workflows/flutter-build.yml
@@ -693,15 +693,20 @@ jobs:
 
       - name: Check sign and import sign key
         if: env.MACOS_P12_BASE64 != null
+        env:
+          MACOS_CODESIGN_IDENTITY: ${{ secrets.MACOS_CODESIGN_IDENTITY }}
+          MACOS_NOTARIZE_JSON: ${{ secrets.MACOS_NOTARIZE_JSON }}
+        shell: bash
         run: |
           security default-keychain -s rustdesk.keychain
           security find-identity -v
-
-      - name: Check if notarization is configured
-        env:
-          MACOS_NOTARIZE_JSON: ${{ secrets.MACOS_NOTARIZE_JSON }}
-        run: |
-          if [ -n "$MACOS_NOTARIZE_JSON" ];then echo "MACOS_NOTARIZE_ENABLE=1" >> $GITHUB_ENV; fi
+          i=
+          if [ -n "$MACOS_NOTARIZE_JSON" ];then i=1; else i=0; fi
+          echo "MACOS_NOTARIZE_ENABLE=$i" >> $GITHUB_ENV
+          if ! [[ "$MACOS_CODESIGN_IDENTITY" =~ ^[A-Za-z0-9+/]*={0,2}$ ]]; then
+            echo "FATAL: Invalid `secrets.MACOS_CODESIGN_IDENTITY` given. If signing key is stored on your Mac, you can run `security find-identity -v -p codesigning` to find out hex format of your identity." >&2
+            exit 128
+          fi
 
       - name: Import notarize key
         if: env.MACOS_P12_BASE64 != null && env.MACOS_NOTARIZE_ENABLE == 1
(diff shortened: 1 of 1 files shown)


.github/workflows/flutter-build.yml
```diff
@@ -846,6 +857,10 @@ jobs:
           codesign --force --options runtime -s ${{ secrets.MACOS_CODESIGN_IDENTITY }} --deep --strict ./flutter/build/macos/Build/Products/Release/RustDesk.app -vvv
           create-dmg --icon "RustDesk.app" 200 190 --hide-extension "RustDesk.app" --window-size 800 400 --app-drop-link 600 185 rustdesk-${{ env.VERSION }}.dmg ./flutter/build/macos/Build/Products/Release/RustDesk.app
           codesign --force --options runtime -s ${{ secrets.MACOS_CODESIGN_IDENTITY }} --deep --strict rustdesk-${{ env.VERSION }}.dmg -vvv
+
+      - name: Notarize dmg
+        if: env.MACOS_NOTARIZE_ENABLE == 1
```
rustdesk: Why not use `env.MACOS_NOTARIZE_JSON != null` directly?

foresee-io: It is my option, not a necessity. MACOS_NOTARIZE_JSON is stored as a "secret", unless you expose that at the beginning of CI in advance.

rustdesk: You can add it to the beginning of CI

.github/workflows/flutter-build.yml
```diff
@@ -693,12 +694,20 @@ jobs:
 
       - name: Check sign and import sign key
         if: env.MACOS_P12_BASE64 != null
+        env:
+          MACOS_CODESIGN_IDENTITY: ${{ secrets.MACOS_CODESIGN_IDENTITY }}
+        shell: bash
         run: |
           security default-keychain -s rustdesk.keychain
           security find-identity -v
+          if ! [[ "$MACOS_CODESIGN_IDENTITY" =~ ^[A-Za-z0-9]+$ ]]; then
```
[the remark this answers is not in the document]
foresee-io: It is NOT intended as a correctness check.

This Regex is only used to filter out special characters, excluding cases where parameters are valid but would be interpreted by Bash as special characters.

.github/workflows/flutter-build.yml
```diff
@@ -49,6 +49,7 @@ env:
   #signing keys env variable checks
   ANDROID_SIGNING_KEY: "${{ secrets.ANDROID_SIGNING_KEY }}"
   MACOS_P12_BASE64: "${{ secrets.MACOS_P12_BASE64 }}"
+  MACOS_NOTARIZE_JSON: "${{ secrets.MACOS_NOTARIZE_JSON }}"
```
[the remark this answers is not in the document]
foresee-io: Merely knowing the macOS Notarize JSON without knowing the decrypted signing key does NOT POSE THREAT to signing key privacy.

Moreover, the syntax here referenced from ANDROID_SIGNING_KEY, and has been acknowledged by the developer.

rustdesk:
Reverted since you make our CI failed. https://github.com/rustdesk/rustdesk/actions/runs/29063717350/job/86271667857

rustdesk:
Let us stop this boring PR, please do not submit again.

rustdesk:
Let us stop this boring PR.

PR #15532, opened 2026-07-07
feat(terminal): add Ctrl and Alt toggles to mobile terminal keyboard

dongrencd:
## Summary

Adds a collapsible third row to the mobile terminal floating keyboard with one-shot `Ctrl` and `Alt` modifiers. The latest reviewed keyboard layout is preserved, with a full-width expand/collapse control at the end of Row2.

## Layout

```text
Row1: Esc   /   |   Home   Up   End   \
Row2: Tab   Ctrl+C   ~   Left   Down   Right   [toggle]
Row3: Ctrl   Alt   -   PgUp   PgDn   [blank]   [blank]
```

Row3 is collapsed by default and its state is persisted locally. Collapsing Row3 clears active modifier locks so hidden state cannot affect later input.

## Input Behavior

- `Ctrl` and `Alt` are one-shot toggles for the next ordinary single-character keyboard input.
- Terminal control bytes (`Esc`, `Tab`, `Enter`, `Ctrl+C`, and `DEL`) and multi-character escape sequences remain unchanged and do not consume a pending modifier.
- `Ctrl` supports letters, space, `[`, `\\`, `]`, `^`, `_`, and `/`; `Alt` prefixes the mapped input with `ESC`.
- Mobile keyboard `LF` is normalized to `CR`, while generated `Ctrl+J` remains `LF`.
- Clipboard paste bypasses keyboard-only modifier and newline transformations, including one-character paste, and honors xterm bracketed-paste mode.
- Hardware `Ctrl+V` and `Cmd+V` use the same paste path as the terminal context-menu paste action.

## Layout Preservation

Every key and placeholder keeps the reviewed 48 dp width with 2 dp gaps, so all three expanded rows remain aligned at 348 dp.

## Validation

- Flutter 3.24.5 full test suite: 40 tests passed.
- Focused static analysis completed successfully; only existing Flutter deprecation notices remain.
- The PR contains no Android keep-alive or foreground-service changes.


rustdesk:
Screenshots please. We won't merge if it does not look pretty for us.

rustdesk:
3 rows is too much. Please add a collapse button on row 2, let users choose (persist the choice) to show it or hide it.

rustdesk:
Please do not modify original two rows. 

Please respect original choice (https://github.com/rustdesk/rustdesk/pull/13876) from another contributor. @alonginwind

dongrencd:
Will provide screenshots after the APK is built and tested. Thanks for the review!

dongrencd:
Done — added a collapse/expand toggle button at the end of Row2. The extra Ctrl row defaults to collapsed (same as original 2-row layout), and the choice is persisted via `terminal-extra-ctrl-keys-expanded` local config option.

dongrencd:
Fixed — Row1 and Row2 are now identical to the original layout. All new Ctrl keys are in the collapsible Row3 only.

dongrencd:
@coderabbitai All nitpick comments have been addressed. Please re-review.

alonginwind:
Since there's already a Ctrl Lock feature, is there still any need for separate buttons like Ctrl+C or Ctrl+X? Wouldn't keeping only the Ctrl interface make it cleaner?

rustdesk:
I prefer to keep ctrl+c

dongrencd:
Agreed — our latest approach removes separate Ctrl+C/Ctrl+X buttons entirely. Row2 replaces Ctrl+C with a single Ctrl toggle. When active, any key sends its control code (e.g. Ctrl active + `c` → `\x03`). No extra rows, no extra buttons.

dongrencd:
The Ctrl toggle now covers Ctrl+C natively: activate Ctrl, press `c` → sends `\x03`. This keeps the layout minimal (only 2 rows unchanged from the original) while supporting all Ctrl combinations.

rustdesk:
I prefer to keep ctrl + C since Ctrl + C is frequently used (also for respecting original contributor),  "activate Ctrl,  press c, deactivate ctrl" is not very convinient. I prefer 3 rows (collpased), because we can add more keys later in the future.

rustdesk:
> I prefer to keep ctrl+c

You misunderstood me, I prefer old two rows including its ctrl c.

alonginwind:
> I prefer to keep ctrl + C since Ctrl + C is frequently used (also for respecting original contributor), "activate Ctrl, press c, deactivate ctrl" is not very convinient. I prefer 3 rows (collpased), because we can add more keys later in the future.

How about this? Press Ctrl once, then press C. Ctrl will automatically unlock after the key combination is sent. Press Ctrl twice to lock it, and pressing other keys won't automatically unlock it.

rustdesk:
> Press Ctrl once, then press C. Ctrl will automatically unlock after the key combination is sent.

Good idea, but please still keep old two rows unchange. 1 press (ctrl c) is still better than two press.

dongrencd:
Good idea — single tap for one-shot, double tap for sticky lock. We will keep the current one-shot behavior for now and can add double-tap lock as a future enhancement.

dongrencd:
Understood. Final layout:
- Row1 + Row2: identical to #13876 (Ctrl+C stays)
- Row3: always visible, contains Ctrl + Alt toggle buttons
- Row3 is reserved for future extensibility (more modifier keys can be added)

No collapse button — it is not visible on small screens.

dongrencd:
@coderabbitai All review comments addressed:
- Extracted common `_buildModifierToggleButton` helper (removed duplicated `_buildCtrlKeyButton`/`_buildAltKeyButton`)
- `Semantics` labels now use `translate()` for localization
- Replaced deprecated `surfaceVariant` with `surfaceContainerHighest`

Please re-review.

dongrencd:
not_visible

The collapse button is not visible on small screens — it gets cut off at the right edge of Row2. Therefore Row3 is always shown instead of using a collapse button.

always_shown

dongrencd:
Final layout verified on device. Row1+Row2 unchanged from #13876. Row3 contains Ctrl, Alt, and `-` keys. 
final_layout

All CodeRabbit suggestions addressed. Ready for review and merge.

dongrencd:
Row3 alignment fixed — changed from `MainAxisAlignment.start` to `MainAxisAlignment.center` to match Row1 and Row2 alignment. The three buttons (Ctrl, Alt, `-`) are now centered consistently with the rows above.

rustdesk:
3 rows looks too wasteful to me

rustdesk:
How about put ... (collapse) button on pagedn (not often used) button, and move pagedn button to 3rd row?

dongrencd:
@rustdesk Updated layout based on your feedback. Please review:

**Layout:**
```
Row1: Esc / Home ↑ End PgUp |       (7 keys, | moved to end)
Row2: Tab Ctrl+C ← ↓ → PgDn [∨/∧]  (7 keys, collapse button aligned with | above)
Row3: Ctrl Alt [-] ~                 (shown when expanded)
```

**Key points:**
- Row1: `|` moved from position 3 to end (position 7), all other keys unchanged
- Row2: `~` removed, collapse button added at position 7 (directly below `|` in Row1)
- Collapse button shows `∨` when collapsed, `∧` when expanded
- Row3 only appears when expanded, contains: Ctrl toggle, Alt toggle, `-`, and `~`
- Ctrl/Alt: single press = one-shot (auto-unlock after one key), double press = lock mode
- Collapse state persisted via local config
- `Ctrl+C` kept in Row2 as requested

Does this layout work for you?

rustdesk:
I still insist to move pagedn to 3rd row because it is more infrequently used.


> PgUp and PgDn are naturally paired, but ~ is a frequently typed terminal character (e.g. ~/, cd ~). Since Row3 is collapsed by default, hiding ~ may hurt usability more than hiding PgDn.

dongrencd:
@rustdesk Good point about ~ being more frequently used. How about this alternative:

**Move `|` to Row3, move `~` to Row1 (aligned with collapse button):**
```
Row1: Esc / Home ↑ End PgUp ~      (7 keys, ~ at end, aligns with ∨/∧ below)
Row2: Tab Ctrl+C ← ↓ → PgDn [∨/∧]  (7 keys, collapse button at end)
Row3: Ctrl Alt [-] |                 (4 keys when expanded)
```

**Why PgUp/PgDn should stay paired:**
- They are naturally used together for scrolling (up/down in man pages, logs, history)
- Every terminal emulator keeps them side by side — splitting them across rows breaks muscle memory
- Users expect navigation keys (↑↓←→ PgUp PgDn) to be grouped together
- PgDn has no alternative when scrolling, but ~ can be replaced

**Why ~ can be substituted:**
- `~/path` → `/home/user/path` (absolute path works the same way)
- `cd ~` → `cd /home/user` (equivalent)
- While ~ is convenient, it is not essential — absolute paths are always available
- `|` is even more substitutable via copy/paste or virtual keyboard

**Result:**
- `~` moves to Row1 for easy access (more frequent than `|`)
- `|` moves to Row3 (less frequent, can be replaced)
- PgUp/PgDn stay paired in Row1/Row2 (intuitive scrolling)
- `~` aligns with `∨/∧` below (visual consistency)

Does this layout work for you?

rustdesk:
Let me ask chatgpt again

rustdesk:
# chatgpt

I still slightly prefer moving **PgDn** to the 3rd row rather than `~`.

My reasoning is that the always-visible rows should be reserved for the most frequently used terminal actions:

* Esc
* Tab
* Ctrl+C
* Home / End
* Arrow keys
* `|`

On mobile terminals, `PgDn` tends to be used less often because users can usually scroll with touch gestures, while `|` is commonly used in shell commands and pipelines.

That said, from a PR/review perspective, keeping the original Row1/Row2 layout unchanged and moving `~` to Row3 is probably the safer choice. It preserves the existing layout from #13876 and minimizes behavioral changes, which makes the patch easier to review and accept.

So:

* **Product preference:** move **PgDn** to Row3.
* **Upstream/maintenance preference:** keep **PgDn** where it is and move **`~`** to Row3.

rustdesk:
I even suggest move both pageup/dn to 3rd, and add another frequently used key.


# chatgpt

For a mobile terminal, the keys I'd consider before **PgUp/PgDn** are:

* `-`
* `_`
* `~`
* `|`
* `\`
* `Ctrl`
* `Alt`

Of those, if I had **one extra always-visible slot**, I'd probably choose **`\`**.

Why `\`?

* Very common in paths, escaping, regexes, shell scripting, and SSH-related commands.
* Much harder to type on many mobile keyboards than letters or numbers.
* Unlike `~`, there isn't usually a convenient alternative.
* Often used together with `|`, which is already on the toolbar.

So my preference would be:

```text
Row1: Esc  /  Home  ↑  End  \  |
Row2: Tab  Ctrl+C  ←  ↓  →  ...
```

If you're optimizing specifically for Linux shell usage, my ranking would roughly be:

```text
\  >  ~  >  -  >  PgUp/PgDn
```

So if you're asking me to replace **PgUp/PgDn** with a single key, I'd vote for **`\`** first.

dongrencd:
I agree that `\` is very useful for Linux shell/SSH usage and is hard to type on mobile keyboards.

My only concern is scope. Replacing PgUp/PgDn with `\` changes the existing toolbar behavior more significantly than just making Row3 collapsible.

For this PR, would you prefer one of these directions?

1. Minimal change: keep the current Row1/Row2 layout and move only PgDn to Row3 for the expand/collapse toggle.

   ```text
   Row1: Esc  /  |  Home  ↑  End  PgUp
   Row2: Tab  Ctrl+C  ~  ←  ↓  →  [∨/∧]
   Row3: Ctrl Alt  [-]  PgDn
   ```

2. Shell-optimized change: make \ always visible and move PgUp/PgDn to Row3.
   ``` text
   Row1: Esc  /  Home  ↑  End  \  |
   Row2: Tab  Ctrl+C  ~  ←  ↓  →  [∨/∧]
   Row3: Ctrl Alt  [-]  PgUp  PgDn
   ```

I’m fine with either direction, but option 1 seems safer for preserving the existing layout from #13876, while option 2 is better if we want to optimize the toolbar for shell input.

rustdesk:
I choose `#2`, but keep `|` in old place.
```
Row1: Esc  /  |  Home  ↑  End  \
Row2: Tab  Ctrl+C  ~  ←  ↓  →  [∨/∧]
Row3: Ctrl Alt  -   PgUp PgDn
```

dongrencd:
Updated the layout to match your latest choice (#2) while keeping `|` in the original position.

Final layout:
```text
Row1: Esc  /  |  Home  ↑  End  \
Row2: Tab  Ctrl+C  ~  ←  ↓  →  [∨/∧]
Row3: Ctrl Alt  -   PgUp PgDn
```

I also kept Row3 width aligned with Row1/Row2 and added a short README note for the mobile terminal floating keyboard.

rustdesk:
# codex

• The patch introduces functional regressions in the mobile terminal keyboard: stale padding after expanding/collapsing, default expansion due option semantics, and rows wide enough to overflow common phone
  screens. It also regresses an existing Italian translation.

  Full review comments:

  - [P2] Recompute keyboard height after row toggles — /Users/hzhou/rustdesk/flutter/lib/mobile/pages/terminal_page.dart:438-441
    When the user expands or collapses Row3, _keyboardHeight remains the value measured at startup, but _calculatePadding() uses that cached height to keep terminal content above the floating keyboard. On
    any toggle, the terminal will reserve too little space after expanding (Row3 covers the bottom of the terminal) or too much after collapsing; update the measured height after this setState completes.

  - [P2] Avoid default-expanding the persisted Row3 state — /Users/hzhou/rustdesk/flutter/lib/consts.dart:180-180
    For users with no saved value yet, this key is read via mainGetLocalBoolOptionSync, and option2bool() treats any option starting with enable- and an empty value as true. That makes the new Row3 expanded
    by default even though _row3Expanded is initialized as collapsed and this is a collapse-state preference, so fresh mobile terminals start in the larger 3-row layout.

  - [P2] Keep Row1 within mobile width — /Users/hzhou/rustdesk/flutter/lib/mobile/pages/terminal_page.dart:359-361
    On portrait phones around 360–390 logical pixels wide, keeping PgUp and appending the new \ button makes Row1 at least 8*48 + 7*2 = 398 px wide, so Flutter will render a horizontal overflow. Since Row3
    also contains PgUp, this should replace the old slot rather than adding an eighth always-visible key.

  - [P2] Replace PgDn instead of appending the toggle — /Users/hzhou/rustdesk/flutter/lib/mobile/pages/terminal_page.dart:380-382
    This keeps PgDn in Row2 while also adding the collapse button, giving Row2 the same 398 px minimum width and causing overflow on common portrait mobile widths. Row3 already includes PgDn, so the collapse
    button needs to occupy the old paging slot rather than being appended.

dongrencd:
Addressed the latest review comments in `3fa4d3181`.

Changes:
- Row1 now keeps 7 slots: `Esc / | Home ↑ End \\`; `\\` replaces the old `PgUp` slot instead of being appended.
- Row2 now keeps 7 slots: `Tab Ctrl+C ~ ← ↓ → [∨/∧]`; the collapse button replaces the old `PgDn` slot instead of being appended.
- Row3 still contains `Ctrl Alt - PgUp PgDn` and keeps two trailing placeholders for row-width alignment.
- Row3 no longer defaults to expanded when there is no saved preference; the persisted option now uses a non-`enable-*` key and raw option read.
- Expanding/collapsing Row3 now remeasures the floating keyboard height on the next frame so terminal padding uses the updated height.
- Reverted the README sentence to avoid translation drift.

Validation:
- `git diff --check` passed.
- `flutter analyze lib/mobile/pages/terminal_page.dart` now has no analyzer errors. It still reports existing warnings/infos in this file, so I also ran `flutter analyze --no-fatal-warnings --no-fatal-infos lib/mobile/pages/terminal_page.dart`, which exits successfully.

rustdesk:
• The modifier mapping order breaks at least Ctrl+J on mobile/web-mobile by converting the generated LF into CR. This is a functional regression in the new Ctrl-lock behavior.

  Review comment:

  - [P2] Normalize mobile Enter before applying modifier locks — /Users/hzhou/rustdesk/flutter/lib/models/terminal_model.dart:95-97
    On mobile/web-mobile, when Ctrl lock is active and the user types j, _applyCtrlKeyMapping correctly turns it into \n for Ctrl+J, but the existing mobile newline normalization that runs afterward rewrites
    that to \r. This makes a standard Ctrl+J shortcut send Enter/CR instead of LF; normalize the original soft-keyboard Enter before applying modifier mappings so generated control codes are not mutated.

dongrencd:
Fixed in `3a7bcb0fc9e4da7a4649e27635459ab1dada274d`.

Changes:
- Normalized the original mobile/web-mobile soft-keyboard Enter (`
` -> `
`) before applying Ctrl/Alt lock mappings.
- Kept modifier-generated control codes untouched, so Ctrl+J now sends LF (`
`) instead of being rewritten to CR (`
`).
- Preserved the existing pasted-text behavior by only normalizing a lone original `
`.

Verification:
- Ran a focused Dart input-order check for mobile Enter, mobile Ctrl+J, and pasted text.
- `flutter analyze lib/models/terminal_model.dart` passes with no issues.
- Full `flutter analyze` was also run, but it still reports existing baseline analyzer errors outside this change.

Actions:
- The upstream PR workflows for this commit were created but are marked `action_required` with no jobs, so they need maintainer approval to run.
- I also started the fork `Full Flutter CI` for the PR branch: https://github.com/dongrencd/rustdesk/actions/runs/29010617919

rustdesk:
@fufesou

rustdesk:
# Codex Review

The modifier implementation can alter pasted data, retain invisible active modifiers, and fails a standard Ctrl mapping. These produce incorrect terminal input in concrete mobile workflows.

## Review comments

### [P2] Bypass modifiers for one-character pastes

`/Users/hzhou/rustdesk/flutter/lib/models/terminal_model.dart:111`

When bracketed-paste mode is disabled, as in plain shells or applications that turn it off, xterm's `paste()` emits the clipboard text directly. Pasting a single `d` therefore enters this branch and, with Ctrl locked, sends `0x04`/EOF instead of the pasted character. Input length cannot distinguish typing from pasting, so paste events need to bypass modifier mapping explicitly.

### [P2] Clear hidden modifiers when collapsing Row3

`/Users/hzhou/rustdesk/flutter/lib/mobile/pages/terminal_page.dart:432`

If Ctrl or Alt is enabled and the user collapses Row3, this only hides the modifier buttons while leaving the lock active. The next character is then modified without any visible indication—for example, a hidden Ctrl lock can turn `d` into EOF and close the shell. Clear active locks when collapsing, or retain a visible indicator/control.

### [P2] Map Ctrl+underscore to the unit-separator code

`/Users/hzhou/rustdesk/flutter/lib/models/terminal_model.dart:78-81`

When Ctrl lock is active and the user enters `_`, this falls through and sends a literal underscore before unlocking. `Ctrl+_` is the standard `0x1f` control sequence and is used for readline undo; the mapping already handles the preceding `[`, `\`, `]`, and `^` range but omits its `_` endpoint.

rustdesk:
# Codex Review

The shortcut predicate both intercepts non-paste key combinations and allows repeated pastes to bypass the source-aware path. Modifier recognition also mishandles valid non-BMP single-character input, producing user-visible terminal input corruption.

## Findings

### [P2] Reject extra modifiers in paste shortcuts

`flutter/lib/models/input_modifier_utils.dart:127-129`

When Ctrl+Alt+V arrives—for example, AltGr+V on layouts that expose AltGr as Ctrl+Alt—this predicate returns true because it never checks Alt or Shift. The event is then handled as paste, whereas xterm's exact Ctrl+V shortcut requires the other modifiers to be released, so users trying to enter an AltGr character can unexpectedly paste their clipboard into the remote shell.

### [P2] Consume paste repeats through the safe path

`flutter/lib/models/input_modifier_utils.dart:127-129`

When Ctrl/Cmd+V is held, Flutter emits `KeyRepeatEvent`, which fails `isKeyDown` and falls through to xterm's default `SingleActivator`; that activator includes repeats and invokes `Terminal.paste()` instead of `pasteText()`. With bracketed paste disabled, a repeated one-character paste is therefore processed as keyboard input again—for example, a locked Ctrl plus clipboard `d` sends EOT—so repeat events must either use the source-aware path or be consumed.

### [P2] Count Unicode characters instead of UTF-16 units

`flutter/lib/models/input_modifier_utils.dart:77-80`

For a supplementary-plane character such as `😀`, Dart reports `data.length == 2`, so this returns false even though the user entered one character. The character is sent unmodified and the one-shot Ctrl/Alt state remains active, causing the following ASCII character to be modified unexpectedly; detect a single Unicode scalar or grapheme rather than a single UTF-16 code unit.

dongrencd:
Addressed the latest Codex Review in `405d9ce5e`.

Changes:
- Paste shortcuts now only handle exact Ctrl+V or exact Meta+V. Extra modifiers such as Ctrl+Alt+V, Ctrl+Shift+V, Meta+Shift+V, and Ctrl+Meta+V are ignored so they can continue through the normal key path.
- Ctrl/Cmd+V key-repeat events are now consumed through the source-aware paste path, preventing repeated pastes from falling back to xterm's default `Terminal.paste()` handling.
- Single-character modifier detection now counts user-visible characters via Flutter's `characters` export instead of UTF-16 code units, so supplementary-plane input such as emoji consumes the one-shot Ctrl/Alt state correctly.
- Added regression coverage for exact paste shortcuts, repeat paste shortcuts, extra modifier rejection, repeat+extra modifier rejection, non-V key rejection, and supplementary-plane character input.

Validation:
- `flutter test test/input_modifier_utils_test.dart test/terminal_keyboard_utils_test.dart` passed: 34/34.
- `flutter analyze --no-fatal-warnings --no-fatal-infos lib/models/input_modifier_utils.dart lib/mobile/pages/terminal_page.dart lib/models/terminal_model.dart` exits 0. It still reports the existing deprecation infos in `terminal_page.dart`.
- `git diff --check` passed.
- Confirmed the diff does not include Android keep-alive, AndroidManifest, wake lock, or foreground-service changes.

Actions:
- Started fork Flutter Nightly Build for this head: https://github.com/dongrencd/rustdesk/actions/runs/29319747453

rustdesk:
• OS/IME paste remains misclassified as keyboard input, and Ctrl mapping corrupts decomposed Unicode graphemes. Both produce incorrect terminal input in supported mobile workflows.

  Full review comments:

  - [P2] Treat IME clipboard commits as paste input — /Users/hzhou/rustdesk/flutter/lib/models/terminal_model.dart:84-84
    When a mobile user pastes from the OS keyboard/IME clipboard UI, such as Gboard's clipboard panel, xterm's CustomTextEdit sends the committed text through Terminal.onOutput, so it
    reaches this branch without _handleTerminalKeyEvent or _pasteClipboardText and is always labeled as keyboard input. With Ctrl locked, a one-character clipboard value such as d is
    therefore sent as 0x04 and consumes the lock, while a lone pasted LF is normalized to CR; the source-aware paste handling still needs to cover IME paste commits.

  - [P2] Keep decomposed graphemes intact under Ctrl — /Users/hzhou/rustdesk/flutter/lib/models/input_modifier_utils.dart:148-153
    For keyboard input normalized as a decomposed grapheme such as é, characters.length accepts the payload as one visible character, but this UTF-16 loop transforms the ASCII e to
    0x05 and appends the combining acute mark. Ctrl+é therefore sends a control byte plus a combining mark while the equivalent precomposed é is left intact; only transform when the
    entire grapheme is exactly one supported ASCII scalar.

dongrencd:
Addressed the decomposed-grapheme part in `7defc03da`.

Changes:
- Ctrl mapping now only transforms when the whole input is one grapheme and exactly one supported ASCII scalar.
- Decomposed graphemes such as `e\u0301`, precomposed non-ASCII characters, and supplementary-plane characters are preserved under Ctrl instead of rewriting only an ASCII base scalar.
- Existing ASCII Ctrl mappings are unchanged, including `_` and `/` -> `0x1f`.
- Added regression coverage for decomposed graphemes and non-ASCII graphemes in both `applyTerminalInputModifiers()` and `prepareTerminalInputPayload()`.

About the IME clipboard commit case:
- I verified the issue path against xterm 4.0.0. `CustomTextEdit.updateEditingValue()` computes a plain text delta and calls `onInsert(textDelta)`, then `TerminalView._onInsert()` falls back to `terminal.textInput(text)`, which reaches RustDesk as the same `Terminal.onOutput('d')` payload as normal text input.
- For a one-character commit, RustDesk's `TerminalModel.onOutput` cannot reliably distinguish typed `d` from Gboard clipboard committed `d` with the current xterm API.
- I did not add a clipboard-content heuristic in `TerminalModel`, because that would misclassify normal typing whenever the current clipboard happens to contain the same character, breaking the intended Ctrl one-shot workflow.
- This part needs an explicit input-source signal from the xterm text input layer, or an upstream/vendor xterm API change, before RustDesk can route IME clipboard commits through the source-aware paste path safely.

Validation:
- `flutter test test/input_modifier_utils_test.dart test/terminal_keyboard_utils_test.dart` passed: 37/37.
- `flutter analyze --no-fatal-warnings --no-fatal-infos lib/models/input_modifier_utils.dart lib/mobile/pages/terminal_page.dart lib/models/terminal_model.dart` exits 0; it still reports the existing deprecation infos in `terminal_page.dart`.
- `git diff --check` passed.
- Confirmed the diff only touches `flutter/lib/models/input_modifier_utils.dart` and `flutter/test/input_modifier_utils_test.dart`; no Android keep-alive, AndroidManifest, wake lock, or foreground-service changes are included.

Triggered fork Flutter Nightly Build for this head:
https://github.com/dongrencd/rustdesk/actions/runs/29384705579

rustdesk:
• IME-originated one-character pastes can still be corrupted by active modifiers, and the shared input path adds avoidable UI-thread work for large desktop pastes.

  Full review comments:

  - [P2] Propagate paste origin for IME clipboard commits — /Users/hzhou/rustdesk/flutter/lib/models/terminal_model.dart:84-84
    When a mobile user pastes one character from an IME clipboard panel such as Gboard, xterm forwards the commit through Terminal.onOutput, so this path always labels it as keyboard
    input instead of reaching pasteText. With Ctrl locked, pasting d is therefore transformed into 0x04 and consumes the lock; the text-input layer needs to propagate an explicit
    paste source for these commits.

  - [P2] Check modifier locks before scanning graphemes — /Users/hzhou/rustdesk/flutter/lib/models/terminal_model.dart:80-81
    When a desktop user pastes a large payload, both modifier callbacks are absent, but this evaluates shouldApplyTerminalInputModifiers before checking the lock booleans, and
    prepareTerminalInputPayload repeats the same ordering. Because Characters.length walks the entire string, a multi-megabyte paste incurs two synchronous Unicode-segmentation passes
    on the UI isolate even though modifiers cannot apply; check the lock booleans first.

dongrencd:
Addressed the large-paste modifier-scan issue in `cfa68231c`.

Changes:
- `TerminalModel._handleInput()` now checks whether Ctrl/Alt locks are active before calling `shouldApplyTerminalInputModifiers()`, so inactive desktop paste/input payloads do not pay the grapheme-scan cost just to decide that modifiers cannot apply.
- `prepareTerminalInputPayload()` uses the same ordering: it checks `ctrlLocked || altLocked` before scanning graphemes.
- Added regression coverage for a large keyboard payload with inactive modifiers to keep the no-modifier behavior unchanged.

IME clipboard source investigation:
- Rechecked xterm 4.0.0 and Flutter's text input APIs. `CustomTextEdit.updateEditingValue()` computes a plain text delta and calls `onInsert(textDelta)`; `TerminalView._onInsert()` then falls back to `terminal.textInput(text)`. The callback has no source/origin parameter.
- Flutter `TextInputClient` exposes `performPrivateCommand`, `commitContent`, and delta updates, but the normal IME `commitText` path used here does not carry a paste cause/source. xterm's `performPrivateCommand` is currently a no-op and is not wired to `onInsert`.
- I also checked the RustDesk Android side and did not find an existing terminal IME paste-origin hook to reuse.
- Because a typed `d` and a one-character Gboard clipboard commit both arrive at RustDesk as the same `Terminal.onOutput('d')` payload, I still avoided adding a clipboard-content heuristic in `TerminalModel`; it would misclassify ordinary typing when the clipboard happens to contain the same character. A complete fix for this part needs a reliable source signal from the text input layer or an upstream/xterm API change that can actually receive such a signal.

Validation:
- `flutter test test/input_modifier_utils_test.dart test/terminal_keyboard_utils_test.dart` passed: 38/38.
- `flutter analyze --no-fatal-warnings --no-fatal-infos lib/models/input_modifier_utils.dart lib/mobile/pages/terminal_page.dart lib/models/terminal_model.dart` exits 0; it still reports the existing deprecation infos in `terminal_page.dart`.
- `git diff --check` passed.
- Confirmed this commit only touches terminal input/modifier Dart files and tests; no xterm fork/vendor, Android keep-alive, AndroidManifest, wake lock, or foreground-service changes are included.

Triggered fork Flutter Nightly Build for this head:
https://github.com/dongrencd/rustdesk/actions/runs/29498901980

rustdesk:
# Review findings

Explicit shortcut and context-menu paste paths bypass keyboard transformations, but OS/IME clipboard commits still traverse keyboard preprocessing and can corrupt pasted data.

## [P2] Propagate IME paste origin before applying modifiers

**File:** `/Users/hzhou/rustdesk/flutter/lib/models/terminal_model.dart:85-85`

When a mobile user pastes through an OS keyboard clipboard panel such as Gboard, xterm delivers the committed text through `Terminal.onOutput` rather than `_pasteClipboardText`, so this hard-coded source misclassifies it as keyboard input. With Ctrl or Alt locked, a one-character paste is transformed and consumes the lock (for example, `d` becomes `0x04`), while a pasted lone LF becomes CR; the text-input layer needs to propagate paste origin before preprocessing.

dongrencd:
I rechecked this against the current xterm 4.0.0 and Flutter text-input path.

Source-path evidence:
- Flutter dispatches normal text commits through `TextInputClient.updateEditingState` / `updateEditingStateWithDeltas`; `TextEditingDeltaInsertion` only carries inserted text, insertion offset, selection, and composing state. It does not carry a paste cause/source.
- xterm's `CustomTextEdit.updateEditingValue()` reduces the committed editing value to `textDelta` and calls `onInsert(textDelta)`.
- `TerminalView._onInsert()` then calls `terminal.textInput(text)` when the text is not consumed as a terminal key.
- `Terminal.textInput()` emits `onOutput(text)`, so RustDesk receives the same `Terminal.onOutput('d')` payload for typed `d` and a one-character IME clipboard commit.
- xterm's `performPrivateCommand()` exists but is currently a no-op and is not wired into `onInsert`; `commitContent` is a separate rich-content path, not the normal plain-text commit path used here.

Because of that, there is no reliable paste-origin signal available at `TerminalModel._handleInput()` today. I did not add a clipboard-content or timing heuristic, because it would misclassify ordinary typing when the clipboard happens to contain the same character and would break the intended Ctrl/Alt one-shot behavior.

I also could not capture a Gboard runtime trace in this environment because no Android device/emulator is available here, but the source path above shows where the source information is lost.

A complete fix needs a reliable source signal from the text input layer before RustDesk can route these commits to `pasteText()`. If there is a specific Flutter/xterm event you expect for Gboard clipboard commits, I can wire that into the source-aware paste path; otherwise this cannot be safely fixed in `TerminalModel` alone.

rustdesk:
@fufesou

fufesou:
Hi @dongrencd, thanks for the PR.


> Hardware `Ctrl+V` and `Cmd+V` use the same paste path as the terminal context-menu paste action.

Sorry, I can't understand it clearly.

I only see this function being called through the chain: `shouldHandleTerminalPasteShortcut()` -> `_handleTerminalKeyEvent()` -> `TerminalView::onKeyEvent`. This function only works when a physical keyboard is plugged in.

Are you referring to a scenario where `d` is in the pasteboard, `Ctrl` is pressed on the mobile screen, and `Ctrl+v` is pressed on the physical keyboard?

I've commented out `flutter/lib/mobile/pages/terminal_page.dart:241` (`//onKeyEvent: _handleTerminalKeyEvent`), and it works fine — `d` is pasted correctly.

In the master branch, `Ctrl+v` works on Android and `Cmd+v` works on iOS, so maybe we should keep the default behavior. Right‑click also works fine.

Another more common issue: if `d` is in the pasteboard and I click the paste button in the IME soft keyboard prompt, then `Ctrl+d` is executed on the remote side.

dongrencd:
I checked this again. The wording in my previous comment was inaccurate.

`shouldHandleTerminalPasteShortcut()` only handles the hardware-keyboard `Ctrl+V` / `Cmd+V` path through `TerminalView.onKeyEvent`. It is not the same path as the terminal context-menu/right-click paste action.

Based on your test and the existing master behavior, I agree we should keep the default hardware paste behavior. I will remove the `onKeyEvent` paste-shortcut interception from this PR and keep the explicit RustDesk paste path only for context-menu/right-click paste.

The IME soft-keyboard paste prompt case you mentioned is a separate issue. I can reproduce the source-path problem in code: that input reaches RustDesk as normal terminal text through `Terminal.onOutput('d')`, so `TerminalModel._handleInput()` cannot distinguish typed `d` from an IME clipboard commit. Without a reliable paste-origin signal from Flutter/xterm, handling it in `TerminalModel` would require heuristics such as comparing clipboard contents, which can misclassify normal typing.

So the next update will narrow this PR by removing the hardware paste shortcut interception. The IME paste prompt issue should stay documented as a current limitation unless we can get a reliable source signal from the text input layer.

fufesou:
> The IME soft-keyboard paste prompt case you mentioned is a separate issue. I can reproduce the source-path problem in code: that input reaches RustDesk as normal terminal text through `Terminal.onOutput('d')`, so `TerminalModel._handleInput()` cannot distinguish typed `d` from an IME clipboard commit. Without a reliable paste-origin signal from Flutter/xterm, handling it in `TerminalModel` would require heuristics such as comparing clipboard contents, which can misclassify normal typing.

Maybe we can ignore this rare case.

Just adding some comments in the source code is also acceptable.

fufesou:
## Testing

Both Android and iPad

### `Ctrl`, `Alt` combinations

```bash
python3 -c 'import os,select,termios,tty; fd=0; old=termios.tcgetattr(fd); tty.setraw(fd);data=bytearray(os.read(fd,1)); exec("while select.select([fd],[],[],0.2)[0]:\n data.extend(os.read(fd,1024))");termios.tcsetattr(fd,termios.TCSADRAIN,old); print("received:",data.hex(" "))'
```

| key | output | name |
| :---: | :----: | :----: |
| Ctrl+Space      |   00  | NUL |
| Ctrl+A          |   01  | SOH |
| Ctrl+B          |   02  | STX |
| Ctrl+C          |   03  | ETX |
| Ctrl+J          |   0a  | LF  |
| Ctrl+M          |   0d  | CR  |
| Ctrl+Z          |   1a  | SUB |
| Ctrl+[          |   1b  | ESC |
| Ctrl+\          |   1c  | FS  |
| Ctrl+]          |   1d  | GS  |
| Ctrl+^          |   1e  | RS  |
| Ctrl+_          |   1f  | US  |
| Ctrl+/          |   1f  | US  |
| Alt+/           | 1b 2f | |
| Ctrl+Alt+/      | 1b 1f | |
| Ctrl+Alt+[      | 1b 1b | |
| Ctrl+Alt+Space  | 1b 00 | |
| Alt+A           | 1b 61 | |
| Alt+B           | 1b 62 | |
| Alt+F           | 1b 66 | |
| Alt+D           | 1b 64 | |
| Ctrl+Alt+B      | 1b 02 | |

###  xterm bracketed-paste mode

```bash
python3 -c 'import os,select,sys,termios,tty; fd=0; old=termios.tcgetattr(fd); sys.stdout.write("\x1b[?2004h");sys.stdout.flush(); tty.setraw(fd); data=bytearray(os.read(fd,1)); exec("while select.select([fd],[],[],0.2)[0]:data.extend(os.read(fd,4096))"); termios.tcsetattr(fd,termios.TCSADRAIN,old); sys.stdout.write("\x1b[?2004l");sys.stdout.flush(); print("received:",data.hex(" "))'
```

paste `d`

- ❌ IME soft keyboard paste -- `64` . It seems hard to 
>  so `TerminalModel._handleInput()` cannot distinguish typed `d` from an IME clipboard commit.  https://github.com/rustdesk/rustdesk/pull/15532#issuecomment-5010999844
- ✅ `Ctrl`/`Cmd`+`V` -- `1b 5b 32 30 30 7e 64 1b 5b 32 30 31 7e`

Explicit clipboard paste paths handled by TerminalPage bypass modifiers and honor xterm bracketed-paste mode.

dongrencd:
Verified from your test results.

The Ctrl/Alt modifier outputs look correct on both Android and iPad, and `Ctrl`/`Cmd`+`V` already works through the default xterm path, including bracketed-paste mode.

I will keep the IME soft-keyboard paste prompt as a documented limitation instead of adding clipboard-content heuristics. I will also remove the hardware `Ctrl+V` / `Cmd+V` interception from this PR, so the default xterm behavior is preserved for that path.

The explicit RustDesk paste path will remain only for context-menu/right-click paste, where TerminalPage can reliably identify the action and route it through `pasteText()`.

dongrencd:
Updated the branch to narrow the paste handling.

The hardware-keyboard `Ctrl+V` / `Cmd+V` interception has been removed, so this PR keeps the default xterm behavior for that path. The explicit RustDesk paste path remains for context-menu/right-click paste, where TerminalPage can reliably identify the action and route it through `pasteText()`.

I also documented the IME soft-keyboard paste prompt limitation in the source code. Since Flutter/xterm exposes that case to RustDesk as normal text input without paste-origin metadata, this PR does not add clipboard-content heuristics.

rustdesk:
# Review

The modifier pipeline misclassifies one-character hardware pastes and can invoke widget-state callbacks after disposal. Both issues can produce unintended terminal input or runtime errors in supported input paths.

## [P2] Preserve one-character hardware paste data

**Location:** `/Users/hzhou/rustdesk/flutter/lib/models/terminal_model.dart:88`

When a virtual Ctrl/Alt lock is active, bracketed-paste mode is off, and xterm's built-in hardware Ctrl/Cmd+V pastes one printable character, `Terminal.paste()` emits that character through `Terminal.onOutput`. Hard-coding this source as `keyboard` applies and consumes the pending modifier—for example, pasting `d` with Ctrl locked sends `\x04` instead of `d`. The hardware paste path therefore also needs to be identified as paste input.

## [P2] Stop modifier callbacks after disposal

**Location:** `/Users/hzhou/rustdesk/flutter/lib/models/terminal_model.dart:94-96`

When an asynchronous xterm action completes after `TerminalPage` is disposed—for example, built-in Ctrl/Cmd+V finishes reading the clipboard—the terminal can still invoke `_handleInput`. If a modifier was locked and the result is one printable character, these callbacks call the page's `setState` after disposal; the `_disposed` check occurs only later in `_sendInputPayload`. Guard `_handleInput` before invoking state callbacks or detach the callbacks during disposal.

dongrencd:
Updated the branch for both P2 findings.

The hardware `Ctrl`/`Cmd`+`V` interception is now limited to the virtual modifier-lock case. When no virtual Ctrl/Alt lock is active, the event is ignored and xterm keeps its default paste behavior. When a virtual lock is active, the shortcut is routed through `pasteText()` so a one-character paste is preserved even when bracketed-paste mode is off.

I also moved the disposed guard to the start of `TerminalModel._handleInput()` and detach the page callbacks during model disposal, so late xterm input cannot invoke modifier callbacks after `TerminalPage` has been disposed.

fufesou:
Verified https://github.com/rustdesk/rustdesk/pull/15532#issuecomment-5042323037  b2f7d35

1. Establish a terminal session
2. Connect a physical keyboard
3. Execute the python command

```bash
python3 -c 'import os,select,sys,termios,tty;fd=0;old=termios.tcgetattr(fd);sys.stdout.write("\x1b[?2004lREADY\r\n");sys.stdout.flush();tty.setraw(fd);data=bytearray(os.read(fd,1));exec("while select.select([fd],[],[],0.2)[0]:data.extend(os.read(fd,4096))");termios.tcsetattr(fd,termios.TCSADRAIN,old);print("\r\nreceived:",data.hex(" "))'
```
4. Copy `d`
5. Lock `Ctrl`
6. `Ctrl`/`Cmd` + `v`

flutter/lib/models/input_modifier_utils.dart
```diff
@@ -36,3 +44,147 @@ bool shouldReleaseStaleMobileShift({
...
+  required bool metaPressed,
+  required bool altPressed,
+  required bool shiftPressed,
+  required bool modifierLockActive,
+}) {
+  if (!modifierLockActive) return false;
+  if (!isKeyDown && !isKeyRepeat) return false;
+  if (logicalKey != LogicalKeyboardKey.keyV) return false;
+  if (altPressed || shiftPressed) return false;
+  return controlPressed != metaPressed;
```
fufesou, line 137: It's better to test`isAndroid && controlPressed || isIOS && metaPressed`, but it's not a big deal here.

fufesou (approved):

PR #15535, opened 2026-07-08
fix: check valid id

fufesou:
LAN discovery accepts peer IDs from unauthenticated peers.

These IDs are later written into a temporary VBScript file when creating a Windows desktop shortcut. If the IDs contain malicious commands, they could be executed unexpectedly. 

To prevent this, reject IDs with unsafe characters before storage, and validate again before shortcut creation.

## Testing

See `untrusted_peer_id_validation()`.

```bash
cargo test -p rustdesk untrusted_peer_id_validation -- --nocapture
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for peer identifiers to reject empty values, overly long IDs, and unsafe characters.
  * Improved handling of LAN discovery responses so invalid peer data is ignored instead of processed.
  * Prevented shortcut creation when a peer ID is not valid, reducing the risk of malformed input being used.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -2623,6 +2623,20 @@ pub fn is_direct_ip_access(peer: &str) -> bool {
     hbb_common::is_ip_str(peer) || hbb_common::is_domain_port_str(peer)
 }
 
+// Align the maximum length of the peer id to the maximum length of the peer id in the server.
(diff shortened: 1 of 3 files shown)


fufesou:
@codex review

fufesou:
@coderabbitai review

PR #15549, opened 2026-07-09
feat: option to only send mouse input to the focused session window (regression risk)

loremipsumpy:
### Problem

With several remote sessions open as separate windows, pointer events are
delivered to whichever window is under the cursor, regardless of OS focus. That
is standard event routing on every desktop platform. Moving the local mouse
across tiled session windows therefore moves the remote cursor, triggers hover
UI, and wheel-scrolls in every session it passes over. Fine for a single
session window, disruptive when many are tiled.

### Change

Adds an opt-in local option, "Only control the remote when its window is
focused" (Settings → General → Other, **default off**, shown on Windows only).
When enabled, a session window drops mouse input while it is not the focused
OS window.

- Gated paths: hover / move / press / wheel, trackpad pan-zoom start and
  update, momentum fling, the pointer-exit edge move. Button-up and
  pan-zoom-end are never gated, so a press or pan started while focused always
  completes on the peer.
- **Windows only.** On Linux (X11, input source 1) RustDesk's own keyboard
  grab makes X emit FocusOut(NotifyGrab)/FocusIn(NotifyUngrab) whenever the
  pointer enters/leaves a session view, so GTK-level focus events do not track
  the active window and the gate would block input in a focused window (this
  is the Ubuntu regression reported in review; the plugin is not at fault). macOS already ignores mouse input to unfocused windows. Linux
  support can follow separately once focus can be derived from
  `_NET_ACTIVE_WINDOW` instead.
- Focus tracking is fail-open: focus events are change-only, so until a
  window has seen its first focus/blur event (fresh window, or after
  restore/un-minimize, where Windows can deliver a trailing blur without a
  focus event) the gate does not block. Blocking on unknown state means dead
  input; failing open means the feature engages after the first real event.
  In practice Windows delivers that first event at window creation and right
  after restore, so the unknown state lasts only an instant.
- After refocus, the first delivered click/wheel is preceded by one cursor
  sync move, so the remote cursor position is current again before the event
  applies (fixes the click-becomes-drag regression and wheel scrolling at a
  stale position). No sync is sent in relative mouse mode, where only deltas
  are meaningful.
- `refreshMousePos` (layout/view-style changes, toolbar dismiss) is gated so
  an unfocused window emits no move without pointer activity.
- The check lives inline in `InputModel` and reads the option only on
  Windows while unfocused.

### Scope / notes (deliberate limitations)

- Purely client-side, no protocol change. Default off.
- If focus is lost mid-drag, the (ungated) release drops at the last position
  the peer saw before the blur. Sending positions from unfocused windows to
  fix this would contradict the option's purpose.
- "Show displays as individual windows": sweeping into the same peer's
  unfocused sibling window is gated like any other window. If a same-peer
  exemption is preferred, that is a small follow-up, opinions welcome.
- Touch/virtual-mouse input paths are not gated (this option covers
  mouse/trackpad, as keyboard stays out of scope too).
- Not related to the report in #13398 (input reaching the remote while the
  pointer is *outside* the window, macOS): that would be a bug and is not
  addressed here.

### Testing

- `flutter analyze` / `dart format`: clean on changed lines; `flutter test`
  passes.
- Linux: option has no effect (gate compiles to false); verified on Ubuntu
  22.04, 24.04 and 26.04: motion still reaches the peer from an unfocused
  window with the option forced on, and the debug build starts cleanly (the
  previous focus seed crashed Linux debug builds at startup via the
  window-frame shadow's reactive read; removed). Branch also built from
  source on 22.04 and run against stock 1.4.9 peers 22.04<->24.04 in both
  directions.
- Gate mechanism (gate condition force-enabled on the Linux build to make it
  testable there): blurred window motion is dropped; after refocus the first
  click/wheel lands at the current pointer position (verified against VM
  peers with cursor-position probes, consistent across GTK 3.24.33/3.24.41).
- Windows (real machine, option on, two sessions tiled): hover/wheel over
  the unfocused session stays local while the focused one keeps working;
  sweeping across the unfocused window and clicking lands as a single clean
  click, no drag selection; wheel right after refocusing scrolls at the
  current pointer position; minimize → restore and un-minimize-to-maximized
  leave the mouse working immediately (focus state is invalidated instead of
  trusting the restore event, which fires without activation on restore-all);
  the toggle applies live and option off is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added a desktop setting: **“Only control the remote when its window is focused.”** (Windows-only)
  * When enabled, remote mouse control is forwarded only while the remote window is focused.
* **Bug Fixes**
  * Improved desktop focus tracking so focus state stays correct across focus/blur, restore, and maximize/un-minimize scenarios.
  * Prevents unfocused pointer input from affecting control, with cursor resynchronization when focus returns.
* **Documentation**
  * Added the new setting label to translations (some locales are currently blank).


flutter/lib/models/input_model.dart
```diff
@@ -490,6 +491,19 @@ class InputModel {
...
 
+  // Opt-in (default off, desktop only): drop mouse input while this session
+  // window is NOT the focused OS window, so a mouse passing over an unfocused
+  // multi-session window won't move/click that remote. Button-up is
+  // intentionally not gated (see onPointUpImage) so a press can always be
+  // released. The option is read lazily, so the read is skipped while the
+  // window is focused (the common case).
+  bool get _blockMouseWhenUnfocused => shouldBlockUnfocusedMouseInput(
+        isDesktop: isDesktop,
+        isWindowFocused: stateGlobal.isFocused.isTrue,
```
rustdesk: Where do you set it to false?

flutter/lib/desktop/pages/desktop_setting_page.dart
```diff
@@ -511,6 +511,12 @@ class _GeneralState extends State<_General> {
           kOptionOpenNewConnInTabs,
           isServer: false,
         ),
+        _OptionCheckBox(
+          context,
+          'Only control the remote when its window is focused',
```
rustdesk: Translations?

loremipsumpy: Done. Added the string to `template.rs` and ran `res/lang.py` to add the key to the other language files, plus the Spanish translation. The rest are left empty for translators.

flutter/lib/models/input_model.dart
loremipsumpy: It's set to `false` in `onWindowBlur` and `true` in `onWindowFocus` in `DesktopTab`, both already there. I only added the `= true` seed in `initState`, since those only fire on focus changes and a window that opens already focused wouldn't get an event, so it'd stay `false` and block input by mistake. Added a comment on that line pointing to the handlers. Can move the seed if you prefer.

Conrad-bit:
I've been needing this feature, thanks!!

flutter/lib/desktop/pages/desktop_setting_page.dart
rustdesk: Please change 'Only control the remote when its window is focused' to something like 'control_remote_when_focus'

rustdesk:
# Codex Review

Several mouse-producing paths still bypass the new focus gate, including the pointer-exit move and trackpad updates. The unconditional focus seed can also classify an initially inactive window as focused, allowing input despite the option.

## Findings

### P1: Gate the pointer-exit mouse move

**Location:** `/Users/hzhou/rustdesk/flutter/lib/models/input_model.dart:1599`

With the option enabled, merely crossing an unfocused session still sends one move: this guard suppresses `onPointMoveImage`, but `RemotePage.leaveView` unconditionally calls `tryMoveEdgeOnExit`, which invokes `handleMouse` and sends a mousemove to the nearest remote edge. Gate the exit path as well, otherwise the exact hover-crossing scenario still disturbs the remote.

### P1: Gate trackpad pan/zoom updates

**Location:** `/Users/hzhou/rustdesk/flutter/lib/models/input_model.dart:1676`

On platforms where two-finger scrolling arrives as `PointerPanZoomUpdateEvent` rather than `PointerScrollEvent`, this guard is never reached; `onPointerPanZoomUpdate` directly sends `trackpad` mouse deltas or touch-pan messages without checking focus. Consequently a trackpad over an unfocused session continues controlling it with the option enabled, so the pan/zoom path needs the same gate while preserving any required end event.

### P2: Initialize focus from the native window state

**Location:** `/Users/hzhou/rustdesk/flutter/lib/desktop/widgets/tabbar_widget.dart:345`

If a `DesktopTab` initializes while its native window is already inactive—for example when several session windows are created quickly or RustDesk starts in the background—this unconditional seed marks it focused. Since the callbacks report only subsequent changes, no blur is guaranteed to correct an already-unfocused window, so `_blockMouseWhenUnfocused` remains false and input is forwarded until a full focus/blur cycle; initialize from the native window's actual focus state instead.

### P2: Gate Linux side-button presses

**Location:** `/Users/hzhou/rustdesk/flutter/lib/models/input_model.dart:1531`

On Linux, back/forward mouse presses bypass `onPointDownImage`: `initSideButtonChannel` routes the hovered `_activeSideButtonModel` straight to `_sendMouseUnchecked`. GDK can deliver that event for the unfocused window under the cursor, so a side-button down is still sent with this option enabled; apply the focus gate to side-button downs while continuing to release any recorded down on up.

flutter/lib/models/input_model.dart
```diff
@@ -490,6 +491,20 @@ class InputModel {
   bool get useEdgeScroll =>
       parent.target!.canvasModel.scrollStyle == ScrollStyle.scrolledge;
 
+  // Opt-in (default off, desktop only): drop mouse input while this session
+  // window is NOT the focused OS window, so a mouse passing over an unfocused
+  // multi-session window won't move/click that remote. Button-up is
+  // intentionally not gated (see onPointUpImage) so a press can always be
+  // released. The option is read lazily, so the read is skipped while the
+  // window is focused (the common case).
+  bool get _blockMouseWhenUnfocused => shouldBlockUnfocusedMouseInput(
```
fufesou: This function is quite simple. Should we add a file for `shouldBlockUnfocusedMouseInput()`?

The comments are also too many. Maybe a link `https://github.com/rustdesk/rustdesk/discussions/13398` is enough here.

I may be wrong, but I feel that `shouldBlockUnfocusedMouseInput()` doesn't really need unit testing.

```dart
bool get _blockMouseWhenUnfocused => isDesktop && stateGlobal.isFocused.isTrue && mainGetLocalBoolOptionSync(kOptionControlFocusedWindowOnly);
```

fufesou:
The main issue (#13393) mentions both keyboard and mouse problems, while #13398 focuses mainly on the keyboard issue.

## The mouse issue

The mouse events are from the Flutter window. It also works fine on my side. The events are triggered only when the mouse is in the window.

Maybe there is a bug in listening to mouse events on the Flutter side.

We should first find and fix it.

## The keyboard issue (main issue)

7551c9ab-1aee-492a-ac63-b2dc9fe2049d

But both work fine on my side.

There might be a potential bug when using "Input source 1", so I think the correct approach is to find and fix the bug. 

### Input source 1

`enterView()` or `leaveView()` is triggered when entering or leaving the window

https://github.com/rustdesk/rustdesk/blob/865fe71c46e197f00272cb43c53ac076d241bfdd/flutter/lib/desktop/pages/remote_page.dart#L648

https://github.com/rustdesk/rustdesk/blob/865fe71c46e197f00272cb43c53ac076d241bfdd/flutter/lib/desktop/pages/remote_page.dart#L556

https://github.com/rustdesk/rustdesk/blob/865fe71c46e197f00272cb43c53ac076d241bfdd/flutter/lib/models/input_model.dart#L1159

So the state should change when entering/leaving the window. The keyboard events are then blocked/enabled.
Nothing about focus here.

### Input source 2

The flutter window will capture the key events.

When the window is focused, it always sends key events.

fufesou:
@loremipsumpy Hi, what are your thoughts on https://github.com/rustdesk/rustdesk/pull/15549#issuecomment-4945214892

I don't think this option is really necessary.

We should first identify and fix the underlying issue and use this option only if there is no better way.

flutter/lib/desktop/pages/desktop_setting_page.dart
loremipsumpy: Done in b813214. Used `control-remote-when-focus`, kebab-case to match the
recently added keys (`conn-e2ee-unavailable-tip`, `password-hidden-tip`). The
English text moved to `en.rs`.

flutter/lib/models/input_model.dart
loremipsumpy: Done in cb059d6. Inlined, and the helper file and unit tests are removed. Kept
the discussion link plus the line noting where `isFocused` is set, since that
was asked earlier in review. One detail about the snippet: it needs
`stateGlobal.isFocused.isFalse`, with `.isTrue` the gate would be inverted and
block while focused.

loremipsumpy:
Three of the four are fixed in 5443edc:

- P1 pointer-exit: `tryMoveEdgeOnExit` now respects the gate, so crossing an
  unfocused window no longer sends the edge mousemove.
- P1 trackpad: `onPointerPanZoomStart/Update` are gated. End events stay
  ungated, like button-up, so a pan started while focused always terminates on
  the peer.
- P2 Linux side buttons: back/forward downs respect the gate. Ups are still
  delivered for any recorded down, so no button is left pressed.

On the P2 focus seed, no code change after checking what is actually available:
`WindowController` (desktop_multi_window) has no `isFocused()`, and the plugin
forwards no engine lifecycle. Its `WM_ACTIVATE` only refocuses the child view,
and focus/blur events come from `WM_NCACTIVATE`. So there is no native state to
initialize from. A window created activated receives the `WM_NCACTIVATE` focus
event immediately, which makes the seed redundant in the normal case. The
mis-state needs a window created without activation, lasts until the first
focus/blur, and the option is off by default. The proper fix is adding
`isFocused()` to the rustdesk-org desktop_multi_window fork.

loremipsumpy:
@fufesou I reread #13393/#13398. The PR description mixed two different things,
so I've updated it.

### The bug in the report

#13398 describes input reaching the remote while the pointer is *outside* an
unfocused window (macOS). If that reproduces, it's a real bug and I agree it
should be fixed, not worked around. I can't reproduce it on Windows or Linux,
and I don't have a Mac to help chase it. This PR doesn't address it, so the
description no longer references #13393. That one can be handled on its own.

### What this PR is

This PR is only about the mouse cursor across session windows. Pointer events
inside an unfocused window follow the cursor. It isn't a bug, it works that way
by design, so it should be considered a feature selected by the user's
preference for their workflow. With several sessions tiled, crossing windows
moves, hovers and scrolls every remote on the way. Windows has the same kind of
setting for native windows ("Scroll inactive windows when I hover over them").
Anything beyond the cursor, like keyboard or the macOS report above, is out of
scope here and can go in separate PRs.

### The keyboard issue

Agreed on all points. This PR doesn't touch keyboard, and #13398 is the right
place for it.

With the description corrected, would you take this as a plain opt-in
preference (default off)?

rustdesk:
> With the description corrected, would you take this as a plain opt-in
preference (default off)?

YES

rustdesk:
# Codex Review

Normal pointer handlers preserve their behavior when the option evaluates false, but initial focus is assumed rather than observed and one mouse-move path bypasses the new gate. Consequently, input can still reach an unfocused remote in specific window-lifecycle and layout scenarios.

## Findings

### [P2] Derive initial focus from the actual window

`flutter/lib/desktop/widgets/tabbar_widget.dart:345`

When a session window is created without being granted focus, such as when the window manager prevents focus stealing, this marks it focused even though no subsequent blur event is guaranteed because it was never focused. With the option enabled, `_blockMouseWhenUnfocused` therefore continues forwarding input until the window is focused and blurred once; the assignment also changes focus-dependent styling when the option is off. Initialize this value from the actual OS focus state instead.

### [P2] Apply the focus gate to refreshMousePos

`flutter/lib/models/input_model.dart:1735-1736`

When the option is enabled and the window is unfocused, `refreshMousePos()` directly above this block remains ungated and calls `handleMouse`. `CanvasModel.updateViewStyle()` invokes it during layout, display, or view-style changes, so an unfocused session can still send a remote mouse-move event without any pointer activity. Apply the same focus check to this refresh path.

fufesou:
## Regression

### Ubuntu 24.04 (critical)

The mouse events can't be sent anymore.

`stateGlobal.isFocused` is buggy; it is always `false`.

`onWindowFocus` is never triggered when the window gains focus. It is triggered when the window has focus, and the mouse leaves the window.

`onWindowBlur` can be triggered correctly.

So when the mouse is in the window, and the window has focus, `stateGlobal.isFocused` is also `false`.
Only when the window has focus, and the mouse leaves the window, `stateGlobal.isFocused` is `true`. But no events will be triggered.

### Windows

There is a regression in which, when the option is checked, a click may trigger a "Down-Move-Up" sequence.

https://github.com/user-attachments/assets/939c6dd1-be6b-4ae4-8468-290ea583b3a4

2e0c7323-4202-4088-913b-0f87b9be0d8e


### Cause (**Windows** as the control/local side)

When clicking in the window:

`onPointDownImage()` and `onPointUpImage()` are triggered only if the Flutter window already has focus.
```
flutter: ====== onPointDownImage
flutter: ====== onPointUpImage
```

`onPointMoveImage()` is also triggered only if the Flutter window already does not have focus.
```
flutter: ====== onPointDownImage
flutter: ====================== onPointMoveImage
flutter: ====== onPointUpImage
```

The master branch does not have this issue, because the mouse is already moved from the edge to the current position.

This PR prevents the mouse move events. Then, after clicking, a selection area may be produced.


## Others

macOS already does not allow mouse events if not focused. This option is useless.

loremipsumpy:
Still working on this, reworking the branch to fix the reported regressions. Will push updates soon.

loremipsumpy:
@fufesou I could reproduce both, thanks for testing. Pushed the fixes and
updated the description.

Ubuntu: the plugin is fine. Those blur/focus events are real GTK events, RustDesk
causes them itself. With input source 1, entering the session view arms the
rdev XGrabKeyboard and X sends FocusOut(NotifyGrab) to the focused window; the
deferred release then sends FocusIn(NotifyUngrab). That's the always-false
isFocused and the focus-on-mouse-leave you saw. Meanwhile _NET_ACTIVE_WINDOW
never changes, and with input source 2 the events track activation fine. Focus
just can't be trusted on X11 while the grab is armed, so I scoped the option
to Windows: the gate is false elsewhere and the checkbox only shows there. If
we want Linux later, deriving focus from _NET_ACTIVE_WINDOW should work since
the grab doesn't touch it.

The click-becomes-drag on Windows was the dropped hovers leaving the remote
cursor stale, so the first click after refocus landed at the old position. I
now send one cursor sync before the first press or wheel delivered after
refocus (skipped in relative mouse mode). The init seed is gone too; it also
crashed Linux debug builds at startup. The gate fails open until the window
has seen a real focus/blur event, and restore/un-minimize reset to that
state; otherwise minimize then restore left the mouse dead.

On Windows with the option on: hover/scroll over an unfocused session stays
local, click after refocus is a clean click, minimize then restore still has
a working mouse. Option off, and on Linux/macOS, nothing changes.

loremipsumpy:
@coderabbitai review

fufesou:
https://github.com/rustdesk/rustdesk/pull/15549#issuecomment-4967740048

@loremipsumpy The issue on Windows persists.

PR #15558, opened 2026-07-10
feat: support simultaneous online on custom and builtin rendezvous servers

hashbk:
## Summary

Currently, when a user configures a `custom-rendezvous-server` via Settings, the client only registers on that custom server, discarding the hardcoded `RENDEZVOUS_SERVERS`. This PR enables the client to be online on **both** the user-configured server and the builtin public servers simultaneously, and accept incoming connections from either.

### `libs/hbb_common/src/config.rs`
- **`Config::get_rendezvous_servers()`**: When `custom-rendezvous-server` is set, now calls `merge_with_builtin_servers()` to return the custom server + hardcoded `RENDEZVOUS_SERVERS` (deduplicated), instead of returning only the custom server.

### `src/rendezvous_mediator.rs`
- **Per-host exit tracking**: Replaced the global `SHOULD_EXIT` cascade with `EXITED_HOSTS: Mutex<HashSet<String>>`. Only when all hosts have exited or a manual restart is triggered does `SHOULD_EXIT` get set, so one server disconnecting doesn't kill the other mediator.
- **Per-host PK mismatch tracking**: Changed `SOLVING_PK_MISMATCH` from `Mutex<String>` to `Mutex<HashSet<String>>`, allowing independent PK mismatch resolution per host without blocking other hosts from registering.
- Updated outdated comment in `start_tcp`.

## UI Impact

None. The `using_public_server()` check remains unchanged — it still returns `false` when a custom server is configured, preserving existing UI behavior.

## Related

- Depends on: https://github.com/rustdesk/hbb_common/pull/570

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when connecting to multiple rendezvous hosts concurrently.
  * Prevented premature shutdown while other configured rendezvous hosts remain active.
  * Improved handling of UUID mismatches across multiple hosts, including safer peer registration behavior.
  * Updated the bundled common library version.

### DIFF
--- a/src/rendezvous_mediator.rs
+++ b/src/rendezvous_mediator.rs
@@ -1,4 +1,5 @@
 use std::{
+    collections::HashSet,
     net::SocketAddr,
     sync::{
         atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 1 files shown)


rustdesk:
Please do not submit again, we won't merge it. NO PLAN TO SUPPORT THIS.

Issue #8106, opened 2024-05-21
Rustdesk service not starting on gnome shell login screen


NicKoehler:
### Bug Description

The service is not starting properly on boot using this specs:

Distro: Arch Linux
Kernel: Linux 6.9.1-arch1-1
DE: Gnome 46

this is what `systemctl status rustdesk` shows

```bash
● rustdesk.service - RustDesk
     Loaded: loaded (/etc/systemd/system/rustdesk.service; enabled; preset: disabled)
     Active: active (running) since Tue 2024-05-21 13:02:13 CEST; 23s ago
   Main PID: 675 (rustdesk)
      Tasks: 19 (limit: 38275)
     Memory: 43.0M (peak: 44.4M)
        CPU: 5.573s
     CGroup: /system.slice/rustdesk.service
             └─675 /usr/bin/rustdesk --service

mag 21 13:02:30 arch rustdesk[5031]: sudo: errore nell'inizializzare il plugin di audit sudoers_audit
mag 21 13:02:32 arch sudo[5882]:     root : unknown user listed. ; PWD=/ ; USER=listed. ;
```

It seems like the error comes from rustdesk parsing an user named "listed." instead of the actual username (maybe root?).

### How to Reproduce

1. install rustdesk
2. enable rustdesk service
3. reboot and don't login
4. try to connect from another client

### Expected Behavior

I should be able to connect

### Operating system(s) on local side and remote side

All -> Linux

### RustDesk Version(s) on local side and remote side

nightly -> nightly

fufesou:
Thanks for your feedback.

Can you please run `longinctl` and paste the output here?
Please hide private info.

NicKoehler:
From what I can tell seems like the error is from this function:

https://github.com/rustdesk/rustdesk/blob/676ee997091f091f68fbaa20de4ee7c376b210be/libs/hbb_common/src/platform/linux.rs#L145



> Thanks for your feedback.
> 
> Can you please run `longinctl` on the server side and paste the output here? Please hide private info.

```bash
SESSION  UID USER SEAT  TTY   STATE  IDLE SINCE    
      2  120 gdm  seat0 tty1  active yes  13min ago
      4 1000 nic  -     pts/0 active no   -        
      5 1000 nic  -     pts/1 active no   -       
```

rustdesk:
Is it currently wayland or x11?

NicKoehler:
Same issue with both, the server doesn't show up as online too.

[...]

PR #5935, opened 2023-10-07
Fix/xwayland server envs

fufesou:
Xwayland works on the controlled side.

https://github.com/rustdesk/rustdesk/discussions/5403

1. Check if Xwayland is running.
2. Add param `envs` in function `run_as_user`.
3. Remove `get_env_tries`.  Query all processes in each loop.

https://github.com/rustdesk/rustdesk/assets/13586388/4cc044b3-a94a-4657-8207-295627ef7c9e

PR #15561, opened 2026-07-10
fix(linux): use Wayland discovery for Wayland sessions with Xwayland

### LINKED ISSUE #15554: [Bug] Wayland session with Xwayland but no XAUTHORITY enters 240-pipeline retry loop
### Bug Description

On an active `systemd-logind` Wayland session with Xwayland present, `rustdesk --service` selects `get_display_xauth_xwayland()`. On compositors that launch Xwayland from inherited listen file descriptors and do not export `XAUTHORITY` (observed with Hyprland), this path can never satisfy its `DISPLAY && XAUTHORITY` success condition even though `DISPLAY`, `WAYLAND_DISPLAY`, and `DBUS_SESSION_BUS_ADDRESS` are all available.

Each failed refresh performs:

- 10 retry rounds
- 6 process patterns (`xdg-desktop-portal`, `Xwayland`, `ibus-daemon`, `goa-daemon`, `kded`, and the RustDesk tray)
- 4 environment variables per process

That is up to **240 calls to the legacy `get_env()` per refresh**. Every call starts a large `ps | grep | tail | awk | xargs cat | tr | grep | tail | sed` shell pipeline. The failed refresh takes several seconds, then the outer service loop repeats it indefinitely.

On the affected host, an idle 10-second sample consumed 9.31 CPU-seconds (93.1% of one core). The long-lived service/server/tray processes together used only about 0.6%; nearly all CPU time belonged to the short-lived discovery children. No remote client connection is needed to trigger the behavior.

This is not a request to remove the baseline `pgrep`/`loginctl` polling discussed in #11959. It is a narrower branch-selection failure: logind reports a Wayland session, but the mere presence of Xwayland routes `Desktop::refresh()` through an X11-authentication path that cannot succeed on this environment. #15520 reported the broader CPU/process-spawning symptom, but did not identify this trigger or validate the existing pure-Wayland branch as a correction. Unlike the baseline polling overhead, this failure path becomes an effectively continuous fork storm.

#### Relevant source paths on current `master`

- Legacy shell pipeline: [`get_env()`](https://github.com/rustdesk/rustdesk/blob/137298e05a10ebac2b12c933b7efebc7c217adec/src/platform/linux.rs#L1375-L1378)
- 10 × 6 × 4 retry loop and `DISPLAY && XAUTHORITY` gate: [`get_display_xauth_xwayland()`](https://github.com/rustdesk/rustdesk/blob/137298e05a10ebac2b12c933b7efebc7c217adec/src/platform/linux.rs#L1686-L1707)
- Wayland session routed into the Xwayland path whenever `pgrep -a Xwayland` succeeds: [`Desktop::refresh()`](https://github.com/rustdesk/rustdesk/blob/137298e05a10ebac2b12c933b7efebc7c217adec/src/platform/linux.rs#L1905-L1942)
- Existing pure-Wayland path already uses native `/proc` discovery and accepts `WAYLAND_DISPLAY && DBUS_SESSION_BUS_ADDRESS`: [`get_display_xauth_wayland()`](https://github.com/rustdesk/rustdesk/blob/137298e05a10ebac2b12c933b7efebc7c217adec/src/platform/linux.rs#L1655-L1684)

The relevant processes had these keys (values intentionally omitted):

```text
Xwayland:           DISPLAY, WAYLAND_DISPLAY, DBUS_SESSION_BUS_ADDRESS
xdg-desktop-portal: DISPLAY, WAYLAND_DISPLAY, DBUS_SESSION_BUS_ADDRESS
rustdesk --tray:    DISPLAY, WAYLAND_DISPLAY, DBUS_SESSION_BUS_ADDRESS
XAUTHORITY:         absent from all of the above
```

### How to Reproduce

1. Log in to a Wayland session with Xwayland enabled. Hyprland is a known reproducer.
2. Confirm that logind reports Wayland, Xwayland is running, and Xwayland has no `XAUTHORITY` key:

   ```bash
   session_id=$(loginctl list-sessions --no-legend | awk -v uid="$UID" '$2 == uid && $4 == "seat0" { print $1; exit }')
   loginctl show-session "$session_id" -p Type -p State -p Seat

   xwayland_pid=$(pgrep -u "$UID" -x Xwayland | head -n 1)
   pgrep -a Xwayland
   tr '\0' '\n' < "/proc/$xwayland_pid/environ" \
     | sed -n 's/=.*//p' \
     | grep -E '^(DISPLAY|XAUTHORITY|WAYLAND_DISPLAY|DBUS_SESSION_BUS_ADDRESS)$'
   ```

   Expected environment-key output for the affected case:

   ```text
   WAYLAND_DISPLAY
   DISPLAY
   DBUS_SESSION_BUS_ADDRESS
   ```

3. Start the system service and let startup settle:

   ```bash
   sudo systemctl restart rustdesk.service
   sleep 15
   ```

4. Measure the service cgroup for 10 seconds:

   ```bash
   unit=rustdesk.service
   cpu0=$(systemctl show "$unit" -p CPUUsageNSec --value)
   time0=$(date +%s%N)
   sleep 10
   cpu1=$(systemctl show "$unit" -p CPUUsageNSec --value)
   time1=$(date +%s%N)

   awk -v c0="$cpu0" -v c1="$cpu1" -v t0="$time0" -v t1="$time1" \
     'BEGIN { printf "RustDesk service CPU: %.1f%% of one core\n", 100 * (c1-c0) / (t1-t0) }'
   ```

5. Optionally confirm the exec loop with `strace` (values such as the numeric UID may appear locally, but need not be included in a report):

   ```bash
   service_pid=$(systemctl show rustdesk.service -p MainPID --value)
   sudo timeout 15 strace -f -e trace=execve -s 300 -p "$service_pid" 2>&1 \
     | grep -E 'pgrep -a Xwayland|ps -u [0-9]+ -f.*(DISPLAY|XAUTHORITY|WAYLAND_DISPLAY|DBUS_SESSION_BUS_ADDRESS)' \
     | head -n 30
   ```

The result is a continuous stream of the `get_env()` shell pipeline, especially searches for the absent `XAUTHORITY` key.

### Expected Behavior

- A logind Wayland session should use Wayland environment discovery even when Xwayland is available for compatibility applications.
- `XAUTHORITY` should be optional when `WAYLAND_DISPLAY` and the session D-Bus address are valid.
- The idle root service should not continually spawn large shell pipelines.
- Successfully discovered session state should be cached until the active session changes, or retries should use bounded backoff.
- Native X11 session discovery should continue to require whatever X11 authentication state it actually needs.

No controlling client is required -> Arch Linux (kernel 7.0.9), Hyprland 0.55.2, Xwayland 24.1.11, xdg-desktop-portal-hyprland 1.3.12

No controlling client is required -> RustDesk 1.4.8. The same code path and legacy loop are still present in 1.4.9 and current `master` at commit `137298e05a10ebac2b12c933b7efebc7c217adec`.

### Screenshots

Not applicable. The CPU measurement and text-only process trace above provide the reproduction evidence.

#### Proposed fix

The smallest candidate correction is to prioritize the known session protocol over the presence of a compatibility Xwayland process:

```rust
// Conceptual change in Desktop::refresh():
if self.is_wayland() {
    self.get_display_xauth_wayland();
} else if is_xwayland_running() {
    self.get_display_xauth_xwayland();
}
```

Apply the same ordering both when initially discovering the seat-0 session and when refreshing an already active session. `get_display_xauth_wayland()` already collects optional `DISPLAY`/`XAUTHORITY` values when available, while correctly using `WAYLAND_DISPLAY && DBUS_SESSION_BUS_ADDRESS` as the success condition. This behavior was validated by forcing RustDesk's existing pure-Wayland branch, rather than by testing a compiled source patch. The source change should therefore also be checked on GNOME and KDE Wayland sessions with Xwayland enabled.

A more conservative alternative would be to migrate `get_display_xauth_xwayland()` from four separate legacy `get_env()` shell calls per candidate to the native `get_envs()` `/proc` implementation, and use this success condition when `self.is_wayland()`:

```rust
has_wayland_display && has_session_bus
```

In that case `DISPLAY` and `XAUTHORITY` can remain optional compatibility values. Successful results should be cached until the logind session ID/state changes; failures should use backoff rather than immediately repeating the full scan.

#### Validation of the branch change

As a diagnostic workaround, only the root RustDesk supervisor was made to receive a negative result for the exact `pgrep -a Xwayland` probe. Xwayland remained enabled globally and the user-owned RustDesk server retained normal process detection.

That forced the existing pure-Wayland branch without changing any capture or input code:

- The environment-search pipelines disappeared completely.
- Idle service usage fell from 93.1% to 7.9% of one core.
- Wayland portal capture, hardware H.264 encoding, and uinput keyboard/mouse initialization continued to work.
- A direct remote desktop session completed successfully.

The remaining approximately 8% is the broader `loginctl`/`ps aux` polling already discussed in #11959; it is separate from the 240-pipeline failure documented here.

#### Suggested regression coverage

1. Wayland session + Xwayland running + no `XAUTHORITY`: select the Wayland discovery path and do not call legacy `get_env()`.
2. Wayland session with optional `DISPLAY`/`XAUTHORITY`: retain those values without making them the success condition.
3. Native X11 session: preserve existing X11 discovery/authentication behavior.
4. Failed discovery: verify bounded retry/backoff and no unbounded shell-process churn.

Related reports: #15520 and discussion #11959.

beltashazzar:
## Problem

`Desktop::refresh()` routes a Wayland session into `get_display_xauth_xwayland()` whenever an Xwayland process is running — which is the common case, since most Wayland desktops keep Xwayland around for X11 compatibility apps. That path uses the deprecated per-variable `get_env()` shell pipeline (`ps | grep | tail | awk | xargs cat | tr | grep | sed`) and only succeeds when **both** `DISPLAY` and `XAUTHORITY` are found.

This was analysed in detail in #15554 (credit to @dougvk, including the proposed fix this PR implements). Two failure modes result:

1. **Xwayland with no `XAUTHORITY` (e.g. Hyprland):** the `DISPLAY && XAUTHORITY` gate can never be satisfied, so every refresh burns the full 10×6×4 retry budget — up to **240 forked shell pipelines per refresh** — and the root service becomes a continuous fork storm. #15554 measured **93.1% of one core** while idle, no client connected.

2. **Xwayland with `XAUTHORITY` present (e.g. GNOME/mutter):** the gate succeeds early, but the successful path still runs several `get_env()` shell pipelines *per refresh* instead of a single native `/proc` scan, and `refresh()` re-runs continuously.

### Independent reproduction of case 2 (GNOME/mutter)

On a separate host from the #15554 report — **RustDesk 1.4.9 (Flutter), Ubuntu 25.04, kernel 6.14, GNOME on Wayland (mutter)** — with Xwayland running and exporting all four keys (`DISPLAY`, `WAYLAND_DISPLAY`, `XAUTHORITY`, `DBUS_SESSION_BUS_ADDRESS`):

- `rustdesk --service` (root) sustains **~23–25% of one core** while idle, no client connected.
- It spawns the `ps -u <uid> -f | grep -E 'xdg-desktop-portal' | … | cat /proc/__/environ | …` pipeline **~82 times per second** (≈ every 86 ms), interleaved with `loginctl show-session`, `pgrep -a Xwayland`, and `ps aux`.
- `systemctl show rustdesk -p CPUUsageNSec` confirms ~25% of one core averaged over hours of uptime.
- Upgrading 1.4.8 → 1.4.9 did not change it (measured 24.4% → 23.2%).

This confirms @dougvk's note that the fix "should also be checked on GNOME and KDE Wayland sessions with Xwayland enabled," and regression item #2 in that issue (retain optional `DISPLAY`/`XAUTHORITY` without making them the success condition).

## Fix

Prefer the session's actual display protocol over the mere presence of a compatibility Xwayland process, at **both** refresh sites (the active-session fast path and initial discovery). A Wayland session now always uses `get_display_xauth_wayland()`, which:

- scans `/proc` natively via `get_envs()` instead of the deprecated per-variable `get_env()` shell pipeline (the migration direction already documented on `get_env` / discussion #11959),
- uses `WAYLAND_DISPLAY && DBUS_SESSION_BUS_ADDRESS` as its success condition, and
- **still collects optional `DISPLAY`/`XAUTHORITY` when present**, so nothing is lost where they exist (GNOME).

X11 sessions are unaffected: they still reach `get_display_x11()` / `get_xauth_x11()`, and `get_display_xauth_xwayland()` / `is_xwayland_running()` remain in use for the non-Wayland branch.

```diff
-                if is_xwayland_running() && !self.is_login_wayland() {
+                if self.is_wayland() {
+                    self.get_display_xauth_wayland();
+                } else if is_xwayland_running() {
                     self.get_display_xauth_xwayland();
                     self.is_rustdesk_subprocess = false;
-                } else if self.is_wayland() {
-                    self.get_display_xauth_wayland();
                 }
```

Expected effect: case 1 (Hyprland) drops from ~93% to the baseline `loginctl`/`ps aux` polling level (~8% per #15554's branch-forcing validation); case 2 (GNOME) eliminates the per-refresh shell pipelines in favour of a single native `/proc` scan.

## Testing

- The change is a control-flow reorder of existing, unchanged function calls (`is_wayland()`, `get_display_xauth_wayland()`, `is_xwayland_running()`, `get_display_xauth_xwayland()`) — no signature or type changes. Delimiter balance and dead-code checks pass; both previously-used functions remain referenced. I was not able to run a full local build (vcpkg/toolchain), so I'm relying on CI to compile-verify.
- The equivalent runtime behaviour was validated in #15554 by forcing the existing pure-Wayland branch (idle service CPU 93.1% → 7.9%, with portal capture, HW H.264 encode, uinput init, and a full remote session all continuing to work).
- Happy to run an `strace -f -e trace=execve` capture or test a patched build on GNOME/Wayland if useful.

Fixes #15554.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved display and environment discovery for Wayland sessions.
  * Ensured Wayland-native handling is used even when Xwayland is also running.
  * Improved display authentication handling during non-active-seat refreshes.

### DIFF
--- a/src/platform/linux.rs
+++ b/src/platform/linux.rs
@@ -1904,12 +1904,18 @@ mod desktop {
 
         pub fn refresh(&mut self) {
             if !self.sid.is_empty() && is_active_and_seat0(&self.sid) {
-                // Xwayland display and xauth may not be available in a short time after login.
-                if is_xwayland_running() && !self.is_login_wayland() {
+                // Prefer the session's actual display protocol over the mere presence of
+                // Xwayland. A Wayland session commonly has Xwayland running for X11
+                // compatibility apps, but its display/xauth must still be discovered via the
+                // Wayland path. Routing it through the Xwayland path makes the legacy get_env()
+                // shell pipeline (deprecated, see get_envs) run every refresh, and where the
+                // Xwayland process exports no XAUTHORITY the DISPLAY && XAUTHORITY success
+                // condition can never be met, turning refresh() into a continuous fork storm.
+                if self.is_wayland() {
+                    self.get_display_xauth_wayland();
+                } else if is_xwayland_running() {
                     self.get_display_xauth_xwayland();
                     self.is_rustdesk_subprocess = false;
-                } else if self.is_wayland() {
-                    self.get_display_xauth_wayland();
                 }
                 return;
             }
@@ -1934,11 +1940,12 @@ mod desktop {
 
             self.get_home();
             if self.is_wayland() {
-                if is_xwayland_running() {
-                    self.get_display_xauth_xwayland();
-                } else {
-                    self.get_display_xauth_wayland();
-                }
+                // Always use Wayland discovery for a Wayland session, regardless of whether
+                // Xwayland is running for X11 compatibility apps. get_display_xauth_wayland()
+                // still collects optional DISPLAY/XAUTHORITY when present, and uses native
+                // /proc scanning (get_envs) instead of the deprecated per-variable get_env()
+                // shell pipeline.
+                self.get_display_xauth_wayland();
                 self.is_rustdesk_subprocess = false;
             } else {
                 self.get_display_x11();


rustdesk:
# Review Summary of Codex

This patch fixes the fork storm that occurs when `XAUTHORITY` is missing, but it also bypasses the protection for GDM login sessions and removes the fallback path introduced by #5935 for discovering controlled-side Xwayland environments. Compatibility logic for both scenarios must be retained before merging.

## Review Comments

### [P1] Preserve the exclusion for GDM Wayland login sessions

**File:** `/Users/hzhou/rustdesk/src/platform/linux.rs:1914-1915`

When seat0 is an active GDM Wayland login session, this change removes the existing `!self.is_login_wayland()` guard. The first refresh returns through the login-session branch below, but subsequent refreshes all enter `get_display_xauth_wayland()`. GDM login sessions typically do not have the user's `xdg-desktop-portal`, so every refresh scans `/proc` 10 times and waits approximately 2.7 seconds.

This special handling originated in [#8111](https://github.com/rustdesk/rustdesk/pull/8111) as a fix for the login-screen service behavior reported in [#8106](https://github.com/rustdesk/rustdesk/issues/8106). The current change would reintroduce continuous discovery overhead and block service refreshes while no user is logged in.

### [P2] Preserve the fallback that obtains DISPLAY and XAUTHORITY from Xwayland processes

**File:** `/Users/hzhou/rustdesk/src/platform/linux.rs:1947`

In a Wayland session, `get_display_xauth_wayland()` checks only `xdg-desktop-portal`. If the portal process contains `WAYLAND_DISPLAY` and D-Bus environment variables, but `DISPLAY` or `XAUTHORITY` exists only in the Xwayland process, the call succeeds while leaving `desktop.display` or `desktop.xauth` empty. As a result, `try_start_server_()` does not pass those variables to `--server`.

The original branch was introduced by [#5935](https://github.com/rustdesk/rustdesk/pull/5935) to support the "Xwayland controlled side" scenario described in [discussion #5403](https://github.com/rustdesk/rustdesk/discussions/5403). It explicitly searches Xwayland and several other processes for these environment variables. Therefore, the claim that "optional values are still collected" is true only when those values are also present in the portal process. After native Wayland discovery completes, the optional X11 environment must still be supplemented from Xwayland.

## Recommended Direction

There are safer and more targeted ways to fix the fork storm than the destructive control-flow change proposed in [#15561](https://github.com/rustdesk/rustdesk/pull/15561). The root cause is the repeated use of the legacy `get_env()` shell pipeline, combined with treating `DISPLAY && XAUTHORITY` as a mandatory success condition. It is not necessary to remove the GDM guard or the Xwayland compatibility path to solve that problem.

A compatible fix should:

1. Preserve the existing `!self.is_login_wayland()` guard so GDM login sessions do not enter user-session discovery.
2. Replace the repeated per-variable `get_env()` shell pipelines in `get_display_xauth_xwayland()` with native `/proc` scanning through `get_envs()`, ideally collecting all relevant variables in one scan per retry.
3. Use `WAYLAND_DISPLAY && DBUS_SESSION_BUS_ADDRESS` as the required success condition for a Wayland session, while treating `DISPLAY` and `XAUTHORITY` as optional values.
4. Supplement only missing `DISPLAY` and `XAUTHORITY` values from Xwayland or the existing fallback process list, without overwriting values already discovered from `xdg-desktop-portal`.
5. Cache successful discovery for the active session, or throttle retries for still-missing optional values, instead of repeating the full discovery loop on every refresh.

This addresses the fork storm at its source while preserving the login-screen behavior from #8111 and the controlled-side Xwayland compatibility introduced by #5935.

rustdesk:
I believe there are clever ways to fix `the fork storm` rather than doing in this destructive way.

> Never underestimate legacy code. One careless change can cause a regression.

dougvk:
One possible middle path: keep the protocol-first routing, but make Wayland discovery layered rather than dropping the Xwayland compatibility lookup:

- Get required `WAYLAND_DISPLAY` + `DBUS_SESSION_BUS_ADDRESS` from the portal via `get_envs()`.
- Fill only missing `DISPLAY`/`XAUTHORITY` from Xwayland/fallback processes, also via one native scan.
- Cache or throttle once the required Wayland values are found; optional X11 values should not keep retries alive.

One nuance on GDM: current master already calls `get_display_xauth_wayland()` for active GDM Wayland on subsequent refreshes, because `!self.is_login_wayland()` only guards the first branch and then falls through to `else if self.is_wayland()`. An explicit early return would fix that existing behavior.

This should preserve #5935 while fixing the fork storm at its source.

rustdesk:
Close since no response.

PR #15603, opened 2026-07-16
fix: android: target API 36

michaeljclarkk:
## Summary

- Set Android `compileSdkVersion` and `targetSdkVersion` to 36.
- Upgrade Android Gradle Plugin from 7.3.1 to 8.10.1.
- Upgrade the Gradle wrapper from 7.6.4 to 8.11.1.
- Move the application package declaration to the Gradle namespace.
- Supply namespaces for legacy Flutter plugins that do not declare one.
- Align each Android module's Kotlin JVM target with its Java target.
- Enable core-library desugaring required by `qr_code_scanner`.
- Add the media-projection foreground-service permission.
- Register a `MediaProjection.Callback` and release capture resources when
  projection is terminated externally.

## Motivation

Google Play requires Android applications to target API 36.

A previous SDK-only update changed `compileSdkVersion` and `targetSdkVersion`
to 36 while retaining Android Gradle Plugin 7.6.4. That toolchain does not
officially support API 36 and can fail while processing the API 36 resource
table.

This change upgrades the Android build toolchain to versions that support API
36 and addresses the resulting compatibility requirements from legacy Flutter
plugins.

## Testing

- Verified the installed Android SDK Platform 36 `android.jar` archive.
- Built an arm64 release APK using:
  - Flutter 3.24.5
  - Java 17
  - Android Gradle Plugin 8.10.1
  - Gradle 8.11.1
  - NDK r28c
- Verified APK metadata reports:
  - `compileSdkVersion=36`
  - `targetSdkVersion=36`
  - `minSdkVersion=22`
- Installed the APK on a Samsung SM-S911B.
- Verified the installed package reports `targetSdk=36`.
- Confirmed a successful cold launch.
- Confirmed no crash was recorded after launch.

## Notes

This PR is independent of #15602, which replaces broad external-storage access
with scoped storage. Both branches currently include Android media-projection
compatibility changes. If one PR merges first, the remaining branch will be
rebased onto the updated `master` and the overlap resolved.

## Summary by CodeRabbit

* **Enhancements**
  * Improved Android compatibility with updated platform and build tooling.
  * Enhanced reliability for microphone, screen-sharing, and remote access sessions.
  * Added clearer foreground-service handling for authorized unattended connections.
  * Improved cleanup, permission checks, and session lifecycle management.
* **Bug Fixes**
  * Prevented unintended service exposure.
  * Handled failed screen-capture and audio initialization more safely.
  * Improved recovery when foreground services or audio recording fail to start.
  * Prevented stale capture states during projection changes.

This PR updates the Android application to target API 36 and upgrades its Gradle, Android Gradle Plugin, and Kotlin toolchain compatibility.
- Adds required Android foreground-service permissions, service types, namespaces, and core-library desugaring.
- Strengthens screen-capture recovery and media-projection lifecycle cleanup.
- Improves audio-recorder startup, switching, failure handling, and shared publisher accounting.
- Updates the playground Android build environment from Java 11 to Java 17.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/MainService.kt | Reworks media-projection replacement, capture recovery, foreground-service typing, and resource cleanup; the previously reported stale-state paths are addressed. |
| flutter/android/app/src/main/kotlin/com/carriez/flutter_hbb/AudioRecordHandle.kt | Adds explicit recorder-start failure handling, deterministic cleanup, and shared audio-frame publisher accounting. |
| .github/workflows/playground.yml | Updates the Android playground build environment to Java 17, resolving the previously reported incompatibility with the upgraded Android Gradle Plugin. |
| flutter/android/settings.gradle | Configures Android Gradle Plugin 8.10.1 and Kotlin 2.1.21 for the upgraded Android build. |
| flutter/android/app/build.gradle | Targets Android API 36, defines the application namespace, and enables core-library desugaring. |
| flutter/android/app/src/main/AndroidManifest.xml | Declares microphone and special-use foreground-service permissions and restricts MainService from external exposure. |

</details>

```mermaid
flowchart TD
  A[Capture requested] --> B{MediaProjection available?}
  B -- No --> C[Return failure]
  B -- Yes --> D[Create or reuse virtual display]
  D -->|Success| E[Start audio when available]
  E --> F[Mark capture active]
  D -->|Failure| G[Release projection and display]
  G --> H[Mark recovery pending]
  H --> I[Request projection permission]
  I -->|Denied or failed| J[Clear pending recovery]
  I -->|Granted| K[Replace projection]
  K --> L[Clear pending recovery]
  L --> A
```
</details>

<sub>Reviews (8): Last reviewed commit: ["fix(android): handle audio and projectio..."](https://github.com/rustdesk/rustdesk/commit/e2e75e9aa7505b27b1204c81f919420626ab7f32) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47862174)</sub>


fufesou:
Hi @michaeljclarkk, thanks for the PR

I'm also working on the same issue. You can merge my changes, and I won't submit a duplicate PR.

https://github.com/fufesou/rustdesk/pull/68

It may also address the storage API usage.

michaeljclarkk:
Ill need to change to API 36 - as googles [latest ](https://developer.android.com/google/play/requirements/target-sdk) has changed since I started this change!

rustdesk:
I’d **request changes before merge**. The API/toolchain bump itself looks sound, but I found one projection-lifecycle correctness issue and two release-validation gaps.

1. **[P2] `releaseMediaProjection()` can leave a `VirtualDisplay` belonging to a stopped projection alive.** On Android 14+, `reuseVirtualDisplay` is enabled, so `stopCapture()` intentionally only detaches the surface and retains `virtualDisplay`. When a new projection permission result arrives, `onStartCommand()` calls `releaseMediaProjection()`, which first unregisters the callback, then calls `MediaProjection.stop()`, and only clears `mediaProjection`. Because the callback was unregistered, its cleanup path cannot release/null `virtualDisplay`. Later, `createOrSetVirtualDisplay()` sees that stale non-null display and calls `resize()`/`setSurface()` instead of creating a display from the **new** `MediaProjection`.  Android explicitly requires projection-owned resources such as the `VirtualDisplay` to be released when a projection session ends, and a stopped virtual display cannot resume. ([[Android Developers](https://developer.android.com/reference/android/media/projection/MediaProjection.Callback?utm_source=chatgpt.com)][1]) I would make the projection-replacement path release and null the existing `VirtualDisplay` before/while stopping the old projection. Keep the current detach-only behavior for ordinary `stopCapture()` if retaining the same projection session is intentional.

2. **[merge blocker] The tests in the PR description no longer test what is being merged.** The description still says API 35, AGP 8.6.1, Gradle 8.7, and verifies `targetSdk=35`, whereas the current diff is `compileSdkVersion/targetSdkVersion 36`, AGP 8.9.1, and Gradle 8.11.1. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/15603)][2]) Both GitHub CI and Full Flutter CI are currently awaiting maintainer approval, including the Android jobs. ([[GitHub](https://github.com/rustdesk/rustdesk/actions/runs/33314053177)][3]) I’d require the current head to pass CI plus an API 34/35/36 smoke test covering projection grant, system-stop/revoke, re-grant, two consecutive remote sessions, orientation change, boot startup, and voice-call microphone transitions. The latter is important because Android 14+ restricts adding a microphone FGS while the app is backgrounded; the new code correctly catches the resulting `SecurityException`, but that can turn into a functional voice-call failure for unattended/background scenarios. ([[Android Developers](https://developer.android.com/develop/background-work/services/fgs/restrictions-bg-start?utm_source=chatgpt.com)][4])

3. **[release blocker] `specialUse` needs corresponding Play Console approval, not just the manifest declaration.** The PR adds `FOREGROUND_SERVICE_SPECIAL_USE`, declares `specialUse` on `MainService`, and uses it whenever the service is foreground on Android 14+. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/15603/files)][5]) That may be appropriate for RustDesk's unattended-host use case, but Google Play requires Android 14+ apps to declare their FGS types in Play Console and supply the use-case description/user impact/demo video; `specialUse` is explicitly reviewed. ([[Google ](https://support.google.com/googleplay/android-developer/answer/16559646?rd=2&utm_source=chatgpt.com)][6]) Since Play compliance is the motivation for this PR, I’d confirm that declaration before treating the change as release-ready.

The **API/toolchain choices themselves look correct**. As of August 31, 2026, Google Play requires new apps and updates to target **API 36**, so the later API-36 commit is necessary and the PR title/description should be updated from “API 35.” ([[Google](https://support.google.com/googleplay/android-developer/answer/11926878?hl=en&utm_source=chatgpt.com)][7]) Google’s current compatibility table lists **AGP 8.9.1 as the minimum for API 36**, and Gradle **8.11.1** is the required Gradle baseline for AGP 8.9, matching this PR. ([[Android Developers](https://developer.android.com/build/releases/about-agp?utm_source=chatgpt.com)][8])

**Verdict: Request changes.** I’d block primarily on the stale `VirtualDisplay` ownership issue and getting the current API-36 head actually built/tested. The namespace/desugaring/plugin compatibility changes otherwise look reasonable from this review.

[1]: https://developer.android.com/reference/android/media/projection/MediaProjection.Callback?utm_source=chatgpt.com "MediaProjection.Callback  |  API reference  |  Android Developers"
[2]: https://github.com/rustdesk/rustdesk/pull/15603 "fix: android: target API 35 by michaeljclarkk · Pull Request #15603 · rustdesk/rustdesk · GitHub"
[3]: https://github.com/rustdesk/rustdesk/actions/runs/33314053177 "fix: android: target API 35 · rustdesk/rustdesk@9cde938 · GitHub"
[4]: https://developer.android.com/develop/background-work/services/fgs/restrictions-bg-start?utm_source=chatgpt.com "Restrictions on starting a foreground service from the background  |  Background work  |  Android Developers"
[5]: https://github.com/rustdesk/rustdesk/pull/15603/files "fix: android: target API 35 by michaeljclarkk · Pull Request #15603 · rustdesk/rustdesk · GitHub"
[6]: https://support.google.com/googleplay/android-developer/answer/16559646?rd=2&utm_source=chatgpt.com "Device and network abuse - Play Console Help"
[7]: https://support.google.com/googleplay/android-developer/answer/11926878?hl=en&utm_source=chatgpt.com "Target API level requirements for Google Play apps - Play Console Help"
[8]: https://developer.android.com/build/releases/about-agp?utm_source=chatgpt.com "About Android Gradle plugin  |  Android Studio  |  Android Developers"

fufesou:
https://github.com/rustdesk/rustdesk/pull/15603#issuecomment-5472727849

1. The stale `VirtualDisplay` problem was fixed in `304bbee`. The latest changes also stop active capture and clear the old projection resources before enabling the foreground-service type for its replacement. A successful replacement installs a per-session callback and restarts capture when needed.

2. GitHub CI built `304bbee` successfully. The follow-up lifecycle changes have not been built or tested yet. I have updated the PR title and testing text to API 36, AGP 8.10.1, and Gradle 8.11.1. The API 34/35/36 projection and microphone matrix is still relevant because this PR changes those lifecycle paths. Voice-call microphone startup failures are returned to the caller and rolled back cleanly; audio frame publishing is enabled only after `AudioRecord.startRecording()` succeeds.

3. **`specialUse`** is needed for the current service lifecycle. `MainService` enters the foreground at startup, including boot and unattended mode, before either `mediaProjection` or `microphone` is active. Without `specialUse`, it would pass `FOREGROUND_SERVICE_TYPE_NONE` on API 34 and later. The source-side requirements are covered by the manifest permission, service type, subtype description, and runtime type. Google Play still requires a separate declaration in Play Console with the use case, user impact, and a demo video. **Someone with access to the app's Play Console must submit that before publishing**.

fufesou:
@coderabbitai review

fufesou:
@codex review

fufesou (approved):

fufesou (approved):

fufesou:
One broken on Android 12, starting a voice call as the control side fails with "Failed to start voice call."

fufesou (approved):

fufesou:
> One broken on Android 12, starting a voice call as the control side fails with "Failed to start voice call."

I investigated this further on the affected Android 12 device.
The failure is real, but the evidence so far does not show that the PR introduced the underlying `AudioRecord` startup failure.
The PR now detects and reports a failure that `master` silently treats as success.

### What the diagnostic logs show

The voice-call recorder is created successfully:

```text
createAudioRecorder begin, sdk=32, inVoiceCall=true, mediaProjectionPresent=false
createAudioRecorder done, state=1, recordingState=1, audioSource=7, sessionId=169, minBufferSize=131072
```

Here, `state=1` is `AudioRecord.STATE_INITIALIZED`, `recordingState=1` is `AudioRecord.RECORDSTATE_STOPPED`, and audio source `7` is `MediaRecorder.AudioSource.VOICE_COMMUNICATION`. This rules out failure while constructing or initializing the `AudioRecord` instance.

The failure happens when Android tries to start the native input:

```text
AudioFlinger: start(70): startInput failed, status -61
AudioRecord: start(28): status -61
startRecording returned, recordingState=1
```

Status `-61` maps to `NOT_ENOUGH_DATA`/`ENODATA`. More importantly, `AudioRecord.startRecording()` returns without throwing, but the recorder remains in `RECORDSTATE_STOPPED`. The `IllegalStateException` in the later application log is raised by the new explicit state check; it is not the original Android framework failure.

The failure is therefore below the Rust frame-publishing layer. Calling `FFI.setFrameRawEnable("audio", true)` before `startRecording()`, as `master` does, cannot make Android's audio input start. That FFI call only controls whether Rust accepts raw frames after Android has produced them.

### Why `master` appears to behave differently

`master` calls `startRecording()`, immediately sets `audioRecordStat = true`, starts the reader thread, and then returns success from `switchToVoiceCall()`. It does not verify that `recordingState` changed to `RECORDSTATE_RECORDING`.

On this device, the same native failure can therefore be hidden by `master`:

- the caller is told that the voice call started;
- the recorder never enters the recording state;
- no microphone frames are delivered;
- the read loop discards non-positive read results without reporting the failure.

The current PR's state check is correct and should remain. It changes a false success into an explicit `false`, which is why the UI now shows "Failed to start voice call." This is not evidence that the check caused the native failure.

The fact that the same problem is present with `master` on this Android 12 device also makes the SDK 36 and foreground-service changes unlikely to be the cause. This control-side path goes directly through the `MainActivity` `AudioRecordHandle`; the failure occurs before any audio frame is published and is not caused by the media-projection foreground-service path.

### Most likely missing setup

Both `master` and this branch create a `VOICE_COMMUNICATION` recorder, but RustDesk does not set `AudioManager.MODE_IN_COMMUNICATION` anywhere and does not currently declare `android.permission.MODIFY_AUDIO_SETTINGS`.

Android documents `MODE_IN_COMMUNICATION` as the mode for an established audio/video chat or VoIP call. `AudioManager.setMode()` informs the audio HAL about the current audio state so that it can select the appropriate routing and communication behavior, and the call requires `MODIFY_AUDIO_SETTINGS`:

- https://developer.android.com/reference/android/media/AudioManager
- https://developer.android.com/reference/android/Manifest.permission
- https://developer.android.com/media/platform/sharing-audio-input

The strongest current explanation is that this Android 12 audio HAL does not activate the `VOICE_COMMUNICATION` input correctly without the process first becoming the communication-mode owner. The recorder can still be built and initialized, but `AudioFlinger::startInput()` fails when the route is actually started.

This is still a diagnosis to verify, not a confirmed fix. The first test should change only the communication-mode setup and retain the detailed logs. If the recorder then reaches `RECORDSTATE_RECORDING` and produces its first frame, that confirms the cause. If status `-61` remains, the mode change should not be presented as the fix; the next variables to isolate would be the selected input route and the stereo/PCM-float recording format. I do not plan to add a silent fallback to `MIC`, because that would hide the failure and may change echo cancellation, noise suppression, routing, and audio quality.

### Proposed test-branch fix

The minimal test fix touches two files:

1. Add `android.permission.MODIFY_AUDIO_SETTINGS` to the application manifest.
2. Manage `AudioManager.MODE_IN_COMMUNICATION` in `AudioRecordHandle` around the lifetime of a voice-call recorder.

The mode handling needs to follow these rules:

- Acquire communication mode before creating or starting a `VOICE_COMMUNICATION` recorder.
- Keep it active for the complete voice-call recording lifetime.
- Release the acquisition immediately if recorder creation fails before a recorder exists.
- For a created recorder, release the per-handle mode ownership from its centralized cleanup/`finally` path. This must cover `startRecording()` failure, normal shutdown, an unexpected capture-thread exit, switching back to playback capture, call closure, and handle destruction without double release.
- Call `setMode(MODE_NORMAL)` only after the last RustDesk voice-call recorder releases it, thereby relinquishing RustDesk's process-level mode request.
- Log the previous, requested, and resulting modes on the test branch.
- If setting the mode throws, log the exception and return an explicit startup failure. Do not continue through a fake-success path.

The last-owner rule is important because the Android process has two `AudioRecordHandle` instances, one owned by `MainActivity` and one by `MainService`. Android globally arbitrates audio mode using requests associated with application processes, so both handles share RustDesk's process-level request. A simple per-instance `setMode(MODE_NORMAL)` on close could break a voice recorder still owned by the other component. The implementation should therefore use a synchronized process-level reference count plus a per-handle ownership flag, similar in scope to the existing shared audio-frame publisher count.

I would not select a particular communication device in this first fix. Hard-coding the speaker, earpiece, or Bluetooth device would change existing routing policy and add a separate regression risk. `MODE_IN_COMMUNICATION` should first allow Android and the device HAL to choose their normal communication route.

### Permission and release impact

`MODIFY_AUDIO_SETTINGS` is a `normal` install-time permission. Android grants it automatically when the app is installed. It does not show a runtime permission dialog, does not grant access to microphone samples, and does not replace the existing `RECORD_AUDIO` permission. Android describes normal permissions as low-risk permissions with minimal impact on user privacy and other applications:

- https://developer.android.com/guide/topics/permissions/overview

Because this is a normal permission and Google Play does not identify it as a high-risk or sensitive permission, I do not expect it to require a separate Permissions Declaration Form. The release owner should still confirm the result after uploading the bundle to Play Console. It may be visible in the app's install-time permission information, as other normal permissions are:

- https://support.google.com/googleplay/android-developer/answer/9214102

I searched the current repository and the reachable history of this non-shallow clone, including all local refs. Neither `MODIFY_AUDIO_SETTINGS` nor `MODE_IN_COMMUNICATION` appears in RustDesk's tracked source history. This would be a new explicit capability for RustDesk. A historical dependency could theoretically have contributed the permission through manifest merging, but that cannot be established from source history without rebuilding the corresponding old artifacts.

### Regression risks

The manifest permission by itself does not change runtime behavior. The meaningful risk comes from calling `AudioManager.setMode()` because audio mode affects routing, communication volume behavior, Bluetooth/earpiece selection, and audio policy decisions.

The main failure cases to avoid are:

- leaving the process in communication mode after the voice call ends;
- restoring `MODE_NORMAL` while the other RustDesk recorder still needs communication mode;
- leaking the mode request on recorder creation or startup failure;
- changing playback-capture behavior when switching out of a voice call;
- unnecessarily overriding Android's selected communication device;
- interacting poorly with a real phone call or another VoIP application's active audio session.

The synchronized shared ownership and complete failure-path cleanup are intended to contain those risks. Outside an active RustDesk voice call, the app should make no mode request and existing screen/playback capture behavior should remain unchanged.

The APIs themselves are not Android 12-specific: `MODIFY_AUDIO_SETTINGS` and `setMode()` have existed since API 1, and `MODE_IN_COMMUNICATION` since API 11. They remain supported on later Android versions. Android 12 and newer versions place more importance on communication-mode ownership when resolving routes, so using the documented VoIP mode is preferable to a device-specific workaround. OEM audio HAL behavior still varies, so success on the affected Android 12 device should be followed by smoke checks on a recent Android version before treating the fix as release-ready.

### Confirmation criteria

For the affected Android 12 device, the retained logs should confirm all of the following:

- the mode request is made before `startRecording()`;
- the native `startInput failed, status -61` message disappears;
- `recordingState` changes from `1` (`STOPPED`) to `3` (`RECORDING`);
- the capture thread logs receipt of its first audio frame;
- the remote side receives microphone audio;
- closing the call releases RustDesk's mode request and, when no other mode owner is active, the observed mode returns to `MODE_NORMAL`;
- a second voice call can start after the first one closes;
- switching between playback capture and voice-call capture still works;
- disabling ordinary audio capture does not disable an active voice-call publisher;
- destroying either `MainActivity` or `MainService` does not release communication mode while the other still owns it.

For higher Android versions, the important smoke cases are starting and closing a call in foreground and background/unattended scenarios, switching speaker/Bluetooth routes, and confirming that ordinary media playback and playback capture use their normal routing after the call closes.

For now, I will keep the verbose diagnostic logs on the test branch and apply the communication-mode change there. Once the Android 12 device confirms the expected mode transition, recording state, and first frame, the working branch can receive the same fix as a small clean change without the temporary diagnostic noise.

PR #15607, opened 2026-07-16
fix: key help tools wasting vertical space on wide screens

33modeling:
## Problem

The mobile remote-page key help tools are laid out for narrow portrait phones. A line break is forced before each key group, so the overlay always occupies three or more rows on wide screens such as unfolded foldables, tablets, and landscape phones, leaving most of each row empty.

## Change

- At widths of at least `kMobilePageConstraints.maxWidth` (600dp), forced group breaks become no-ops and the existing `Wrap` lays keys out naturally in fewer rows.
- The existing **More keys** (`...`) toggle stays separate from the expanded key group and is pinned to the right on wide screens.
- Screens narrower than 600dp keep the existing layout unchanged.

The terminal-keyboard portion was removed from this PR because #15532 redesigns that area. This PR now modifies only `remote_page.dart`, and a test merge with the current #15532 head completes without conflicts.

## Screenshots

Captured on an Android 14 emulator connected to a Linux peer. "Before" is the official 1.4.9 release APK and "after" is this PR build.

| | Before | After |
|---|---|---|
| **Foldable unfolded (about 830dp wide)** |  |  |
| **Phone (about 393dp wide), unchanged** |  |  |

### DIFF
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -986,6 +986,14 @@ class _KeyHelpToolsState extends State<KeyHelpTools> {
       return Offstage();
     }
     final size = MediaQuery.of(context).size;
+    // On wide screens (foldables, tablets, phones in landscape) the key groups
+    // fit in one or two flowing rows, so keep them on the same row instead of
+    // forcing a line break before each group. Narrow (portrait phone) screens
+    // keep the forced breaks so each group starts on its own row.
+    final isWideScreen = size.width >= 600;
+    // A zero-height box wider than any screen makes `Wrap` start a new run.
+    final groupBreak =
+        isWideScreen ? const SizedBox.shrink() : const SizedBox(width: 9999);
 
     final pi = gFFI.ffiModel.pi;
     final isMac = pi.platform == kPeerPlatformMacOS;
@@ -1037,7 +1045,7 @@ class _KeyHelpToolsState extends State<KeyHelpTools> {
           active: _more),
     ];
     final fn = <Widget>[
-      SizedBox(width: 9999),
+      groupBreak,
     ];
     for (var i = 1; i <= 12; ++i) {
       final name = 'F$i';
@@ -1046,7 +1054,7 @@ class _KeyHelpToolsState extends State<KeyHelpTools> {
       }));
     }
     final more = <Widget>[
-      SizedBox(width: 9999),
+      groupBreak,
       wrap('Esc', () {
         inputModel.inputKey('VK_ESCAPE');
       }),
@@ -1093,7 +1101,7 @@ class _KeyHelpToolsState extends State<KeyHelpTools> {
       wrap('Enter', () {
         inputModel.inputKey('VK_ENTER');
       }),
-      SizedBox(width: 9999),
+      groupBreak,
       wrap('', () {
         inputModel.inputKey('VK_LEFT');
       }, icon: Icons.keyboard_arrow_left),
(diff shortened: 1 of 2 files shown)


rustdesk:
Please show sceenshots for different screen sizes.

33modeling:
Here is the current behavior (v1.4.9) of the **terminal page key bar** on an unfolded Galaxy Z Fold inner display (cropped to the bar area). The two fixed rows are sized for a portrait phone; on this wide screen all 14 keys fit in a single row, so the second row only pushes the terminal content up:

terminal key bar on unfolded foldable, before

I've extended the PR with the same single-row treatment for this bar (second commit). Will follow up with after-screenshots and the remote-page key help tools at phone/foldable sizes.

33modeling:
Screenshots at different screen sizes, captured on an Android 14 emulator (Genymotion-free stock AVD) connected to a Linux peer. "Before" is the official 1.4.9 release APK, "after" is this PR's build. Narrow-screen layouts are unchanged by design.

### Terminal key bar

| | Before | After |
|---|---|---|
| **Foldable unfolded (2176×1812, 420dpi ≈ 830dp wide)** |  |  |
| **Phone (1080×2340, 440dpi ≈ 393dp wide) — unchanged** |  |  |

On the unfolded foldable the two fixed rows collapse into a single row, returning one row's height to the terminal. On phones the layout is identical to before.

### Remote page key help tools (pinned, '...' expanded)

| | Before | After |
|---|---|---|
| **Foldable unfolded** |  |  |
| **Phone — unchanged** |  |  |

On the unfolded foldable the forced per-group line breaks (3 sparse rows with the right half empty) become 2 fully-packed rows. On phones the layout is identical to before.

rustdesk:
`...` looks wierd now.

And this PR https://github.com/rustdesk/rustdesk/pull/15532 is also adding this `...`, it will be wierd also.

33modeling:
The `...` is the existing **More keys** toggle; it is blue in the screenshot because the expanded state is active. This PR did not add that control, but I agree that packing it inline with the expanded keys on wide screens makes its placement look odd. I will revise the layout so the modifier/toggle controls remain visually separate from the expanded key group, and I will also check the interaction with #15532 to avoid repeating the same issue.

33modeling:
The `...` placement issue raised above was fixed in `0d000de44`: on wide screens, the existing **More keys** toggle is kept separate from the expanded key group and pinned to the right. Narrow-screen layouts remain unchanged.

Compatibility with the current #15532 has now been checked and the overlap has been removed:

- #15532 no longer adds the same `...` toggle; its latest version uses a `∨` / `∧` expand-collapse control.
- The terminal-keyboard changes were removed from this PR in `b9dd342ac`, leaving this PR scoped to `remote_page.dart`.
- A test merge of the current heads of #15607 and #15532 now completes successfully with no conflicts.

**The requested fix is complete, and compatibility with #15532 has been checked and confirmed.**

33modeling:
@rustdesk Gentle ping — just following up on this PR.

The `...` placement you flagged was addressed in 0d000de44: on wide screens the **More keys** toggle is now kept separate from the expanded key group and pinned to the right, so it no longer packs inline with the keys. Narrow-screen layouts are unchanged, and compatibility with #15532's overlap toggle has been verified (screenshots in the comment above).

No conflicts with master and CodeRabbit passes. Please take another look when you have a chance — happy to make further adjustments if anything still looks off.

rustdesk:
Thanks, https://github.com/rustdesk/rustdesk/pull/15532 has been merged. You can make some change for it.

PR #15616, opened 2026-07-17
Enable configurable connection management notification window

mynetfan:
## Summary by CodeRabbit

- **New Features**
  - Added a Security setting to hide or show the connection management window when password protection is enabled.
  - The setting now takes effect automatically when verification or approval modes change.

- **Bug Fixes**
  - Corrected startup and password-update handling so the connection management window consistently reflects the configured preference.
  - Prevented the window from being hidden in permanent-password mode.


mynetfan:
ok

Issue #15601, opened 2026-07-15
[Linux/Wayland] Remote mouse/click position is offset  when controlling a KDE Plasma 6 Wayland host — multi-monitor; RustDesk 1.4.9 via RPM


Abel-12345678:
### Bug Description

When my laptop (running Linux Fedora 44, KDE plasma) is controlled, the remote mouse
cursor is offset from where input actually lands — I know it is relative but it appears roughly
5–10 cm to the LEFT of where it should be — so clicks hit the wrong place.

- With multiple monitors shared, cursor is offset every time, from every client I tested except for one case: Windows 11 client running rustdesk 1.4.8 and me using an older LG monitor
- With a single monitor shared it is inconsistent: it was accurate from
  one client (Arch/X11)  when  monitors were connected but only the main screen (laptop's screen) was shared but a Windows 11 client still could not position the cursor correctly on my screen (external monitors were connected but only main was shared). 

### How to Reproduce

1. Controlled host: Fedora, KDE Plasma 6, **Wayland** session, laptop's internal
   HiDPI panel (scaled 155%) plus one or two external 1080p monitors (scale 100%).
2. Accept an incoming connection and share the monitor(s).
3. From a client, move the mouse and click around the remote desktop.

### Expected Behavior

The remote cursor lands where the controller points; clicks hit the target.

### Operating system(s) on local (controlling) side and remote (controlled) side

Windows 11 -> Linux Fedora 44 (KDE, Wayland), Arch Linux (dwm, Xorg) -> Linux Fedora 44 (KDE, Wayland),

### RustDesk Version(s) on local (controlling) side and remote (controlled) side

1.4.8/1.4.9->1.4.7, 1.4.8->1.4.9, 1.4.6->1.4.9

### Screenshots

<img width="1765" height="880" alt="Image" src="https://github.com/user-attachments/assets/d5c0d5d8-3784-4769-a49b-45118b2825f1" />

In the image on the right side there is the Arch (client) on the left Fedora 44. 

### Additional Context

I looked into this together with an AI assistant. The following are **hypotheses /
leads from that investigation, not a confirmed diagnosis** — sharing in case they
help whoever picks this up. These are the assumptions of the AI not mine.

From the host logs, the multi-display coordinate space seems to be built from each
output's **logical position** but its **physical (unscaled) pixel size**, and the
mouse-input space always spans the whole desktop regardless of what is shared. Two
captures (IPs / session handles scrubbed; only the display-handling lines shown):

```
[...]

PR #15619, opened 2026-07-18
fix(wayland): warn when advertised multi-display geometry is incoherent

VibeCodyH:
Partial fix for #15601.

On Wayland multi-monitor hosts with a fractionally scaled output, if the output's logical size is not detected, the server advertises the display as logical origin + physical size with `scale = 1.0`. Neighboring display rects then overlap in the client's coordinate space, and absolute input lands offset (clicks hit the wrong monitor). The client cannot detect or correct this, and the existing detection-failure paths are silent (`try_fix_logical_size` skips with a debug log; `get_displays` failures fall back to physical size), so there is nothing in the host logs to point at the cause.

This does not change any advertised value. It adds detection and logging so the failure is diagnosable:

- `check_update_displays` now checks whether the client-effective rects (`x`, `y`, `width / scale`, `height / scale`) overlap, and logs a warning with the full geometry of the offending pair when they do. Sub-pixel contact at shared edges from fractional rounding is ignored.
- The `try_fix_logical_size` skip on a mismatched portal size is raised from `debug` to `warn`, since on a scaled output that skip is what leads to the misplacement.

Why not advertise logical sizes and fix it outright: I tried that on the same rig. Input becomes correct for every client, but `DisplayInfo.width/height` is also the video buffer contract, so rendering breaks into overlapping strips. A real fix needs logical desktop geometry carried separately from physical video dimensions, which is a larger protocol-facing change; this warning is a safe first step that makes the field failures identifiable.

Tested (Fedora 44, KDE Plasma 6 Wayland, host RPM, one 2560x1440 monitor at 125% next to one at 100%, cursor position verified with kdotool):
  - Healthy multi-monitor session (logical sizes detected): no warning logged, input accurate
  - Broken geometry (native 1.4.0 client against the same host): clicks land offset by exactly the 1.25 scale factor onto the wrong monitor; the overlap this patch detects is the advertised-geometry shape that produces it
  - `cargo test -p rustdesk display_service`: 5 passed (includes the reporter's 155% panel layout from the issue)
  - `cargo fmt --check` clean on touched files, clippy clean on touched code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of display layout issues on Linux when fractional scaling is enabled.
  * Added clearer warnings for physical-size mismatches and overlapping display rectangles, skipping logical-size correction when appropriate.
* **Tests**
  * Added Linux unit tests covering overlapping and non-overlapping display scenarios, including fractional rounding edge cases.

### DIFF
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -1035,7 +1035,9 @@ pub fn try_fix_logical_size(shared_displays: &mut Vec<crate::Display>) {
                         {
                             // If "Full Workspace" is selected in the portal dialog,
                             // the physical size reported by portal may not match the display info.
-                            debug!(
+                            // Warn rather than debug: skipping here on a fractionally scaled
+                            // output leads to clients misplacing input (#15601).
+                            warn!(
                             "Physical size of capturable ({:?}) does not match display info: ({:?}) - ({:?}). Skipping logical size fix.",
                             capturable.position,
                             capturable.physical_size,
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -1033,16 +1033,24 @@ pub fn try_fix_logical_size(shared_displays: &mut Vec<crate::Display>) {
                         if capturable.physical_size.0 != wd.width as usize
                             || capturable.physical_size.1 != wd.height as usize
                         {
-                            // If "Full Workspace" is selected in the portal dialog,
-                            // the physical size reported by portal may not match the display info.
-                            // Warn rather than debug: skipping here on a fractionally scaled
-                            // output leads to clients misplacing input (#15601).
-                            warn!(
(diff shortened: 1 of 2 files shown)


VibeCodyH:
Second commit is from the same #15601 dig. Reproduced on Fedora 44 / Plasma 6, two 2560x1440 outputs with the left one at 150%:

- Full Workspace share: one combined 5120x1440 capturable at scale 1.0. The portal gives no per-output logical size, so try_fix_logical_size can't pair it, scale stays 1.0, and input is offset on the scaled output.
- Individual screens: per-output capturables that pair 1:1 and get the right logical size/scale (1.5 and 1.0) straight from the portal, no fixup needed.

So full-workspace is a distinct trigger for this offset. The commit detects the spanning capturable and turns the silent skip into an actionable warning (share individual screens or use integer scaling). Logging only, generic message kept for other mismatches.

rustdesk:
I do not think the warn is useful.

Issue #14603, opened 2026-03-23
macOS: 2.9 GB memory leak via MALLOC_SMALL after 4 days (23.5M leaked allocations)


clemlesne:
## Summary

RustDesk v1.4.6 on macOS leaks **~11.6 MB per connection session** (~92,000 allocations of ~132 bytes each) that are never freed. Over 4 days and 256 connections, this accumulates to **2.9 GB in `MALLOC_SMALL`** (23.5M total allocations). The leak compounds with a CVDisplayLink crash on child process exit and a connection accept race condition that creates orphaned connection objects.

## Environment

- **RustDesk version:** 1.4.6 (build 64)
- **OS:** macOS 26.3 (25D125), ARM64
- **Hardware:** MacBookPro18,4 (M1 Max), 32 GB RAM
- **Display:** 5120×2880 (Retina)
- **Codec in use:** AOM AV1 software encoder (hardware HEVC VT fails, see #13069)
- **Config flags:** `av1-test = 'Y'`, `use-texture-render = 'Y'`

## Observed behavior

After ~4 days of uptime (PID running since Mar 19), `vmmap` shows:

| Metric | Value |
|---|---|
| Physical footprint | 3.1 GB (peak 3.6 GB) |
| `MALLOC_SMALL` swapped | 2.9 GB across 825 regions |
| Total allocations in `DefaultMallocZone` | 23,532,992 |
| Average allocation size | 132 bytes |
| Fragmentation | 2% |
| RSS | 45 MB (rest compressed/swapped by macOS) |

```
MALLOC ZONE                          SIZE       SIZE       SIZE       SIZE      COUNT  ALLOCATED  FRAG SIZE  % FRAG   COUNT
DefaultMallocZone_0x10475c000        3.1G      3456K      3456K       3.0G   23532960       2.9G      44.7M      2%     862
```

### Bug 1 — Per-connection memory leak (~11.6 MB / session)

High-level resources are properly cleaned up (verified via logs):

| Resource | Created | Destroyed | Delta |
|---|---|---|---|
| Video service | 302 | 302 | 0 |
| Clipboard subscriptions | 239 | 239 | 0 |
| Wakelocks | 239 | 239 | 0 |

But **~92,000 small allocations (~132 bytes each, totaling ~11.6 MB) leak per connection** and are never freed. Over 256 connections: 256 × 11.6 MB ≈ 2.9 GB — matching the observed leak exactly.

[...]

Issue #8899, opened 2024-07-31
OSX memory leak


pnill:
### Bug Description

If left on/connected to a host for a long period there is a memory leak where it continually eats ram until having to be forcefully closed.

### How to Reproduce

Leave it running overnight connected to a host (I was connected to 3), and you'll notice it consumes a ton of RAM.

### Expected Behavior

It to only use what RAM it needs and to cleanup after itself.

### Operating system(s) on local side and remote side

OSX Sonoma -> Windows 10

### RustDesk Version(s) on local side and remote side

1.2.7 (osx) -> 1.2.7 (windows)

### Screenshots

<img width="413" alt="Screenshot 2024-07-30 at 11 20 11 AM" src="https://github.com/user-attachments/assets/344a4581-e395-4620-a27d-623d5bf6f93d">

Issue #7747, opened 2024-04-17
Memory leaking  MAC OS 13


lovitus:
### Bug Description

reaching more than 2G memory usage after few days  (only standby , no remote control connected before )

### How to Reproduce

open Rustdesk, keep it in background running .

### Expected Behavior

drain reasonable memory usage 

### Operating system(s) on local side and remote side

Macos 13

### RustDesk Version(s) on local side and remote side

1.2.3(39)

### Screenshots

![image](https://github.com/rustdesk/rustdesk/assets/14361541/d44350b7-68c6-4d94-90da-d44a047d6a2d)

rustdesk:
try out https://github.com/rustdesk/rustdesk/releases/nightly please

lovitus:
> try out https://github.com/rustdesk/rustdesk/releases/nightly please

problem solved in nightly version .

PR #15623, opened 2026-07-18
fix: macOS server memory growth, wrap BackingScaleFactor in autoreleasepool

Kuksgauzen:
## What this PR does

Wraps the body of `BackingScaleFactor()` (`src/platform/macos.mm`) in `@autoreleasepool`.

Fixes the unbounded memory growth of the macOS `--server` process reported in #14603 (and matching #13277, #8899, #7747).

## Root cause

`BackingScaleFactor()` iterates `[NSScreen screens]` and calls `[screen deviceDescription]` on each screen. Both return autoreleased objects. The function is reached via scrap's `Display::scale()` (which `width()`/`height()` also call internally) from the display service loop, which runs every 300 ms for as long as a peer is subscribed, on a plain Rust thread that never drains an autorelease pool. The ObjC runtime accumulates the autoreleased dictionaries in implicit pool pages that would only drain at thread exit, so they accumulate for the lifetime of the process.

On a host with 10 days of uptime this reached 16 GB of phys_footprint: `heap` shows ~20.9M live `NSDeviceDescriptionKey` dictionaries (448 B storage each on macOS 26, where AppKit returns them Swift-bridged), 2 `NSValue`s per dictionary, and ~137k autorelease pool pages pinning them. `leaks` reports nothing because everything is still reachable from the pools, which is likely why #14603 was labeled unreproducible. Full analysis with numbers in https://github.com/rustdesk/rustdesk/issues/14603#issuecomment-5012040507.

Every other NSScreen helper in the codebase already runs under a pool (`get_focused_display`, `get_cursor`, `get_cursor_data` are wrapped in `src/platform/macos.rs`); this was the only unwrapped path.

## Verification

Manual, on macOS 26.3 / MacBook Pro M4 Max: `vmmap --summary` allocation count of `DefaultMallocZone` grows by tens per second during an active session on unpatched builds (~130–140 MB/h) and is flat when idle. No automated test added: the leak lives in an ObjC helper driven by wall-clock service loops, so it isn't reachable by unit tests; measurement methodology and full numbers are in the linked issue comment.

## Symptoms this explains

- Memory grows only while a peer is connected, proportional to connection time; idle is completely flat.
- Growth lives in MALLOC_SMALL, average allocation ~132 B.
- Invisible to `leaks`; visible in `vmmap`/`heap`/footprint.
- Growth rate differs between macOS versions (Swift-bridged dictionaries on macOS 26 are ~3× larger than the ObjC ones on earlier systems).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved display-scale lookup stability on macOS.
  * Preserved accurate scaling for matched displays, with a safe default when no match is found.

### DIFF
--- a/src/platform/macos.mm
+++ b/src/platform/macos.mm
@@ -118,16 +118,18 @@ extern "C" bool MacCheckAdminAuthorization() {
 
 // https://gist.github.com/briankc/025415e25900750f402235dbf1b74e42
 extern "C" float BackingScaleFactor(uint32_t display) {
-    NSArray<NSScreen *> *screens = [NSScreen screens];
-    for (NSScreen *screen in screens) {
-        NSDictionary *deviceDescription = [screen deviceDescription];
-        NSNumber *screenNumber = [deviceDescription objectForKey:@"NSScreenNumber"];
-        CGDirectDisplayID screenDisplayID = [screenNumber unsignedIntValue];
-        if (screenDisplayID == display) {
-            return [screen backingScaleFactor];
+    @autoreleasepool {
+        NSArray<NSScreen *> *screens = [NSScreen screens];
+        for (NSScreen *screen in screens) {
+            NSDictionary *deviceDescription = [screen deviceDescription];
+            NSNumber *screenNumber = [deviceDescription objectForKey:@"NSScreenNumber"];
+            CGDirectDisplayID screenDisplayID = [screenNumber unsignedIntValue];
+            if (screenDisplayID == display) {
+                return [screen backingScaleFactor];
+            }
         }
+        return 1;
     }
-    return 1;
 }
 
 // https://github.com/jhford/screenresolution/blob/master/cg_utils.c


Kuksgauzen:
Some verification data. I've been running 1.4.9 rebuilt with this patch on the machine where I originally hit the leak (~16 GB footprint, see #14603). I sample `RustDesk --server` every 10 minutes with vmmap/ps: DefaultMallocZone size, allocation count, RSS, plus whether a remote session is active.

Official 1.4.9, last ~12h before I switched builds: DefaultMallocZone at 15.5-15.8 GB with ~127M live allocations, growing ~90-135 MB/h whenever a session was active, frozen while idle.

Patched build, 48h so far (~25h of it with an active session): zone stays between ~5 MB idle and ~50 MB during a session, allocation count sits around 36k, memory drops back to baseline after each session ends.

Each sample is basically:

```
vmmap --summary $(pgrep -f 'RustDesk --server') | grep '^DefaultMallocZone'
```

plus `ps -o rss=` for RSS and an `lsof` check for an established session.

Chart attached (blue = active remote session). Can share the raw log if useful.
изображение

PR #15629, opened 2026-07-19
fix(macos): prevent remote keyboard focus leaks

fufesou:
This PR may fix the following issues on macOS.

https://github.com/rustdesk/rustdesk/discussions/13398

Also some reports:

> 1. rustdesk is effectively incompatible with fullscreen on mac. I tend to put my heavily used apps on their own fullscreen desktop to avoid distraction while working, but the way that rustdesk pulls focus means that even if I don't have that desktop open, my typing might end up on whatever PC I am remoted into instead of slack or outlook, or my local IDE or anywhere else. If you are like me and might be controlling robots or instruments where bad settings can cause damage, this is pretty scary so it seems like best practice is to never fullscreen rust desk and fully quit and close connections if you are going to be working in another window for a while
> 2. Some users report keyboard input occasionally remaining directed to the remote session after switching away, while others report the opposite behavior, where keyboard input doesn't transfer to the active remote session. Other users with similar workflows cannot reproduce either issue. 
> 3. it appears to preferentially pull the keyboard but not the mouse. Sometimes I can get away with typing a sentence in another window by "typing" with copy/paste so I don't need to use my mouse
> 4. Huh, I tend to do the same (keeping a bunch of full-screen RustDesk sessions on Mac, each on their own fullscreen desktop), but I've never had an issue with it pulling the focus for a desktop that is not currently being shown on my screen before. I did encounter a kind of an oppositeds problem, though -- sometimes it refuses to move the focus to the host system that is currently shown on screen. Perhaps there're specific keyboard/accessibility settings or other apps that break it…

I can only reproduce a similar issue with the following steps:
1. Use two displays on macOS.
2. Connect, use "Input source 1".
3. In the first display, open TextEdit, and run `(sleep 8; open -a TextEdit) &` to gain focus after 8 seconds.
4. Put the remote window in the second display.
5. Enter full screen.
6. 8 seconds later, TextEdit gains focus, but what I type is still sent to the remote side.

### 1. Derive native keyboard ownership from the complete macOS context

`RemotePage` now tracks the macOS-specific facts needed to synchronize Flutter focus with native input:

- current app lifecycle state;
- whether local/window focus was lost;
- whether native remote input is currently active;
- whether the page has been permanently suppressed for close or transfer;
- whether fullscreen focus recovery is pending;
- whether the current `FocusNode` request was an explicit remote-input request;
- whether this page is the selected remote tab.

`_syncMacOSKeyboardGrab()` is the central reconciliation point. It enables native remote input only when all applicable conditions are satisfied:

- the lifecycle permits input, or the action is an explicit pointer/fullscreen
  recovery for a visible secondary window;
- the RustDesk window is focused and not blurred;
- this remote page is the selected tab;
- the pointer is over the remote image;
- local RustDesk focus has not taken ownership;
- the page has not been suppressed for close or transfer;
- the remote image's `FocusNode` has primary focus.

The native hook is disabled before the Flutter focus node is released. A `reassert` path is retained for cases where the native input source needs to be refreshed even though the derived boolean state did not change.

### 2. Release input on every ownership-loss boundary

On macOS, remote keyboard input is synchronized or released when:

- the window loses focus;
- the app lifecycle leaves `resumed`;
- the pointer leaves the remote image;
- the raw-key `FocusNode` loses primary focus;
- the page becomes an unselected tab;
- the window is minimized;
- the session is closed;
- a tab is transferred to another window.

Window focus alone does not authorize normal remote keyboard input. The pointer must explicitly enter or press the remote image before the page can reacquire input. This prevents a window or Space switch from silently restoring the native hook while the user is typing elsewhere.

### 3. Recover input after native fullscreen transitions

Entering or leaving native fullscreen queues a macOS focus recovery operation. Recovery waits until the next rendered frame and then advances one event-loop turn because macOS can deliver `FocusNode` loss after the fullscreen callback.

The recovery is preserved while lifecycle state is `hidden`, because secondary Flutter engines can temporarily remain hidden while their remote window is visible. It is cancelled for a blurred window, an unselected tab, a minimized or detached lifecycle, or a page that no longer contains the pointer.

When the page remains the valid owner, the input source is refreshed and both Flutter focus and native input are reasserted. This allows typing to continue after fullscreen entry or exit without requiring an unrelated app switch or reconnect.

### 4. Handle relative mouse mode separately

Normal pointer mode deliberately keeps the local-focus-loss latch after an ordinary window-focus event. Relative mouse mode is different: while capture is active, the pointer cannot leave the remote window, so normal `PointerExit`/`PointerEnter` ownership transitions do not occur. Blur temporarily releases the relative-mode constraints, and restoring them when window focus returns still does not generate a new `PointerEnter` event.

When a macOS window regains focus while relative mouse mode is active, the change clears that latch before running the existing synchronization logic. Pointer presence, selected-tab state, window focus, and primary Flutter focus are still required, so the exception is limited to the path that cannot produce its own pointer-enter event.

### 5. Make tab selection and tab-to-window transfer ordered

Each macOS remote page observes selected-tab changes. The old tab releases input synchronously. Acquisition by the newly selected page is deferred to a microtask, ensuring the previous page cannot disable a hook just acquired by the new page.

Before a source tab is removed during a tab-to-window transfer, it explicitly releases and suppresses its native input. Its later retained-page disposal then does not issue another leave operation that could disable input already acquired by the destination window.

### 6. Keep toolbar input-source changes local

Changing the macOS keyboard input source from the toolbar explicitly releases native remote input while the toolbar owns focus. Returning to the remote image uses the explicit pointer path, which refreshes the input source and reasserts the native state.

## Note

Important macOS cases include:

- Switching Spaces or blurring a window may not emit `PointerExit`.
- A fullscreen transition may report `hidden` or `inactive` even while a secondary Flutter window is visible.
- `FocusNode` loss may arrive after the fullscreen callback and after the next rendered frame.
- A new Flutter engine may initially have no lifecycle state.
- A page retained during tab-to-window transfer may be disposed of after the destination page has already acquired native input.
- Relative mouse mode confines the pointer to the remote window while capture is active. Restoring that capture after a blur/focus cycle does not emit a new `PointerEnter` event.

## Testing

This PR is macOS only.

### macOS

- [x] On the first connection, type in the remote image in windowed mode.
- [x] Enter fullscreen for the first time and type immediately without clicking the remote image again.
- [x] Leave and re-enter fullscreen repeatedly, including a second fullscreen cycle, and confirm typing remains available.
- [x] Put one remote session in a fullscreen Space, switch to TextEdit without moving the pointer, and confirm all typing remains local.
- [x] Put two remote sessions in separate fullscreen Spaces and confirm that only the visible, selected session receives its marker. 
- [x] Switch among RustDesk, local applications, Mission Control, and Spaces with both mouse and trackpad workflows.
- [x] Open a RustDesk toolbar menu, dialog, chat field, or other local control and confirm typing is not sent remotely.
- [x] Return to the remote image with PointerEnter and PointerDown paths and confirm input is restored exactly once.
- [x] Minimize and restore the window while the pointer is over the remote image;  confirm local typing does not leak while minimized.
- [x] Close and reconnect a session and confirm no stale focus state survives.
- [x] Swith between "Input source 1" and "Input source 2". 
- [x] Switch rapidly between two remote tabs while the pointer remains over the page area; only the selected tab may receive input.
- [x] Move the active tab into a new window and confirm the source releases before the destination acquires input.
- [x] Close a selected tab and a multi-tab remote window while the pointer is inside the image; subsequent local typing must remain local.

### Windows and Linux

- [x] Windows: enter/leave the remote image, Alt-Tab, switch tabs, close a tab,  and transfer a tab to another window.
- [x] Linux: enter/leave the remote image, blur/refocus, switch tabs, close a tab, and transfer a tab to another window.

## Changed behavior

"Input source 2" also needs the pointer inside the window to send keyboard events.

In the master commit, the mouse doesn't need to be in the window—it only needs focus.

## Summary of Updates

* **Bug Fixes**
  * Improved macOS keyboard/input focus consistency across tab switching, window focus/blur, and fullscreen transitions.
  * Added a macOS reset when changing the keyboard input source to prevent stuck/held input.
  * Improved reliability of remote tab/session closing, including safe handling when session data isn’t available.
  * Ensured macOS remote input is released and synchronized before completing tab/session transfer.

* **New Features**
  * Introduced generation-based macOS fullscreen focus recovery for more reliable focus restoration timing.

### DIFF
--- a/flutter/lib/desktop/pages/macos_full_screen_focus_recovery.dart
+++ b/flutter/lib/desktop/pages/macos_full_screen_focus_recovery.dart
@@ -1,0 +1,24 @@
+class MacOSFullScreenFocusRecovery {
+  int _generation = 0;
+  int? _pendingGeneration;
+
+  int? get pendingGeneration => _pendingGeneration;
+
+  int queue() {
+    _generation += 1;
+    _pendingGeneration = _generation;
+    return _generation;
+  }
+
+  void cancel() {
+    _pendingGeneration = null;
+  }
+
+  bool isCurrent(int generation) => _pendingGeneration == generation;
+
+  bool consume(int generation) {
+    if (!isCurrent(generation)) return false;
+    _pendingGeneration = null;
+    return true;
+  }
+}
--- a/flutter/lib/desktop/pages/remote_page.dart
+++ b/flutter/lib/desktop/pages/remote_page.dart
@@ -507,7 +507,9 @@ class _RemotePageState extends State<RemotePage>
     // Restore relative mouse mode constraints when window regains focus.
     if (_ffi.inputModel.relativeMouseMode.value) {
       if (isMacOS) {
-        // Relative mode does not emit PointerEnter after window focus returns.
+        // Native relative mode retains pointer capture and does not emit
+        // PointerEnter after window focus returns, so restore both latches.
+        _cursorOverImage.value = true;
         _macOSLocalFocusLost = false;
       } else {
         _rawKeyFocusNode.requestFocus();
--- a/flutter/lib/desktop/pages/remote_page.dart
+++ b/flutter/lib/desktop/pages/remote_page.dart
@@ -22,6 +22,7 @@ import '../../utils/image.dart';
 import '../widgets/remote_toolbar.dart';
 import '../widgets/kb_layout_type_chooser.dart';
 import '../widgets/tabbar_widget.dart';
+import 'macos_full_screen_focus_recovery.dart';
 
 import 'package:flutter_hbb/native/custom_cursor.dart'
     if (dart.library.html) 'package:flutter_hbb/web/custom_cursor.dart';
@@ -94,7 +95,7 @@ class _RemotePageState extends State<RemotePage>
   bool _macOSLocalFocusLost = false;
   bool _macOSInputActive = false;
   bool _macOSInputSuppressed = false;
-  bool _macOSFullScreenFocusRecoveryPending = false;
+  final _macOSFullScreenFocusRecovery = MacOSFullScreenFocusRecovery();
   bool _macOSExplicitFocusRequestPending = false;
   StreamSubscription<DesktopTabState>? _tabStateSubscription;
   final _cursorOverImage = false.obs;
@@ -271,7 +272,7 @@ class _RemotePageState extends State<RemotePage>
 
   void _onMacOSTabStateChanged(DesktopTabState _) {
     if (!_isSelectedTab) {
-      _macOSFullScreenFocusRecoveryPending = false;
+      _macOSFullScreenFocusRecovery.cancel();
       _syncMacOSKeyboardGrab();
       return;
     }
@@ -285,7 +286,7 @@ class _RemotePageState extends State<RemotePage>
   }
 
   void _releaseMacOSRemoteInput() {
-    _macOSFullScreenFocusRecoveryPending = false;
+    _macOSFullScreenFocusRecovery.cancel();
     _macOSExplicitFocusRequestPending = false;
     _macOSInputSuppressed = true;
     _macOSLocalFocusLost = true;
@@ -330,7 +331,8 @@ class _RemotePageState extends State<RemotePage>
     final lifecycleAllowsInput = allowInactiveLifecycle ||
         _macOSLifecycleState == null ||
         _macOSLifecycleState == AppLifecycleState.resumed;
-    // Window focus alone must not enable the global hook without pointer presence.
+    // Input stays pointer-gated except for focused fullscreen recovery, which
+    // compensates when macOS omits PointerEnter during a Space switch.
     final shouldFocus = lifecycleAllowsInput &&
         _isMacOSKeyboardContextActive &&
         !_macOSInputSuppressed &&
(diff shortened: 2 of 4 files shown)


fufesou:
@codex review

fufesou:
@codex review

fufesou:
@coderabbitai review

rustdesk:
# Review

The macOS relative-mouse restore path loses its pointer-presence prerequisite during minimization and cannot reacquire the keyboard grab automatically. This introduces a functional regression for minimized relative-mode sessions.

## Review comment

- **[P2] Re-establish pointer presence after minimizing relative mode** — `flutter/lib/desktop/pages/remote_page.dart:581`

  When a macOS session is minimized while relative mouse mode is active, this clears the cursor latch. `onWindowFocus()` only clears `_macOSLocalFocusLost`, while `_syncMacOSKeyboardGrab()` still requires `_cursorOverImage`; relative capture also does not emit a new `PointerEnter`. After restoring, keyboard input therefore remains disabled until an extra pointer action, so this latch should be preserved or restored for relative mode.

fufesou:
@codex review

PR #15637, opened 2026-07-20
feat(flutter): decode clipboard server config at startup

tianjinli:
Try to decode the clipboard text as a ServerConfig when the desktop or mobile home page is initialized. If the clipboard contains a valid server config (with non-empty idServer), show the server settings dialog pre-filled with the decoded values.

This allows users to copy a server config string and have RustDesk automatically offer to apply it on startup, instead of having to manually open the server settings page and paste it.

## Summary by CodeRabbit

* **New Features**
  * Automatically checks the clipboard for valid server configuration data when the home screen opens (mobile and desktop).
  * If a valid configuration is found, opens the server settings dialog with the fields prefilled.
  * Safely ignores empty/invalid clipboard content, suppresses decoding errors, and clears the clipboard after a successful read.
  * Applies platform-specific adjustments to the configuration as needed and performs an elevated-permission check when required.

### DIFF
--- a/flutter/lib/common.dart
+++ b/flutter/lib/common.dart
@@ -3656,6 +3656,25 @@ ColorFilter? svgColor(Color? color) {
   }
 }
 
+/// Try to decode clipboard text as ServerConfig.
+/// Returns the decoded config if valid (has idServer), otherwise null.
+ServerConfig? tryDecodeClipboardServerConfig(String? text) {
+  if (text == null || text.trim().isEmpty) return null;
+  try {
+    final sc = ServerConfig.decode(text.trim());
+    if (sc.idServer.isNotEmpty) {
+      // Clear relayServer on iOS and Web as the field is hidden in settings UI.
+      if (isIOS || isWeb) {
+        sc.relayServer = '';
+      }
+      return sc;
+    }
+  } catch (e) {
+    // Not a valid server config
+  }
+  return null;
+}
+
 // ignore: must_be_immutable
 class ComboBox extends StatelessWidget {
   late final List<String> keys;
--- a/flutter/lib/common/widgets/clipboard_server_config.dart
+++ b/flutter/lib/common/widgets/clipboard_server_config.dart
@@ -1,0 +1,45 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_hbb/common.dart';
+import 'package:flutter_hbb/consts.dart';
+import 'package:flutter_hbb/mobile/widgets/dialog.dart' show showServerSettingsWithValue;
+import 'package:flutter_hbb/models/platform_model.dart';
+
+/// Check clipboard for valid server config on startup and show settings dialog if found.
+///
+/// This should be called from the home page's initState.
+/// The [isMounted] callback is used to check if the widget is still mounted after async operations.
+/// The [setState] callback is passed to the dialog for UI updates.
+void checkAndShowClipboardServerConfig({
+  required bool Function() isMounted,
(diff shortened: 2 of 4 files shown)


rustdesk:
# Review

The startup path bypasses existing authorization and deployment restrictions for server settings. It also mishandles hidden iOS relay values and can retain callbacks to disposed widget state.

## [P1] Enforce the network-settings authorization gate

**Location:** `/Users/hzhou/hbbs/flutter/lib/desktop/pages/desktop_home_page.dart:873`

On an installed desktop, ID/Relay changes are gated by `callMainCheckSuperUserPermission()` or the unlock PIN, but this unconditional startup call opens the same editor without either check. An unprivileged user can place a valid config on the clipboard, relaunch, and click OK to invoke `setServerConfig`; hidden or disabled custom-client settings are similarly bypassed. Apply the existing authorization and visibility checks before showing this dialog.

## [P2] Clear hidden relay values on iOS

**Location:** `/Users/hzhou/hbbs/flutter/lib/common.dart:3666`

When an iOS clipboard config contains a relay server, the settings dialog hides that field but retains its controller and submits the hidden value when OK is pressed. The existing `importConfig` path explicitly clears `sc.relayServer` on iOS, so this path can silently apply a value the user cannot inspect or edit; normalize it before returning the decoded config.

## [P2] Check mounted before opening the delayed dialog

**Location:** `/Users/hzhou/hbbs/flutter/lib/mobile/pages/home_page.dart:54-57`

If the home page is disposed during the 500 ms delay or clipboard read, this still opens a global dialog with the disposed state's `setState` callback; accepting the dialog then raises `setState() called after dispose`. Check `mounted` after the asynchronous work and before showing the dialog in both the mobile and desktop implementations.

tianjinli:
Done! All items ([P1] & [P2]) addressed and squashed into the main commit. Ready for review, thanks!

rustdesk:
# Review

The startup path bypasses established settings-visibility policies and mishandles the existing unlock-PIN authorization flow. It also introduces unsolicited iOS paste permission prompts during startup.

## Review comments

### [P1] Honor all server-settings visibility gates

`flutter/lib/common/widgets/clipboard_server_config.dart:17-19`

When a custom build sets `disable-settings=Y` or `hide-network-settings=Y` without also setting `hide-server-settings`, the normal mobile and desktop settings pages omit the ID/Relay editor, but this startup path still opens it from the clipboard. On mobile and portable builds no authorization check follows, so pressing OK writes the server options despite the policy; gate this path with `bind.isDisableSettings()` and `kOptionHideNetworkSetting` as the existing settings pages do.

### [P2] Preserve unlock-PIN authorization for installed clients

`flutter/lib/common/widgets/clipboard_server_config.dart:30-32`

On an installed desktop where Unlock with PIN is configured, the existing network-settings lock uses `checkUnlockPinDialog` instead of requesting administrator privileges. This unconditional admin check means a non-admin operator with the valid unlock PIN can edit network settings normally but cannot use clipboard startup import when the elevation request is denied; reuse the same PIN-versus-admin authorization decision here.

### [P2] Avoid unsolicited iOS pasteboard reads at startup

`flutter/lib/common/widgets/clipboard_server_config.dart:25`

On iOS 16+ when clipboard content originated in another app, this startup `Clipboard.getData` call triggers the system Allow Paste prompt even when the content is unrelated or invalid. Since invalid content is retained, users can receive the prompt again on every cold start; exclude iOS from automatic reads or use a user-initiated paste control.

rustdesk:
> [P2] Avoid unsolicited iOS pasteboard reads at startup

Considering this, I would not consider to merge this PR because of security concern. I am sorry.

PR #15638, opened 2026-07-20
Fix Niri NVIDIA PipeWire capture

kristofferR:
## What changed

- keep the PipeWire buffer pool enabled so DMA-BUF memory reaches the importer
- import modifier-backed DMA-BUF frames through `glupload` using `external-oes`
- convert them to RGBA 2D GL textures and download to system memory before the existing `videoconvert`/appsink path
- remove stale source `VideoMeta` after the direct external-OES import while preserving other metadata

## Why

On Niri with the proprietary NVIDIA driver, the portal supplies block-linear DMA-BUF frames with DRM modifiers. The existing system-memory `videoconvert` pipeline cannot consume those frames. In addition, `pipewiresrc always-copy=true` disables the PipeWire buffer pool while leaving DMA-BUF caps attached to copied memory, which prevents valid GL import.

The result is a successful RustDesk connection that remains stuck waiting for its first image.

## User impact

Wayland hosts using Niri and proprietary NVIDIA can produce frames instead of hanging after connection. The existing BGRx/RGBx appsink contract is unchanged.

## Validation

- `rustfmt --check libs/scrap/src/wayland/pipewire.rs`
- `git diff --check`
- `VCPKG_ROOT=<rustdesk-vcpkg> cargo check -p scrap --features wayland`
- installed and exercised as RustDesk 1.4.9 on CachyOS, Niri, and an RTX 4090 with the proprietary NVIDIA driver
- verified both GStreamer pipelines reached PLAYING and continuously used the direct DMA-BUF external-OES import path
- verified live 5120x1440 HEVC NVENC streaming to the macOS client

This is intentionally a draft because broader testing across other Wayland compositors and GPU drivers would be useful.

### DIFF
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -272,9 +272,10 @@ impl PipeWireRecorder {
         src.set_property("path", &format!("{}", capturable.path))?;
         src.set_property("keepalive_time", &1_000.as_raw_fd())?;
 
-        // For some reason pipewire blocks on destruction of AppSink if this is not set to true,
-        // see: https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/982
-        src.set_property("always-copy", &true)?;
+        // Keep PipeWire's buffer pool enabled so DMA-BUF memory reaches the GL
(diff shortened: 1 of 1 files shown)


kristofferR:
Closing because this work is intended for review in the fork first.

PR #15654, opened 2026-07-22
fix(macos): fall back to installed package prefixes when a vcpkg dependency is missing

### LINKED ISSUE #15653: scrap/build.rs ignores installed macOS system packages when a vcpkg dependency is missing
### Summary

On macOS, `libs/scrap/build.rs` treats `VCPKG_ROOT` as an all-or-nothing source for native dependencies and otherwise falls back only to a hardcoded Homebrew Cellar path under `/opt/homebrew/Cellar`.

In a mixed-source setup, this breaks builds even when the missing package is already installed elsewhere on the machine.

### Reproduced case

- `libvpx` and `aom` were present in `vcpkg`
- `libyuv` was already installed via MacPorts at:
  - `/opt/local/include/libyuv/convert.h`
  - `/opt/local/lib/libyuv.a`
- `VCPKG_ROOT` was set

The current build script still accepted `VCPKG_ROOT`, assumed `libyuv` was there too, and then failed later during bindgen/header resolution instead of falling back to a valid macOS system prefix.

### Steps to reproduce

1. On macOS arm64, install `libvpx` and `aom` in `vcpkg`.
2. Install `libyuv` somewhere outside `vcpkg`, for example MacPorts under `/opt/local`.
3. Set `VCPKG_ROOT` and run `cargo run -- --version`.

### Actual result

The build fails because `libyuv` is looked up only in the `vcpkg` include tree, and the fallback logic does not consider MacPorts or other standard macOS prefixes for that missing package.

### Expected result

If a specific dependency is missing from `vcpkg`, the macOS build script should fall back to installed system prefixes such as MacPorts or local standard prefixes instead of hard-failing on that package.

### Notes

- I checked upstream for existing duplicates before opening this:
  - no matching open issue
  - no matching open PR
- Historical PR `#5539` touched MacPorts support but was closed, unmerged, and is not present in current `master`.

Soundcreates:
Fixes #15653

### What changed

`libs/scrap/build.rs` now validates whether a requested package actually exists in `vcpkg` before accepting that path on macOS.

If a package is missing there, the build script falls back per package to:

1. `/opt/local`
2. `/opt/homebrew/Cellar`
3. `/usr/local`

### Why

The previous logic treated `VCPKG_ROOT` as an all-or-nothing source and otherwise only knew how to look in Homebrew's Cellar.

That breaks mixed-source macOS setups where, for example, `libvpx` and `aom` come from `vcpkg` but `libyuv` is already installed in MacPorts.

### Validation

- `rustfmt --edition 2021 --check libs/scrap/build.rs`
- `git diff --check`
- `env VCPKG_ROOT=/Users/shantanavmukherjee/Desktop/vcpkg cargo run -- --version`

Result: the branch built successfully in the reproduced mixed-source setup and `target/debug/rustdesk --version` printed `1.4.9`.

### DIFF
--- a/libs/scrap/build.rs
+++ b/libs/scrap/build.rs
@@ -74,19 +74,62 @@ fn link_vcpkg(mut path: PathBuf, name: &str) -> PathBuf {
     include
 }
 
+fn include_dir_name(name: &str) -> &str {
+    match name {
+        "libvpx" => "vpx",
+        _ => name,
+    }
+}
+
+fn has_static_lib(lib_dir: &Path, name: &str) -> bool {
+    lib_dir
+        .join(format!("lib{}.a", name.trim_start_matches("lib")))
+        .exists()
+}
+
+fn emit_package(name: &str, lib_dir: &Path, include_dir: PathBuf) -> PathBuf {
+    println!(
+        "cargo:rustc-link-lib=static={}",
+        name.trim_start_matches("lib")
+    );
+    println!("cargo:rustc-link-search={}", lib_dir.to_str().unwrap());
+    println!("cargo:include={}", include_dir.to_str().unwrap());
+    include_dir
+}
+
+fn try_link_vcpkg(path: PathBuf, name: &str) -> Option<PathBuf> {
+    let include = link_vcpkg(path, name);
+    let lib_dir = include.parent()?.join("lib");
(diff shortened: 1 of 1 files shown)


rustdesk:
Thank you very much, but I am sorry I can not accept and spend review time on boring PR like this.

PR #15141, opened 2026-05-29
Update Flutter from 3.24.5 to 3.27.4

dennisameling:
In the journey towards Windows arm64 support (#15139), some dependencies like Flutter will need to be updated. I'd like to do this in small, easily reviewable PRs. I think this version bump, while small, is already meaningful enough to warrant its own, dedicated PR. It looks like quite some patches and workarounds can already be removed because of this.

## Release Notes

* **Chores**
  * Upgraded Flutter SDK from version 3.24.5 to 3.27.4 across all build platforms and CI environments.
  * Standardized dependency resolution in build workflows.

* **Refactor**
  * Updated theme system definitions to align with current Flutter theme APIs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rustdesk/rustdesk/pull/15141?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #15668, opened 2026-07-24
fix(macOS): use modern Flutter to avoid semantics crash

xanderdunn:
## Summary

- build macOS with Flutter 3.44.8 instead of 3.24.5
- reuse the existing modern-Flutter source patch and matching bridge artifact
- keep legacy platforms on Flutter 3.24.5
- make the shared patch script portable to BSD `sed` on macOS

## Root cause

RustDesk 1.4.9's Flutter 3.24.5 engine asserts in
`FlutterViewController::updateSemantics` when a queued semantics update arrives
one run-loop turn after semantics are disabled. This is the exact crash reported
in [flutter/flutter#148279](https://github.com/flutter/flutter/issues/148279).
Flutter removed the assertion in
[flutter/engine#54364](https://github.com/flutter-team-archive/engine/pull/54364)
because this event ordering is expected.

Flutter 3.44.8 contains that fix. The Flutter 3.24.5 macOS framework contains
the `Semantics must be enabled.` assertion; the 3.44.8 framework does not.

## Scope

Windows ARM64 already uses the modern Flutter path. This change shares that path
with macOS. iOS, Android, Linux, Windows x64, and web remain unchanged.

This is deliberately narrower than the earlier repo-wide upgrade attempted in
[rustdesk/rustdesk#15141](https://github.com/rustdesk/rustdesk/pull/15141).

## Validation

- Flutter 3.44.8 `flutter pub get`
- restored the matching `bridge-artifact-flutter-3.44` from a successful
  upstream nightly build
- Flutter analyzer: zero errors (existing warnings remain)
- `flutter test --no-pub`: 19 passed
- patch script: `bash -n`, ShellCheck, and execution with macOS BSD `sed`
- `git diff --check`

### DIFF
--- a/.github/patches/apply_flutter_3.44_source_patches.sh
+++ b/.github/patches/apply_flutter_3.44_source_patches.sh
@@ -2,13 +2,13 @@
 # Applies the Flutter 3.44-only source/pubspec changes on the fly, in CI only.
 #
 # Windows arm64 needs Flutter >= 3.44 (the first stable release shipping an arm64 Dart SDK +
-# engine), which renamed DialogTheme/TabBarTheme -> *Data and needs newer extended_text/
-# google_fonts. Every other platform is still on Flutter 3.24.5, where the old names/versions
-# are required, so these changes are kept OUT of the committed sources and applied here instead.
+# engine), and macOS needs a newer engine to avoid flutter/flutter#148279. Flutter 3.44 renamed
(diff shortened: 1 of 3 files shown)


xanderdunn:
I think this PR is slop and can be closed.
The root issue is a segfault in the macOS app: https://github.com/flutter-team-archive/engine/pull/54364
I just need to wait for the fix to get released.

Issue #15686, opened 2026-07-26
Severe Excel Copy Crash Bug


xrf108:
### Bug Description

RustDesk crashes instantly when copying selected cells in Excel.

This critical bug is preventing all staff from carrying out their work. We hope the developers can fix it as soon as possible.

rivacy Mode 2 and Screen Off Mode work fine on iOS, but throw errors in Windows-to-Windows connections.

### How to Reproduce

Windows-to-Windows: RustDesk crashes on copying Excel selected cells

rivacy Mode 2 and Screen Off Mode work fine on iOS, but throw errors in Windows-to-Windows connections.

### Expected Behavior

RustDesk works fine when copying selected cells in Excel.

Windows-to-Windows: Privacy Mode 2 & Screen Off Mode work as expected

### Operating system(s) on local (controlling) side and remote (controlled) side

Windows-to-Windows

### RustDesk Version(s) on local (controlling) side and remote (controlled) side

rustdesk-1.4.9-x86_64.exe

### Screenshots

<img width="126" height="96" alt="Image" src="https://github.com/user-attachments/assets/d6bea081-15ab-495a-ac31-d13e3ac8ae02" />

fufesou:
Hi @xrf108, thanks for the feedback.

Do you use Tencent WeType IME? Is it similar to the following discussions?

https://github.com/rustdesk/rustdesk/issues/12559
https://github.com/rustdesk/rustdesk/issues/9302 
https://github.com/rustdesk/rustdesk/discussions/12562
https://github.com/rustdesk/rustdesk/issues/15291#issuecomment-4718318927

xrf108:
> 你好，谢谢你的反馈。
> 
> 你用腾讯WeType吗？这和以下讨论类似吗？
> 
> [#12559](https://github.com/rustdesk/rustdesk/issues/12559) [#9302](https://github.com/rustdesk/rustdesk/issues/9302) [#12562](https://github.com/rustdesk/rustdesk/discussions/12562) [#15291（评论）](https://github.com/rustdesk/rustdesk/issues/15291#issuecomment-4718318927)

I'm using WeType, but I'm not sure if it's the culprit. This issue didn't occur when I tried out other remote tools.

rustdesk:
Close since duplicate issue.

xrf108:
> 因为重复问题就关闭了。



> Close since duplicate issue.

Privacy Screen 2 failed to activate. Could this also be caused by WeType?

[...]

Issue #9302, opened 2024-09-09
use wechat screenshot,crash.


shizenghua:
### Bug Description

use wechat screenshot,crash.
if i restart,always crash.
if i copy a text,restart again,it's normal.

### How to Reproduce

use wechat screenshot,crash.

### Expected Behavior

use wechat screenshot,crash.

### Operating system(s) on local side and remote side

windows 11

### RustDesk Version(s) on local side and remote side

1.3.0

### Screenshots

use wechat screenshot,crash.

martint028:
I also encountered the same problem

fufesou:
@shizenghua @martint028 Thanks for your feedback.

It seems the app crashes when "copy&paste" the image data.

But I've tried both `1.3.0` -> `1.3.0` and `1.3.1` -> `1.3.1`, both Windows. It does not crash.

Could you please:
1. Try `1.3.1` -> `1.3.1`
2. Connect, then use "PrtScr" to "copy&paste" screenshot

![image](https://github.com/user-attachments/assets/d27b2db2-3ce2-47b0-bb60-338cc02d8ffe)

martint028:
Is it because of using the clipboard synchronization feature of the WeType input method in WeChat that all my Windows 11 computers experience the same crash? Even after upgrading to version 1.3.1, the issue remains the same.

fufesou:
> Is it because of using the clipboard synchronization feature of the WeType input method in WeChat

Maybe

@martint028 I've tried 1.3.1 -> 1.3.1 again. But failed to reproduce this issue. Could you please give a short video, thanks?

rustdesk:
Close since no response and we can not reproduce.

fufesou:
@shizenghua @martint028 could u please test this build on both sides? https://github.com/fufesou/rustdesk/releases/tag/test-clipboard-win-crash

PR #15692, opened 2026-07-27
fix(clipboard): win, populate file formats

fufesou:
https://github.com/rustdesk/rustdesk/discussions/11134
https://github.com/rustdesk/rustdesk/issues/15686
https://github.com/rustdesk/rustdesk/discussions/12562
https://github.com/rustdesk/rustdesk/issues/9302
https://github.com/rustdesk/rustdesk/issues/15291

## Possible cause

Some apps set the clipboard again after copying.

Then `count = CountClipboardFormats();` is `1`, causing the following crash

```c
		formats[index++].formatId = fsid;
		formats[index++].formatId = fcid;
```

## Changes

Introduce `wf_cliprdr_populate_file_formats()` to always return 2 formats for `CF_HDROP`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows clipboard interoperability with safer file transfers, format handling, and data validation.
  * Strengthened clipboard resource cleanup, ownership management, and failure handling for greater stability.
  * Improved clipboard synchronization when connections close or clipboard contents are cleared.
  * Added safeguards for malformed clipboard data, invalid ranges, encoding issues, and incomplete transfers.
* **Diagnostics**
  * Updated clipboard processing logs to display viewing and stopping states more clearly.

This PR fixes a buffer-overflow crash in the Windows clipboard file-transfer path where `CountClipboardFormats()` could return 1 (after a secondary clipboard write by another app), causing two format IDs to be written into a single-element array. The fix ensures at least `WF_CLIPRDR_FILE_FORMAT_COUNT` (2) slots are always allocated before populating the descriptor and contents format IDs.

- **Core crash fix**: `numFormats` is now clamped to a minimum of 2 before the `formats[]` array is allocated, preventing the out-of-bounds write for `CF_HDROP` clipboard data.
- **Threading**: A new `SRWLOCK format_map_lock` protects format-map replacement by Tokio callbacks against concurrent reads from the clipboard STA thread; `get_remote_format_id` acquires a shared lock, and `wf_cliprdr_server_format_list` acquires an exclusive lock with goto-based cleanup on every error path.
- **Clipboard ownership/cleanup**: `wf_do_empty_cliprdr` is refactored to post `OLE_EMPTYCLIPBOARD` to the window message queue instead of calling COM/clipboard APIs directly (avoiding re-entrancy from a `WM_RENDERFORMAT` handler), and several memory leaks in error paths of `wf_cliprdr_server_format_list` are plugged. Also fixes a pre-existing log label/argument mismatch in `io_loop.rs`.
</details>

Safe to merge — the changes are targeted fixes for real crashes and resource leaks with careful lock-based thread safety and proper COM reference counting throughout.

The primary change (clamping numFormats to at least 2 before allocating the format array) is a minimal, clearly correct fix for the documented crash. The supporting changes — the SRWLOCK, the message-queue-based wf_do_empty_cliprdr, the FindClose leak fix, and the DROPFILES bounds checking — all follow established patterns in the file and are well-guarded. No pre-existing invariants appear broken. The only notable gap is the missing `static` on wf_cliprdr_format_data_size_valid, which is a link-time style issue rather than a runtime concern.

**Files Needing Attention:** No files require special attention beyond the style nit in wf_cliprdr.c.
</details>

| Filename | Overview |
|----------|----------|
| libs/clipboard/src/windows/wf_cliprdr.c | Core bug fix: ensures at least 2 format slots are allocated for CF_HDROP to prevent a buffer-overflow crash when CountClipboardFormats() returns 1. Also adds SRWLOCK for format-map cross-thread safety, refactors wf_do_empty_cliprdr to post to the window message queue (avoiding COM calls in the wrong thread), fixes FindClose handle leaks in directory traversal, adds DROPFILES bounds-checked parsing, and plugs several memory leaks in format-list error paths. One minor style issue: wf_cliprdr_format_data_size_valid is missing `static`. |
| src/client/io_loop.rs | Fixes a pre-existing log-format bug where the positional label order (stop/is_stopping_allowed/view_only) did not match the argument order (view_only/stop/is_stopping_allowed), causing misleading debug output. |

</details>

```mermaid
sequenceDiagram
    participant App as External App
    participant CB as Windows Clipboard
    participant STA as Clipboard STA Thread
    participant Tokio as Tokio/RDP Callback
    participant Peer as RDP Peer

    App->>CB: SetClipboard (CF_HDROP, 1 format)
    CB->>STA: WM_DRAWCLIPBOARD
    STA->>STA: "cliprdr_send_format_list()<br/>count=CountClipboardFormats()=1<br/>numFormats=max(count,2) [fix]"
    STA->>STA: alloc formats[2], populate fsid+fcid
    STA->>Peer: Send CLIPRDR_FORMAT_LIST (2 formats)

    Peer->>Tokio: wf_cliprdr_server_format_list()
    Tokio->>Tokio: AcquireSRWLockExclusive(format_map_lock)
    Tokio->>Tokio: clear_format_map + populate
    Tokio->>Tokio: ReleaseSRWLockExclusive(format_map_lock)
    Tokio->>STA: PostMessage(OLE_SETCLIPBOARD)

    STA->>STA: wf_create_file_obj() → IDataObject
    STA->>CB: OleSetClipboard(IDataObject)

    Peer->>Tokio: wf_cliprdr_server_format_data_request()
    STA->>Tokio: "get_remote_format_id()<br/>AcquireSRWLockShared(format_map_lock)"

    note over Peer,Tokio: Connection closes
    Tokio->>STA: "wf_do_empty_cliprdr()<br/>PostMessage(OLE_EMPTYCLIPBOARD)"
    STA->>STA: "wf_empty_cliprdr_on_sta()<br/>OleIsCurrentClipboard + EmptyClipboard"
```
</details>

<sub>Reviews (5): Last reviewed commit: ["fix(clipboard): missing frees"](https://github.com/rustdesk/rustdesk/commit/d907144bc152cf8531db258f96d9a45a10fc5da2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47862132)</sub>


rustdesk:
• The primary crash fix appears sound, but the new capability-count validation rejects supported multi-set capability messages and can regress clipboard negotiation with compatible
  peers.

  Review comment:

  - [P2] Allow multiple clipboard capability sets `/Users/hzhou/rustdesk/libs/clipboard/src/windows/wf_cliprdr.c:2690-2692` When a peer sends a capabilities PDU containing more than one set, this guard returns `ERROR_INTERNAL_ERROR` before examining them, even though the protocol and the loop below support multiple sets. This breaks interoperability with peers that send a general capability set alongside another set; validate the pointer whenever the count is nonzero and apply a reasonable upper bound instead of limiting the count to one.

rustdesk:
# Review: rustdesk/rustdesk#15692 — harden Windows clipboard (cliprdr) memory handling

**Base:** `master` · **Size:** 14 commits, 2 files, +433 / −170
**Files:** `libs/clipboard/src/windows/wf_cliprdr.c` (all of it) + a one-line log reorder in
`src/client/io_loop.rs`

---

## Overview

A systematic memory-safety pass over the Windows RDP clipboard redirection channel. The work
falls into three groups: COM interface contracts and their error paths, bounds-checking of
peer-influenced data, and object lifetime across the clipboard STA thread and the Tokio callback
thread.

## What it fixes

These were verified individually against the pre-change code — they are real defects, not
theoretical ones:

| Defect | Location | Class |
|---|---|---|
| `DVTARGETDEVICE` is variable-length (`tdSize`), but was allocated as `sizeof()` and copied whole | `cliprdr_format_deep_copy` | **heap OOB read + write** |
| DROPFILES name list walked with no bounds, trusting only the double-NUL terminator | `..._format_data_request` | **OOB read** |
| `calloc(count)` followed by two unconditional writes; `count` can be 0 or 1 | `cliprdr_send_format_list` | **heap OOB write** |
| Function pointer checked for NULL, then called anyway | `cliprdr_send_response_filecontents` | **NULL deref** |
| Error path `free()`s members, then calls `_Delete()` which frees them again | `CliprdrDataObject_New` | **double free** |
| `hmem` not cleared after `SetClipboardData` takes ownership; freed again later | `cliprdr_proc` (×2) | **double free** |
| `m_nIndex + (LONG)celt` signed overflow | `CliprdrEnumFORMATETC_Skip` | out-of-range index |
| `dataObj` not released when `GetData` fails | `..._format_data_request` | reference leak |
| Early returns skip `FindClose(hFind)` | `wf_cliprdr_traverse_directory` | handle leak |
| `(int)GlobalSize()` truncation (`size` is `size_t`) | `..._format_data_request` | truncation |
| `clipboard->data_obj` passed to OLE with no reference and no lock | `is_set_by_instance` | **use-after-free** |

The DROPFILES rewrite is particularly solid: `remaining` is tracked in bytes, lengths come from
`StringCchLengthW/A` with an explicit cap, the wide path validates `pFiles % sizeof(WCHAR)`
alignment, and the result is accepted only once the empty terminator is actually reached. The
`remaining -= (len + 1) * sizeof(WCHAR)` step cannot underflow, since a successful
`StringCchLengthW` guarantees `len < cchMax`.

The new locking is coherent. The only writer is `wf_cliprdr_server_format_list` (callback thread,
serialized by ContextSend) holding the lock exclusively; the STA-side reader `get_remote_format_id`
— reached from `CliprdrDataObject_GetData` and `cliprdr_proc` — takes it shared. The two unlocked
reads inside the callback (`file_transferring`, and the `format_ids` fill) are on the same thread
as the writer. Nothing inside the locked region re-enters the lock, so the non-recursive SRWLOCK
cannot self-deadlock.

---

## Suggestions

### 1. `wf_cliprdr_server_format_list_fail()` releases a lock it did not acquire

```c
/* Called only while format_map_lock is held exclusively. */
static UINT wf_cliprdr_server_format_list_fail(wfClipboard *clipboard)
{
        ...
        ReleaseSRWLockExclusive(&clipboard->format_map_lock);
```

All eight call sites are correct today. The concern is shape, not behaviour: unlocking inside a
callee is held together only by a comment, so the day someone adds a ninth early return and
forgets to route it through this helper, the lock leaks and every later STA reader blocks
forever — with nothing to catch it at compile time or in testing.

Restructuring to a single `goto fail;` exit would put the acquire and release in the same
function where they can be paired by eye.

### 2. Rejecting `cCapabilitiesSets > 1` outright may be too strict

```c
capabilities->cCapabilitiesSets > 1 || ...
    return ERROR_INTERNAL_ERROR;
```

The loop below already iterates over `cCapabilitiesSets` and handles more than one.
MS-RDPECLIP permits multiple capability sets, so this turns "extra sets are ignored" into
"the whole message fails".

If the peer is always RustDesk this is harmless, but it is an interop-narrowing change that the
memory-safety goal doesn't require. Worth either a note on why, or relaxing it to bound the
count and keep iterating.

### 3. `size_t size` narrowed into `response.dataLen` (UINT32) — nit

On the non-file path `size = GlobalSize(hClipdata)` is unbounded, so above 4 GB `dataLen`
truncates while `buff` still holds the full allocation. An absurd input, and already far better
than the previous `(int)` cast — but since this pass is tightening types anyway, an explicit
cap would be tidier.

### 4. Test coverage: none

433 lines of memory-safety-critical C with no tests. COM and clipboard code is genuinely hard to
unit test — but **the DROPFILES parser is a pure function**: bytes in, file names out. Extracting
it (`parse_dropfiles(const void *buf, SIZE_T size, ...)`) would make it unit-testable and
fuzzable, and it is exactly the code that had a real OOB read and that handles data the peer can
influence.

Highest-value cases to cover: `pFiles` past the end, unaligned `pFiles`, missing terminator, both
`fWide` branches, and `nFiles` reaching `WF_CLIPRDR_MAX_STREAMS`.

---

## Risks

The largest semantic change is `wf_do_empty_cliprdr` switching from synchronous work to a
`PostMessage` onto the STA: callers go from "cleared on return" to "queued". The stated rationale
is sound — it keeps `OpenClipboard` out of a `WM_RENDERFORMAT` handler and releases ContextSend
promptly — but it does assume no caller depends on completion.

Worth exercising on real hardware: repeated copy/disconnect cycles across multiple connections,
confirming the connID match (`connID != 0 && instance->m_connID != connID` skips) never leaves a
stale clipboard behind.

## Verdict

High quality. The defects fixed are real, the bounds checks are carefully constructed, and the
COM contract work (rollback in `Next`, `New` returning `HRESULT`) follows the interface rules
properly. Suggestions 1 and 2 are small; 4 is worth its own follow-up.

fufesou:
https://github.com/rustdesk/rustdesk/pull/15692#issuecomment-5189846025

> [P2] Allow multiple clipboard capability sets /Users/hzhou/rustdesk/libs/clipboard/src/windows/wf_cliprdr.c:2690-2692

This is not a P2 in the current integration.

1. RustDesk’s ClipboardFile transport has no capabilities message, so `ServerCapabilities` has no production call path.
2. `capabilitySets` is an unbounded pointer to variable-length sets; applying only a count limit would not make `capabilitySets[index]` safe.
3. The current `> 1` validation is intentional fail-closed behavior. Supporting multiple sets would require a length-aware representation and parser rather than relaxing this guard.

fufesou:
https://github.com/rustdesk/rustdesk/pull/15692#issuecomment-5189966103

> 1. `wf_cliprdr_server_format_list_fail()` releases a lock it did not acquire

Removed in ecbb4dd, `goto fail` is used instead.

> 2. Rejecting `cCapabilitiesSets > 1` outright may be too strict

Not an issue. https://github.com/rustdesk/rustdesk/pull/15692#issuecomment-5190126037

> 3. `size_t size` narrowed into `response.dataLen` (UINT32) — nit

Addressed in ecbb4dd

PR #15772, opened 2026-08-05
fix(linux): keep-awake does nothing on a GNOME login screen

### LINKED ISSUE #15741: keep-awake does nothing on a gnome login screen (linux/wayland)
on linux, keeping the host awake during an incoming session goes through keepawake:
org.freedesktop.ScreenSaver on the session bus for the display part, logind on the
system bus for idle and sleep.

that name does not exist on a gdm greeter bus:

    $ busctl --user call org.freedesktop.ScreenSaver /org/freedesktop/ScreenSaver \
        org.freedesktop.ScreenSaver Inhibit ss "RustDesk" "incoming session"
    Call failed: The name org.freedesktop.ScreenSaver was not provided by any .service files

two consequences:

- keepawake asks for the screensaver inhibit first and abandons the whole request if it
  fails, so the logind idle and sleep inhibits are lost with it. on a host whose logind
  IdleAction is not the default, the machine can suspend during an active session.
- WakeLock::new discards the error with .ok(), so nothing reports any of it.

on that same greeter org.gnome.SessionManager is on the bus and its idle inhibit works.
with it held the output was still on after a couple of minutes of idle. without it the
compositor disabled the output after about half a minute. disabled, not blanked: the
connector loses its crtc, so there is nothing left for any capture backend to read either.

i have a fix ready if you want it, one file in src/platform/linux.rs: log the keepawake
error instead of dropping it, re-ask keepawake without the display part so the logind
inhibits are still taken, and get the display half from org.gnome.SessionManager with
org.freedesktop.PowerManagement as a fallback. only the failure path changes, a session
where the screensaver inhibit already works is untouched.

branch: https://github.com/fxd0h/rustdesk/tree/fix/gnome-greeter-idle-inhibit

verified end to end with a live client on the greeter. the PowerManagement entry is not
verified, i have no kde or xfce box, and where nothing answers it logs every name tried.

happy to open a PR.

fxd0h:
Fixes #15741. One file, one commit, independent of any other work.

## The problem

`WakeLock` keeps the display on through the `keepawake` crate, which inhibits `org.freedesktop.ScreenSaver`. That is the right interface inside a desktop session.

It is **not present on a GDM login screen**. Measured on a GNOME/Wayland greeter, that bus name answers `was not provided by any .service files` and cannot be activated either, so `keepawake::Builder::create()` fails as a whole and the wake lock is silently a no-op. The screen then blanks under a live incoming connection, which on Wayland means the compositor eventually disables the connector and there is nothing left to capture.

## The fix

When `keepawake` fails, fall back to a session-bus inhibit, trying the interfaces a greeter actually offers:

- `org.gnome.SessionManager.Inhibit` with flag 8 (idle only). **Verified on a GDM greeter:** with the inhibit held the output stayed on 129.9 s against 30.3 s without it, which matches the greeter idle timer exactly.
- `org.freedesktop.PowerManagement.Inhibit`, which is what powerdevil and xfce4-power-manager implement. **Not verified** — no KDE or XFCE machine here. It costs one failed method call where it is absent, and the log names every interface tried, so a user report is enough to learn what a given desktop wants.

Only the idle flag is requested. The others (logout, switch-user, suspend) would take away actions the person at the machine should keep.

`org.freedesktop.ScreenSaver` is deliberately not in the list: if it had answered, this fallback would not have run.

## Scope

Only reached when the existing path fails, so a normal desktop session behaves exactly as before. No new dependency: the D-Bus client is already in the tree.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux wake-lock reliability with fallback inhibition methods when the primary request fails.
  * Added support for desktop session idle inhibition through GNOME and freedesktop power-management services.
  * Wake locks now report inhibition errors and cleanly release all active inhibitors.

The PR adds Linux session-bus idle-inhibit fallbacks for environments where the existing ScreenSaver-based wake lock fails.
- Tries GNOME SessionManager and freedesktop PowerManagement inhibition in order.
- Retains separate logind idle/sleep protection by retrying keepawake without display inhibition.
- Holds and releases the selected session inhibitor for the wake-lock lifetime.
</details>

The PR appears safe to merge.

No blocking failure remains within the eligible follow-up review scope.
</details>

| Filename | Overview |
|----------|----------|
| src/platform/linux.rs | Adds an ordered session-bus idle-inhibition fallback and preserves idle/sleep inhibition when the primary display-inhibit request fails. |

</details>

```mermaid
flowchart TD
    A[Create Linux wake lock] --> B{keepawake succeeds?}
    B -->|Yes| C[Hold keepawake handle]
    B -->|No| D[Retry idle or sleep inhibit without display]
    D --> E{Display inhibit requested?}
    E -->|Yes| F[Try GNOME SessionManager]
    F -->|Refused| G[Try freedesktop PowerManagement]
    F -->|Accepted| H[Hold system and session inhibitors]
    G -->|Accepted| H
    G -->|Refused| I[Hold system inhibitor if available]
    E -->|No| I
    C --> J[Drop WakeLock]
    H --> J
    I --> J
    J --> K[Release handles and close D-Bus connection]
```
</details>

<sub>Reviews (2): Last reviewed commit: ["fix(linux): stop losing every inhibitor ..."](https://github.com/rustdesk/rustdesk/commit/1dbfed1b9cdbb1f50daad0eabe67e87095d8beb9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50479388)</sub>

### DIFF
--- a/src/platform/linux.rs
+++ b/src/platform/linux.rs
@@ -1982,18 +1982,167 @@ mod desktop {
     }
 }
 
-pub struct WakeLock(Option<keepawake::AwakeHandle>);
+/// A session-bus idle-inhibit interface, tried in order; the first that answers wins.
+/// `org.freedesktop.ScreenSaver` is absent on purpose: that is the one `keepawake` already tried.
+struct SessionInhibitTarget {
+    dest: &'static str,
+    path: &'static str,
+    iface: &'static str,
+    /// GNOME takes `(app_id, xid, reason, flags)`; PowerManagement takes `(app, reason)`.
+    gnome_shape: bool,
+    uninhibit: &'static str,
+}
+
+/// `org.gnome.SessionManager.Inhibit` flag 8 = idle only; logout/switch-user/suspend would take
+/// away actions the person at the machine should keep.
+const GNOME_INHIBIT_IDLE: u32 = 8;
+
+const SESSION_INHIBIT_TARGETS: &[SessionInhibitTarget] = &[
+    // Measured on a GDM greeter: output held 129.9 s with the inhibit, 30.3 s without.
+    SessionInhibitTarget {
+        dest: "org.gnome.SessionManager",
+        path: "/org/gnome/SessionManager",
+        iface: "org.gnome.SessionManager",
+        gnome_shape: true,
+        uninhibit: "Uninhibit",
+    },
+    // What powerdevil and xfce4-power-manager implement. NOT tested here; costs one failed call
+    // where absent, and the log below names every interface tried.
+    SessionInhibitTarget {
+        dest: "org.freedesktop.PowerManagement",
+        path: "/org/freedesktop/PowerManagement/Inhibit",
+        iface: "org.freedesktop.PowerManagement.Inhibit",
+        gnome_shape: false,
+        uninhibit: "UnInhibit",
+    },
+];
+
+/// Idle inhibit for the case `keepawake` cannot serve: it inhibits `org.freedesktop.ScreenSaver`,
+/// which a GDM greeter bus neither provides nor can activate, so its `create()` fails outright.
+/// The connection is kept because the inhibit is bound to it: dropping it releases the inhibit.
+struct SessionIdleInhibit {
+    conn: dbus::blocking::Connection,
+    target: &'static SessionInhibitTarget,
+    cookie: u32,
+}
+
+impl SessionIdleInhibit {
+    fn new(reason: &str) -> Option<Self> {
+        let conn = match dbus::blocking::Connection::new_session() {
+            Ok(conn) => conn,
+            Err(err) => {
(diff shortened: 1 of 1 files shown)


fxd0h:
the slop flag was fair. this branch predates the comment cleanup that went into #15420, so it still carried the dense style: 23 percent of the added lines were comments, against 7 percent there after that pass.

condensed to 13 percent in `1dbfed1b9`. what went is the narration and the rationale; every measured fact stayed, and I checked that rather than assumed it: the 129.9 s vs 30.3 s greeter measurement, the flag-8 choice, why ScreenSaver is not in the list, which desktops implement the PowerManagement interface, that the logind inhibits live on the system bus, and that the session manager ties the inhibit to the caller bus name are all still present. code unchanged, still one file.

PR #15774, opened 2026-08-06
Custom client no rebuild

rustdesk:
# Build Windows custom clients without recompiling

Paired with **rustdesk/hbbs#custom-client-no-rebuild**. Both must land together, and this one first — see [Rollout](#rollout).

## Problem

Every custom client ran a full `cargo build` plus a brotli pass over the whole ~100MB payload, though customers differ only by `custom.txt`, a few icons, an app name and a re-branded runner exe.

The waste is structural rather than incidental. `data.bin` is baked into the packer with `include_bytes!`, so a payload that is byte-identical across customers has to be recompressed and relinked for each one. The msi side is worse: `preprocess.py` rewrites the WiX sources per customer, which forces a clean rebuild of the native CustomActions dll and a fresh cabinet for the entire payload.

## Approach

Build templates once per release; assemble each customer from them.

**exe.** The packer now also reads an `RDPKG` RCDATA resource holding a second payload in the same format as `data.bin`, and folds it over the compiled-in one. A build injects that resource into a prebuilt template instead of running cargo. The executable to launch comes from the package trailer, and the extraction directory follows its stem, which replaces the `sed` of `APP_PREFIX`.

Where the executable itself is not customized — sciter x86 has never supported icon customization — it stays in the generic payload and is only renamed, so one merge covers both shapes.

**msi.** CI builds an additional msi with the placeholder app name `RDAPPNAM`, unsigned, and ships it in the unsigned tarball. A build rebrands that database instead of running WiX. The files a customer replaces live in their own cabinet, so a rebrand rebuilds a few hundred KB rather than recompressing the payload.

`custom.txt` keeps being written to disk next to the app throughout. That is what the client reads at startup and what the updater stages, so a customization survives an upgrade to a stock build — the property the whole design is built around.

## Changes

| Area | What |
|---|---|
| `libs/portable` | Load a per-customer payload from a PE resource; merge and rename rules; delete files a later package drops |
| `libs/portable/generate.py` | `--package` and `--exclude-exe`; fixes `os.curdir` being restored instead of the previous working directory |
| `res/msi/preprocess.py` | `--template` mode: second cabinet for per-customer files, conditional components for optional branding assets |
| `res/msi/CustomActions` | Printer and port names arrive through `CustomActionData` instead of being compiled in, so one dll serves every custom client |
| `.github/workflows` | Ship `windows-aarch64` in the unsigned tarball; build and publish the msi templates |

## Notable details

**ARM.** CI already built Windows arm64 but never shipped it in the unsigned tarball. It does now, and the arm64 msi template is built on the native arm64 runner — which is what makes ARM custom clients possible at all, since `preprocess.py` runs the packaged exe to read its version and build date, and the build agents are x64. Patching runs no exe, so an x64 agent can rebrand the arm64 template.

**The placeholder is eight characters and a valid 8.3 name.** WiX derives a short name for anything longer, and a patch cannot rewrite a truncated placeholder, so short names would keep pointing at it. `RustDesk` itself cannot serve as the placeholder: it also names payload that must never be renamed — `librustdesk.dll`, `drivers\RustDeskPrinterDriver` — and `preprocess.py` is safe today only because it never replaces globally.

**Branding assets are optional and a stock build ships none of them.** There is no `icon.ico`, `icon.png` or `logo*.png` in `flutter/assets`, only `icon.svg`. The template therefore carries placeholders so the File rows exist, and each installs only when the customer supplied one — otherwise a customer without a logo would get a placeholder image, where today a missing asset means no logo at all.

**Compression.** Both `-l 5` settings can go to 11: the payload is now compressed once per release rather than once per customer, so every customer's download gets smaller.

## Public msi

Behaviour is unchanged; the file is not byte-identical. `preprocess.py --template` is fully guarded, so the shipped msi keeps its single cabinet and gains no conditions. The one real difference is the printer custom action: for a stock build `ProductName` is `RustDesk`, so the printer and port are named `RustDesk Printer` exactly as the compiled-in strings did.

## Verification

Verified on macOS, without Windows:

- Unit tests for the payload format, malformed blobs, the rename and override rules, backslash paths (which `generate.py` emits on Windows), and the path guard on the dropped-file sweep.
- The cross-language contract: the Rust parser reads a real `generate.py` blob — paths, hex md5, brotli round-trip.
- End to end with the Windows-only steps stubbed, across x86_64, aarch64 and x86, with the real artifacts fed back through the merge.

**The msi path has never run on Windows.** It needs a build machine and the first CI-built template. `MsiFileHash`/`FileVersion` behaviour, `msidb`/`makecab`, and whether the CustomActions change compiles are all open.

The change set went through five rounds of adversarial review; thirty findings were fixed, and the fifth round found nothing in the core.

## Rollout

1. Merge here, run a nightly, confirm the unsigned tarball contains `msi-template/` and `windows-aarch64/`.
2. On the build agent, patch a template by hand and inspect the result with Orca.
3. Then merge the hbbs side, which consumes those artifacts.
4. Real-machine matrix before any customer sees it: fresh install, uninstall, same-brand upgrade, repair, silent install, and **two differently branded clients installed side by side then each uninstalled** — the case component reference counting gets wrong.

https://claude.ai/code/session_01Q7fBdTwziR5BHTkSz7Tzcm

## Summary by CodeRabbit

- **New Features**
  - Added unsigned Windows MSI templates for x64 and ARM64 builds.
  - Added support for customer-specific branding and files in MSI templates.
  - Portable packages can exclude executables and include customized contents.
  - Remote printers now use the application name for printer and port identification.
  - Windows build archives now include ARM64 binaries and MSI templates.

- **Bug Fixes**
  - Improved portable package extraction, path validation, stale-file cleanup, and malformed-package handling.
  - Added safer support for customized application executables and package files.

The PR introduces reusable Windows executable and MSI templates so customer-specific clients can be assembled without recompiling the full application.
- Adds an RCDATA customer package that is merged with the portable executable’s embedded payload.
- Adds customer-specific extraction directories and stale branding-file cleanup.
- Builds MSI templates with a separate cabinet for customer-specific files and runtime-configured printer names.
- Publishes ARM64 Windows artifacts and MSI templates in the unsigned release archive.
</details>

The PR appears safe to merge based on the eligible follow-up review findings.

No blocking failure remains in the available follow-up scope.
</details>

| Filename | Overview |
|----------|----------|
| libs/portable/src/bin_reader.rs | Adds safe parsing of an optional RDPKG resource and merges customer files over the embedded portable payload. |
| libs/portable/src/main.rs | Selects extraction directories from the packaged executable name and records customer files for stale-asset cleanup. |
| libs/portable/generate.py | Adds package-only generation and executable exclusion while correctly restoring the original working directory. |
| res/msi/preprocess.py | Adds template generation with a dedicated cabinet and conditional components for optional branding assets. |
| res/msi/CustomActions/RemotePrinter.cpp | Derives printer and port names from the MSI product name while retaining the common RustDesk driver identity. |
| .github/workflows/flutter-build.yml | Builds and publishes x64 and ARM64 MSI templates and includes ARM64 binaries in the unsigned archive. |

</details>

```mermaid
flowchart LR
    A[Release build] --> B[Portable executable template]
    A --> C[MSI template]
    D[Customer files and branding] --> E[RDPKG resource injection]
    B --> E
    E --> F[Custom portable client]
    D --> G[Rebuild customer cabinet and patch MSI]
    C --> G
    G --> H[Custom MSI client]
```
</details>

<sub>Reviews (9): Last reviewed commit: ["fix, remove useless file"](https://github.com/rustdesk/rustdesk/commit/0d846dca6699cde46d6e3d6354ea61e346d16fba) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50703686)</sub>


fufesou:
@coderabbitai review

fufesou:
@coderabbitai review

fufesou:
@codex review

fufesou:
@codex review

fufesou (approved):

fufesou (approved):

rustdesk:
I’d **request changes** on PR #15774 before merge. The portable side looks substantially reviewed and tested; the MSI side still has a merge-blocking validation gap.

### Findings

1. **[P1] The new MSI path is not exercised by PR CI**

   In `.github/workflows/flutter-build.yml`, `Build msi template for custom clients` is guarded by `if: env.UPLOAD_ARTIFACT == 'true'`. On the current PR head, both the x64 and ARM64 Windows jobs report **Build msi**, **Build msi template for custom clients**, and template upload as **skipped**.

   That means the changes to `preprocess.py`, WiX sources, and `CustomActions` can merge without even being compiled together. This is particularly important because the PR itself says the MSI path “has never run on Windows” and explicitly lists CustomActions compilation and MSI metadata behavior as open questions.

   I would change the workflow so **building/validating the MSI template runs on PRs**, while only artifact upload/publishing remains behind `UPLOAD_ARTIFACT`. At minimum, run:

   * `preprocess.py --template`
   * `msbuild -t:clean`
   * `msbuild` for x64 and ARM64 where supported

2. **[P1, cross-repo blocker] Verify that the patcher replaces every customer-specific MSI identity**

   `preprocess.py` historically makes custom MSI identity customer-specific. In template mode, however, this now happens once for the placeholder `RDAPPNAM`: `gen_pre_vars()` generates the `UpgradeCode`, `gen_upgrade_info()` generates another upgrade GUID, and `replace_component_guids_in_wxs()` replaces every Component GUID.

   Consequently, every customer produced from one template will inherit the **same component identities unless the hbbs transformer rewrites them**. That is exactly the situation where two branded clients installed side-by-side can corrupt Windows Installer component reference counting. The PR itself calls that scenario out as something that must be tested.

   Before approval, I’d require the paired transformer to demonstrably rewrite at least the effective:
   `ProductCode`, `UpgradeCode`, generated `Upgrade.Id` values, and customer-specific `Component.ComponentId` GUIDs. An integration test should create two brands from one template, install both, uninstall each independently, and verify the survivor.

3. **[P1/P2, cross-repo blocker] Verify MSI metadata is regenerated when cab2 contents are replaced**

   The template deliberately creates placeholder `custom.txt`/branding assets and puts those plus the application executable into the customer cabinet.  `preprocess.py` creates real MSI `File` rows for those placeholders and uses `Checksum="yes"`.

   Rebuilding `cab2.cab` alone is therefore insufficient: the transformer needs to update the corresponding MSI database metadata for the replacement files—particularly `FileSize`, version/language data where applicable, and `MsiFileHash` for unversioned files. Otherwise install costing, repair, overwrite/version rules, or integrity checks can describe the placeholder instead of the actual customer file.

   This isn't hypothetical uncertainty: the PR description explicitly says **`MsiFileHash`/`FileVersion` behaviour is still open**.  I’d treat that as a pre-merge contract test, not a post-merge rollout item.

### Other observations

The portable changes are much stronger. The parser now bounds accesses, distinguishes missing versus malformed `RDPKG`, tests merge/rename behavior, and the stale-package cleanup has a lexical path-escape guard plus retry handling.   I didn’t find a portable-side defect strong enough to block this PR.

One additional reason to rerun the MSI work now: PR #15891 merged on **August 22, 2026**, after #15774’s current August 18 head, and significantly changed the same MSI registry/preprocessing area. GitHub can currently synthesize a clean merge, and the combined `preprocess.py` contains both sets of changes, but “clean merge” does not validate the resulting MSI semantics.

**Verdict: Request changes.** I’d be comfortable reconsidering once the MSI template actually builds in PR CI and there is an end-to-end transformer test covering MSI identity + file metadata + two-brand side-by-side install/uninstall.

fufesou:
https://github.com/rustdesk/rustdesk/pull/15774#issuecomment-5436497940

### 1. MSI build coverage

**Conclusion: This is not a valid blocker for the intended release workflow, and no workflow change is needed.**

The Flutter Nightly Build workflow covers the MSI path. `flutter-nightly.yml` calls `flutter-build.yml` with `upload-artifact: true`. For both x86_64 and ARM64, it builds the regular MSI, resets `res/msi`, runs `preprocess.py --template`, cleans the MSI solution, and builds the template.

The `rustdesk-unsigned-msi-template-*` artifacts are collected under `msi-template/` in `rustdesk-1.4.9-unsigned.tar.gz`, together with the unsigned Windows application directories. This is the archive used by the hbbs template update. A failure in `preprocess.py`, CustomActions, or WiX fails the Nightly job and prevents the archive from being published. The Nightly run for the PR head is the build validation for this path.

### 2. MSI identities

**Conclusion: This is partly valid. The identity rewrite is already implemented, so no code change is needed; the side-by-side Windows validation is still missing.**

The paired hbbs branch already rewrites the MSI identities for every generated customer package. `new_product_guids()` creates a new `ProductCode`, derives the `UpgradeCode` from the final app name, and rewrites the effective `Upgrade` table rows to use that code. `new_component_guids()` replaces every non-empty `ComponentId`, and `new_package_code()` writes a new package code to Summary Information. All three are called before the patched database is committed.

One detail here is that the GUID generated by `gen_upgrade_info()` is not part of the effective package. The `UpgradesFile` reference is commented out in `Package.wxs`; the active rows come from `<MajorUpgrade>`. Those are the rows rewritten by the hbbs patcher.


### 3. MSI file metadata

**Conclusion: This is partly valid. The metadata update is already implemented, so no code change is needed; validation with a real patched MSI is still missing.**

The hbbs patcher does not only replace `cab2.cab`. Before rebuilding the cabinet, `update_file_metadata()` updates `FileSize` and `File.Version` from the replacement files. For unversioned files it recalculates and inserts or updates `MsiFileHash`; for a versioned file it removes any hash row. It also refreshes the installed-size registry value before committing the database and replacing the cabinet stream.

`File.Language` is left unchanged because the only versioned replacement is copied from the same stock executable and keeps the same language. `custom.txt` and the branding assets are unversioned and use `MsiFileHash` instead.

rustdesk:
I re-reviewed the PR at the current head, **`47a8127`**. The previous issue with reconfiguring the Rust capture list has been fixed: when topology changes in **All Displays** mode, the code now calls `sessionSwitchDisplay()` with the new display indexes. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/15974/commits)][1])

I still see **one P2 issue** in the new fix.

### [P2] `reconcileDisplays()` preserves frames for the wrong monitor when display indexes shift

The new code does:

```dart
_pi.displays.value = newDisplays;

if (_pi.currentDisplay == kAllDisplayValue) {
  parent.target?.imageModel.reconcileDisplays(newDisplays.length);
  ...
}
```

But `reconcileDisplays()` only removes image entries whose index is outside the new display count:

```dart
void reconcileDisplays(int newDisplayCount) {
  final invalidKeys =
      _images.keys.where((k) => k >= newDisplayCount).toList();

  for (final k in invalidKeys) {
    final img = _images.remove(k);
    if (img != null) {
      _scheduleDispose(img);
    }
  }
}
```



That assumes an existing display index still refers to the same physical monitor after a topology change. It doesn't necessarily.

For example:

```text
Before:
index 0 -> Monitor A -> image A
index 1 -> Monitor B -> image B

Monitor A is unplugged.

After:
index 0 -> Monitor B
```

`reconcileDisplays(1)` removes `_images[1]`, but **keeps `_images[0]`**, which still contains Monitor A's last frame.

The painter then does exactly this:

```dart
final img = images[i];
final display = displays[i];
```

so the stale frame from Monitor A is drawn using Monitor B's new geometry until a fresh frame for display 0 arrives. 

The same problem is even clearer for a same-count topology change:

```text
[A, B] -> [B, C]
```

`newDisplayCount` stays `2`, so `reconcileDisplays(2)` removes **nothing**. Initially:

```text
images[0] = old A, displays[0] = new B
images[1] = old B, displays[1] = new C
```

Both displays can therefore briefly show the wrong monitor.

There is a second route to the same stale-cache problem: reconciliation currently happens **only when `currentDisplay == kAllDisplayValue`**. If monitors change while the user is viewing one display, old `_images` entries remain. Later, entering All Displays on mobile does not clear them—the existing `clearImage()` in `openMonitorInTheSameTab()` is desktop-only. 

#### Review comment I'd leave

> **[P2] Invalidate cached frames by topology, not just by display count.**
>
> `reconcileDisplays(newDisplays.length)` only removes indexes that are now out of range, but display indexes can be reassigned when a monitor is removed/reordered. For example, `[A, B] -> [B]` keeps `_images[0]` (A), after which `MultiDisplaysImagePainter` pairs that stale image with the new `displays[0]` (B) until a fresh frame arrives. A same-count change such as `[A, B] -> [B, C]` removes nothing at all.
>
> The cache can also remain stale if topology changes while a single display is selected, since reconciliation is only performed in the All Displays branch.
>
> Since `Display` has no stable monitor identity, I think the safest approach is to invalidate all per-display cached images whenever the display topology changes, using the existing deferred `_scheduleDispose()` mechanism. The new `sessionSwitchDisplay()` logic can remain conditional on currently being in All Displays mode.

### Suggested implementation

Something along these lines would be safer:

```dart
final oldDisplays = List<Display>.of(_pi.displays);
final topologyChanged = !listEquals(oldDisplays, newDisplays);

if (topologyChanged) {
  parent.target?.imageModel.invalidateDisplayImages();
}

_pi.displays.value = newDisplays;
_pi.displaysCount.value = newDisplays.length;

if (_pi.currentDisplay == kAllDisplayValue) {
  bind.sessionSwitchDisplay(
    isDesktop: isDesktop,
    sessionId: sessionId,
    value: Int32List.fromList(
      List.generate(newDisplays.length, (index) => index),
    ),
  );
  sessionRefreshVideo(sessionId, _pi);
  updateCurDisplay(sessionId);
}
```

`invalidateDisplayImages()` should remove the `_images` entries and pass their images through `_scheduleDispose()` rather than synchronously disposing them, preserving the paint/dispose protection already added in this PR. The `Display` class already implements value equality based on its geometry/size/cursor properties, so `listEquals()` can be used for the topology comparison. 

**Current verdict: Request changes — 1 P2 remains.**

The earlier Flutter 3.24.5 blocker is fixed, and the new `sessionSwitchDisplay()` handling fixes the capture-side part of my previous topology review. I did not find another P1/build blocker in the current `47a8127` head.

[1]: https://github.com/rustdesk/rustdesk/pull/15974/commits "feat(mobile): add multi-display side-by-side view support by lapunik · Pull Request #15974 · rustdesk/rustdesk · GitHub"

rustdesk:
Re-reviewing the current HEAD after the explanation above: **I no longer think the two printer findings are merge blockers.**

1. **Shared printer driver lifecycle — withdraw**

I previously flagged `uninstallPrinter()` because every branded client shares `RustDesk v4 Printer Driver`, while the printer and port names are brand-specific.

On closer inspection, Windows provides the required protection here. `DeletePrinterDriverExW()` cannot remove a driver while any printer still references it; the spooler returns `ERROR_PRINTER_DRIVER_IN_USE`. RustDesk's `uninstallDriver()` returns immediately when that deletion fails, so it also does not proceed to delete the driver package.

Therefore, with two branded clients installed:

* uninstalling A deletes `A Printer`;
* deletion of the shared driver fails while `B Printer` still uses it;
* B's printer and driver remain intact;
* A's brand-specific port can then be removed.

So the original claim that uninstalling one branded client could tear down the driver underneath another client was incorrect.

There is a minor logging issue: `uninstallPrinter()` logs `"Uninstalled the printer driver"` regardless of whether `uninstallDriver()` succeeded, but that is not a correctness blocker.

2. **`UninstallPrinter` running during installation — not a blocker for this PR**

The WiX condition is indeed broad:

`Condition="VersionNT >= 603"`

so `UninstallPrinter` can execute during a fresh install rather than only during removal.

However, this behavior predates this PR. The current master branch has the same unconditional `UninstallPrinter` scheduling; this PR mainly adds `UninstallPrinter.SetParam` so the deferred action receives the runtime `ProductName`.

The new branding behavior also does not create the cross-brand failure I previously described. The printer being deleted is now brand-specific, and an attempt to delete the shared driver is protected by the Windows spooler if another branded printer still uses it.

The condition could still be cleaned up separately if the intended semantics are strictly "run only on uninstall/upgrade removal", but I would not require that cleanup as part of #15774.

### Updated conclusion

The author's responses to the earlier MSI concerns are technically convincing:

* the Nightly workflow exercises the real x64/ARM64 MSI-template build path;
* the paired hbbs patcher regenerates effective MSI product/upgrade/component/package identities;
* the patcher updates replacement-file metadata and hashes rather than merely swapping `cab2.cab`.

After also re-checking the two printer findings, I do **not** see a remaining code-level issue strong enough to request changes on the current PR.

I would still strongly recommend completing the Windows end-to-end matrix described in the PR, especially:

1. generate two differently branded MSIs from the same template;
2. install both with printer support enabled;
3. verify both printers;
4. upgrade/repair each independently;
5. uninstall A and verify B still prints;
6. uninstall B.

That is validation of the new packaging model rather than evidence of a known defect.

**Verdict: Approve, with the Windows side-by-side MSI test retained as release validation rather than a code-change blocker.**

fufesou (approved):

rustdesk:
# Fable 5.1 Review of rustdesk/rustdesk#15774 — Custom client no rebuild

Reviewed at head `8e1e9a33b` (fix: simple path fix in generate.py), cross-checked against
every comment on the PR and against the paired hbbs branch `custom-client-no-rebuild`
(`jobs/msi_patch.py`, `jobs/build_agent.py`, `jobs/update_windows_template.sh`,
`jobs/win_res.py`).

**Verdict: no merge blocker.** Five findings remain, all small. Two of them were already
raised by Copilot/CodeRabbit and have not been answered. Three findings from the first
pass are withdrawn below with the evidence that refuted them.

## Findings that stand

### 1. A corrupt or truncated `RDPKG` resource makes the client exit silently

`libs/portable/src/main.rs:296` — `BinaryReader::new()?` propagates the new package
errors out of `main() -> Result<(), String>`. The crate is `#![windows_subsystem =
"windows"]`, so the `Error: "RDPKG resource is invalid"` line goes to a stderr that does
not exist, and `ui::setup()` is only reached inside `setup()` after the reader exists.

Consequence: a customer double-clicks the exe and nothing happens at all. This is not a
regression — master's `expect` panicked just as silently — but the PR introduces a new
per-customer, post-build input (the injected resource) with no diagnostic path, and the
hbbs side (`win_res.inject`) does not read the resource back after `EndUpdateResource`.

Suggested fix: on the `Err` path, show a `#[cfg(windows)]` modal (native-windows-gui is
already linked) or write a log file next to the exe before exiting.

### 2. `generate.py` still accepts an executable outside the folder

`libs/portable/generate.py:134` — the new `os.path.exists(os.path.join(folder,
options.executable))` probe patches over the separator-less `startswith` checks on lines
128 and 137, which the author's own comment calls "incorrect".

- Typo case, deterministic: `-f rustdesk -e rustdesk-typo.exe` passes `startswith`,
  the probe misses, line 137 passes, and the trailer becomes `.-typo.exe`. The packer
  extracts fine and then fails to spawn.
- CI case, needs a coincidence: with `-f ../../rustdesk/ -e ../../rustdesk/rustdesk.exe`
  (flutter-build.yml:367) the probe resolves to `<repo-root>/rustdesk.exe`; if that
  file ever exists the trailer becomes `..exe`.

Suggested fix: change line 137 to `exe.startswith(os.path.abspath(folder) + os.sep)`.
Every case above then hits the existing "must locate in source folder" error, and the
probe and its comment can go. (The commit that was reverted on Aug 18 changed the
`--exclude-exe` comparison, not this check, so this has not been tried.)

### 3. `normalized` duplicates `bin_reader::normalize_path`

`libs/portable/src/main.rs:87` — byte-for-byte copy of `normalize_path`
(`bin_reader.rs:95`). Both define the path identity that decides which extracted files
the dropped-file sweep deletes and which generic entries a package shadows; if one is
tweaked the other silently diverges. Make `normalize_path` `pub(crate)` and reuse it.

Already raised by CodeRabbit as a nitpick; no reply.

### 4. New `expect` in production code

`libs/portable/src/bin_reader.rs:143` — `parse(BIN_DATA).expect("bin file is not
valid!")` in `read_embedded`, while the caller `BinaryReader::new()` already returns
`Result<Self, String>` (the PR made it fallible for exactly this purpose). AGENTS.md:
"Avoid `unwrap()` / `expect()` in production code"; the only exceptions are tests and
lock poisoning. `.ok_or_else(|| "bin file is not valid".to_owned())?` matches
`parse_package_blob` two functions below.

Already raised by Copilot and CodeRabbit; no reply.

### 5. Placeholder file list restated in the workflow

`.github/workflows/flutter-build.yml:428` — the PowerShell list (`custom.txt` on 425 plus
`icon.ico, icon.png, logo.png, logo_light.png, logo_dark.png`) restates the keys of
`PER_CUSTOMER_FILES` (`res/msi/preprocess.py:112`). `preprocess.py --template` already
receives `-d` and could create the placeholders itself.

Value is limited: the same list also lives in hbbs (`msi_patch.py` `OPTIONAL_ASSETS` and
the asset tuple in `build_agent.py`), and the cross-repo contract forces at least two
copies. Consolidating here removes one of four.

## Withdrawn after checking the hbbs patcher

### Manufacturer / Publisher placeholder — withdrawn

Claimed: the template build passes no `--manufacturer`, so the default "Purslane Tech
Pte. Ltd." is baked in with nothing for the patcher to key on.

Refuted: `msi_patch.py` `set_publisher` (called unconditionally before commit) sets
`Property.Manufacturer` via `set_property` and rewrites the Registry row `Name =
'Publisher'`, asserting exactly one row. It keys on names, not on a placeholder. The
premise that the old flow used `-m` was also wrong: `msi_build` never passed it.

Side observation for the hbbs side: the patcher sets Publisher to the customer's app
name, whereas the old flow left the Purslane default. Confirm that is intended.

### `CC_HAS_*` Property rows not emitted — withdrawn

Claimed: template conditions reference `CC_HAS_*` but no Property rows exist, so the
patcher must INSERT; if skipped, the customer silently gets no logo.

Refuted: `msi_patch.py` `set_property` explicitly inserts when the row is absent, and its
docstring records that the template defines no row. `enable_supplied_assets` calls it
for every asset the customer supplied. The contract is implemented and documented on
the consuming side. Emitting `Value="0"` rows in the `--template` branch remains an
optional readability improvement, not a correctness item.

### Legacy extraction directory left behind — withdrawn

Claimed: deriving the directory from the exe stem moves existing custom clients from
`%LOCALAPPDATA%\rustdesk` to `%LOCALAPPDATA%\<app>` and orphans the old tree.

Refuted: the old `build_agent.py` ran `sed -i 's/"rustdesk"/"<app>"/g' src/main.rs`,
which rewrote `APP_PREFIX`, and master's `setup` uses `dir.join(APP_PREFIX)`. Old custom
clients therefore already extracted to `%LOCALAPPDATA%\<app>`; `app_dir_name("./<app>.exe")`
yields the same directory. There is no migration and nothing is orphaned.

## Dropped in the first pass (unchanged)

- Shared Component GUID / UpgradeCode: real, but answered on the PR (Aug 27) and
  confirmed in code — `new_product_guids`, `new_component_guids`, `new_package_code`
  all run before commit.
- 32-bit `parse` length overflow: wraps to `None` in release, no crash.
- Repair removes the logo: components are not Transitive and the cached MSI carries the
  patched Property table.

## PR comment state at review time

- Inline threads: Copilot (2) and CodeRabbit (5); none answered by the author.
- Aug 27 (rustdesk): request changes on MSI CI coverage, identity rewrite, file metadata.
- Aug 27 (fufesou): nightly workflow builds the templates; patcher rewrites identities
  and file metadata. Verified in `msi_patch.py`.
- Sep 1 (rustdesk): withdrew both printer findings, approved; Windows side-by-side
  install matrix kept as release validation.
- None of these replies address findings 1–5 above.

rustdesk:
# Review of rustdesk/rustdesk#15774 at `0d846dca6`

No blocker. The MSI/native-ARP mismatch is fixed in hbbs `ca7957a21`, the generate.py
path check and the `normalize_path` duplicate are fixed in `09a75482a`. Verified on
macOS: `cargo test` 9/9; `generate.py --package` correct for every caller shape.
What follows is only what still warrants a change.

## 1. A corrupt `RDPKG` resource exits with no message

`libs/portable/src/main.rs:291` — `BinaryReader::new()?` returns from `main` before
`ui::setup()` runs, and the crate is `#![windows_subsystem = "windows"]`, so the
`Err` text goes to a stderr that does not exist. A customer double-clicks a client
whose injected resource is truncated and nothing happens.

Fix: on the `Err` path, show a `#[cfg(windows)]` modal (native-windows-gui is already
linked) or write a log file next to the exe before exiting.

## 2. Dead-code warnings on non-Windows builds

Since `14a7088e2`, the non-Windows `read_embedded` returns `Ok(Default::default())`
without calling `parse`, so `cargo check` reports 7 new warnings: `LENGTH`,
`IDENTIFIER`, `IDENTIFIER_LENGTH`, `MD5_LENGTH` (`bin_reader.rs:18-21`), `read_u32`
(`:99`), `parse` (`:105`), `parse_package_blob` (`:149`). Master has none.

Fix: `#[cfg(any(windows, test))]` on those items.

## 3. `app_metadata.toml` is not ignored

`libs/portable/.gitignore` covers `/target`, `*.exe`, `*.bin` only. `generate.py` in
normal mode writes `libs/portable/app_metadata.toml`; it was committed in `09a75482a`
and removed in `0d846dca6`. Pre-existing gap, one ignore line prevents a repeat.

## 4. Placeholder file list duplicated in the workflow (optional)

`.github/workflows/flutter-build.yml:415-419` restates the keys of
`PER_CUSTOMER_FILES` (`res/msi/preprocess.py:112`). `preprocess.py --template` already
receives `-d` and could write the placeholders itself. Low value on its own, since the
same list also has to exist on the hbbs side.

PR #15781, opened 2026-08-07
feat(rdp): title the mstsc window after the peer instead of "localhost"

rustdesk:
The RDP tunnel launched `mstsc /v:localhost:<port>`, so with several sessions open every window is titled "localhost" and servers cannot be told apart.

mstsc titles the session window after the launched .rdp file's base name, so write a temp .rdp file (containing only the tunnel address) named after the peer alias, cached hostname, or id, and launch that instead. Falls back to the old /v: form when no usable name remains after filename sanitization or the file cannot be written. Credential handling is unchanged: cmdkey targets "localhost", which is still the host mstsc resolves credentials against.

Fixes rustdesk/rustdesk#15775 (discussion)

| Directly from RDP Client | RustDesk Release | PR Build |
|--------------------------|------------------|----------|
| rdpnative |  Screenshot 2026-08-13 143523 | image |

### Known limitation

After MSTSC enters full-screen mode, the connection bar still displays `localhost`. This is expected because the RDP connection is made to the local tunnel endpoint (`localhost:<port>`).

This change uses `SetWindowTextW` to update only the top-level MSTSC window title. The full-screen connection bar is rendered separately by the MsTscAx ActiveX control and is not affected by the window title.

Microsoft provides [`IMsRdpClientNonScriptable3::ConnectionBarText`](https://learn.microsoft.com/en-us/windows/win32/termserv/imsrdpclientnonscriptable3-connectionbartext), but it is only available to applications that host the MsTscAx control themselves and must be set before entering full-screen mode. It cannot be applied to an independently launched `mstsc.exe` process.

## Where the hostname comes from, and when it is available

1. Origin — the controlled peer reports its own OS hostname. After a successful login, the controlled side fills PeerInfo (src/server/connection.rs:1874): on desktop platforms pi.hostname = crate::whoami_hostname() (the machine's actual computer name via gethostname/GetComputerName, src/common.rs:833); on Android it is the device name. This is sent to the controlling side in the LoginResponse — it is the real "TEST01" the reporter wants to see.
2. Cache — the controlling side persists it per peer. Every successful login runs LoginConfigHandler::handle_peer_info (src/client.rs:2563), which stores the hostname into that peer's config file and updates the in-memory lc.config (save_config, src/client.rs:1958). All session types refresh this cache — normal remote sessions, file transfer, and RDP tunnel sessions themselves (connect_and_login calls handle_peer_info at src/port_forward.rs:162).
3. Read — rdp_display_name reads that cache (lc.info.hostname via the Deref to PeerConfig).

## Timing implications

- The first mstsc launch happens when the tunnel listener starts, i.e. before this session's login — so it uses the hostname cached by a previous session. Any prior connection to the peer (remote / file transfer / RDP) is enough to have it populated.
- For a peer that has never been connected before, the cache is empty and the title falls back to the peer id — still far more useful than "localhost".
- Once the first login inside the RDP session completes, the cache is refreshed in memory and on disk, so "New RDP window" in the same session and all later launches show the up-to-date hostname.
- The value reflects the hostname as of the last login; if the remote machine is renamed, the title catches up on the next login — at most one session behind.
- A user-set alias takes priority over the hostname, since that is the label the user chose for the peer; the order is alias → cached hostname → peer id.

## Summary by CodeRabbit

* **New Features**
  * RDP windows now display a recognizable peer name, using the selected alias or available peer details.
  * Connection credentials are scoped to the specific local connection, improving reliability when using multiple RDP sessions.
  * RDP sessions continue launching directly for a faster, more consistent connection experience.
  * Window titles are safely sanitized and limited for clear display.

The PR replaces peer-named temporary RDP launch files with direct `/v:localhost:<port>` launches and applies the peer identity to the corresponding MSTSC window.
- Tracks each launched MSTSC process independently and reapplies its title while it remains active.
- Builds the title from the peer alias or ID, optionally including the cached hostname.
- Eliminates the shared temporary file that caused session redirection and the persistent peer-named artifacts reported previously.
</details>

The PR appears safe to merge because both previously reported temporary-file issues are eliminated and no blocking replacement failure was established.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| src/port_forward.rs | Launches MSTSC directly with the tunnel port and derives a peer-specific display name without creating shared or persistent RDP files. |
| src/platform/windows.rs | Adds process-scoped window enumeration and title updates that terminate when the launched MSTSC process exits. |

</details>

<sub>Reviews (9): Last reviewed commit: ["feat(rdp): show peer identity with hostn..."](https://github.com/rustdesk/rustdesk/commit/9604309abf92f38686f5fac3438fc1b027923104) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51026902)</sub>

### DIFF
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -2642,13 +2642,9 @@ pub fn wide_string(s: &str) -> Vec<u16> {
         .collect()
 }
 
-// This changes only mstsc's top-level window title. The full-screen connection
-// bar is rendered separately by the MsTscAx ActiveX control, so SetWindowTextW
-// cannot change its displayed target. Microsoft exposes
-// IMsRdpClientNonScriptable3::ConnectionBarText for hosts that create the
-// ActiveX control themselves, but it must be set before entering full screen
-// and cannot be applied to an independently launched mstsc.exe process:
-// https://learn.microsoft.com/windows/win32/termserv/imsrdpclientnonscriptable3-connectionbartext
+// This only changes mstsc's top-level window title. The full-screen connection
+// bar is rendered separately and cannot be customized when mstsc.exe is
+// launched as an independent process.
 pub fn set_rdp_window_title(mut child: std::process::Child, name: String) {
     let name: String = name.chars().filter(|c| !c.is_control()).take(120).collect();
     if name.is_empty() {
--- a/src/platform/windows.rs
+++ b/src/platform/windows.rs
@@ -2642,6 +2642,95 @@ pub fn wide_string(s: &str) -> Vec<u16> {
         .collect()
 }
 
+// This changes only mstsc's top-level window title. The full-screen connection
+// bar is rendered separately by the MsTscAx ActiveX control, so SetWindowTextW
+// cannot change its displayed target. Microsoft exposes
+// IMsRdpClientNonScriptable3::ConnectionBarText for hosts that create the
+// ActiveX control themselves, but it must be set before entering full screen
+// and cannot be applied to an independently launched mstsc.exe process:
+// https://learn.microsoft.com/windows/win32/termserv/imsrdpclientnonscriptable3-connectionbartext
+pub fn set_rdp_window_title(mut child: std::process::Child, name: String) {
+    let name: String = name.chars().filter(|c| !c.is_control()).take(120).collect();
+    if name.is_empty() {
+        return;
+    }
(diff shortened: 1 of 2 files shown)


21pages:
@greptileai review

21pages (approved):

Issue #580, opened 2022-05-21
Chang the name of Windows executable


JohnCoconut:
The name of Windows executable contains "putes" (reverse of "setup"), which is unfortunately a vulgar term in several languages(French, Austrian etc). See https://en.wiktionary.org/wiki/putes.  

Could it be changed to something else?

I guess the reason for avoiding the filename is due to this, https://stackoverflow.com/questions/11573444/why-is-windows-asking-for-system-administrator-privileges-for-running-executable. One answer provides a solution to use "setup" in the filename. 

Sorry I don't have a widows machine around, so I couldn't test the solution myself.

rustdesk:
Thanks for you reminder. We can not use setup, because WIndows will change the icon. But I will find another word. Maybe setpu etc.

youhers:
You can try my submission on rusdesk.iss

youhers:
https://github.com/rustdesk/rustdesk/pull/583 你可以看看这个

SoftwareMagicIT:
You can use install instead of setup for icon changing issue!

PR #15792, opened 2026-08-07
fix(linux): serve the Wayland login screen the DRM backend was built for

### LINKED ISSUE #15791: drm capture does not work at a Wayland login screen (gdm/sddm)
reported by fufesou on #15420 after it was merged. he is right, and it is worse than the one guard
he found.

what happens: connecting to a host sitting at a Wayland greeter is refused with "Wayland login
screen is not supported". comment out that refusal and the connection fails instead, because the
display list comes back empty.

the cause is one premise that the drm backend invalidates. `get_values_of_seat0` is
`_get_values_of_seat0(.., ignore_gdm_wayland = true)`, so a gdm/sddm Wayland session is skipped by
construction and `get_display_server` falls back to x11. that was the right call while the portal
was the only backend, since the portal cannot serve a greeter at all. the drm path never talks to
the compositor, which is exactly why it can serve one, so at a login screen every x11-vs-Wayland
decision in the tree now answers x11 and the whole backend switches itself off.

six things follow from it, and they hide behind each other, so fixing only the first one looks like
progress and changes nothing a user can see:

- the refusal in `on_remote_authorized` fires even when something can capture
- the service pre-warm gives up with "session still reads as X11 after 30s"
- the consumer warm logs "found no producer at startup" and the display list is empty
- the display service skips the drm displays
- the uinput rect has no compositor to ask, so a 2880x1800 panel is programmed as 1920x1080 and the
  pointer lands far from where it was aimed
- enigo keeps its own copy of that answer, decided once in `Enigo::default`, and routes every key
  and mouse event to xdo. with no x server the xdo context is null and libxdo drops them without an
  error, so the uinput devices are created, the compositor opens them, and nothing is ever written
  to them

the #15420 description says this works at the login screen. it does not, and that is on me.

verified on an sddm Plasma Wayland greeter, 2880x1800: the greeter renders,
typing from the client enters characters in the password field, and a click at an absolute
coordinate opens the greeter session combo. **gdm under Wayland is not proven yet**, only sddm.

one limitation i have not solved: non-ASCII characters still cannot be typed at a greeter. that path
goes through the clipboard, and the clipboard on this path is X11 only.

and one premise from #15420 that turned out to be wrong in a useful way. i had said the greeter
`--server` does not need the compositor variables, since the drm path never talks to the compositor.
the capture does not, but the desktop layout does, and blanking them left the layout unavailable at
a login screen. the compositor is running and its socket is in the runtime directory of the same
user, so it can simply be found. that is what the last commit does, and it removes more code than it
adds.

fxd0h:
Fixes #15791.

The login screen support in #15420 never worked on a real greeter. fufesou found it: the session is refused, and with the refusal commented out the client gets a failed connection instead of a screen.

One premise under all of it. `get_values_of_seat0` is `_get_values_of_seat0(.., ignore_gdm_wayland = true)`, so a gdm/sddm Wayland session is skipped by construction and `get_display_server` falls back to x11. That was correct while the portal was the only backend, since the portal cannot serve a greeter at all. The DRM path never talks to the compositor, which is precisely why it can serve one, so the premise stops holding there and every x11-vs-Wayland decision in the tree answers x11 at a login screen.

## What changes

The central one is the memoised `IS_X11`: when it reads x11 and seat0 is a Wayland greeter, answer Wayland. That covers fifteen routing sites at once. It is under `cfg(feature = "drm")`, so a build without the backend keeps the current answer exactly.

- `connection.rs` — the login-screen refusal now stands only while nothing can capture the screen.
- `platform/linux.rs` — corrected `IS_X11`, plus `is_x11_for_drm()` for the two retry loops that must keep asking while a boot is still naming the session, and a memoised accessor scoped to per-frame callers in the per-session `--server`.
- `ipc/drm.rs`, `drm_capturer.rs` — the service pre-warm and the consumer warm use the loop-safe form.
- `wayland.rs` — with no compositor to ask, the uinput rect comes from the DRM display list. Those are the same displays being captured, so the coordinate space matches by construction.
- `libs/enigo` + `input_service.rs` — this was the last layer and it sat outside all of the above. `Enigo` decides x11-vs-Wayland once in `Default::default()`, from the same seat0 lookup, and on "x11" routes every key and mouse event to xdo; with no X server that context is null and libxdo drops them without an error. The uinput devices were created, the compositor opened them, and nothing was ever written to them. `set_is_x11` is now called where the custom devices are installed, which is only reached once `!is_x11()` is already established. The unit test pins both directions, since a one-directional test passes against the bug.
- `display_service.rs`, `drm_capturer.rs` — telling the truth about a greeter makes four compositor-probing paths reachable where the probe cannot answer. All four already treat an empty output list as "nothing to do", so they skip it: 11818 "Could not find wayland compositor" warnings in one greeter session became 1.

## Tested

sddm Plasma Wayland greeter, MacBook T2, 2880x1800 panel:

- the greeter renders, session combo and power buttons and all
- typing from the client enters characters in the password field
- a click at an absolute coordinate opens the greeter's own session combo
- service pre-warm primes in 994 us instead of timing out at 30 s
- the privileged service still maps no EGL during a live capture

Both feature configurations build: `flutter,drm,drm-wake` and `flutter` alone.

**Not proven on gdm under Wayland**, only sddm.

## Known limitations

- Non-ASCII characters cannot be typed at a greeter. That path goes through the clipboard, and the clipboard here is X11 only.
- At a multi-monitor greeter the pointer reaches the first display only. DRM reports every output at origin (0,0) on Wayland, so a union of overlapping rectangles collapses to one display and there is no arrangement to derive.

Both of those come from the same place: the greeter's `--server` is started with the compositor variables blank, so nothing can ask the compositor anything. That turns out to be avoidable, and rustdesk/hbb_common#580 is the follow-up — the compositor is running and its socket is in the runtime directory of the same user, so the output layout can simply be found. With it the second limitation goes away and the four probe guards above become unnecessary. Kept out of this PR because it lives in the submodule.

## Correction

The #15420 description says this works at the login screen. It does not, and that is on me. I am leaving that description as it stands, since it is the record of what was reviewed, and correcting it here instead.

## Summary by CodeRabbit

* **New Features**
  * Added support for capturing supported Wayland login screens through the DRM backend.
  * Improved display detection and geometry handling for login-screen sessions.
  * Added safer Linux keyboard and mouse input routing across display servers.

* **Bug Fixes**
  * Fixed display-server detection and compositor probing during Wayland greeter sessions.
  * Improved fallback behavior when display geometry is unavailable or incomplete.
  * Prevented unnecessary compositor queries during DRM-based login-screen capture.
  * Improved remote login availability when DRM can support the greeter.
  * Improved detection of changing headless-session states.

The PR enables the DRM backend to recognize and serve Wayland greeters while preserving existing behavior when DRM is disabled.
- Corrects display-server and seat-owner detection for Wayland login screens.
- Routes greeter input through custom uinput devices rather than the unavailable X11 path.
- Uses DRM display information when compositor geometry is unavailable and avoids blocking compositor probes on routing paths.
- Adds tri-state DRM availability handling and asynchronous seat-state refreshes.
</details>

The PR appears safe to merge because no blocking failure remains.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| src/platform/linux.rs | Adjusts cached and retry-safe display-server detection so DRM builds recognize Wayland greeters. |
| src/platform/linux_desktop_manager.rs | Refreshes seat ownership outside the manager lock and represents DRM-capable greeters as supported sessions. |
| src/server/drm_capturer.rs | Introduces tri-state availability and augments DRM display geometry with compositor data when available. |
| src/server/connection.rs | Permits Wayland greeter connections only when cached DRM availability confirms capture support. |
| src/server/input_service.rs | Forces non-X11 dispatch when installing uinput or RDP custom input devices. |
| src/server/wayland.rs | Computes uinput geometry asynchronously and falls back to the DRM display rectangle. |
| libs/enigo/src/linux/nix_impl.rs | Adds an explicit display-server override and bidirectional dispatch coverage for custom mouse input. |

</details>

```mermaid
flowchart TD
    A[Detect seat0 session] --> B{Wayland greeter?}
    B -- No --> C[Use normal X11 or Wayland routing]
    B -- Yes --> D{DRM available?}
    D -- No --> E[Refuse unsupported login-screen session]
    D -- Yes --> F[Capture greeter through DRM]
    F --> G[Resolve geometry from compositor]
    G -- Unavailable --> H[Fall back to DRM display rectangle]
    G -- Available --> I[Use compositor layout]
    H --> J[Inject input through uinput]
    I --> J
```
</details>

<sub>Reviews (8): Last reviewed commit: ["fix: respect RUSTDESK\_FORCED\_DISPLAY\_SER..."](https://github.com/rustdesk/rustdesk/commit/f89bafe73a64d9db5924d944d6890a807bfe7026) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51304341)</sub>

**Context used:**

- Context used - AGENTS.md ([source](https://github.com/rustdesk/rustdesk/blob/master/AGENTS.md))


fufesou:
| OS | VM (?) | issues | remark |
| :----: | :----: | :----: | :---- | 
| Archlinux / gdm | Physical | ✅ | |
| Ubuntu 24.04 / gdm3 | Physical | ✅ |  |  |
| KUbuntu 26.04 / gddm | VM | ❌ | The cursor does not show. </br> After logging in, the same cursor issue occurs. |
| Ubuntu 24.04 / gdm3 | VM | ❌ | The cursor does not show. </br> After logging in, the same cursor issue occurs. |
| Ubuntu 22.04 / gdm3 | VM | ❌ | The cursor does not show. </br> After logging in, the same cursor issue occurs. |

It seems the VM might be different. I need to test it on more physical machines, and I will update this later.

## Screenshots

fxd0h:
thanks for the matrix. all three rows have a cause now, and one of them turned out not to be in this
PR at all.

the x-session dialog on physical ubuntu 24.04: reproduced at a real gdm3 greeter. it needs
`allow-linux-headless = Y`, which is why arch passes with the same code. `DesktopManager::new` reads
seat0 through `get_values_of_seat0`, which skips a gdm/sddm wayland session by construction, so at a
greeter it finds no session at all, `get_supported_display_seat0_username` returns None from its
empty-username arm, and `is_headless()` goes true. the corrected `IS_X11` never reaches this because
it asks who owns seat0, not which display server runs. fixed by asking again with the greeter
visible, at query time.

the dead input at the greeter is mine, from #15420. the uinput authorizer takes a fresh seat0 lookup
on purpose, and that read hides the greeter too, so it rejected the very `--server` that serves the
greeter:

    Rejected unauthorized connection on uinput ipc channel:
    postfix=_uinput_control, peer_uid=Some(120), active_uid=None

uid 120 is gdm, owner of the only active seat0 session. that side gets ECONNRESET out of
`setup_uinput`, no uinput device is created, and nothing reaches the greeter. the cache-based gates
never saw it because `Desktop::refresh` fills the cache through the greeter-visible read, which is
also why capture worked at a greeter while input did not. after the fix, on ubuntu 24.04 + gdm3, a
click selects the user, gdm opens the password field, and five keystrokes are five dots in it. i
verified this on 24.04; our deb cannot load on 22.04 itself (glibc 2.35 there against the 2.39 floor
of the build), but nothing in that path is release specific.

the cursor is a libdrmtap bug, not a greeter one, which is why it follows you into the session.
since kernel 6.6 a para-virtualized driver (virtio-gpu, vmwgfx, qxl, vboxvideo) hides its cursor
plane from a client that set `DRM_CLIENT_CAP_ATOMIC` without also setting
`DRM_CLIENT_CAP_CURSOR_PLANE_HOTSPOT`. libdrmtap sets the atomic cap to read the `CRTC_ID` of a
connector and never declared the other one, so the cursor plane was invisible to it and every read
reported the cursor hidden. reproduced on virtio-gpu: `drmModeGetPlaneResources` returns two planes
without the atomic cap, one with it, and two again with the hotspot cap; in the live service, 811 of
811 cursor reads found no plane, and with the cap the client draws the host cursor again. on i915
the cap is refused with `EOPNOTSUPP` and the plane count is unchanged, so bare metal loses nothing.
released today as libdrmtap 0.5.4, tagged and on crates.io; it reaches this PR through a pin bump
once rustdesk-org/libdrmtap takes the release (the pin is still on 0.5.2).

worth knowing while testing: "show remote cursor" ships unchecked in the client, so on a first
connection the cursor is absent for that reason alone. with it checked, ubuntu 24.04 + gdm3 and a
kde sddm greeter both show the greeter here with input working and the cursor drawn, on bare metal
and in a virtio-gpu VM, at the login screen and after logging in.

one question, because it decides whether the cursor fix covers your 22.04 VM: what does `uname -r`
say there? i measured both kernels. on the GA kernel (5.15.0-187) virtio-gpu shows two planes with
the atomic cap and two without, nothing is hidden, and the hotspot cap is rejected with EINVAL
because it does not exist yet, so 0.5.4 changes nothing and a missing cursor there is a different
bug i have not found. on 6.6 and later, which includes the 22.04 HWE kernel, the hiding exists and
the fix applies.

fufesou:
> what does `uname -r` say there?

6.8.0-124-generic

fxd0h:
6.8.0-124 is the 22.04 hwe kernel, so it has DRM_CLIENT_CAP_CURSOR_PLANE_HOTSPOT (added in 6.6), and that matches the invisible cursor on your three vms: a paravirtualized driver hides its cursor plane from an atomic client that does not declare that cap, so the drm capture side never sees the cursor at all, greeter or not. i reproduced it on a 24.04 qemu vm with virtio-gpu on the same 6.8 series and fixed it in libdrmtap 0.5.4 (fxd0h/libdrmtap#49): declaring the cap brings the plane back, plane count goes 2 -> 1 -> 2 with and without it.

this pr still pins libdrmtap 0.5.2, which predates that fix, so the cursor stays invisible in vm tests of this branch for now. once the rustdesk-org mirror picks up 0.5.4 i will bump the pin here. the missing input on the 22.04 greeter is the other two layers this pr already fixes, so with the pin bump the three vm rows of your matrix should be covered by this branch plus 0.5.4

fufesou:
This PR works perfectly on my physical machine now, except for the multiple-screen scenario.

1. The display arrangement.
2. The cursor positions.


https://github.com/user-attachments/assets/c7dcd8bf-8692-4787-aaf4-225a9a67046a


image


image

fufesou:
These two reviews seem reasonable https://github.com/rustdesk/rustdesk/pull/15792#pullrequestreview-4889330502

fxd0h:
thanks for running that 5 config matrix, it caught what my test setup could not.

both coderabbit findings are addressed in cd80c3dee: the headless decision now asks the probing form of the drm check, and only after the cheap seat0 read says a wayland greeter is actually there, so an unsettled warm up can no longer route login to x11 over a live greeter. and the seat0 values are read fresh on every query instead of trusting what DesktopManager::new cached, so a logout or fast user switch cannot hand back a stale owner. regression tested on a real sddm wayland greeter: capture works, the rustdesk password prompt is the only prompt, and typed characters land in the greeter password field.

the multi screen arrangement and the cursor positions turned out to be one bug with two faces, and i reproduced both on a two head virtio vm at the gdm greeter. drm has no origins, so every display was advertised at (0,0), and the uinput pointer range was taken from the union of the drm modes. the code did that at greeters on purpose, from before the socket fallback existed: it assumed a login screen has no compositor to ask. it does, and the fallback reaches it with no environment variables. cd80c3dee drops that gate, and on the vm the pointer range went from the overlapped (0, 5120) x (0, 2160) to the compositor arranged (0, 6400) x (0, 2160), with the greeter view filling the canvas instead of floating cut inside an oversized one.

two notes for retesting multi screen: the layout piece needs rustdesk/hbb_common#580 as well, since that pr is what provides the compositor answer at a greeter, and without it this branch just degrades to the old behavior. and the invisible cursor on your vms stays until the libdrmtap pin moves to 0.5.4 as discussed above

fufesou:
• The update improves both reviewed areas, but neither original finding is fully resolved at PR head 144f4f89.

  1. P2 — The blocking-probe finding remains on the fallback path.

     The first lookup and DRM fallback now run inside spawn_blocking() at src/server/wayland.rs:163, which is good.

     However, after a failed compositor lookup, snapshot_layout() calls get_display_rects_for_uinput() outside that task at
     src/server/wayland.rs:181. Failed lookups are not cached, so this can synchronously repeat the compositor/socket probe
     on the current-thread executor.

     The fallback can now perform up to three probes:
      - Initial compositor lookup.
      - drm_desktop_rect_for_uinput() → get_display_infos() → augmentation.
      - The later layout snapshot.

     The layout must be returned from the blocking task or omitted when only raw DRM geometry is available.

  2. P2 — The exact single-output case remains unchanged.

     src/server/drm_capturer.rs:1122 still returns raw DRM geometry when Wayland reports only one output. Thus the original
     one-DRM/one-Wayland-output example with a non-zero compositor origin is still unhandled.

     The update correctly handles one DRM connector among multiple compositor outputs. It does not fully implement the
     original review (https://github.com/rustdesk/rustdesk/pull/15792#discussion_r3749838461). An origin-only single-output
     path would need to preserve the intentional scale-1 behavior.

  3. P3 — Adjacent documentation is stale.

     src/server/drm_capturer.rs:1090 still says augmentation declines below two connectors, although it now runs with one
     DRM connector.

fxd0h:
all three are addressed in 2b17ba950.

the layout baseline is now computed inside the same blocking task, from the snapshot the successful lookup just cached. when only the raw drm union was available it is omitted, set empty, which keeps the 15601 remap inactive exactly where origins are unknown, instead of re-running the probe to learn nothing. so the fallback does at most the two probes that were already inside the task and nothing on the executor.

the single output case now adopts the origin: one connector against one output takes x,y from the output and keeps scale 1, matching how a single display is advertised at physical size everywhere else. more connectors than the one output stays unaugmented on purpose - the layout order fallback in the assignment would plant that origin on a guess, and a wrong origin is worse than the old (0,0).

the stale doc at get_primary_index is refreshed too.

on verification: the 2 head vm greeter regression passed (probe subprocess answers, uinput rect (0,6400),(0,2160)). the one output non zero origin case itself i could not measure, no machine here produces a lone output away from (0,0), so that path is code reviewed and compile verified only and your multi screen retest remains the real test.

fufesou:
## Findings

### [P1] Keep fresh seat and DRM probes out of pre-auth connection setup

Location: `src/platform/linux_desktop_manager.rs:299` (called from `src/server/connection.rs:514` and `src/server/connection.rs:6571`)

With `allow-linux-headless=Y`, constructing a connection calls `is_headless()` synchronously. That happens before RustDesk password authentication. The call holds the global `DESKTOP_MANAGER` mutex while `get_supported_display_seat0_username()` runs fresh `loginctl` commands. At a Wayland greeter it can also call `drm_capturer::is_available()`, whose display query can wait for `HANDSHAKE_WAIT_MS`—15.5 seconds with the current constants.

This is easy to miss because the outer `Connection -> is_headless()` call is old. The expensive work is not. At the merge base, `get_supported_display_seat0_username()` only read fields cached in `DesktopManager`. Commit `07f83eb64` added a fresh greeter lookup, and `e18d06ae5` made every query fresh and replaced the cached DRM check with the blocking one.

A client that reaches `Connection::start` can therefore occupy a Tokio worker before authenticating. Concurrent connections also serialize on the same standard mutex. If the DRM producer is unavailable, repeated probes can keep doing this while the state remains unsettled. Connection setup should stay cache-only; refresh the state on a blocking worker and do not hold `DESKTOP_MANAGER` across subprocess or DRM IPC waits.

Classification: **introduced by this PR**.

### [P1] Do not treat an unsettled DRM probe as “unavailable”

Location: `src/platform/linux_desktop_manager.rs:244`

The comment here says the probing form gives a definitive verdict, but `is_available()` still does not. If another probe is already in flight while `DRM_STATE` is `Unknown`, `src/server/drm_capturer.rs:853` returns `false` immediately. A failed query also returns `false` for the first four attempts while deliberately leaving the state as `Unknown` (`src/server/drm_capturer.rs:876`).

`drm_login_screen_seat0_username()` turns every one of those transient `false` results into `None`. During OS-login handling, `try_start_x_session()` can then continue into the Xorg startup path even though a live Wayland greeter may still be present and the DRM verdict has not settled.

The boolean behavior of `is_available()` predates this branch. The regression is the new use in commit `e18d06ae5`: this is the first time an unsettled result is used to decide that starting X11 is safe. This decision needs three states, or equivalent waiting/retry behavior. Only a definitive `Unavailable` result should allow Xorg startup.

Classification: **introduced by this PR**; the underlying two-state helper is historical, but this unsafe use of it is new.

### [P2] Preserve an explicit forced X11 selection

Location: `src/platform/linux.rs:61`

`hbb_common::platform::linux::get_display_server()` gives `RUSTDESK_FORCED_DISPLAY_SERVER` priority over auto-detection (`libs/hbb_common/src/platform/linux.rs:135`), and the root service deliberately forwards that variable to the per-user server (`src/platform/linux.rs:746`). The new greeter correction runs after that decision and changes the result to Wayland whenever seat0 looks like a Wayland greeter. `is_x11_for_drm()` does the same at `src/platform/linux.rs:244`.

As a result, `RUSTDESK_FORCED_DISPLAY_SERVER=x11` no longer forces X11 for routing: `get_display_server()` still reports X11, while `platform::is_x11()` and the DRM retry gates behave as Wayland. That leaves capture and input routing internally inconsistent and ignores an explicit operator override. The greeter correction should only adjust an auto-detected X11 result, not a forced one.

Classification: **introduced by this PR**, starting with commit `4c7e85cfa`.

fxd0h:
all three are addressed: the two P1 in ccf76a388, the P2 in f89bafe73, plus 980011ac6 and 7032a9079 with what an adversarial pass over my own fixes then found.

the pre-auth path is cache only now. is_headless() answers from a snapshot seeded at server start and refreshed off thread, so constructing a connection runs no loginctl and never waits on the drm probe. the fresh lookup became a free function and every caller invokes it with DESKTOP_MANAGER released - get_username and try_start_x_session included, so nothing holds that mutex across subprocess or ipc waits anymore. and the login request path itself no longer reaches the probing form either: greeter adoption reads a cached tri-state that never blocks (unknown kicks the probe off thread and answers unsettled), so an unauthenticated peer cannot park a worker on the probe deadline through try_start_desktop any more than through construction. two more corrections from the adversarial pass: an empty snapshot answers not-headless instead of consulting the boot time seat0 cache (that cache is blank at a wayland greeter because the loginctl wrapper skips greeter sessions, so the first connection of a fresh server process would have read headless=true at the exact screen this pr serves), and wait_desktop_cm_ready re-reads at call time instead of trusting a bool stored pre-auth that can lag one seat0 transition behind.

the probe is a tri-state now: Available, Unavailable, Unsettled. the in-flight-elsewhere case and the below-threshold failures return Unsettled, and the login decision routes to x11 only on a definitive Unavailable. the comment that claimed a definitive verdict was wrong and is gone. one more correction from the adversarial pass: a settled Unavailable used to be erased at its ttl expiry (state back to Unknown, failure counter to zero), which reopened the unsettled window every 30 seconds forever on a permanently helper-less box - your 22.04 GA class - and kept adopting a greeter nothing can serve. the verdict now stays Unavailable while an off thread re-probe re-verifies it; a failed or empty re-probe just restamps the no, only a non-empty display list flips it.

the forced display server case: IS_X11 and is_x11_for_drm() follow the override, and the adversarial pass showed that alone was incomplete - the seat0 adoption path still probed drm and connection admission still accepted greeter sessions whose capture and input then routed to x11, which is your "internally inconsistent" surviving at the sites the first commit did not touch. greeter adoption now yields to an operator forced x11 too, so a forced-x11 host at a wayland greeter degrades to upstream behavior and refuses the connection at the login screen instead of admitting a session that cannot work. i also dropped the seat0 fields of DesktopManager, dead since the fresh-read change.

verified on the 2 head vm greeter at the final head: capture serves end to end and the uinput rect is the compositor union at session init. the unsettled windows, the pre-auth timing and the forced-x11 matrix i could not measure on demand - those are code reviewed and compile verified.

fxd0h:
regarding Greptile trial, Greptile offers +2000 credits for OSS software, is the "trial ended" message a reminder that the 2k+ credits got depleted?

fufesou:
## Findings

No blocking issues found.

### [P3] Avoid fresh seat0 subprocess work on the unauthenticated login path

Location: `src/platform/linux_desktop_manager.rs:330`

With Linux headless support enabled, `get_username()` still calls `supported_display_seat0_username()` synchronously. That helper runs `get_values_of_seat0()`, and at a greeter it may also run `get_values_of_seat0_with_gdm_wayland()` and `get_display_server_of_session()`. These end up starting several `loginctl` subprocesses.

This work happens on the `LoginRequest` path before RustDesk password validation. A request with OS credentials can reach it more than once, while a request without them still reaches it through `try_start_desktop()`. The merge base served these lookups from fields populated once in `DesktopManager::new()`, so the per-request `loginctl` work was introduced by this PR.

The earlier review overstated the likely failure mode. The pinned `hbb_common` implementation does use `Command::output()` without its own deadline, but the relevant `loginctl` queries use sd-bus method calls. Unless configured otherwise, sd-bus has a [default 25-second method-call timeout](https://www.freedesktop.org/software/systemd/man/latest/sd_bus_set_method_call_timeout.html). I found no evidence that `loginctl list-sessions` or `show-session` commonly hangs on a healthy host, and no direct mechanism has been shown for a remote peer to make `systemd-logind` stop responding. The merge base also already ran `which Xorg` and `ls /usr/share/xsessions/` synchronously on part of this path.

It would still be cleaner to keep the pre-auth path cache-only or move the fresh lookup to a bounded blocking task. As written, an unhealthy logind can delay a login worker, and repeated subprocess creation is avoidable. However, practical worker-pool exhaustion has not been demonstrated, and this does not justify blocking the PR.

Classification: **introduced by `e18d06ae5`**.

fufesou:
> regarding Greptile trial, Greptile offers +2000 credits for OSS software, is the "trial ended" message a reminder that the 2k+ credits got depleted?

I don't know either.

It seems coderabbitai supports more reviews.

Please also address the findings https://github.com/rustdesk/hbb_common/pull/580#issuecomment-5248239422

fufesou:
@coderabbitai review

fufesou:
@codex review

fxd0h:
thanks. addressed in a136bfc54:

- your P3 (loginctl on the login request path): try_start_desktop now runs on spawn_blocking, so the loginctl reads and PAM when a session must start no longer sit on an async request worker - a slow logind ties up a blocking-pool thread instead of a request one. agreed it was not a blocker, but it is cheap to keep that path off the executor and it also covers the repeated-subprocess concern you raised.

- coderabbit also flagged: make drm_can_serve_login_screen treat Unsettled as servable, to match the seat0 gate. i left it Available-only on purpose and put the reason in the code. the two gates are deliberately not symmetric: seat0 adoption only hands the seat to Xorg on a definitive Unavailable (never over a maybe-live greeter), and admission only accepts on a definitive Available (never a greeter nothing can yet capture). both wait through an unsettled probe - refuse-and-retry - and admitting there would black-screen a client on a permanently helper-less box, which is the case the sticky-negative verdict exists to protect.

- and a guard on the seat0 refresh flag, so a panic in that thread cannot freeze is_headless on a stale snapshot.

the #580 findings are addressed too, in f696487 there - process group kill, privilege drop before parsing the socket, and the uninspectable-vs-empty distinction.

fxd0h:
unrelated to the review, a heads up on the cursor: the libdrmtap 0.5.4 mirror on rustdesk-org is still at 0.5.2 (653de8c, six commits behind). 0.5.4 is the one that declares the cursor-hotspot cap so a paravirtual driver (virtio-gpu, vmwgfx, qxl, vbox) stops hiding its cursor plane from an atomic client - the exact thing showing up on the vms. the pin cannot move to it until the mirror catches up.

could you sync rustdesk-org/libdrmtap to fxd0h main (5da68a3)? once it is there i will bump the pin in the three places (build.py, the drm workflow, the scrap version comment).

rustdesk:
@fufesou will finish remaining part (clean this PR)

fxd0h:
thanks for wiring it up. i built your head (73eff0368) and ran it on the two head gdm greeter vm here, since that is the setup i can exercise:

- the probe answers through your global_init dispatch: 2 outputs, and the uinput rect is the compositor union (0,6400),(0,2160)
- libdrmtap 0.5.4 loads and the first frame comes over dma-buf in 1.8 ms
- libEGL stays 0 in the root service after a real capture, libdrmtap 5 regions
- no new errors in the greeter server log; the ones there are the pre existing audio and clipboard noise from having no pulse or x11 for gdm

one behavior change worth a look, P3: the new early return in try_start_desktop answers SESSION_NOT_READY whenever os_login is empty, so linux_desktop_manager::try_start_desktop is never called with an empty username any more and its empty username arm is now unreachable. that arm was the one that ran detect_headless(), so on a box where xorg or /usr/share/xsessions is missing, a client that connects without os credentials used to get XORG_NOT_FOUND (info-nocancel with the headless docs link) on the first attempt, and now gets the session-login prompt instead - the real cause only shows up after the user supplies credentials that cannot help. properly configured headless boxes are unaffected, both paths answer SESSION_NOT_READY there.

if you want the old diagnosis back, detect_headless is two bounded run_cmds so it could run inside the same spawn_blocking task for the empty username case; if you would rather keep the flow as is, the dead arm in linux_desktop_manager could go. your call, i did not touch the branch.

fxd0h:
follow up on the multi screen case: i got a second monitor onto the sddm greeter box, so this is measured on real hardware now instead of the vm.

the box has one card with two connectors, the built in retina panel at 2880x1800 and an external at 1920x1080, side by side at a wayland sddm greeter. running your head, 73eff0368.

the peer info the client receives:

    DisplayInfo { x: 0,    width: 2880, height: 1800, name: "eDP-1", original_resolution: 1986x1241, scale: 1.4501 }
    DisplayInfo { x: 1987, width: 1920, height: 1080, name: "DP-6",  original_resolution: 1920x1080, scale: 1.0 }

so the external really lands at a non zero origin, each display carries its own scale (2880/1986 = 1.4501, fractional hidpi), and the connectors matched by name with no layout-order fallback. raw drm would have reported both at (0,0) at physical size.

the part i wanted to confirm for your original report is the agreement: the client desktop is 1987 + 1920 = 3907 wide, and the uinput range came out (0,3907),(0,1241). same coordinate space, so pointer positions line up across both screens.

capture works there too: first frame over dma-buf in 5.9 ms, libEGL stays 0 in the root service, and the box enumerates three drm devices (one is the touch bar, no render node) without trouble.

i also put your head on a normal logged in wayland desktop, not a greeter, since should_start_desktop and the in-flight guard sit on every linux connection and not just the login screen path. multi monitor session there, first frame in 2.3 ms, libEGL 0, nothing odd in the login flow.

rustdesk:
# Review

Fresh `loginctl` and Wayland probes remain in retry loops and async paths, allowing child-process churn or Tokio worker stalls under failure conditions. The patch also regresses first-attempt diagnostics on misconfigured headless hosts.

## Findings

### P1: Keep `loginctl` out of the CM retry loop

`src/platform/linux_desktop_manager.rs:336`

With `allow-linux-headless=Y`, `start_ipc()` invokes this getter at `connection.rs:6245/6258` from a Tokio task while waiting for a desktop username. If `headless_cm` has been selected but no username becomes available—for example, a just-started X session exits—and the connection drops, `_rx_desktop_ready.recv()` returns `None` immediately; the existing loop then calls this newly fresh getter without delay, spawning several synchronous `loginctl` processes indefinitely. The merge-base getter was cache-only, so break on the closed channel and keep this retry cache-only or offloaded as required by [AGENTS.md:53](AGENTS.md#L53).

### P2: Back off failed greeter geometry probes

`src/server/drm_capturer.rs:1247`

During an active DRM greeter session, if the compositor cannot be enumerated because there is no usable socket or it stops answering, failures from `get_displays()` are not cached. This call is reached through `get_display_infos()` on every nominal 300-ms display-service poll, causing repeated probe-process churn, and from async `get_displays_and_primary()`, which then calls `get_primary_index()` and can repeat the same two-second deadline on the Tokio worker. Add failure backoff and perform the login geometry/primary lookup in one blocking task; see [AGENTS.md:53](AGENTS.md#L53).

### P3: Preserve headless prerequisite diagnostics

`src/server/connection.rs:6723-6725`

On a headless host with Xorg missing or `/usr/share/xsessions` empty, an initial request without an OS username now exits here before `linux_desktop_manager::try_start_desktop("", "")` can run `detect_headless()`. The client receives the generic session-login prompt instead of the actionable Xorg/no-desktop error and documentation link, and only sees the real cause after entering credentials that cannot fix it. Run the prerequisite diagnosis in the same blocking task before returning not-ready.

fufesou:
Headless is not tested.

Because Linux headless has stopped working since 1.2.3 https://github.com/rustdesk/rustdesk/releases/tag/1.2.3
When the `--server` process was changed to the user process.

I will file a new PR to test and fix the headless feature.

Issue #15695, opened 2026-07-28
linux/wayland: a mid session resolution decrease keeps encoding into the old canvas


fxd0h:
on wayland, dropping the resolution of a captured output while a client is connected keeps the session running with the smaller image encoded into the previous canvas, and the stale right and bottom edges stay visible until the connection is torn down. an increase recovers on its own, a decrease does not.

three pieces line up, all of them in code shared by every capture backend on linux:

1- the encoder is sized once from the capturer info when the capturer is built, src/server/video_service.rs around line 597, where width and height come from CapturerInfo.

2- check_display_changed returns none unconditionally when the session is not x11, src/server/display_service.rs around line 223, with the comment that wayland does not support changing display for now, so the periodic display changed broadcast never fires.

3- convert_to_yuv only bails when the source is larger than the destination, libs/scrap/src/common/convert.rs around line 28. a smaller source passes and is written into the old canvas.

so a smaller frame is converted and encoded rather than being treated as a topology change.

what i verified and what i did not. i traced the three pieces above in the current tree and reproduced the wedge on the drm capture backend in #15420. i did not verify what the pipewire backend does when the mode changes under it, so i do not know whether it yields a smaller frame or fails the session outright. the three pieces are backend agnostic, so if pipewire does yield a smaller frame the outcome should be the same.

a bounded fix on the capturer side is to record the dimensions the session was built with and return an error from frame when a dequeued frame differs, which routes a modeset through the same rebuild path that a size increase already takes. the general fix is a wayland display changed signal, which is what point 2 turns off today.

i am shipping the bounded version for the drm path in #15420 so that path is correct on its own. filing this separately because points 2 and 3 are not specific to that backend.

fxd0h:
update after measuring this today, because two of the three paths turned out to be covered and one is not.

with a client connected i dropped the mode of the captured output from 1280x800 to 1024x768 in a live session:

1- x11: `check_display_changed` fires, the log shows `switch to refresh` and a new capturer, and the encoder comes back at 1024x768. it recovers on its own, as described above.

2- wayland with the drm backend: the capture stream ends (`drm capture stream ended: stopped`), the video service restarts, and the new encoder is 1024x768. that is the bounded fix from #15420 working as intended, the reader records the dimensions the session was built with and errors when a dequeued frame differs.

[...]

PR #15804, opened 2026-08-09
fix: route a mid session resolution decrease through capturer rebuild (#15695)

fxd0h:
as discussed in #15695. the check compares each PixelBuffer frame against the size the capturer was built with and bails on a mismatch, which routes a decrease through the same rebuild an increase already takes today, since convert_to_yuv refuses oversized sources. a smaller frame currently passes that check and is written into the old canvas, which leaves the stale right and bottom edges the issue shows.

on the VideoCropMeta question: yes, a cropped frame with a size different from the capture size is legal, and our own pipewire path already produces one. libs/scrap/src/wayland/pipewire.rs reads VideoCropMeta, copies the crop region and reports the cropped width and height upward, so when a compositor shrinks the content while the stream still runs at the old buffer size, video_service receives a frame smaller than the capture size. that is exactly the case this check catches. a crop equal to the full frame is normalized away in the recorder, so it never reaches the comparison.

the bail is not a dead end for a legal crop. run() tears down, the teardown clears the wayland display caches, and the rebuilt session renegotiates at the current size. check_display_changed returns None on wayland by design, so without a per frame check nothing else notices the mismatch on the pipewire path.

the comparison is against the capture size and not the encoder format on purpose: vpx_img_wrap aligns the encoder width up, so a legitimate frame is routinely narrower than dst_fmt.w.

what i measured and what i could not:

- x11: a resolution decrease under a live session recovers already, check_display_changed broadcasts and the capturer is rebuilt before a mismatched frame is dequeued. the new check stays silent there
- wayland with the drm backend of #15420: the reader ends the stream on a dimension change first, the check stays silent there too
- wayland with pipewire is the exposed path, and i do not have a portal screencast setup here to drive a live repro, so the crop behavior above is verified in code, not live. validation there belongs to whoever can run a portal session with a mode switch

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved video capture reliability by rejecting frames with unexpected dimensions.
  * Automatically rebuilds the capture session when an invalid frame is detected.

This PR detects pixel-buffer dimensions that differ from the active capture dimensions and rebuilds the capture session rather than encoding into a stale canvas.
- Compares each pixel-buffer frame against the capturer’s configured width and height.
- Aborts the current capture run on mismatch so the existing teardown and rebuild path can renegotiate the dimensions.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| src/server/video_service.rs | Adds a per-frame dimension guard that routes mismatched pixel buffers through capture-session rebuilding. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix: reject a frame that does not match ..."](https://github.com/rustdesk/rustdesk/commit/1fb4ed4ba1906a6a339a7d5ddf99050c4b1773bb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51527746)</sub>

### DIFF
--- a/src/server/video_service.rs
+++ b/src/server/video_service.rs
@@ -724,6 +724,20 @@ fn run(vs: VideoService) -> ResultType<()> {
         let res = match c.frame(spf) {
             Ok(frame) => {
                 repeat_encode_counter = 0;
+                // Rebuild when a mode change hands back a frame that no longer matches the size
+                // this capturer/encoder pair was built with: convert_to_yuv only refuses LARGER
+                // sources, and a smaller one lands in the old canvas leaving stale edges.
+                if let scrap::Frame::PixelBuffer(f) = &frame {
+                    if f.width() != capture_width || f.height() != capture_height {
+                        bail!(
+                            "frame {}x{} is not the {}x{} this capturer was built with; rebuilding",
+                            f.width(),
+                            f.height(),
+                            capture_width,
+                            capture_height
+                        );
+                    }
+                }
                 if frame.valid() {
                     let screenshot_key = (vs.source, display_idx);
                     let screenshot = SCREENSHOTS.lock().unwrap().remove(&screenshot_key);


fufesou:
This patch improves recovery for a single active Wayland capture, but during concurrent multi-display capture it can turn a degraded-but-live resized stream into a restart loop.

**[P1] Rebuild the failed capturer when another video service is active**

This `bail!` only recreates the Wayland capturer when the failing service is the last active video service. If another display or camera is still running, the cleanup guard leaves `CAP_DISPLAY_INFO` intact. The service loop then starts this display again with the same `CapturerPtr` and old `rect`, so every smaller frame exits here again. A multi-display session therefore cannot recover one resized display until all other video services stop. Please invalidate and recreate the affected capturer, or route the geometry change through a restart path that can safely refresh all active displays.

fxd0h:
your reading of the restart loop is right, i verified each step: the guard only clears CAP_DISPLAY_INFO when the failing service is the last one, check_init returns early on a non empty map, and the loop reuses the same CapturerPtr and rect. if the bail fires with another service alive, the loop happens as you describe.

but before fixing that i measured whether the bail can fire at all, and i could not reach it on any path. with a client connected i dropped the captured output resolution mid session and counted encoder rebuilds, restarts and bail hits, reading the crtc mode back from sysfs after each change as the control that the modeset really happened:

| path | stream size after the modeset | bail |
|---|---|---|
| pipewire, kde plasma | unchanged, the portal keeps the negotiated size | 0 |
| pipewire, gnome 50 | unchanged | 0 |
| pipewire, gnome 46 aarch64 | unchanged even on an increase | 0 |
| drm | changes, via the topology restart: 5 restarts, 5 right sized encoders | 0 |

window capture cannot reach this code either, select_sources asks for types 1, monitors only. and x11 does not need it, check_display_changed works there.

so the guard never sees a mismatched frame: pipewire scales inside the negotiated stream, and the drm path already routes a modeset through the rebuild that #15420 added, which is what the second half of #15695 said. the loop you found needs the bail to fire first, so it is not reachable either.

closing this instead of fixing the loop, as written the patch is a no-op for the trigger its own issue names. if you know a compositor or a path where the stream really hands the capturer a smaller frame, that revives both the patch and your finding, and i would want to know.

PR #15854, opened 2026-08-13
refact: remove feature plugin-framework

fufesou:
## Summary by CodeRabbit

* **Removed Features**
  * Removed plugin management, installation, configuration, and settings interfaces.
  * Removed plugin controls from the desktop home page and remote toolbar.
  * Removed plugin event handling, synchronization, session integrations, and peer-selection dialogs.
  * Removed plugin installation and management commands and related web integrations.
* **Localization**
  * Removed plugin- and peer-selection translation entries from supported languages.
* **Refactor**
  * Simplified startup, connection handling, and desktop settings by eliminating plugin functionality.

The PR removes the plugin framework from Rust and Flutter, including plugin lifecycle hooks, IPC handling, UI surfaces, dependencies, and localization entries.
- Deletes the Rust and Flutter plugin implementations.
- Removes plugin-related session, connection, startup, settings, and toolbar integration.
- Removes the feature flag, unused archive dependency, example workspace exclusion, and plugin translations.
</details>

The PR appears safe to merge, with no concrete regressions identified in the plugin-framework removal.

The removed feature, modules, UI integrations, IPC handlers, dependency, and translations are consistently eliminated, and no remaining repository references or non-plugin behavior dependencies were found.
</details>

| Filename | Overview |
|----------|----------|
| Cargo.toml | Removes the plugin feature, plugin-only zip dependency, and obsolete custom-plugin workspace exclusion. |
| src/server/connection.rs | Removes plugin request dispatch, connection lifecycle events, and plugin-controlled input blocking without affecting established non-plugin branches. |
| src/ipc.rs | Removes the plugin IPC data variant and handler alongside all in-repository producers and consumers. |
| src/flutter_ffi.rs | Removes Flutter bridge functions that exposed plugin management and configuration. |
| flutter/lib/models/model.dart | Removes Flutter-side plugin event routing and model dependencies. |
| flutter/lib/desktop/pages/desktop_setting_page.dart | Removes the plugin settings tab and corrects cleanup of the typed settings-tab observable. |
| flutter/lib/desktop/widgets/remote_toolbar.dart | Removes plugin-provided remote-toolbar display items. |
| src/lang/template.rs | Removes translation keys belonging exclusively to the deleted plugin interfaces. |

</details>

```mermaid
flowchart LR
  A[Application startup] --> B[Core Rust services]
  B --> C[Flutter UI]
  C --> D[Desktop and remote-session features]
  X[Plugin framework] -. removed .-> A
  X -. removed .-> B
  X -. removed .-> C
  X -. removed .-> D
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix: delete settings tab observable with..."](https://github.com/rustdesk/rustdesk/commit/c3751e242c35ba441c99f35af37175c4a6256b18) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52967761)</sub>


rustdesk:
PR #15854 is a near-pure deletion (5,150 lines removed, 7 lines added) removing the plugin_framework feature. I verified the removal mechanically: dart analyze lib on the PR head reports 0 errors / 0 warnings (238 info-only, all pre-existing), and cargo check --offline --locked --features flutter --lib produces a byte-identical warning/error set on the PR head and on its parent commit — so no dangling references and no new unused imports on either side. Cargo.lock is structurally self-consistent after the zip drop (verified programmatically: 0 dangling/ambiguous dep refs). plugin_framework was never in default and never passed by build.py or any CI workflow, so no shipped binary changes behavior.

The findings that survive are about the sweep being incomplete (schema, translations, packaging, on-disk state) plus a couple of pre-existing defects in files this PR rewrites.

```
[
  {
    "file": "flutter/lib/desktop/pages/desktop_setting_page.dart",
    "line": 169,
    "summary": "dispose() deletes the tab-key observable under the wrong type argument — `Get.delete<RxInt>(tag: _kSettingPageTabKeyTag)` vs the `Get.put<Rx<SettingsTabKey>>(selectedTab, tag: _kSettingPageTabKeyTag)` at line 131 — so the GetX key never matches and the instance is never removed.",
    "failure_scenario": "GetX keys are `type.toString() + tag`, so `RxInt|settingPageTabKey` never matches the registered `Rx<SettingsTabKey>|settingPageTabKey`. Open Settings, close the tab (the PageController at line 168 is deleted correctly; the Rx is not), reopen Settings: `Get.put` uses putIfAbsent, so the registry still holds the first page's dead observable. `DesktopSettingPage.switch2page` (line 98, reached from the Security shortcut in desktop_home_page.dart) then writes `selected.value = page` into a disposed Rx nothing observes, and every Settings open leaks one Rx for the process lifetime. Pre-existing, but this PR rewrites the enum, `tabKeys`, and both switches in this same class."
  },
  {
    "file": "libs/hbb_common/protos/message.proto",
    "line": 852,
    "summary": "The removal stops at the submodule boundary: `PluginRequest`/`PluginFailure` (lines 806/811) and `Misc` oneof fields `plugin_request = 25` / `plugin_failure = 26` stay in the wire schema with zero producers and zero consumers, undeprecated and unreserved; the hbb_common pointer (f124c0a5d) is unchanged.",
    "failure_scenario": "The PR deletes the only sender (`Session::send_plugin_request`, src/ui_session_interface.rs) and all three receive arms (src/client/io_loop.rs, src/server/connection.rs), yet protobuf codegen keeps emitting both message types, their oneof variants and `set_plugin_request`/`set_plugin_failure` into every Rust and Dart build. This proto has no `reserved` markers anywhere and the repo's convention for retired members is a `// deprecated` comment (see fields 24 and 27) — 25/26 got neither, so a future contributor can legitimately reuse tag 25 or 26 and have those bytes silently decoded by an older peer, landing in the catch-all `_ => {}` with no log. Same for `PrivacyModeState.PrvOnFailedPlugin = 6` at line 761."
  },
  {
    "file": "src/client/io_loop.rs",
    "line": 2263,
    "summary": "A plugin-specific special case survives the removal of the general mechanism: the server still emits `PrvOnFailedPlugin` (src/server/connection.rs:5014) and the client still pops a msgbox reading \"Please install plugins\", which is now unactionable.",
    "failure_scenario": "Any privacy-mode impl whose `turn_on_privacy` returns `Ok(false)` (e.g. src/privacy_mode/win_topmost_window.rs:103) makes connection.rs:5014 send `PrivacyModeState::PrvOnFailedPlugin`; io_loop.rs:2263-2266 then shows \"Privacy mode / Please install plugins\". After this PR there is no plugin concept, no plugin settings tab, no `--plugin-install` handler and no loader in the binary, so the user has no way to comply. The PR touched all 50 locale files to drop `(\"Plugins\", …)` but walked past `(\"Please install plugins\", …)` at line 297 of those very same files."
  },
  {
    "file": "src/lang/template.rs",
    "line": 486,
    "summary": "Only the `Plugins` key was swept; six sibling keys whose sole consumers were the deleted `flutter/lib/plugin/**` files remain in template.rs and all 50 locale files.",
    "failure_scenario": "Pre-PR the only consumers were flutter/lib/plugin/utils/dialogs.dart:28 (`translate(singleSelection ? \"Select peers\" : \"Select a peer\")`) and flutter/lib/plugin/widgets/desktop_settings.dart:108/129/197 (`'Uninstall'`, `isEnabled ? 'Disable' : 'Enable'`, `'Options'`). A repo-wide grep of the post-PR tree finds zero consumers outside src/lang. All six survive at template.rs:486-488 and 490-492 plus the same block in 50 locale files (~300 dead tuples) that stay compiled into every binary and keep appearing for translators. Note line 489 (`Update`) must stay — desktop_home_page.dart:436 still uses it."
  },
  {
    "file": "src/lang/pt_PT.rs",
    "line": 1,
    "summary": "The PR edits src/lang/pt_PT.rs, but that file is not declared as a module in src/lang.rs and is never compiled — the mod list has only `mod ptbr;`, and both \"br\" and \"pt\" resolve to `ptbr::T` (src/lang.rs:182-183).",
    "failure_scenario": "`grep -rn pt_PT` across the whole tree returns nothing outside the file itself. A pt-PT user gets the Brazilian strings; every entry in pt_PT.rs — including the `Plugins` line this PR just removed — is unreachable. Editing it gives the false impression the European Portuguese table is live, and the file will keep accumulating maintenance for no effect."
  },
  {
    "file": "src/core_main.rs",
    "line": 743,
    "summary": "Deleting `init_plugins` also deletes the `--service` call to `crate::plugin::remove_uninstalled()` — the only code that ever removed `Config::path(\"plugins\")/<id>` directories — with no replacement cleanup step, and the `--plugin-install`/`--plugin-uninstall` CLI verbs go with it.",
    "failure_scenario": "The deleted src/plugin/config.rs wrote to `HbbConfig::path(\"plugins\")` (per-plugin dirs, `shared.toml`, per-peer `<peer>.toml`, `manager.toml`). Post-PR, grepping `\"plugins\"` across src/ and libs/hbb_common/src returns nothing, so any install that ever loaded a plugin keeps that tree on disk forever: the uninstall path is gone and the CLI verbs that could trigger it are gone. Mitigation that lowers the severity: `plugin_framework` was never in `default` and never passed by build.py or any workflow, so only self-built/forked binaries can be in this state."
  },
  {
    "file": "res/msi/Package/Language/Package.en-us.wxl",
    "line": 24,
    "summary": "The packaging layer was not swept: the WiX localization file still defines `F_Client_Plugins` / `F_Client_Plugins_Desc` (lines 24-25) and `F_Server_Plugins` / `F_Server_Plugins_Desc` (lines 38-39) for MSI installer features that no longer exist.",
    "failure_scenario": "No `.wxs` under res/msi and nothing in res/msi/preprocess.py references `!(loc.F_*_Plugins)` — these are dead loc entries baked into the MSI string table, and after this PR they are the only textual trace of \"Plugins\" left outside src/lang. The PR reached Rust core, FFI, web bridge, Dart UI, IPC enum, Cargo features and 50 translation files but never touched res/, leaving the next person editing the MSI feature tree to wonder whether a Plugins feature was overlooked."
  },
  {
    "file": "src/lang/sl.rs",
    "line": 1,
    "summary": "src/lang/sl.rs is tracked with mode 100755 — the only file under src/ carrying the executable bit — and this PR rewrites the file without normalizing it.",
    "failure_scenario": "`git ls-files -s src/ | grep '^100755'` returns exactly one row: `src/lang/sl.rs`. The bit propagates into source tarballs and distro packaging (res/rpm-flutter.spec, res/PKGBUILD, the appimage builders), where rpmlint/lintian flag `spurious-executable-perm` on a non-executable source file, and it produces spurious `old mode 100755 / new mode 100644` churn in any future diff from a contributor whose tooling normalizes permissions. Editing the file is the cheapest moment to fix it."
  },
  {
    "file": "flutter/lib/desktop/widgets/remote_toolbar.dart",
    "line": 1479,
    "summary": "With `pluginItem` gone, `_DisplayMenu`'s 7-line constructor is a pure pass-through (`: super(key: key);`) that should collapse to `const _DisplayMenu({super.key, required this.id, required this.ffi, required this.state, required this.setFullscreen});`.",
    "failure_scenario": "All four remaining fields are plain finals assigned by `required this.x`, and the initializer list is now nothing but `super(key: key)` — exactly what Dart's `super.key` shorthand gives for free. Cost: boilerplate the analyzer flags under `prefer_const_constructors_in_immutables`, and the widget permanently forfeits const-canonicalization at its single call site (line 839). (The nearby `ffi.connType == ConnType.defaultConn` checks are NOT redundant — several other uses remain.)"
  },
  {
    "file": "flutter/lib/desktop/pages/desktop_setting_page.dart",
    "line": 176,
    "summary": "Removing one `SettingsTabKey` variant required four coordinated edits — the enum, `tabKeys`, and two exhaustive `switch (tab)` statements over the identical list (lines 176 and 213) — which is duplicated structure that should collapse into one table.",
    "failure_scenario": "The `plugin` variant had to be deleted from the enum (~line 56), from `tabKeys` (~line 74), from the label/icon switch at 177, and from the page-widget switch at 214. The two switches are already aligned one-to-one across all seven remaining keys, so the label, icon pair and page widget could live in a single `Map<SettingsTabKey, …>` or one switch returning a record, letting the enum exhaustiveness-check the table. Cost: every future tab add/remove can silently desync the enum from one of the two switches, and a missing case throws at runtime rather than failing to compile — this PR is the demonstration that the duplication is paid on every change."
  },
  {
    "file": "src/server/connection.rs",
    "line": 1222,
    "summary": "The `MessageInput` match cleanup leaves a nested `#[cfg(not(any(target_os = \"android\", target_os = \"ios\")))]` inside `handle_input`, which already carries that identical cfg on the fn itself (line 1158).",
    "failure_scenario": "The inner attribute at line 1222 guarding `if block_input_mode { let _ = crate::platform::block_input(true); }` is unconditionally true wherever the function compiles. The deleted `BlockOnPlugin`/`BlockOffPlugin` arms used the same doubled-cfg idiom, so the pattern was plugin-era noise; leaving one copy behind forces a reader of the recv-timeout path to re-derive that the guard can never be false. (`MessageInput::BlockOn`/`BlockOff` and `block_input_mode` itself do still earn their keep — privacy-mode senders at lines 4882/4885.)"
  },
  {
    "file": "flutter/lib/desktop/pages/desktop_home_page.dart",
    "line": 95,
    "summary": "The `children` list this PR edited still wraps a synchronously-available widget in `FutureBuilder<Widget>(future: Future.value(Obx(...)))`, so the entry rebuilds through a pointless async round-trip.",
    "failure_scenario": "`Obx(() => buildHelpCards(stateGlobal.updateUrl.value))` is evaluated eagerly as the argument to `Future.value(...)`, so the widget exists before the FutureBuilder is constructed. Every rebuild of `buildLeftPane` allocates a fresh Future and a fresh FutureBuilder element whose initState schedules a microtask, and because a new `future` identity is supplied each build the builder re-enters the `!hasData` branch and paints `const Offstage()` for one frame before settling — pure churn for a value that could be inlined. This is the same list `buildPluginEntry()` was removed from."
  }
]
```
Notable candidates I refuted rather than reported, in case they come up: Data::Plugin's removal is not an IPC upgrade hazard (src/ipc.rs:1538/1565 use serde_json, name-tagged, and the variant never existed in a shipped build); both misc::Union matches keep _ => {} catch-alls; _DisplayMenu's menuChildren can never be empty; both SettingsTabKey switches are exhaustive; the zip drop is clean and also sheds bzip2-sys/pbkdf2/password-hash from every build; removing examples/custom_plugin from workspace exclude is inert (the path was already absent on master); and the deleted per-frame self.hooks.read().unwrap() loop in on_rgba_soft_render was an unconditional RwLock acquisition on a map nothing could populate without plugin_framework — its removal is a small win, not a lost extension point.

fufesou:
https://github.com/rustdesk/rustdesk/pull/15854#issuecomment-5283197053

A few points are valid, but there are also several unrelated cleanups mixed into the review.

The two translation keys that were specific to the removed plugin UI, `Select a peer` and `Select peers`, have been removed in `0960f4427`. `Plugins` was already removed by the original commit. I deliberately kept `Uninstall`, `Enable`, `Disable`, and `Options`. They may have been added during the plugin framework work, but they are common UI terms and useful translations to keep around. `Update` is also still used by the desktop update card.

`Please install plugins` must stay as well. It is not part of `plugin_framework`; it belongs to the Windows privacy-mode flow. When `WindowInjection.dll` is missing, `win_topmost_window.rs` returns `Ok(false)`, the server sends `PrvOnFailedPlugin`, and the client displays that message. The enum name is old and a little confusing, but this code predates the general plugin framework and is still live.

`PluginRequest`, `PluginFailure`, and fields 25/26 in `hbb_common` are unused after this removal. I'm keeping the `hbb_common` submodule unchanged in this PR, so those entries will stay for now.

The old on-disk plugin directories are also intentionally left alone. This feature was never enabled in the default build or CI.

The MSI localization entries are not connected to this plugin framework. There is no corresponding WiX feature, and nothing references them. They are existing, unrelated package-localization leftovers, so I don't think they belong in this PR.

For the remaining points:

- `pt_PT.rs` not being wired up and the executable bit on `sl.rs` are both pre-existing repository issues. This PR only applies the same translation deletion across the language tables.
- The nested `cfg` and the `FutureBuilder` were already there and are unrelated to removing the framework.
- Making `_DisplayMenu`'s constructor `const` would be a style cleanup. Its current call site passes runtime values, so there is no const canonicalization to recover there.
- Replacing the settings-tab switches with a table would be a broader refactor, not part of this removal. The existing enum switches are exhaustive, so removing a variant is checked by the analyzer.

So the only code change I took from this review was removing the two plugin-specific selection strings. The protobuf cleanup is deliberately outside this PR, and I don't see the other items as blockers for removing `plugin_framework`.

PR #15855, opened 2026-08-13
fix(drm): hold off after a failed compositor lookup instead of re-probing every poll

fxd0h:
follow up to the P2 in the #15792 review. the "one blocking task, one roundtrip" half landed with the merge; this is the failure backoff half.

`scrap::wayland::display::get_displays()` caches a successful enumeration but not a failed one, and the display service polls about three times a second. so on a host where the compositor cannot be enumerated - a greeter whose socket is not answering, a drm build with no compositor to reach - every poll pays another probe process, each able to wait out the socket probe deadline.

the lookup now holds off for 5 s after an empty answer.

the part worth reviewing is what it serves during that window. the obvious answer, an empty list, is wrong: `augment_with_wayland_geometry_from` reads empty as "no compositor here" and leaves every monitor at DRM's (0, 0), so answering empty would publish a stacked desktop for 5 s where the un-backed-off code published it for one 300 ms tick. so it serves the last layout the compositor really gave instead. the poll only sends changes, so repeating the layout the client already has is silence, which is the right thing to say while you do not know. until there is a layout to repeat it just asks.

master consolidated the enumeration into two call sites, which is why this is one helper and not a per-caller split.

tested on a two head gdm greeter vm: two sessions with an abrupt disconnect between them, capture over dma-buf, probe answering two outputs, libEGL still 0 in the root service, and the uinput rect still the compositor union (0,6400),(0,2160) rather than the drm stack - so the augmentation is unaffected by the hold-off. cpu measured from /proc/pid/stat deltas, idle after the sequence. builds with and without the drm feature, 33 drm tests green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Wayland display geometry handling when compositor information is unavailable or empty.
  - Reuses the most recent valid display layout during temporary lookup failures.
  - Reduces repeated failed geometry checks and prevents empty layout information from being shown.
  - Refreshes cached display geometry when the connected display configuration changes.

### DIFF
--- a/src/server/drm_capturer.rs
+++ b/src/server/drm_capturer.rs
@@ -743,6 +743,16 @@ enum ProbeState {
 }
 
 static DRM_STATE: Mutex<ProbeState> = Mutex::new(ProbeState::Unknown);
+/// When the compositor could not be enumerated, hold off before asking again: `get_displays()`
+/// does not cache that case, and the display service polls three times a second, so a greeter
+/// whose compositor stops answering would otherwise pay a probe per poll.
+static WAYLAND_GEOMETRY_FAILED_AT: Mutex<Option<Instant>> = Mutex::new(None);
+/// The last layout the compositor actually gave, served during that hold-off. Deliberately never
+/// an empty list: `augment_with_wayland_geometry_from` reads empty as "no compositor here" and
+/// leaves every monitor at DRM's (0, 0), so answering empty would publish a stacked desktop.
+static WAYLAND_GEOMETRY_LAST_GOOD: Mutex<Option<std::sync::Arc<scrap::wayland::display::Displays>>> =
+    Mutex::new(None);
+const WAYLAND_GEOMETRY_BACKOFF: Duration = Duration::from_secs(5);
 const NEGATIVE_TTL: Duration = Duration::from_secs(30);
 const POSITIVE_TTL: Duration = Duration::from_secs(15);
 
@@ -1216,7 +1226,7 @@ pub(super) fn get_display_infos_and_primary() -> Option<(Vec<DisplayInfo>, usize
         ProbeState::Available(_, list) => list.clone(),
         _ => return None,
     };
-    let wl = scrap::wayland::display::get_displays();
+    let wl = wayland_displays_for_geometry();
     let assignment = assign_wayland_outputs(&list, &wl.displays);
     let mut infos = augment_with_wayland_geometry_from(&list, &wl, &assignment);
     mark_demoted_displays(&list, &mut infos);
@@ -1242,9 +1252,36 @@ pub(super) fn get_display_infos() -> Option<Vec<DisplayInfo>> {
 /// cannot answer, the list comes back empty and everything stays unaugmented, which is what the
 /// old is-login-screen gate produced unconditionally.
 fn augment_with_wayland_geometry(drm: &[DrmDisplayInfo]) -> Vec<DisplayInfo> {
-    let wl = scrap::wayland::display::get_displays();
+    let wl = wayland_displays_for_geometry();
     let assignment = assign_wayland_outputs(drm, &wl.displays);
     augment_with_wayland_geometry_from(drm, &wl, &assignment)
+}
+
+/// The compositor layout, without re-probing a failure on every call.
+///
+/// A failed lookup is not cached below, so the display-service poll paid one probe per turn where
+/// there is no answer to be had. While the hold-off runs, the last layout the compositor really
+/// gave is served instead: the poll only publishes CHANGES, so repeating the layout the client
+/// already has is the quiet answer, where an empty one would advertise the raw DRM stack. Until
+/// there is a layout to repeat, ask.
+fn wayland_displays_for_geometry() -> std::sync::Arc<scrap::wayland::display::Displays> {
+    let holding_off = matches!(
+        *WAYLAND_GEOMETRY_FAILED_AT.lock().unwrap(),
+        Some(failed_at) if failed_at.elapsed() < WAYLAND_GEOMETRY_BACKOFF
+    );
+    if holding_off {
+        if let Some(last_good) = WAYLAND_GEOMETRY_LAST_GOOD.lock().unwrap().clone() {
+            return last_good;
+        }
+    }
+    let wl = scrap::wayland::display::get_displays();
+    if wl.displays.is_empty() {
+        *WAYLAND_GEOMETRY_FAILED_AT.lock().unwrap() = Some(Instant::now());
+    } else {
+        *WAYLAND_GEOMETRY_FAILED_AT.lock().unwrap() = None;
+        *WAYLAND_GEOMETRY_LAST_GOOD.lock().unwrap() = Some(wl.clone());
+    }
+    wl
 }
 
 fn augment_with_wayland_geometry_from(
--- a/src/server/drm_capturer.rs
+++ b/src/server/drm_capturer.rs
@@ -750,7 +750,11 @@ static WAYLAND_GEOMETRY_FAILED_AT: Mutex<Option<Instant>> = Mutex::new(None);
 /// The last layout the compositor actually gave, served during that hold-off. Deliberately never
 /// an empty list: `augment_with_wayland_geometry_from` reads empty as "no compositor here" and
 /// leaves every monitor at DRM's (0, 0), so answering empty would publish a stacked desktop.
-static WAYLAND_GEOMETRY_LAST_GOOD: Mutex<Option<std::sync::Arc<scrap::wayland::display::Displays>>> =
+///
+/// Kept with the `DRM_STATE_GEN` it was captured under and dropped when that moves, because the
+/// assignment matches connectors to outputs by name and resolution: replaying a layout from a
+/// different topology could hand a new connector another monitor's origin.
+static WAYLAND_GEOMETRY_LAST_GOOD: Mutex<Option<(u64, std::sync::Arc<scrap::wayland::display::Displays>)>> =
     Mutex::new(None);
 const WAYLAND_GEOMETRY_BACKOFF: Duration = Duration::from_secs(5);
 const NEGATIVE_TTL: Duration = Duration::from_secs(30);
@@ -1265,21 +1269,35 @@ fn augment_with_wayland_geometry(drm: &[DrmDisplayInfo]) -> Vec<DisplayInfo> {
 /// already has is the quiet answer, where an empty one would advertise the raw DRM stack. Until
 /// there is a layout to repeat, ask.
 fn wayland_displays_for_geometry() -> std::sync::Arc<scrap::wayland::display::Displays> {
+    let gen = DRM_STATE_GEN.load(Ordering::Acquire);
+    let last_good = || {
+        WAYLAND_GEOMETRY_LAST_GOOD
+            .lock()
+            .unwrap()
+            .as_ref()
+            .filter(|(captured_gen, _)| *captured_gen == gen)
+            .map(|(_, wl)| wl.clone())
+    };
     let holding_off = matches!(
         *WAYLAND_GEOMETRY_FAILED_AT.lock().unwrap(),
         Some(failed_at) if failed_at.elapsed() < WAYLAND_GEOMETRY_BACKOFF
     );
     if holding_off {
-        if let Some(last_good) = WAYLAND_GEOMETRY_LAST_GOOD.lock().unwrap().clone() {
-            return last_good;
+        if let Some(wl) = last_good() {
+            return wl;
         }
     }
     let wl = scrap::wayland::display::get_displays();
     if wl.displays.is_empty() {
         *WAYLAND_GEOMETRY_FAILED_AT.lock().unwrap() = Some(Instant::now());
+        // This answer would leave every monitor at (0, 0) too, so the previous layout stands here
+        // as well, not only on the calls the hold-off covers.
+        if let Some(wl) = last_good() {
+            return wl;
+        }
     } else {
         *WAYLAND_GEOMETRY_FAILED_AT.lock().unwrap() = None;
-        *WAYLAND_GEOMETRY_LAST_GOOD.lock().unwrap() = Some(wl.clone());
+        *WAYLAND_GEOMETRY_LAST_GOOD.lock().unwrap() = Some((gen, wl.clone()));
     }
     wl
 }


rustdesk:
# Code review — PR #15855

https://github.com/rustdesk/rustdesk/pull/15855 — `fix/drm-geometry-backoff`, 2 commits, one file:
`src/server/drm_capturer.rs`

Two independent review passes were run on this PR. This document merges them; findings marked
*(both passes)* were reached independently by each, which is corroborating signal.

---

## The two that undercut the PR's own premise

### 1. The replay hands `get_display_infos()` logical origins with physical sizes, mis-programming uinput
`src/server/drm_capturer.rs:1295`

`src/server/wayland.rs:114` documents `drm_desktop_rect_for_uinput` as being "in scanout pixels: no
compositor here applied a scale, so unlike `desktop_rect_of` there is no logical size to handle." It
is reached (`wayland.rs:172`) only after `get_desktop_rect_for_uinput()` returned `None` — i.e.
exactly when the compositor is unreachable, which is precisely when the new replay kicks in.

Two 4K monitors at scale 2, previously enumerated so `WAYLAND_GEOMETRY_LAST_GOOD` holds logical
origins (0,0) and (1920,0), DRM list unchanged so same gen: `get_display_infos()` now returns
`x=0/w=3840` and `x=1920/w=3840` → the rect is `(0,5760,0,2160)` while the compositor's logical space
is 3840×1080. `update_mouse_resolution` programs that range and **every injected pointer coordinate
lands at ~2/3 scale**.

Before this PR the same path produced the pure DRM union `(0,3840,0,2160)` with all origins at 0.
The same arm also passes `Vec::new()` as the #15601 layout baseline, so the drift remap stays
disabled exactly when the published geometry is stale.

### 2. `get_displays()` *does* cache `Ok(vec![])` — the doc comment's premise holds only for `Err`
`src/server/drm_capturer.rs:1291`

`libs/scrap/src/wayland/display.rs:197-202` stores `Ok(vec![])` in its `DISPLAYS` cache; only the
`Err` arm is uncached. So `wl.displays.is_empty()` conflates "could not reach the compositor" with the
cached, authoritative "compositor answered with zero outputs."

A compositor that is reachable but momentarily exposes no `wl_output` global (a race at compositor
start) therefore latches an empty list in the scrap cache. From then on every call here is cheap —
there is no probe to save — yet it still stamps `FAILED_AT` and replays a stale layout for as long as
that cache entry lives, which nothing clears.

The premise in the doc comment at line 746 ("`get_displays()` does not cache that case") is only true
for the `Err` path.

---

## The hold-off does not fire on the hosts it targets

### 3. No last-good layout means no suppression at all
`src/server/drm_capturer.rs:1285` *(both passes)*

The hold-off only short-circuits when a previous successful layout exists. On a greeter or DRM build
where `get_wayland_displays()` never succeeds — the PR's stated case, "a drm build with no compositor
to reach" — `WAYLAND_GEOMETRY_LAST_GOOD` is never populated, `last_good()` returns `None`,
`holding_off` is true but control falls straight through to
`scrap::wayland::display::get_displays()` on every call.

`display_service.rs:395` polls every 300 ms, so `wayland_displays_from_runtime_dir` forks the full
rustdesk binary and waits up to `RUNTIME_DIR_PROBE_TIMEOUT` (2 s) before SIGKILLing it, ~3 forks per
second. **CPU cost after the patch is identical to before it.**

The stated rationale ("answering empty would publish a stacked desktop for 5 s where the un-backed-off
code published it for one 300 ms tick") does not hold here either: with no last-good layout the
un-backed-off code was *already* publishing the stacked desktop on every tick, and
`SYNC_DISPLAYS.check_changed` suppresses the repeat — so serving the empty answer during the hold-off
would be client-visibly identical at zero probe cost.

---

## Staleness: the invalidation contract is broken in three ways

### 4. LAST_GOOD is a second layout cache that `clear_wayland_displays_cache()` cannot reach
`src/server/drm_capturer.rs:1285` / `1295` *(both passes)*

`wayland::clear()` (`src/server/wayland.rs:405-411`) clears the scrap cache on every capturer teardown
specifically so "the next session re-reads fresh geometry … and self-heals" after a
hotplug/rotation/scale change; `check_init` (`wayland.rs:262`) and `pipewire::close_session` do the
same for rustdesk#15601.

A scale or rotation change does not alter the DRM connector list, so `DRM_STATE_GEN` does not move. If
a failure was recorded less than `WAYLAND_GEOMETRY_BACKOFF` ago, the very next `get_display_infos()`
returns at line 1287 without ever consulting the cache that was just deliberately emptied, serving the
pre-change layout instead.

Corollary: because `get_displays()` caches every success in `DISPLAYS` and never evicts except via
`clear_wayland_displays_cache()`, `wl.displays.is_empty()` can only be observed *after* an explicit
clear — so the empty-answer fallback exists solely to keep re-serving the layout that was just
explicitly invalidated. The client is published stale origins and `get_capturer_info` computes a stale
capture origin, offsetting all input.

### 5. Nothing ages LAST_GOOD out, and every expiry re-arms the hold-off
`src/server/drm_capturer.rs:1292` / `757` *(both passes)*

The only invalidation is a `DRM_STATE_GEN` change, which tracks DRM connectors, not the compositor
layout.

User logs out of a Wayland session; the greeter's compositor cannot be enumerated but the same
monitors stay attached, so the generation never moves. Every 5 s the hold-off expires,
`get_displays()` fails, line 1292 **re-stamps** `WAYLAND_GEOMETRY_FAILED_AT` and line 1295 serves the
ended session's layout again — forever. The client is told the previous user's origins, `scale` and
`original_resolution` for the whole greeter session, and input is mapped through them. Because
`check_get_displays_changed_msg` only broadcasts *changes*, the client is never corrected.

A wall-clock TTL on the entry, independent of the generation, would bound this.

### 6. `DRM_STATE_GEN` is a publish counter, not a topology identity
`src/server/drm_capturer.rs:1278` *(both passes)*

`refresh_outcome` returns `RefreshOutcome::Publish` for any non-empty probe result (line 145), and
`refresh_available_async` then calls `publish_probe_state` unconditionally (line 1067) — it computes
`changed` only to decide whether to clear the scrap cache. `refresh_displays_for_login` (line 1157)
and `refresh_unavailable_async`'s restamp (line 1022) do the same.

So every POSITIVE_TTL (15 s) refresh with an identical connector list bumps the generation,
`last_good()` starts returning `None` on the gen filter, and `wayland_displays_for_geometry` reverts to
a full probe on every 300 ms poll even though the monitors never changed.

Worse variant: `wayland::clear()` on a capturer teardown clears the scrap cache without bumping the
generation; the poll probes, fails, and correctly serves LAST_GOOD; the capture rebuild then calls
`is_available()`, whose POSITIVE_TTL staleness check fires `refresh_available_async`, which
re-publishes an IDENTICAL list and bumps the generation; the next failing poll now finds
`captured_gen != gen`, drops the fallback, and publishes every monitor at DRM's (0,0) — a stacked
desktop — even though the topology never changed.

The validity key should be the DRM list identity (both refresh paths already compute `changed`), not
the republish counter.

---

## Races

### 7. The generation is sampled before a multi-second blocking probe
`src/server/drm_capturer.rs:1272` / `1300` *(both passes)*

**Pre-hotplug layout tagged with the post-hotplug generation.** `Data::DrmDisplaysChanged` does
`swap_available_displays(list)` (gen++) and only *then* `clear_wayland_displays_cache()`
(`drm_capturer.rs:621-622`). A poll thread that loads `gen` in the window between those two statements
reads the NEW gen, calls `get_displays()`, gets the still-cached PRE-hotplug layout, and stores
`(new_gen, old_layout)`. That entry passes the gen filter for the whole life of the new topology and
is replayed on every failure hold-off, handing a new connector another monitor's origin — the exact
hazard the doc comment at line 754 says the generation prevents.

**Fresh layout discarded.** `gen = DRM_STATE_GEN.load()` reads 5 at line 1272; `get_displays()` blocks
up to 2 s in the probe child plus up to 2 s in `get_primary_monitor()`'s external commands; a hotplug
lands meanwhile and bumps to 6; the probe returns the NEW, correct layout, stored as `(5, wl)`. The
next call loads 6, `filter(|(captured_gen, _)| *captured_gen == gen)` rejects it, and the fallback is
empty again.

**Two-thread clobber.** Thread A (sampled gen 5) blocks on scrap's `DISPLAYS` mutex behind thread B
(sampled gen 6); B stores `(6, wl)`, A wakes, gets the same cached `Arc` and overwrites with
`(5, wl)`, destroying the valid entry.

The store needs a re-load-and-compare of `DRM_STATE_GEN` (or the whole read under one generation), not
the pre-probe sample.

### 8. Three separate short lock scopes let a failure arm the hold-off right after a success
`src/server/drm_capturer.rs:1299`

The display-service poll thread and a video-service thread both enter
`wayland_displays_for_geometry` (both reach it via `get_display_infos` /
`get_display_infos_and_primary`). Thread B starts first and its probe fails; thread A's probe
succeeds, sets `FAILED_AT = None` (line 1299) and stores LAST_GOOD (line 1300); B then returns and
executes line 1292, setting `FAILED_AT = Some(now)`. The hold-off is now armed even though a valid
enumeration just landed, so the next 5 s of polls replay the cached layout instead of reading the
compositor.

The mirror case — B succeeding later with an older `wl` and overwriting A's newer entry under the same
gen — is equally reachable, since the compare is only on gen, never on recency.

### 9. `get_capturer_info` augments a list the generation does not describe
`src/server/drm_capturer.rs:1509`

`get_capturer_info` builds `displays` from a live `_drm` handshake, and its own comment (line 1457)
states it deliberately does NOT publish that list into `DRM_STATE` and that it can be ordered
differently (`wire_idx != display_idx`). It then calls `augment_with_wayland_geometry(&displays)`,
which can now serve a LAST_GOOD layout validated against DRM_STATE's generation — a generation that
says nothing about the handshake list in hand.

When the handshake list is newer than DRM_STATE (which is why it is re-read: the service wakes
sleeping displays and the list changes), `assign_wayland_outputs` matches a new connector set against
the old compositor outputs by name/resolution with the gen check passing, handing a connector another
monitor's origin. The capture `origin` published to `video_service` is then wrong and all injected
coordinates land offset.

---

## Altitude

### 10. The negative cache is at the wrong layer
`src/server/drm_capturer.rs:1271` *(both passes)*

The failure that costs the probe process lives in `scrap::wayland::display::get_displays()`, which
already owns a positive cache and an invalidation entry point. A short negative TTL there would cover
every caller, stay consistent with the clear, and would not need the DRM generation at all.

As placed, the sibling paths on the same failing host keep re-probing:

- `display_service.rs:112` calls `get_layout_for_uinput_live()` every `WAYLAND_LAYOUT_CHECK_INTERVAL`
  (1500 ms) while a client is subscribed, bypassing the cache entirely and calling
  `get_wayland_displays()` directly — one 2 s probe child per 1.5 s.
- `get_desktop_rect_for_uinput` / `get_display_rects_for_uinput` (`wayland.rs:166`), whose own comment
  already documents this cost ("a failed lookup is not cached, so asking for the rects afterwards
  would rerun the whole socket probe synchronously") and whose caller then calls straight into
  `get_display_infos()`, so one hotplug pays two full 2 s probes back to back.
- `check_init` (`wayland.rs:262`), `display_service.rs:516`, `flutter_ffi.rs:1639`.

And `display_service::wayland_has_compositor` (`display_service.rs:493-497`) already works around the
identical premise with a completely different mechanism (`is_login_screen_wayland_cached`).

Because the new static lives outside scrap, `clear_wayland_displays_cache()` also stops being a
complete invalidation — see finding 4.

### 11. Eight reachable branches, no test, in a file whose idiom is the opposite
`src/server/drm_capturer.rs:1271` *(both passes)*

`refresh_outcome(probe, failures) -> RefreshOutcome` (line 142) exists precisely so this class of
policy can be tested without globals, and
`refresh_keeps_a_verdict_through_one_failure_and_gives_it_up_after_a_run` (line 1842) covers it.

The new state machine — hold-off armed/expired × gen match/mismatch × empty/non-empty × last-good
present/absent — sits entirely on three process-global statics plus a direct call to the real
`scrap::wayland::display::get_displays()`, so none of its eight reachable combinations can be
exercised without a live compositor. None of the 33 tests the PR reports green touch it. Splitting the
decision into a pure function over `(failed_at, now, gen, cached_gen, probe_result)` would have caught
findings 1, 3 and 6.

---

## Comments, logging, style

### 12. The doc comment of the touched function now states the opposite of what the code does
`src/server/drm_capturer.rs:1256`

Lines 1252-1257 still say "Where that fallback cannot answer, the list comes back empty and everything
stays unaugmented, which is what the old is-login-screen gate produced unconditionally." After this PR
a failed lookup returns the previous layout and everything IS augmented.
`drm_desktop_rect_for_uinput`'s own doc at `wayland.rs:110-111` makes the same now-false assumption —
which is how finding 1 slipped through.

### 13. A stale substitution is never logged
`src/server/drm_capturer.rs:1296`

A host stuck serving a pre-invalidation layout produces logs identical to a healthy one apart from
scrap's generic "Failed to get wayland displays" warning, which predates this change and does not say
a substitution happened. A user reporting "monitor positions are wrong after I rescaled a display"
gives a log in which nothing distinguishes the stale-replay path from a fresh read. A one-line
`log::debug!` on the *first* substitution after a success (not per poll) would make the state visible.

### 14. The new comments break the AGENTS.md comment rules
`src/server/drm_capturer.rs:750` *(both passes)*

AGENTS.md / Editing Hygiene / Comments (referenced by `CLAUDE.md`):

> Keep them short: one line by default, three at most.
> A comment must never be longer than the code it describes.
> Do not document rejected alternatives, past bugs, measurements, or how you arrived at the code.
> That belongs in the commit message or the PR.

- Lines 750-756: a 7-line doc comment over a 2-line static (757-758).
- Lines 1264-1270: a 7-line doc comment that spends most of its length on the rejected alternative
  ("where an empty one would advertise the raw DRM stack") and on past behavior ("the display-service
  poll paid one probe per turn").
- Lines 746-748: a 3-line doc comment for a 1-line static. Lines 1293-1294: same pattern.

### 15. Duplicated early-return and a `holding_off` computed too early
`src/server/drm_capturer.rs:1281`

Lines 1285-1289 and 1293-1297 are the identical `if let Some(wl) = last_good() { return wl; }`, and
`holding_off` is computed before it is known whether there is anything to hold off with — on a host
with no last-good layout the mutex is taken and the elapsed time compared on every one of the ~3 polls
per second for a value that cannot change the outcome. Reading LAST_GOOD once at the top and branching
on `(last_good, holding_off)` expresses the whole policy in one place and makes finding 3's behavior
visible rather than implicit.

### 16. `std::sync::Arc` written fully qualified
`src/server/drm_capturer.rs:757`

Line 13 is `use std::sync::{Arc, Condvar, Mutex};` and the rest of the file uses the bare `Arc` (e.g.
line 1537). The fully-qualified form in the static (757) and in the return type of
`wayland_displays_for_geometry` (1271) is what pushes line 757 past the width rustfmt can wrap, and
diverges from the surrounding style AGENTS.md asks to keep consistent.

---

## Suggested order

1. **#2** first — it invalidates the premise behind the whole empty-answer branch.
2. **#1** — the only finding with direct user-visible input breakage.
3. **#3** — without it the PR does not achieve its stated goal.

**#10** would subsume #3, #4 and #6 if moving the negative cache into
`scrap::wayland::display::get_displays()` is acceptable.

~/rustdesk
> cat review.md
# Code review — PR #15855

https://github.com/rustdesk/rustdesk/pull/15855 — `fix/drm-geometry-backoff`, 2 commits, one file:
`src/server/drm_capturer.rs`

Two independent review passes were run on this PR. This document merges them; findings marked
*(both passes)* were reached independently by each, which is corroborating signal.

---

## The two that undercut the PR's own premise

### 1. The replay hands `get_display_infos()` logical origins with physical sizes, mis-programming uinput
`src/server/drm_capturer.rs:1295`

`src/server/wayland.rs:114` documents `drm_desktop_rect_for_uinput` as being "in scanout pixels: no
compositor here applied a scale, so unlike `desktop_rect_of` there is no logical size to handle." It
is reached (`wayland.rs:172`) only after `get_desktop_rect_for_uinput()` returned `None` — i.e.
exactly when the compositor is unreachable, which is precisely when the new replay kicks in.

Two 4K monitors at scale 2, previously enumerated so `WAYLAND_GEOMETRY_LAST_GOOD` holds logical
origins (0,0) and (1920,0), DRM list unchanged so same gen: `get_display_infos()` now returns
`x=0/w=3840` and `x=1920/w=3840` → the rect is `(0,5760,0,2160)` while the compositor's logical space
is 3840×1080. `update_mouse_resolution` programs that range and **every injected pointer coordinate
lands at ~2/3 scale**.

Before this PR the same path produced the pure DRM union `(0,3840,0,2160)` with all origins at 0.
The same arm also passes `Vec::new()` as the #15601 layout baseline, so the drift remap stays
disabled exactly when the published geometry is stale.

### 2. `get_displays()` *does* cache `Ok(vec![])` — the doc comment's premise holds only for `Err`
`src/server/drm_capturer.rs:1291`

`libs/scrap/src/wayland/display.rs:197-202` stores `Ok(vec![])` in its `DISPLAYS` cache; only the
`Err` arm is uncached. So `wl.displays.is_empty()` conflates "could not reach the compositor" with the
cached, authoritative "compositor answered with zero outputs."

A compositor that is reachable but momentarily exposes no `wl_output` global (a race at compositor
start) therefore latches an empty list in the scrap cache. From then on every call here is cheap —
there is no probe to save — yet it still stamps `FAILED_AT` and replays a stale layout for as long as
that cache entry lives, which nothing clears.

The premise in the doc comment at line 746 ("`get_displays()` does not cache that case") is only true
for the `Err` path.

---

## The hold-off does not fire on the hosts it targets

### 3. No last-good layout means no suppression at all
`src/server/drm_capturer.rs:1285` *(both passes)*

The hold-off only short-circuits when a previous successful layout exists. On a greeter or DRM build
where `get_wayland_displays()` never succeeds — the PR's stated case, "a drm build with no compositor
to reach" — `WAYLAND_GEOMETRY_LAST_GOOD` is never populated, `last_good()` returns `None`,
`holding_off` is true but control falls straight through to
`scrap::wayland::display::get_displays()` on every call.

`display_service.rs:395` polls every 300 ms, so `wayland_displays_from_runtime_dir` forks the full
rustdesk binary and waits up to `RUNTIME_DIR_PROBE_TIMEOUT` (2 s) before SIGKILLing it, ~3 forks per
second. **CPU cost after the patch is identical to before it.**

The stated rationale ("answering empty would publish a stacked desktop for 5 s where the un-backed-off
code published it for one 300 ms tick") does not hold here either: with no last-good layout the
un-backed-off code was *already* publishing the stacked desktop on every tick, and
`SYNC_DISPLAYS.check_changed` suppresses the repeat — so serving the empty answer during the hold-off
would be client-visibly identical at zero probe cost.

---

## Staleness: the invalidation contract is broken in three ways

### 4. LAST_GOOD is a second layout cache that `clear_wayland_displays_cache()` cannot reach
`src/server/drm_capturer.rs:1285` / `1295` *(both passes)*

`wayland::clear()` (`src/server/wayland.rs:405-411`) clears the scrap cache on every capturer teardown
specifically so "the next session re-reads fresh geometry … and self-heals" after a
hotplug/rotation/scale change; `check_init` (`wayland.rs:262`) and `pipewire::close_session` do the
same for rustdesk#15601.

A scale or rotation change does not alter the DRM connector list, so `DRM_STATE_GEN` does not move. If
a failure was recorded less than `WAYLAND_GEOMETRY_BACKOFF` ago, the very next `get_display_infos()`
returns at line 1287 without ever consulting the cache that was just deliberately emptied, serving the
pre-change layout instead.

Corollary: because `get_displays()` caches every success in `DISPLAYS` and never evicts except via
`clear_wayland_displays_cache()`, `wl.displays.is_empty()` can only be observed *after* an explicit
clear — so the empty-answer fallback exists solely to keep re-serving the layout that was just
explicitly invalidated. The client is published stale origins and `get_capturer_info` computes a stale
capture origin, offsetting all input.

### 5. Nothing ages LAST_GOOD out, and every expiry re-arms the hold-off
`src/server/drm_capturer.rs:1292` / `757` *(both passes)*

The only invalidation is a `DRM_STATE_GEN` change, which tracks DRM connectors, not the compositor
layout.

User logs out of a Wayland session; the greeter's compositor cannot be enumerated but the same
monitors stay attached, so the generation never moves. Every 5 s the hold-off expires,
`get_displays()` fails, line 1292 **re-stamps** `WAYLAND_GEOMETRY_FAILED_AT` and line 1295 serves the
ended session's layout again — forever. The client is told the previous user's origins, `scale` and
`original_resolution` for the whole greeter session, and input is mapped through them. Because
`check_get_displays_changed_msg` only broadcasts *changes*, the client is never corrected.

A wall-clock TTL on the entry, independent of the generation, would bound this.

### 6. `DRM_STATE_GEN` is a publish counter, not a topology identity
`src/server/drm_capturer.rs:1278` *(both passes)*

`refresh_outcome` returns `RefreshOutcome::Publish` for any non-empty probe result (line 145), and
`refresh_available_async` then calls `publish_probe_state` unconditionally (line 1067) — it computes
`changed` only to decide whether to clear the scrap cache. `refresh_displays_for_login` (line 1157)
and `refresh_unavailable_async`'s restamp (line 1022) do the same.

So every POSITIVE_TTL (15 s) refresh with an identical connector list bumps the generation,
`last_good()` starts returning `None` on the gen filter, and `wayland_displays_for_geometry` reverts to
a full probe on every 300 ms poll even though the monitors never changed.

Worse variant: `wayland::clear()` on a capturer teardown clears the scrap cache without bumping the
generation; the poll probes, fails, and correctly serves LAST_GOOD; the capture rebuild then calls
`is_available()`, whose POSITIVE_TTL staleness check fires `refresh_available_async`, which
re-publishes an IDENTICAL list and bumps the generation; the next failing poll now finds
`captured_gen != gen`, drops the fallback, and publishes every monitor at DRM's (0,0) — a stacked
desktop — even though the topology never changed.

The validity key should be the DRM list identity (both refresh paths already compute `changed`), not
the republish counter.

---

## Races

### 7. The generation is sampled before a multi-second blocking probe
`src/server/drm_capturer.rs:1272` / `1300` *(both passes)*

**Pre-hotplug layout tagged with the post-hotplug generation.** `Data::DrmDisplaysChanged` does
`swap_available_displays(list)` (gen++) and only *then* `clear_wayland_displays_cache()`
(`drm_capturer.rs:621-622`). A poll thread that loads `gen` in the window between those two statements
reads the NEW gen, calls `get_displays()`, gets the still-cached PRE-hotplug layout, and stores
`(new_gen, old_layout)`. That entry passes the gen filter for the whole life of the new topology and
is replayed on every failure hold-off, handing a new connector another monitor's origin — the exact
hazard the doc comment at line 754 says the generation prevents.

**Fresh layout discarded.** `gen = DRM_STATE_GEN.load()` reads 5 at line 1272; `get_displays()` blocks
up to 2 s in the probe child plus up to 2 s in `get_primary_monitor()`'s external commands; a hotplug
lands meanwhile and bumps to 6; the probe returns the NEW, correct layout, stored as `(5, wl)`. The
next call loads 6, `filter(|(captured_gen, _)| *captured_gen == gen)` rejects it, and the fallback is
empty again.

**Two-thread clobber.** Thread A (sampled gen 5) blocks on scrap's `DISPLAYS` mutex behind thread B
(sampled gen 6); B stores `(6, wl)`, A wakes, gets the same cached `Arc` and overwrites with
`(5, wl)`, destroying the valid entry.

The store needs a re-load-and-compare of `DRM_STATE_GEN` (or the whole read under one generation), not
the pre-probe sample.

### 8. Three separate short lock scopes let a failure arm the hold-off right after a success
`src/server/drm_capturer.rs:1299`

The display-service poll thread and a video-service thread both enter
`wayland_displays_for_geometry` (both reach it via `get_display_infos` /
`get_display_infos_and_primary`). Thread B starts first and its probe fails; thread A's probe
succeeds, sets `FAILED_AT = None` (line 1299) and stores LAST_GOOD (line 1300); B then returns and
executes line 1292, setting `FAILED_AT = Some(now)`. The hold-off is now armed even though a valid
enumeration just landed, so the next 5 s of polls replay the cached layout instead of reading the
compositor.

The mirror case — B succeeding later with an older `wl` and overwriting A's newer entry under the same
gen — is equally reachable, since the compare is only on gen, never on recency.

### 9. `get_capturer_info` augments a list the generation does not describe
`src/server/drm_capturer.rs:1509`

`get_capturer_info` builds `displays` from a live `_drm` handshake, and its own comment (line 1457)
states it deliberately does NOT publish that list into `DRM_STATE` and that it can be ordered
differently (`wire_idx != display_idx`). It then calls `augment_with_wayland_geometry(&displays)`,
which can now serve a LAST_GOOD layout validated against DRM_STATE's generation — a generation that
says nothing about the handshake list in hand.

When the handshake list is newer than DRM_STATE (which is why it is re-read: the service wakes
sleeping displays and the list changes), `assign_wayland_outputs` matches a new connector set against
the old compositor outputs by name/resolution with the gen check passing, handing a connector another
monitor's origin. The capture `origin` published to `video_service` is then wrong and all injected
coordinates land offset.

---

## Altitude

### 10. The negative cache is at the wrong layer
`src/server/drm_capturer.rs:1271` *(both passes)*

The failure that costs the probe process lives in `scrap::wayland::display::get_displays()`, which
already owns a positive cache and an invalidation entry point. A short negative TTL there would cover
every caller, stay consistent with the clear, and would not need the DRM generation at all.

As placed, the sibling paths on the same failing host keep re-probing:

- `display_service.rs:112` calls `get_layout_for_uinput_live()` every `WAYLAND_LAYOUT_CHECK_INTERVAL`
  (1500 ms) while a client is subscribed, bypassing the cache entirely and calling
  `get_wayland_displays()` directly — one 2 s probe child per 1.5 s.
- `get_desktop_rect_for_uinput` / `get_display_rects_for_uinput` (`wayland.rs:166`), whose own comment
  already documents this cost ("a failed lookup is not cached, so asking for the rects afterwards
  would rerun the whole socket probe synchronously") and whose caller then calls straight into
  `get_display_infos()`, so one hotplug pays two full 2 s probes back to back.
- `check_init` (`wayland.rs:262`), `display_service.rs:516`, `flutter_ffi.rs:1639`.

And `display_service::wayland_has_compositor` (`display_service.rs:493-497`) already works around the
identical premise with a completely different mechanism (`is_login_screen_wayland_cached`).

Because the new static lives outside scrap, `clear_wayland_displays_cache()` also stops being a
complete invalidation — see finding 4.

### 11. Eight reachable branches, no test, in a file whose idiom is the opposite
`src/server/drm_capturer.rs:1271` *(both passes)*

`refresh_outcome(probe, failures) -> RefreshOutcome` (line 142) exists precisely so this class of
policy can be tested without globals, and
`refresh_keeps_a_verdict_through_one_failure_and_gives_it_up_after_a_run` (line 1842) covers it.

The new state machine — hold-off armed/expired × gen match/mismatch × empty/non-empty × last-good
present/absent — sits entirely on three process-global statics plus a direct call to the real
`scrap::wayland::display::get_displays()`, so none of its eight reachable combinations can be
exercised without a live compositor. None of the 33 tests the PR reports green touch it. Splitting the
decision into a pure function over `(failed_at, now, gen, cached_gen, probe_result)` would have caught
findings 1, 3 and 6.

---

## Comments, logging, style

### 12. The doc comment of the touched function now states the opposite of what the code does
`src/server/drm_capturer.rs:1256`

Lines 1252-1257 still say "Where that fallback cannot answer, the list comes back empty and everything
stays unaugmented, which is what the old is-login-screen gate produced unconditionally." After this PR
a failed lookup returns the previous layout and everything IS augmented.
`drm_desktop_rect_for_uinput`'s own doc at `wayland.rs:110-111` makes the same now-false assumption —
which is how finding 1 slipped through.

### 13. A stale substitution is never logged
`src/server/drm_capturer.rs:1296`

A host stuck serving a pre-invalidation layout produces logs identical to a healthy one apart from
scrap's generic "Failed to get wayland displays" warning, which predates this change and does not say
a substitution happened. A user reporting "monitor positions are wrong after I rescaled a display"
gives a log in which nothing distinguishes the stale-replay path from a fresh read. A one-line
`log::debug!` on the *first* substitution after a success (not per poll) would make the state visible.

### 14. The new comments break the AGENTS.md comment rules
`src/server/drm_capturer.rs:750` *(both passes)*

AGENTS.md / Editing Hygiene / Comments (referenced by `CLAUDE.md`):

> Keep them short: one line by default, three at most.
> A comment must never be longer than the code it describes.
> Do not document rejected alternatives, past bugs, measurements, or how you arrived at the code.
> That belongs in the commit message or the PR.

- Lines 750-756: a 7-line doc comment over a 2-line static (757-758).
- Lines 1264-1270: a 7-line doc comment that spends most of its length on the rejected alternative
  ("where an empty one would advertise the raw DRM stack") and on past behavior ("the display-service
  poll paid one probe per turn").
- Lines 746-748: a 3-line doc comment for a 1-line static. Lines 1293-1294: same pattern.

### 15. Duplicated early-return and a `holding_off` computed too early
`src/server/drm_capturer.rs:1281`

Lines 1285-1289 and 1293-1297 are the identical `if let Some(wl) = last_good() { return wl; }`, and
`holding_off` is computed before it is known whether there is anything to hold off with — on a host
with no last-good layout the mutex is taken and the elapsed time compared on every one of the ~3 polls
per second for a value that cannot change the outcome. Reading LAST_GOOD once at the top and branching
on `(last_good, holding_off)` expresses the whole policy in one place and makes finding 3's behavior
visible rather than implicit.

### 16. `std::sync::Arc` written fully qualified
`src/server/drm_capturer.rs:757`

Line 13 is `use std::sync::{Arc, Condvar, Mutex};` and the rest of the file uses the bare `Arc` (e.g.
line 1537). The fully-qualified form in the static (757) and in the return type of
`wayland_displays_for_geometry` (1271) is what pushes line 757 past the width rustfmt can wrap, and
diverges from the surrounding style AGENTS.md asks to keep consistent.

---

## Suggested order

1. **# 2** first — it invalidates the premise behind the whole empty-answer branch.
2. **# 1** — the only finding with direct user-visible input breakage.
3. **# 3** — without it the PR does not achieve its stated goal.

**# 10** would subsume # 3, # 4 and # 6 if moving the negative cache into
`scrap::wayland::display::get_displays()` is acceptable.

fxd0h:
point 3 sinks it - on the host this pr names there is never a prior good layout, so the hold-off never fires and the patch does not do what it claims. no argument on the rest, the replay cache in particular should not have existed.

the sibling caller list in 10 settled it. closing this. the negative ttl belongs in get_displays next to the positive cache like you said, and i will send that as a small pr, with the decision split into a pure function plus a test.

PR #15860, opened 2026-08-14
fix(wayland): wait before retrying an enumeration that just failed

fxd0h:
the negative cache where you said it belongs in #15855 point 10, replacing that pr.

`get_wayland_displays` has exactly two callers, both in this file: `get_displays`, which caches a success and is reached by the ~300 ms display-service poll, and `get_layout_for_uinput_live`, which bypasses the cache and runs every 1.5 s while a client is subscribed. neither caches a failure, so where there is no compositor to reach both fork the probe subprocess and wait out its deadline to learn the same thing again. wrapping the shared function covers both, which the drm-side attempt did not.

a refused probe reports a failure. that is what both callers already do with one, so nothing downstream sees a state it did not see before, and the layout cache above keeps serving the last good answer exactly as it did - no replay, no second cache, no generation key.

`clear_wayland_displays_cache` drops the stamp with the layout, so an explicit invalidation still takes a fresh read.

scope: the delay is armed by the `Err` arm only. `Ok(vec![])` still caches as before - i am not changing enumeration semantics here, only what happens when the lookup fails.

measured on a two head gdm greeter vm with the compositor socket moved out of the scan path the probe uses, one client subscribed, 60 s windows:

| | failed lookups | refused by the delay | probe forks |
|---|---|---|---|
| master | 162 | 0 | 162 |
| this | 201 | 188 | 13 |

the call count goes up because a refused lookup returns immediately and the pollers turn faster; the fork count is what drops. restoring the socket returns it to zero refusals.

the decision is a pure function over (failed_at, now, delay) with tests, including a stamp from the future, which saturating_duration_since turns into a wait rather than a panic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Wayland display detection reliability by preventing repeated retries after failed attempts.
  * Display detection now retries automatically after a short cooldown.
  * Resetting display information immediately allows fresh detection attempts.
  * Prevented outdated in-progress detection results from overriding newer display state.

<details open><summary><h3>Greptile Summary</h3></summary>

The PR adds a five-second negative cache for failed Wayland display enumeration and generation-based coordination with explicit cache invalidation. The attempted coordination still leaves a check-to-write race that can restore a stale failure after invalidation.
- Routes both cached and live display enumeration through the shared retry delay.
- Clears the negative-cache timestamp during explicit invalidation.
- Adds generation checks intended to discard failures from probes overlapping invalidation.
- Adds unit tests for retry timing and generation comparisons.
</details>

[...]

PR #15865, opened 2026-08-14
fix(wayland): back off the polling display lookups after a failure (drm)

rustdesk:
In drm builds an enumeration that fails **with no endpoint named in the environment** falls back to the socket probe, which forks a child bounded by seconds, and the display service asks again every 300 ms -- at a greeter with no reachable compositor that is a probe child per turn, forever. Such a failure now stamps a shared 5 s backoff, and **only the polling callers honor it**: the 300 ms displays-changed check skips its turn, and the 1.5 s live layout poll returns no answer for that turn.

Two scoping rules keep the backoff exactly where the fork cost is:

- **Only the failure that would fork stamps.** A session server is spawned with `WAYLAND_DISPLAY` set, so its failed connect bails in-process before any fork (`wayland_probe` refuses the fallback for named endpoints). Stamping there would buy nothing and cost recovery latency (the #15601 drift-remap refresh would stretch from ~1.5 s to ~6.5 s) -- so live sessions keep master's behavior exactly.
- **The stamp survives `clear_wayland_displays_cache`.** It describes the seat, not the cache: the ~1/s capturer rebuild loop clears on every teardown, and dropping the stamp with the cache would let that loop defeat the backoff (a fork per second again) and would turn every post-hotplug failure into a "first" one, withholding new topology indefinitely during a flap. A unit test pins this.

The displays-changed check weighs the backoff against what is already published:

- **Nothing synced yet** → always populate: an unaugmented DRM list beats the empty broadcast the send path would otherwise emit (`SyncDisplaysInfo::default()` would broadcast "No displays").
- **Synced layout, suppressed turn** → keep it.
- **Synced layout, fresh first failure** → keep it for one backoff: a transient failure never tears down a good layout.
- **Failure persisting across a backoff** → replace with the DRM list: a hotplug at a failing seat converges within one backoff, and a logged-out seat never serves the dead session's layout forever.

One-shot callers -- session init, pipewire stream setup, capturer info -- keep probing fresh through `get_displays`, whose failure semantics are unchanged: replaying a transient failure there would latch an empty answer into session-long state (stream positions at (0,0), logical scale stuck at 1.0). This poll-only scoping is what distinguishes this from the earlier attempts (#15855, #15860), which suppressed every caller.

Non-drm builds compile none of this (`cfg`), and the backoff unit tests are `cfg(any(test, feature = "drm"))` so they run under the plain `cargo test --workspace` CI job.

Known, accepted bounds: a lookup in flight across a concurrent success is last-write-wins on the stamp (bounded at one 5 s window, self-healing); the probe module's busy-bail stamps like a real failure at unnamed-endpoint seats (bounded the same way); and the pre-existing double probe in `update_uinput_resolution`'s fallback is untouched.

Verified: backoff boundaries, a future-dated stamp, stamp-survives-clear, and the env gate (named / unnamed / empty-valued endpoint) compile and behave correctly on the MSRV toolchain (1.75); rustfmt-clean against the file's baseline. Needs a Linux CI run for the full build and test pass.

This PR adds a DRM-specific five-second backoff after unnamed Wayland display lookup failures while keeping one-shot callers on fresh enumeration.
- Preserves synchronized display state during suppressed polling turns.
- Allows persistent failures to converge to DRM-provided topology after a backoff period.
- Adds tests for timing boundaries, future timestamps, and cache-clear behavior.
</details>

The PR appears safe to merge because no blocking failure remains.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| libs/scrap/src/wayland/display.rs | Adds DRM-scoped failure timestamping, polling suppression, warning throttling, and unit coverage while retaining fresh enumeration for one-shot callers. |
| src/server/display_service.rs | Integrates polling suppression with synchronized-layout preservation and eventual DRM topology replacement. |

</details>

```mermaid
flowchart TD
    A[Display poll] --> B{DRM cache available?}
    B -- No --> G[Return current display message]
    B -- Yes --> C{Lookup suppressed?}
    C -- Yes and layout synced --> G
    C -- No --> D[Enumerate display information]
    D --> E{Lookup succeeds?}
    E -- Yes --> F[Update synchronized layout]
    E -- No --> H[Record five-second failure stamp]
    H --> I{First failure with layout synced?}
    I -- Yes --> G
    I -- No --> F
    F --> G
```
</details>

<sub>Reviews (9): Last reviewed commit: ["fix(wayland): reset lookup warning after..."](https://github.com/rustdesk/rustdesk/commit/889cfeb1ccf86170147a8c4f67501316cb1edefa) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53298222)</sub>

### DIFF
--- a/libs/scrap/src/wayland/display.rs
+++ b/libs/scrap/src/wayland/display.rs
@@ -19,6 +19,14 @@ static MISSING_LOGICAL_SIZE_WARNED: std::sync::atomic::AtomicBool =
 
 const COMMAND_TIMEOUT: Duration = Duration::from_millis(1000);
 
+// drm builds only: an unnamed-endpoint failure there forks the probe child, and the pollers
+// turn every few hundred milliseconds. Every other failure is one cheap in-process error.
+#[cfg(any(test, feature = "drm"))]
+const FAILED_LOOKUP_BACKOFF: Duration = Duration::from_secs(5);
+
+#[cfg(any(test, feature = "drm"))]
+static LAST_FAILED_LOOKUP: Mutex<Option<Instant>> = Mutex::new(None);
+
 pub struct Displays {
     pub primary: usize,
     pub displays: Vec<WaylandDisplayInfo>,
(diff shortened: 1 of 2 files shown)


rustdesk:
@fxd0h please help review this one.

fxd0h (approved):
went through it with the four cases from the body traced against the code, plus a build and test pass on this head: the three new tests pass (18 in the module), scrap checks clean with drm, the full lib checks clean with flutter plus drm, and a non drm check stays clean too.

the design closes both holes my attempts had. one shot callers keep probing fresh through get_displays, so a transient failure cannot latch into session state, and keeping the stamp across clear_wayland_displays_cache resolves the invalidation race without any generation counter. reading endpoint_named before the connect that consumes WAYLAND_SOCKET is a catch i had missed entirely.

two observations, neither blocking:

- the body says a transient failure never tears down a good layout, but that guarantee only covers the unnamed endpoint case: a named endpoint failure does not stamp, so one failed 300 ms turn with a synced layout still swaps in the drm stack and rebroadcasts. the old code did that on every turn, so this is strictly narrower than before, just narrower than the body reads too. maybe worth one line there.

- when the drm list itself is empty, check_changed keeps SYNC_DISPLAYS empty, synced stays false, and the not synced arm enumerates every turn with the backoff never consulted, forks included. reaching it needs a drm available seat with zero active displays and an unreachable compositor at the same time, which i could not produce on any of my boxes, so this is structural rather than demonstrated. gating that arm on the backoff as well would close it if you think the state is worth the line.

if numbers for the body are useful i can run this head on the bench from #15860: greeter vm with the compositor socket out of the scan path, 60 s windows. master measured 162 probe forks a minute there.

Issue #3129, opened 2023-02-08
[Bug] Prevents local printer installation after Rustdesk is installed


dasruralgeek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug Description

Operating System: KDE neon 5.26
KDE Plasma Version: 5.26.5
KDE Frameworks Version: 5.102.0
Qt Version: 5.15.8
Kernel Version: 5.19.0-17.2-liquorix-amd64 (64-bit)
Graphics Platform: X11
Processors: 8 × Intel® Core™ i7-7700 CPU @ 3.60GHz
Memory: 31.2 GiB of RAM
Graphics Processor: Mesa Intel® HD Graphics 630
Manufacturer: Dell Inc.
Product Name: OptiPlex 7050

Rustdesk Version 1.1.9  Prevents local printer installation after Rustdesk is installed. System will locate network printers. But wiil not allow printer database to load or manually browse PPD file. Issue effects both GUI printer interface and CUPS web interface. (localhost:631).  This may also effect Thunderbirds add account wizard. Does not effect printers installed prior to installing Rustdesk. 
Current workaround: Remove Rustdesk, install printer, reinstall Rustdesk.
also , Thunderbird account addition works normally when Rustdesk is removed  

Issue was also reproduced on Various HP laptops and desktops HP840G5, HP650G9, HP800G1AiO

### How to Reproduce

Rustdesk Version 1.1.9  Prevents local printer installation after Rustdesk is installed. System will locate network printers. But wiil not allow printer database to load or manually browse PPD file. Issue effects both GUI printer interface and CUPS web interface. (localhost:631).  This may also effect Thunderbirds add account wizard. Does not effect printers installed prior to installing Rustdesk. 

### Expected Behavior

System will locate network printers. But wiil not allow printer database to load or manually browse PPD file. Issue effects both GUI printer interface and CUPS web interface. (localhost:631). 

### Operating system(s) on local side and remote side

Linux to Linux 

### RustDesk Version(s) on local side and remote side

1.1.9-1.1.9

### Additional Context

Current workaround: Remove Rustdesk, install printer, reinstall Rustdesk.

rustdesk:
Sadly, totally no clue yet.

Kundanagrawalofficial:
Ok will try on it if you give me chance

DeCiZoR:
I have reproduced same bug with:
**Ubuntu 22.04 JAMMY** 
(gnome-shell 42.5)
[...]

PR #15866, opened 2026-08-14
refact: remove linux headless

fufesou:
This PR requires https://github.com/rustdesk/hbb_common/pull/584 to be merged first.

Headless works in 1.2.3 but fails in 1.2.4 and later.  https://github.com/rustdesk/rustdesk/releases/tag/1.2.3

https://github.com/rustdesk/rustdesk/wiki/FAQ#headless-linux-support----dropped-as-of-version-124

## Testing

- [x] Github CI https://github.com/fufesou/rustdesk/actions/runs/31783671795
- [x] Greeter screen, logged-in screen, X11, and Wayland.   Input, audio, clipboard.

## Summary by CodeRabbit

- **Changes**
  - Removed Linux headless mode and automatic desktop-session management.
  - Simplified Linux login flows by removing OS-account setup, account remembering, and session re-login prompts.
  - Connection Manager startup now uses the standard connection flow.

- **Packaging**
  - Linux packages and build instructions no longer require PAM components.

- **Documentation**
  - Updated multilingual Linux build instructions to reflect the revised dependencies and packaging requirements.

The PR removes Linux headless desktop-session management and its PAM, packaging, protocol, and UI integration.
- Removes the Linux virtual desktop and OS-account login flows.
- Simplifies Connection Manager startup and login handling.
- Removes PAM dependencies and headless resources from supported package formats.
- Updates Flutter, Sciter, localization, and build documentation accordingly.
</details>

The PR appears safe to merge because no blocking failure remains in the eligible follow-up scope.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| src/server/connection.rs | Removes Linux headless authentication and desktop-start coordination while routing accepted sessions through standard Connection Manager IPC. |
| src/platform/linux.rs | Removes virtual headless-session detection, subprocess cleanup, and related service-loop behavior. |
| src/platform/linux_desktop_manager.rs | Deletes the obsolete PAM-backed Linux virtual desktop session manager. |
| src/client.rs | Removes desktop-session login errors and limits OS-login payloads to terminal connections. |
| flutter/lib/common/widgets/dialog.dart | Removes stored OS-account controls while retaining terminal administrator credential dialogs. |
| build.py | Stops packaging PAM configuration and virtual X-session resources. |
| Cargo.toml | Removes the PAM crate dependency associated with Linux headless login. |
| src/server/drm_capturer.rs | Retains non-blocking tri-state DRM availability routing with updated visibility and documentation. |

</details>

```mermaid
flowchart LR
    A[Incoming connection] --> B[Authenticate RustDesk credentials]
    B --> C[Start standard Connection Manager IPC]
    C --> D[Use active Linux desktop session]
    D --> E[Capture and input services]
    F[Removed headless flow] -. no longer starts .-> G[Virtual X session and PAM OS login]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix(linux): remove unreachable session c..."](https://github.com/rustdesk/rustdesk/commit/5bed065544c7accb6fe0f79427a561e20f58049f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53315357)</sub>


fufesou:
@coderabbitai review

fufesou:
@codex review

fufesou:
@codex review

fufesou:
@coderabbitai review

rustdesk:
# Code Review — PR #15866 `refact: remove linux headless`

**Target:** https://github.com/rustdesk/rustdesk/pull/15866
**Range:** merge-base `d829d14` → `pr15866`
**Findings:** 7 (ordered most-severe first)

Every finding below was verified against the real diff: master's behavior at
`d829d14` and the PR's behavior were each read from source for the described
scenario, and only differences **caused by removing headless support** are kept.
Three candidates were dropped during that pass as pre-existing on master or
unrelated to headless, and are not reported here.

## Scope verified

The removal sweep itself is clean. Symbol-level sweeps found no leftovers for
`linux_desktop_manager`, `is_headless_allowed`, `OPTION_ALLOW_LINUX_HEADLESS`,
`--cm-no-ui`, `LOGIN_MSG_DESKTOP_*`, `startwm.sh`, `xorg.conf`, `pam`, or
`main_show_option`. All 89 lang files removed exactly the same 10 keys. The
`hbb_common` submodule bump to `b2b1ac4` is the tip of that repo's `main`, so the
dependency PR is merged.

`allow-linux-headless` was opt-in and default-off. That matters throughout: on a
default-configured host, most of the deleted machinery was already a no-op, so
several of the findings below are transition-only rather than steady-state.

## Summary

| # | Site | Class | Impact |
| --- | --- | --- | --- |
| 1 | `src/client.rs:3343` | Cross-version compat | Dead-end error against an older headless-enabled peer |
| 2 | `src/platform/linux.rs:937` | Upgrade transition | Orphaned Xorg / CM survive until reboot |
| 3 | `src/client.rs:3618` | Cross-version compat | Cannot supply OS credentials to an older peer |
| 4 | `src/core_main.rs:715` | Upgrade transition | Stray control-panel window in the upgrade window |
| 5 | `src/server/drm_capturer.rs:832,848,883` | Stale docs | Misleading; no runtime effect |
| 6 | `src/server/connection.rs:446` | Misleading comment | Comment false; behavior unchanged |
| 7 | `src/platform/linux.rs:1064` | Cadence note | +200ms worst case; arguably the intended behavior |

---

## 1. `LOGIN_MSG_DESKTOP_*` dropped from `LOGIN_ERROR_MAP`

**`src/client.rs:3343`** · cross-version compatibility

The diff deletes the `LOGIN_MSG_DESKTOP_*` constants, every corresponding
`LOGIN_ERROR_MAP` entry, and the `session-login*` msgbox dispatch in
`flutter/lib/models/model.dart`, `src/ui/common.tis` and `src/ui/msgbox.tis`.

The client-side handling on master was **unconditional** — it did not depend on
the local client's own `allow-linux-headless`, because the opt-in lives on the
*peer*. So a new client loses the ability to handle an older peer that still has
headless enabled. On master, `handle_login_error` found the entry and raised the
OS-account prompt with `try_again: true`; on the PR the string misses the map,
falls into the final `else` → `msgbox("error", "Login Error", err, "")` →
`try_again = false`. The user sees a raw, untranslated English string and a
dead end.

**Which cases actually regress** (narrower than "cannot connect" — if the old
peer already has a live desktop session, `try_start_desktop` returns an empty
`err_msg` and login still succeeds normally):

- `SESSION_NOT_READY`, `..._PASSWORD_EMPTY`, `..._PASSWORD_WRONG`,
  `XSESSION_FAILED` — master offered an OS-login retry; the PR dead-ends.
- `XORG_NOT_FOUND`, `NO_DESKTOP` — master showed a localized tip plus
  `LINK_HEADLESS_LINUX_SUPPORT` and `try_again: true`.
- `SESSION_ANOTHER_USER` — unchanged in outcome (master was already
  `try_again: false`); only the message becomes raw English. Cosmetic.

**Judgment call for the maintainer.** Dropping controlling-side support for
headless peers is arguably exactly what "remove linux headless" implies, so this
may be intended. If you want a softer landing, note that a *display-only* shim
(keep the map entries so the client still shows a localized, actionable error)
would **not** restore login: see finding 3 — `os_login` is now populated only for
`ConnType::TERMINAL`, so completing such a login would additionally require
re-adding the OS-credential send path.

## 2. `stop_subprocess()` removal orphans Xorg and CM processes

**`src/platform/linux.rs:937`** · upgrade transition

`stop_subprocess()` is deleted along with all three call sites (startup in
`start_os_service()`, plus both restart branches). It SIGKILLed exactly two
process kinds: anything whose cmdline contains `/etc/rustdesk/xorg.conf` (the
`-config` path `DesktopManager::start_x_server` passes) and `rustdesk +--cm-no-ui`.
The surviving `stop_rustdesk_servers()` only matches the regex `<app> +--server`,
so neither is reaped, and no package script or one-shot migration replaces it.

Because `--server` is killed with `kill -9` (no destructors, no cgroup teardown),
its Xorg/wm children are orphaned to init — `stop_subprocess()` was the reaper for
precisely that.

**The finding understates one thing:** the same file deletion also removed
`linux_desktop_manager::recover_orphaned_session()` and its boot-id-tagged
`headless-orphaned-session` marker, which reaped the leaked logind session scope
and X lock files after a `--server`/service crash (rustdesk#15183). That is a
second reaper gone with no replacement.

**Window.** Not a defect in the default configuration — on a headless-off host
`stop_subprocess()` matched zero processes, and post-PR nothing can spawn them.
The exposure is exactly one transition: an in-place upgrade over a
headless-enabled install that has a live managed Xorg / `--cm-no-ui` when the new
service starts. Those orphans then hold the dummy X display, `/tmp/.X<n>-lock`,
and the logind session scope until reboot. It is one-shot and self-clearing at the
next boot, not a persistent hazard. The stale `headless-orphaned-session`
LocalConfig key is also left behind, but is inert.

**Cheap fix if you want it:** retain a one-time
`pkill -f '/etc/rustdesk/xorg.conf'` / `--cm-no-ui` in `start_os_service()` for a
release or two, or in the package postinst.

## 3. `handle_hash` hard-codes empty OS credentials

**`src/client.rs:3618`** · cross-version compatibility

`handle_hash` now passes `String::new(), String::new()` to `send_login`, and
`create_login_msg` populates `os_login` only when
`self.conn_type == ConnType::TERMINAL`. `showSetOSAccount` — the only writer of
`os-username` — is deleted, leaving no reader or writer of that key anywhere in
Rust or Dart.

**Most of this is not a loss.** On master the only non-terminal consumer of
`os_login` was `linux_desktop_start_credentials`, which returns `None` on its
first line when `!is_headless_allowed`. So on default-configured hosts master
already read the saved credentials and discarded them — no behavior change here at
all, and arguably an improvement, since master shipped a stored OS password over
the wire to servers that threw it away.

**The genuine residual is cross-version only:** a post-PR client talking to an
older/unpatched server with `allow-linux-headless=Y` can no longer supply OS
credentials at all. Combined with finding 1 (which removes the prompt that would
collect them), headless peers become unreachable in the failure cases listed
there. Worth a release-note mention.

**Two corrections to how this was originally framed.** "Orphaned in every peer
config" is wrong — `os-username` was persisted only when the user checked remember
on the headless `session-login*` dialog or used the `pi.isHeadless`-gated toolbar
item (the terminal callers pass `canRememberAccount: false`), so the stale key
exists only in configs of peers that were actually headless Linux hosts, and it is
an inert entry in the generic options map. And `os-password` remaining live via
`should_auto_login()` is not an inconsistency: it is a separate non-headless
feature whose writer `showSetOSPassword` is deliberately retained; only the
headless co-write inside `_connectDialog` was deleted.

## 4. `--cm-no-ui` falls through to the main control panel

**`src/core_main.rs:715`** · upgrade transition

The `--cm-no-ui` dispatch arm is deleted and `IS_CM` in `src/common.rs` no longer
matches it, so the flag now reaches `return Some(flutter_args)` with
`flutter_args` empty — the "no args" case, i.e. the main control-panel window.

**This is a complete, self-consistent removal, not a dangling handler.** The
producer (`args = vec!["--cm-no-ui"]` inside the `if headless_cm {` block in
`server/connection.rs`), `start_cm_no_ui()` in `flutter.rs`, the pkill reaper in
`platform/linux.rs`, and all of `linux_desktop_manager.rs` are removed in the same
diff; `git grep` finds zero residual references. And falling through to the
control panel is the generic, pre-existing behavior for *any* unrecognized flag —
`rustdesk --bogus` did the same on master. What is new is only that `--cm-no-ui`
joined the unrecognized set, which is the intended consequence.

**Residual risk (marginal, transient).** During an in-place upgrade on a host that
had `allow-linux-headless=Y`, an already-running old `--server` executing old
in-memory code can spawn the *new* on-disk binary with `--cm-no-ui` via
`run_as_user`. That process now opens a Flutter control-panel window on the
session user's desktop instead of a headless CM. It self-resolves at the next
service restart.

**Belt-and-braces fix:** `else if args[0] == "--cm-no-ui" { return None; }` — a
no-op stub, not a restoration of the mode.

## 5. Three stale comments in `drm_capturer.rs`

**`src/server/drm_capturer.rs:832`, `:848`, `:883`** · stale documentation

`src/server/drm_capturer.rs` is untouched by the diff, yet three of its comments
were invalidated from outside it. On master, `availability_cached()` had exactly
one caller — `drm_login_screen_seat0_username()` at
`linux_desktop_manager.rs:360` — which refused only on a definitive `Unavailable`,
so the doc's "the login decision treats Unsettled as a possibly servable greeter"
was accurate. The PR deletes that caller and re-points
`drm_can_serve_login_screen()` at `availability_cached()`, making it the sole
consumer — and it admits on `Available` only.

Three comments now describe a decision that no longer exists:

- `:832` — on the `Availability` enum: "the login-screen headless decision must
  not read it as one" (names the deleted decision outright).
- `:848` — justifies the `stale_no` sticky-negative behavior in `availability()`
  with "the login decision reads Unsettled as a possible greeter".
- `:883-884` — the `availability_cached()` doc contract itself.

**Zero runtime impact.** Master's gate already refused on `Unsettled` too
(`is_available_cached()` is `matches!(state, ProbeState::Available(..))`), so
admission is `Available`-only on both revisions. The `stale_no` logic is still
correct; only its stated rationale is obsolete. Fix is a three-comment edit
restating that the sole login-path consumer is now admission-gated
`Available`-only.

## 6. Routing-gate comment is false, and the gate now mutates probe state

**`src/server/connection.rs:446`** · misleading comment

`drm_can_serve_login_screen()` switched from `is_available_cached()` — a
side-effect-free read, documented at `drm_capturer.rs:826` as "Never probes, never
blocks: the form the ROUTING gates must use" — to
`availability_cached() == Availability::Available`, which on `Unknown` spawns a
`drm-avail-probe` thread and on a stale verdict spawns
`refresh_available_async()` / `refresh_unavailable_async()`.

**The boolean result is bit-identical in all three states** (`Available` → true,
`Unavailable` → false, `Unknown` → `Unsettled` → false), so no client is admitted
or refused differently.

**The actual defect is the new comment.** "This routing gate never blocks an
unauthenticated request" describes a path no unauthenticated peer reaches: the
sole call site is `connection.rs:1941`, inside `send_logon_response_and_keep_alive`
(line 1767), and every caller of that function is post-authorization — the
`ipc::Data::Authorize` handler at line 675, and lines 2896/2922/2945/3013, all of
which sit after `is_recent_session` / empty-password / terminal /
`validate_password` succeeded.

**Counterpoint worth weighing before "fixing" it.** The added side effect is
single-flight-guarded by `DRM_PROBE_IN_FLIGHT` (it cannot fan out threads per
connection) and never blocks the caller. And with the headless seat0 prober
deleted, this call site becomes the *only* thing that can lift a cold `Unknown`
state on a greeter box whose DRM producer came up after `warm_availability()` gave
up — on master with headless off, such a box would refuse with
`LOGIN_SCREEN_WAYLAND` forever. Read that way the swap repairs a latent master
defect. If so, only the comment needs correcting.

## 7. No-session poll cadence: 300ms → 500ms

**`src/platform/linux.rs:1064`** · note, low impact

`Desktop::is_headless()` and `is_rustdesk_subprocess` were deleted with the rest of
headless, taking the `if !desktop.is_headless() { sid = ... }` guard with them.
`sid` is now assigned unconditionally, so it can return to `""`.

**Exact delta:** while no seat0 session exists *and* at least one existed earlier
in this service lifetime, the loop-tail sleep goes 300ms → 500ms. Cold boot with no
session was already 500ms on both revisions.

**Master's 300ms was an accident, not a design.** `keeps_headless = sid.is_empty()
&& desktop.is_headless()` was clearly written to grant the 500ms anti-dbus-spam
throttle (added for #3129) to the no-session state — it just never fired after the
first session, because sid-pinning guaranteed `sid.is_empty()` was false. The PR
makes the no-session state actually take the throttle the code intended.

Impact is bounded at +200ms worst case in noticing a new login, on a path that is
already coarse: the same iteration's `force_stop_server()` already sleeps
`SERVICE_INTERVAL` (300ms) plus several `ps | grep | xargs kill` shell-outs, so
per-iteration cost goes ~600ms → ~800ms. Listed for the record because the change
is invisible in a PR titled "remove linux headless"; no action needed.

fufesou:
https://github.com/rustdesk/rustdesk/pull/15866#issuecomment-5313995909

For 1–4, those cases can happen, but they all involve an old headless-enabled peer or a headless process that is still running during an upgrade.
Linux headless is being removed by this PR, and keeping compatibility with those older peers and processes is intentionally out of scope. Because Linux headless is not supported since `1.2.4`.

5 and 6 are good catches. The DRM behavior itself is intentional: a cold cache starts one background probe, and a login-screen session is admitted only after an `Available` result. The comments still described the old headless decision and incorrectly called this an unauthenticated request, so I've updated both comments.

No change for 7. That polling interval is intentional while there is no active session.

rustdesk:
# Code Review — PR #15866 `refact: remove linux headless`

Effort: **xhigh**. The author's dismissals in the PR thread were taken into account (see
[Author's explanations](#authors-explanations)). 7 findings, most severe first.

## Verdict

The removal sweep itself is clean. Verified there are no leftovers for `linux_desktop_manager`,
`is_headless_allowed`, `--cm-no-ui`, `LOGIN_MSG_DESKTOP_*`, `startwm.sh` / `xorg.conf`, `pam`, or
`main_show_option`; all 51 lang files dropped the same 10 keys; packaging scripts, specs and CI have
no dangling references; `Cargo.lock` is consistent. Findings 5 and 6 of the previous review round
were fixed in `40f3718`.

What remains is one functional break on the `--cm-no-ui` path, one dead-code substitution in
`should_start_server`, and five cleanups.

<a name="authors-explanations"></a>
## Author's explanations, and how they were applied

The author declared three things out of scope. They are honoured below, not re-litigated:

| Author's position | How this review treats it |
|---|---|
| Cross-version compatibility with headless-enabled peers is out of scope | Not counted as a finding. |
| Orphan processes during the upgrade window are out of scope | Finding 1 is still recorded — see the note there — because the consequence is broader than a stray window, not because the scoping call is wrong. |
| The 500 ms poll cadence is intentional | Not counted as a finding. |

Findings 5 and 6 of the prior round (comment/doc fixes) were accepted and landed in `40f3718`;
finding 5 below is the same class and was simply missed in that pass.

---

## 1. `--cm-no-ui` now launches the full main app, leaving the session with no CM

`src/core_main.rs:715` — **correctness**

Dropping the `--cm-no-ui` branch makes that argument fall through to `return Some(flutter_args)`
with an **empty** arg vector, so the process launches the full Flutter control panel while `IS_CM`
and `IS_MAIN` are both false.

`flutter_args` is only populated for `--connect` / `--play` / `--file-transfer` / etc., so
`rustdesk --cm-no-ui` yields `Some(vec![])` and Flutter runs `runMainApp`. Meanwhile `IS_MAIN` is
false (the arg starts with `--`) and `IS_CM` is false, so the process is neither — main-only init is
skipped. The old `--server` that spawned it then loops 20 × 300 ms in `start_ipc` and bails with
`"Failed to connect to connection manager"`, leaving that incoming session with **no CM at all**: no
accept dialog, no session list.

> Reported previously as finding 4 and dismissed as an upgrade-window issue. Recorded here only
> because the consequence is broader than a cosmetic stray window; whether that is still acceptable
> is the author's call.

## 2. `os_login` is silently dropped for every non-TERMINAL connection

`src/client.rs:2730` — **correctness**

`os_login` is now a `MessageField::none` for every non-TERMINAL connection, so OS credentials handed
to `handle_login_from_ui` on a default / file-transfer / port-forward session are silently discarded
instead of sent.

Any current or future caller of `gFFI.login(osUsername, osPassword, ...)` / `handle_login_from_ui`
on a `ConnType::DEFAULT_CONN` session — the sciter `handler.login(res.osusername, res.ospassword, ...)`
shape is still present in the tis API surface — has the credentials dropped at `create_login_msg`
with no log and no error. The request goes out with the field absent and the server reads `""` via
`MessageField::deref`.

This is the deliberate consequence of removing headless (previously reported as finding 3 and
dismissed), but the drop is now *silent* rather than gated: `send_login`'s doc at
`src/client.rs:3648-3656` still documents both parameters as meaningful, and `handle_hash`
(line 3618) hard-codes empty strings — so the two parameters are live on only one of three call
paths.

## 3. `sid.is_empty()` branch is unreachable; `uid` goes stale across logout

`src/platform/linux.rs:874` — **correctness**

`if desktop.sid.is_empty()` in `should_start_server` is now unreachable dead code, so the
`*uid = ""` "from having a monitor to not having a monitor" reset can never fire.

`Desktop::refresh()` (~line 2089) either sets `sid` / `uid` / `username` all non-empty or resets the
whole struct to `Default`, so `sid.is_empty()` is exactly equivalent to `username.is_empty()`. Both
call sites are guarded by `desktop.username == "root" || desktop.is_login_wayland()` and
`desktop.username != ""`, which both require a non-empty username.

On master the branch was reachable only through the `is_rustdesk_subprocess` half of `is_headless()`,
which this PR deletes. Replacing `desktop.is_headless()` with `desktop.sid.is_empty()` **kept the
half that was already dead and dropped the half that was live**. Consequence: `uid` retains a stale
value across a logout, and the branch, its comment and the `should_kill` path are dead weight that
reads as live logic.

## 4. The login path now spawns a DRM probe thread while `DRM_STATE` is `Unknown`

`src/server/connection.rs:444` — **efficiency**

`drm_can_serve_login_screen` now starts a DRM probe from the post-auth connection task on every call
while `DRM_STATE` is `Unknown` — new work master never did on this path.

On a host where the `_drm` producer is absent or flaky, `probe_and_publish` leaves `DRM_STATE` at
`Unknown` for the first four failures (`DRM_PROBE_MAX_FAILURES == 5`), so each Wayland-greeter login
request spawns a fresh `drm-avail-probe` thread that blocks for
`DRM_CONNECT_TIMEOUT_MS + DISPLAY_LIST_TIMEOUT_MS`.

`DRM_PROBE_IN_FLIGHT` single-flights it so it is bounded to one thread at a time and self-limits
after 5 failures, and `warm_availability()` (`src/server.rs:616`) already warms the cache at
`--server` startup — so the practical exposure is small, but the login path now owns work that
previously lived entirely off it.

## 5. `is_available_cached`'s doc contradicts the new routing gate

`src/server/drm_capturer.rs:824` — **docs**

The doc on `is_available_cached` still claims it is "the form the ROUTING gates must use", but this
PR moved the login-screen routing gate off it onto `availability_cached`, which *does* spawn a probe
thread.

Commit `40f3718` updated the docs on `Availability` (line 830) and `availability_cached` (line 879)
but not this one. A future reader following `is_available_cached`'s stated contract will see
`drm_can_serve_login_screen` (`src/server/connection.rs:444`) — a routing gate — using the other
function, and cannot tell which invariant is authoritative. Same class as the two comment fixes the
author already accepted.

## 6. The `Availability` tri-state no longer has a consumer that distinguishes it

`src/server/drm_capturer.rs:881` — **simplification**

With `linux_desktop_manager` deleted, both remaining consumers of the `Availability` tri-state
collapse it to `== Available`, so the tri-state API and its `pub(crate)` visibility now exist for no
caller that distinguishes `Unsettled` from `Unavailable`.

`drm_can_serve_login_screen` (`connection.rs:444`) and `is_available` (`drm_capturer.rs:962`) both
compare `== Availability::Available`. The only caller that ever needed the third state —
`drm_login_screen_seat0_username`, which acted only on a definitive `Unavailable` — is deleted by
this PR. `availability_cached` and `Availability` are still `pub(crate)` when `pub(super)` now
suffices (both consumers live under `crate::server`).

The author's own `fd77ac8f` had already collapsed this to `is_available_cached_or_probe() -> bool`
before `9fe68b1` reverted it for diff minimality. Cost: a wider-than-needed API and a documented
contract with no consumer to keep it honest.

## 7. `start_listen_ipc()` is now a single-caller indirection

`src/flutter.rs:1593` — **simplification**

`start_listen_ipc()` exists only because it used to take a `new_thread: bool`.

After removing `start_cm_no_ui` and `start_listen_ipc_thread`, `cm_init()` is the only caller and
does nothing but call `start_listen_ipc()`. Two functions and a `#[cfg]` duplication remain where
one body inside `cm_init` would do; the extra hop makes the CM startup path look like it still has
variants it no longer has.

fufesou:
https://github.com/rustdesk/rustdesk/pull/15866#issuecomment-5323338922

1. The failure mode is real, but it still requires an old headless-aware process to survive an in-place upgrade and launch the new binary with `--cm-no-ui`. The current code never produces that argument. Linux headless support does not work since `1.2.4`, so I haven't added a compatibility stub.

2. Dropping `os_login` outside terminal sessions is intentional. The remaining UI paths that send non-empty OS credentials are the terminal admin-login dialogs, normal login sends empty values, and the server only consumes these credentials for terminal OS login. The generic signature could be narrowed, but there is no current supported non-terminal flow losing credentials here.

3. The `sid.is_empty()` branch is indeed unreachable now. Both call sites require a non-empty username, and `Desktop::refresh` clears the whole snapshot when there is no sid. I've removed that branch. The uid staying populated across logout is not introduced by this PR: the no-session branch did not call `should_start_server` before this change either.

4. The login-side probe is intentional. It gives a Wayland greeter a recovery path when startup warming leaves `DRM_STATE` as `Unknown`. The probe runs off-thread, is single-flight, and settles to `Unavailable` after the failure threshold. Removing it would make a late DRM producer impossible to discover from the greeter path.

5. Agreed and addressed in 5bed065

6. It is true that the current callers collapse the result to `Available` or not available. I kept the tri-state because it still describes the probe state directly and avoids expanding this removal PR with a behavior-neutral refactor. The visibility could be narrowed, but that is not a correctness issue.

7. I prefer to keep `start_listen_ipc` as the platform-gated implementation behind the public `cm_init` entry point. It now contains the actual setup rather than being a forwarding wrapper, so inlining it would not remove obsolete behavior or change the flow.

PR #15880, opened 2026-08-16
fix(windows): restore keyboard focus when the cursor re-enters the remote image

palmoni5:
### Problem

On Windows, entering a window that is already connected to a peer leaves the keyboard dead — keystrokes go nowhere until the user clicks on the remote image.

`onWindowBlur()` deliberately unfocuses `_rawKeyFocusNode` so the OS handles keys while the window is inactive, but nothing requests focus back:

- `onWindowFocus()` only calls `requestFocus()` in relative mouse mode.
- `enterView()` skipped Windows entirely (`else if (!isWindows)`).

So the only remaining path was `onPointerDown` -> `requestFocus()`, i.e. an actual click inside the canvas. The same applies after switching to another tab in the window.

### Change

`enterView()` now requests focus on Windows too, matching the Linux behaviour, but gated so a background window can never take over the keyboard:

- `!_isWindowBlur` — Windows delivers hover events to inactive windows as well; without this guard, moving the mouse across a background remote window would start the grab-key thread and steal system keys (Alt+Tab, Win) from the focused app.
- `_isSelectedTab` — a non-selected tab must not own the keyboard.
- `middleBlocked.isFalse` — a local dialog keeps input.

`enterOrLeave(true)` is intentionally not called here: on Windows it is already driven by `RawKeyFocusScope.onFocusChange`, and calling it twice would reset modifiers twice.

`leaveView()` is unchanged, so focus is not dropped when the cursor moves to the toolbar — same as the existing behaviour after a click. The `onPointerDown` path stays as a fallback for the known cases where Windows does not deliver a focus event and `_isWindowBlur` stays stale.

Tab switching within a window still needs pointer movement over the image (there is no tab-state listener on Windows, only on macOS); that is left for a separate change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard and mouse input focus handling on Windows when the app regains focus.
  * Restored remote-input focus more reliably when the selected remote view is active and unobstructed.
  * Prevented focus from being reclaimed while another tab, overlay, window, or control is active.
  * Improved focus restoration when entering remote views and re-enabling relative mouse mode.
  * Preserved existing input activation behavior on other platforms.

This PR restores Windows keyboard focus when an active remote image regains window focus or receives pointer entry.
- Centralizes selected-tab and blocking-overlay focus guards.
- Restores focus after window activation when the cursor is already over the remote image.
- Applies the same guards to relative mouse mode and pointer-entry focus paths.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| flutter/lib/desktop/pages/remote_page.dart | Adds guarded Windows keyboard-focus restoration and brings the focus rationale comment within repository limits. |

</details>

<sub>Reviews (4): Last reviewed commit: ["refactor(windows): share one focus predi..."](https://github.com/rustdesk/rustdesk/commit/d77367383c841862ebca024f4c4b25b4a477907c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53795607)</sub>

### DIFF
--- a/flutter/lib/desktop/pages/remote_page.dart
+++ b/flutter/lib/desktop/pages/remote_page.dart
@@ -835,7 +835,17 @@ class _RemotePageState extends State<RemotePage>
       _macOSLocalFocusLost = false;
       stateGlobal.getInputSource(force: true);
       _syncMacOSKeyboardGrab(reassert: true, allowInactiveLifecycle: true);
-    } else if (!isWindows) {
+    } else if (isWindows) {
+      // Blur unfocuses this node and nothing restores it, so the keyboard stayed
+      // dead until a click. Focus only while the window is really active, or a
+      // background window would grab system keys. onFocusChange does enterOrLeave.
+      if (!_isWindowBlur &&
+          _isSelectedTab &&
+          _blockableOverlayState.middleBlocked.isFalse &&
+          !_rawKeyFocusNode.hasFocus) {
+        _rawKeyFocusNode.requestFocus();
+      }
+    } else {
       if (!_rawKeyFocusNode.hasFocus) {
         _rawKeyFocusNode.requestFocus();
       }
--- a/flutter/lib/desktop/pages/remote_page.dart
+++ b/flutter/lib/desktop/pages/remote_page.dart
@@ -273,6 +273,11 @@ class _RemotePageState extends State<RemotePage>
         tabState.tabs[selected].key == widget.id;
   }
 
+  // Every Windows requestFocus() must pass this, or a blocking dialog or an
+  // inactive tab could hand remote input to this page.
+  bool get _windowsCanFocusRemoteInput =>
+      _isSelectedTab && _blockableOverlayState.middleBlocked.isFalse;
+
   bool get _isMacOSKeyboardContextActive {
     return stateGlobal.isFocused.value && !_isWindowBlur && _isSelectedTab;
   }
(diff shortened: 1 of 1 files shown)


workturnedplay:
I didn't test if this PR works, but the issue is definitely something I encounter constantly.

fufesou:
This PR does not work for me, though I have commented `!_isWindowBlur` and added `_ffi.inputModel.enterOrLeave(true);`.

https://github.com/user-attachments/assets/e145e4be-28c0-4654-8ce0-70e252cd1ed1

Please tell me what I was missing.

palmoni5:
Thanks for testing — the video and the log made the problem clear.

The video shows hovering into the RustDesk window while VS Code is still the foreground window: every log line prints `window active:false`. That scenario is intentionally out of scope:

1. Windows is click-to-focus. Hovering never gives the window OS keyboard focus, so keystrokes keep going to the foreground app no matter what the Flutter side does.
2. `!_isWindowBlur` is the guard that keeps a background session from grabbing the keyboard. Even with it commented out, `onFocusChange` still sees `_isWindowBlur == true` and immediately unfocuses — that's the `true` → `false` pair right after `enterView - requestFocus` in your log, and it's why the extra `enterOrLeave(true)` didn't help either. Enabling input there would capture keys typed into the foreground app.

What this PR fixes is the case where the RustDesk window is already the foreground window and the keyboard used to stay dead until you clicked the remote image:

- the cursor leaves the remote image (toolbar, edge) and comes back → hover restores the keyboard (874718768);
- returning to the window via Alt+Tab or the taskbar with the cursor already inside the image, where no PointerEnter is emitted → handled in `onWindowFocus` (1af7cc2f6, pushed after your test);
- d77367383 routes every Windows `requestFocus()` through one predicate so a blocking dialog or an unselected tab can never take remote input (from the CodeRabbit finding).

Could you re-test with the latest branch, making sure the RustDesk window is the active window first (Alt+Tab to it or click its title bar), then move the cursor over the image and type?

If activating input by hover from a background window is the UX you'd like, that would need focus-follows-mouse behavior (`SetForegroundWindow` on hover), which is a much more invasive change — happy to explore it separately if you think it's worth it.

fufesou (approved):

PR #15889, opened 2026-08-18
fix(drm): deliver a rotated output upright (#15886)

### LINKED ISSUE #15886: linux/wayland drm capture: a 90 deg rotated monitor arrives rotated in the viewer
reported by @tekmaven on x (https://x.com/tekmaven/status/2089010821892374848) and relayed by zhou in #15420, filing it here so it has a home outside a merged thread.

reproduced on real hardware (gnome/mutter, i915), with the transform read back before and after each capture: with a 90 deg transform the compositor renders the rotated desktop sideways into the landscape scanout. the plane itself is not rotated, so the raw scanout is exactly what the drm path captures and ships. the physically rotated panel is what straightens it for the local user.

no topology event fires because the framebuffer size does not change, so nothing re-advertises. second order: rotating one monitor shifts the logical x offsets of every monitor after it, so input mapping across the desktop is off while the rotation holds.

the portal path is unaffected, it delivers the logical image.

fix shape: only the compositor knows the output transform, so it belongs in the wayland enumeration the drm path already does, plus a rotate in the unprivileged converter and the swapped advertised size. working on it.

fxd0h:
closes #15886, the first user report after the launch: a monitor rotated 90 deg arrives rotated
in the viewer.

the mechanism, measured on gnome/mutter with the transform read back from DisplayConfig before
and after every capture: the compositor renders the rotated desktop sideways into the landscape
scanout and the physically turned panel straightens it locally. the plane is not rotated and the
framebuffer never changes size, so the drm path ships the sideways scanout and nothing ever
rebroadcasts. same blindness, second symptom: scale was physical width over the already swapped
logical width, so a rotated 1:1 monitor advertised scale 1.78.

what this does:

- the capturer resolves the output transform once per session from the wayland enumeration
(rustdesk/hbb_common#586 carries the field; the socket probe child inherits it) and turns
accepted frames upright into its own buffer, sizing the session in rotated dimensions. the turn
direction is anchored to a measurement rather than protocol text: mutter transform=1 carries the
top bar down the scanout left edge, and left column to top row is what reads upright.
- the advertised list swaps width and height for 90/270 outputs and computes scale from the post
swap width, which also makes a mid session rotation a topology change.
- that alone did nothing mid session, and the bench caught it: the augmentation reads the cached
wayland snapshot and nothing invalidated it while connected (transform applied and held, desktop
layout changed logged, zero new encoders). the 1.5 s live layout poll now drops that cache when
it sees an actual drift, so the probe cost stays tied to real layout events.
- a non 4 byte format on a rotated session is a hard error that degrades through the existing
health path instead of shearing the image.

measured before and after on the same bench (drm path, HDMI-1 1920x1080 rotated via
DisplayConfig, the transform verified held across each window because mutter auto reverts it):

| | master | this branch |
|---|---|---|
| mid session rotate 0 to 90 | nothing: same encoder, sideways viewer | restart within a poll turn, 1080x1920 encoder, upright viewer |
| rotate back to 0 | nothing | restart, 1920x1080 encoder |
| fresh session on a rotated output | sideways, scale 1.78 advertised | 1080x1920 encoder from the start, upright |

consumers of the values this touches, and how long each holds them:

| consumer | on this change | retention |
|---|---|---|
| session guard / encoder sizing | rotated dims at build | one session, rebuilt on change |
| advertised list (check_changed) | swapped dims per poll | one 300 ms turn |
| cached wayland snapshot | dropped on live drift | until the next real layout event |
| uinput live layout poll | reads live already | one 1.5 s turn |
| login screen rect (no compositor) | unchanged, no transform source exists there | unchanged |

scope: everything sits inside the drm path (cfg gated); non drm builds compile byte identical.
verified here: 90 both directions mid session and a fresh session on a rotated output, on
gnome/mutter, plus unit tests for the three rotations, padded strides and the rotated session
guard. not verified here: input mapping on a rotated output beyond the rect update firing, kde
(the transform arrives by the same wl_output event, but i did not measure there), live 180, and
flipped variants, which map to their rotation.

the submodule is pinned to the two rustdesk/hbb_common#586 commits atop the current pin rather than main tip: main also carries an unrelated config-keys refactor the app has not adopted yet, and both commits are reachable upstream through the merge.

## Summary by CodeRabbit

* **New Features**
  * Improved Wayland capture for rotated displays, including correct dimensions, orientation, scaling, frame layout, and cursor positioning.
  * Capture sessions now automatically rebuild when display rotation, geometry, or layout snapshots change.

* **Bug Fixes**
  * Improved handling of changing Wayland display layouts and input regions.
  * Fixed capture of padded frame buffers and transformed cursor hotspots.
  * Improved consistency between DRM capture and PipeWire dimensions for rotated outputs.
  * Improved fallback handling when display dimensions are reported in different orientations.

The PR corrects rotated Wayland DRM capture and strengthens live layout-change detection.
- Carries output transforms into display geometry, frame rotation, cursor rotation, and advertised dimensions.
- Adds generation-based capturer rebuilding and state that preserves layout edges across concurrent initialization and failed enumeration.
- Updates fallback health handling and tests rotation, stride, assignment, and layout-race behavior.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| libs/scrap/src/wayland/display.rs | Adds transform-aware rectangles, oriented dimensions, and snapshot-generation helpers. |
| src/server/display_service.rs | Introduces persistent layout-edge bookkeeping and generation promotion across initialization races. |
| src/server/drm_capturer.rs | Rotates captured frames and cursors, tracks snapshot generations, and tightens fallback health handling. |
| src/server/wayland.rs | Integrates the updated Wayland display geometry and DRM capturer metadata. |
| libs/hbb_common | Updates the shared submodule pin to provide Wayland output-transform data. |

</details>

```mermaid
sequenceDiagram
    participant Poll as Wayland layout poll
    participant Layout as WaylandLayout
    participant Cache as Display snapshot cache
    participant Capture as DRM capturer

    Poll->>Layout: Compare live layout with seen build layout
    alt Layout changed
        Layout-->>Poll: Edge detected
        Poll->>Cache: Clear snapshot and bump generation
        Poll->>Layout: Promote live layout to baseline
        Capture->>Cache: Check recorded generation
        Cache-->>Capture: Generation changed
        Capture->>Capture: Rebuild with current transform
    else Layout unchanged
        Layout-->>Poll: No edge
    end
```
</details>

<sub>Reviews (17): Last reviewed commit: ["chore: bump hbb\_common to main tip"](https://github.com/rustdesk/rustdesk/commit/1e1348782d3d0c6ebdba71c246d995afb0662271) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54225938)</sub>


rustdesk:
# Code Review — PR #15889 `fix(drm): deliver a rotated output upright (#15886)`

Effort: **max**. 10 finder angles plus verifiers, then re-verified line by line against
`3d54b0eea` — the fix commit the author pushed after the first pass. `hbb_common` is out of
scope by request, so `rustdesk/hbb_common#586` and the `transform_degrees` mapping are not
reviewed here. 15 findings, most severe first.

## Verdict

The rotation itself is correct: `unrotate_bgra`'s direction is anchored to a measurement, the
`rotated_dims` guard is applied to both sides of the frame-size check, and `3d54b0eea` fixed the
advertise/deliver split that made the headline scenario a black screen.

What is left is that the PR rotates **pixels** without rotating everything measured in the same
space. Input mapping (1), the hardware cursor (11), the scale/original-resolution pair (12) and
the PipeWire fallback's geometry check (8) all still read the unrotated numbers. And the second
commit — dropping the snapshot when the layout drifts — is inert in the two configurations it
was written for: a single monitor (4) and a login screen (6).

Finding 1 is the one to fix before merge: on the single rotated monitor from #15886 the video is
now upright and correctly sized, but the pointer cannot reach the right or bottom ~44% of it.

## Fixed by `3d54b0eea` since the first pass, not re-listed below

- The 90/270 swap sat below the `origin_only` cut, so a lone rotated output advertised unrotated
  dimensions while the capturer delivered rotated frames — every frame dropped at
  `src/flutter.rs:465`. The swap now precedes the cut and a test pins it.
- `transform` and the augmented origin came from two separate `get_displays()` reads inside the
  capturer build. `transform_and_origin` now takes one snapshot. (The *cross-path* half of this
  survives — see finding 10.)
- Both over-length comments flagged against AGENTS.md were trimmed. A new one appeared: finding 15.

---

## 1. The uinput absolute range stays unrotated, so the pointer cannot reach 44% of a rotated screen

`libs/scrap/src/wayland/display.rs:335` — **correctness**

`desktop_rect_of`'s single-display branch returns `(d.x, d.x + d.width, d.y, d.y + d.height)`
from the wl_output mode, which rotation never changes. After this PR the client is told that
display is 1080x1920 and sends coordinates in that space, while the injected device's range is
still 0..1920 by 0..1080.

The compositor maps the absolute device across its logical desktop, which for a 1920x1080 panel
rotated 90 is 1080 wide. A client x of 1080 (the right edge) is 1080/1920 = 0.5625 of the device
range and lands at logical x 607; everything past it is unreachable. Client y above 1080 is
clipped at the device maximum, so the bottom of the screen is unreachable too.

This is on the preferred path, not a fallback: `update_uinput_resolution`
(`src/server/wayland.rs:165`) asks the compositor first via `get_desktop_rect_for_uinput` ->
`desktop_rect_of`, and only falls back to `drm_desktop_rect_for_uinput` when that lookup fails.
The irony is that the fallback is now the correct one — it builds its rect from
`drm_capturer::get_display_infos()`, which this PR swapped. Its doc at `src/server/wayland.rs:111`
("in scanout pixels") is stale as of this PR.

Multi-monitor is unaffected: that branch uses `logical_size`, which arrives already rotated.
Only the single-display carve-out drops the rotation — the configuration in the bug report.

## 2. A 180-degree output that the compositor rotates in hardware is delivered upside down

`src/server/drm_capturer.rs:432` — **correctness**

`unrotate_bgra` runs for every non-zero transform, 180 included. But mutter refuses hardware plane
rotation for 90/270 while *accepting* 180 (`meta-kms-plane.c`:
`case MTK_MONITOR_TRANSFORM_180: return plane->rotations & META_KMS_PLANE_ROTATION_ROTATE_180`).

On that path `calculate_view_transform` returns NORMAL, the desktop is rendered upright into a
mode-sized onscreen, and `apply_transform` sets the plane rotation — while `wl_output` still
advertises transform 180. The capturer reads 180, the geometry guard passes (180 does not change
dimensions), and `unrotate_bgra` reverses both axes on an already-upright frame.

Master delivers this configuration correctly; the PR delivers it upside down, with no size
mismatch to trip any guard. The compositor-side reasoning here is from the review's trace of
mutter, not re-measured on a 180 host — worth confirming before choosing the fix, but the
asymmetry (90/270 software, 180 hardware) is the documented mutter behaviour.

## 3. `transform` is frozen at capturer build and nothing on this path can rebuild it

`src/server/drm_capturer.rs:314` — **correctness**

The comment above it says "A rotation change rebuilds this capturer." Nothing does.

`check_display_changed` returns `None` at `src/server/display_service.rs:245` for every non-x11
Linux session ("wayland do not support changing display for now"), so the SWITCH broadcast never
fires on this path. The other rebuild trigger is the frame-size guard at
`src/server/drm_capturer.rs:394`, and it cannot fire either: rotating an output leaves the CRTC
mode and the scanout framebuffer at 1920x1080, so `rotated_dims(self.transform = 0, ..)` still
equals `session_size`.

Rotate a monitor mid-session on a multi-monitor host and the advertise half of this PR does land
— the cache clear fires, the list flips to 1080x1920, the client calls `sessionSetSize` — while
the capturer keeps delivering 1920x1080 with `transform == 0`. `on_rgba` then drops every frame
until the user reconnects. The advertise half without the capture half is worse than neither.

## 4. The drift check can never fire on a single monitor, so the second commit is inert there

`libs/scrap/src/wayland/display.rs:384` — **correctness**

`logical_rects_of` has `let single = displays.len() == 1;` and then `(d.width, d.height)` for that
case; `desktop_rect_of:333` does the same. Both are physical mode dimensions, which rotation does
not change, and the origin stays (0,0).

So for a 1920x1080 monitor rotated to 90: `live_rects == baseline` (`drifted` false) and `rect` is
unchanged (`range_ok` true). The `if drifted || !range_ok` guard at `display_service.rs:140` is
never taken, `clear_wayland_displays_cache()` never runs, and the cached snapshot keeps the old
transform for the life of the session.

Commit 2 of this PR exists to make a mid-session rotation visible. On the single-monitor host of
#15886 it cannot see one.

## 5. The drift test compares rectangles, which carry no transform

`src/server/display_service.rs:127` — **correctness**

`DisplayRect` is `{name, x, y, w, h}`. Rotating an output 0 -> 180, or 90 -> 270, leaves the
origin and both the logical and physical sizes identical, so `layout.baseline != live_rects` is
false and `desktop_rect_of` is unchanged.

The cached snapshot keeps the stale transform, any later capturer build reads it from that cache,
and the stream stays upside down (180) or turned the wrong way (90 <-> 270) until the session is
torn down. This is independent of finding 4 — it holds on multi-monitor hosts too.

Fix: compare the full `WaylandDisplayInfo` snapshot including `transform`, not the derived rects.

## 6. The cache clear sits below the login-screen early return

`src/server/display_service.rs:105` — **correctness**

`refresh_wayland_uinput_rect_if_changed` returns at line 100 (`is_x11() || !wayland_use_uinput()`)
and again at line 105 (`is_login_screen_wayland_cached()`), both before the new clear at 140.

A greeter is where DRM/KMS capture is the intended backend, and it is where the socket probe does
supply a transform. There, nothing in the process ever invalidates `DISPLAYS`: the only other
clear fires on `Data::DrmDisplaysChanged`, which a rotation does not produce — connectors, CRTCs
and modes are all unchanged. `get_display_infos()` serves the pre-rotation snapshot for the whole
session.

The second commit is therefore inert exactly where the first one matters most.

## 7. Clearing the cache mid-session double-applies the origin delta

`src/server/display_service.rs:141` — **correctness**

Two monitors swap positions (baseline A@0/B@1920, live A@1920/B@0). No DRM hotplug fires —
`ipc/drm.rs` only emits `DrmDisplaysChanged` when connectors, CRTCs or modes differ — so
`set_wayland_layout_baseline` (its only call sites are `wayland.rs:190` and `:288`) never runs and
the baseline stays L0.

The poll sets `drifted = true`, clears the cache, and stores `WAYLAND_LAYOUT_DRIFTED = true` at
line 185. Within 300 ms the advertise re-enumerates from the fresh snapshot and the client rebases:
`input_model.dart:2039` sends `x += rect.left` using the new L1 origin. `remap_wayland_uinput_coord`
still maps from the stale L0 baseline, matches baseline B, and maps into live B@0 — the pointer
lands on the other monitor.

Before this PR the advertised list stayed on the old snapshot, so client and remap agreed. This
regresses the #15601 fix in the scenario it targets.

## 8. A rotated display demoted from DRM can never fall back to PipeWire

`src/server/wayland.rs:540` — **correctness**

The consistency check raw-compares `advertised.width/height` against `rect.1/rect.2`, and its own
comment states why that is safe: "The advertised DRM geometry is physical too
(`augment_with_wayland_geometry` sets x/y/scale and deliberately leaves width/height as the DRM
mode)." This PR falsifies that sentence.

`rects` is still built from `d.width()/d.height()`, which is `physical_size` and unrotated
(`common/wayland.rs:82-88`). On a multi-monitor host with one rotated display, any demotion — a
rapid-rebuild flap, a zero-frame streak, or the PR's own format hard-error — makes `consistent`
false and `bail!("drm display N demoted with no geometry-consistent PipeWire stream")` fires on
every attempt. That display becomes permanently uncapturable instead of falling back.

Single-display is safe: the size half is carved out when `single_display && cap_display_info.num == 1`.

The comment also records that this guard was already twice "corrected" into breakage by readers
reasoning from it. Leaving it asserting something the PR makes false invites a third.

## 9. `drifted` and `!range_ok` are latched levels, not edges

`src/server/display_service.rs:140` — **efficiency**

The added comment says probes stay "tied to real layout events". The code does the opposite.

`baseline` is only rewritten by `set_wayland_layout_baseline` — session init and DRM hotplug. A
compositor-only move leaves `baseline != live_rects` forever, so `drifted` stays true and
`clear_wayland_displays_cache()` runs on **every poll for the rest of the session**. Likewise
`!range_ok` stays true forever if `update_mouse_resolution` keeps failing, since the rect is only
recorded on success (line 171).

Each cold `get_displays()` re-runs `get_primary_monitor()`, which forks `xrandr`, then
`kscreen-doctor -o` and `gdbus` under 1 s timeouts each — on the 300 ms display-service loop. And
a failed enumeration is not cached at all (`get_displays` only writes `*lock` in the `Ok` arm), so
after any failure the full probe repeats every 300 ms.

## 10. Transform silently degrades to 0 on a lookup failure, and still comes from a different read than the advertise path

`src/server/drm_capturer.rs:151` — **correctness**

`transform_and_origin` returns `(0, None)` when `wl.displays.is_empty()`, with no log line and no
health signal. `get_displays()`'s `Err` arm returns an empty `Displays` without caching it, and
`LOOKUP_FAILURE_WARNED` has already latched, so nothing is printed anywhere.

If the probe fails at build time — the process-wide `RUNTIME_DIR_PROBE_BUSY` bails with "an
earlier probe has not returned", or the deadline is hit — the session is sized and delivered in
scanout orientation for its whole life, sideways, with nothing logged and nothing retrying.

`3d54b0eea` fixed the two-reads problem *within* the capturer build, but the advertise path is
still a separate read: `IpcDrmCapturer::new` calls `get_displays()` at line 313, while
`get_display_infos_and_primary` calls it at 1333 and `augment_with_wayland_geometry` at 1359 — on
different threads and schedules, with this PR's own new cache invalidation free to land between
them. When they disagree the advertised `DisplayInfo` is swapped and `CapturerInfo.width/height`
is not, and `on_rgba` drops every frame. Only the cache makes this rare.

## 11. The hardware cursor is still delivered in scanout orientation

`src/server/drm_capturer.rs:697` — **correctness**

The cursor arrives on a separate plane (`Data::DrmCursor` -> `set_drm_cursor`), is served verbatim
by `platform/linux.rs:471 get_cursor_data` (colors, width, height, hotx, hoty), and
`display_info_from_drm` sets `cursor_embedded: false` — so the client composites it over the video.

On a rotated output the compositor pre-rotates the bitmap it programs into the plane. After
`unrotate_bgra` turns the desktop upright, the cursor alone stays turned 90 degrees, and its
hotspot is transposed because hotx/hoty were measured in the rotated bitmap — so the click point
is displaced by the cursor's dimensions.

## 12. The swap is unconditional; `scale` and `original_resolution` are not

`src/server/drm_capturer.rs:1396` — **correctness**

The swap runs at 1396, but `scale` and `original_resolution` are only rewritten inside
`if let Some((lw, lh)) = w.logical_size` at 1401. With no xdg-output, or the field unset,
`info.width/height` become 1080x1920 while `scale` stays 1.0 and `original_resolution` keeps what
`display_info_from_drm` computed from the unrotated mode — 1920x1080. The client is told the
display is 1080x1920 but its "original resolution" is 1920x1080, which reads as a third-party
resolution change.

Independently of `logical_size`: `crate::platform::resolutions(&display.name)`
(`src/platform/linux.rs:1738`) in `make_display_changed_msg` still returns unrotated DRM modes, so
on any rotated display the advertised current size is absent from the list of resolutions offered
to the user.

## 13. The transform is resolved from the wire list while the advertise path uses the client list

`src/server/drm_capturer.rs:146` — **correctness**

`transform_and_origin` is called with the handshake `displays` (line 314), and the advertise path
assigns against the DRM_STATE list (line 1333). `get_capturer_info`'s own doc notes these can be
ordered differently ("when `wire_idx != display_idx`").

`assign_wayland_outputs` falls back to `taken.iter().position(|t| !t)` — pure layout order — when
a connector matches neither by normalized name nor by a unique resolution (line 1439). Two
same-resolution monitors whose DRM names do not normalize to the compositor's, one rotated: the
two assignments can disagree, and the capturer can rotate the upright monitor's frames while
leaving the rotated one sideways.

The fallback does log a warning, so it is diagnosable. But the doc comment at line 141 claims this
function carries "the same one-output-many-connectors guard ... a layout-order fallback there
would pin a rotation on a guess" — and a layout-order fallback is exactly what the assignment it
calls performs.

## 14. `unrotate_bgra` hand-rolls a rotation libyuv already provides

`src/server/drm_capturer.rs:95` — **efficiency**

The 90/270 arms copy 4 bytes per iteration with the row index innermost
(`let s = (h - 1 - dx) * stride + dy * PX;`), touching a new cache line every step. Benchmarked by
the review at 2.11 ms/frame at 1080p and 9.79 ms/frame at 4K — about 14x a row-wise memcpy — all
inside `TraitCapturer::frame`, which the video loop blocks on. That is roughly 59% of the frame
budget at 4K60.

`libs/scrap/src/dxgi/mod.rs:445` and `gdi.rs:156` already call `crate::common::ARGBRotate` for
display rotation. The bindings come from `include!(concat!(env!("OUT_DIR"), "/yuv_ffi.rs"))` in
`common/convert.rs:7`, gated only against iOS, and `lib.rs:9`'s `pub use common::*` re-exports
them — so `scrap::ARGBRotate` / `scrap::RotationMode` should be reachable on Linux (worth
confirming at build time). The mapping is exact: 90/180/270 -> `kRotate90/180/270` all match the
hand-written index math.

Also on this path: `self.cur` leaves the recycle pool whenever the frame is rotated, holding an
extra 8.3 MB at 1080p and 33 MB at 4K for the session.

## 15. The old doc comment was left stacked above the new one

`src/server/drm_capturer.rs:140` — **docs**

`wayland_transform_for` became `transform_and_origin`, but its three-line doc was not removed —
lines 140-142 (the old one) now sit directly above lines 143-145 (the new one), giving the
function a six-line doc comment whose first half describes a function that returns only a
transform.

`3d54b0eea`'s message says "comments trimmed to the three-line guideline", and it did trim the two
that were flagged. This one goes the other way, and it breaks AGENTS.md "Comments — Keep them
short: one line by default, three at most" in the same commit.

---

## Notes beyond the numbered findings

**One candidate refuted.** The 90/270 plane-rotation demotion theory: `rotated_dims` is applied
with the same transform to both sides of the session-size guard, so the predicate is mathematically
identical to master's, and mutter refuses hardware 90/270 rotation outright. Refuting it is what
surfaced finding 2 — the 180 case, which mutter does hardware-handle.

**Dead code, verified.** The non-4-byte-format hard error at `drm_capturer.rs:417` is unreachable:
the only publish path is line 663, which hardcodes `Pixfmt::BGRA`. Its comment asserts a safety
property ("the health path degrades it to PipeWire") that can never engage. `unrotate_bgra`'s `_`
arm is likewise only reachable from tests.

**Also cut for space.** A transient enumeration error can publish the unaugmented DRM list —
`get_displays()`'s `Err` arm returns an empty, uncached `Displays`, and `enumerate_displays:225`
skips the backoff stamp when `WAYLAND_DISPLAY` is set (`named` true), so `wayland_lookup_suppressed()`
stays false. Every re-probe forced by the new clear is a fresh chance for `check_changed` to
broadcast a spurious "Displays changed" that the next 300 ms turn flips back. Also: the session-size
guard rotation is a provable no-op.

**Test coverage.** `3d54b0eea` added the two tests that pin its own fix, and they are the right
ones. Still uncovered: 180 (finding 2), the `logical_size == None` pair (finding 12), and any
drift detection at a non-zero transform (findings 4 and 5) — `logical_rects_of` is never exercised
against a rotated `WaylandDisplayInfo`.

**Out of scope by request.** `libs/hbb_common` and `rustdesk/hbb_common#586`, including the
`transform_degrees` mapping and the `#[serde(default)]` on `WaylandDisplayInfo`.

rustdesk:
# Code Review — PR #15889 `fix(drm): deliver a rotated output upright (#15886)`

Against head `3d54b0eea`. hbb_common out of scope. 15 findings, most severe first.
Each was then adversarially re-checked against the code; none was refuted, four were corrected
(6, 9, 10, 14 — noted inline).

The rotation itself is correct. What is left: the PR rotates pixels without rotating everything
measured in the same space (1, 11, 12, 8), and the mid-session-drift commit is inert on a single
monitor (4) and at a login screen (6).

Already fixed by `3d54b0eea`: the swap below the `origin_only` cut, the two `get_displays()` reads
inside the capturer build, both over-length comments.

## 1. uinput range stays unrotated — `libs/scrap/src/wayland/display.rs:335`

`desktop_rect_of`'s single-display branch returns the unrotated mode, so the range is 0..1920 x
0..1080 while the client now sends coordinates in 1080x1920. Client x=1080 lands at logical 607;
y above 1080 is clipped. Right and bottom ~44% unreachable.

This is the preferred path: `update_uinput_resolution` (`wayland.rs:165`) asks the compositor
first and only falls back to `drm_desktop_rect_for_uinput` — which, after this PR, is the correct
one. Multi-monitor is fine (uses `logical_size`, already rotated).

Verified: `WaylandDisplayInfo.width/height` come from `mode.dimensions` (hbb_common
`platform/linux.rs:451`), i.e. the raw scanout mode, which rotation never changes; `transform`
appears nowhere in `input_service.rs` or `display_service.rs`; `wayland_use_uinput()` is
`!is_x11() && is_server()`, so this is on for every Wayland server session.

## 2. 180 degrees is delivered upside down — `src/server/drm_capturer.rs:432`

Mutter refuses hardware plane rotation for 90/270 but accepts 180, so a 180 output already scans
out upright while `wl_output` still advertises transform 180. `unrotate_bgra` then flips a correct
frame. No size change, so no guard trips. Master gets this right. (Mutter behaviour from the
review's trace, not measured here; the rest of the mechanism follows from it.)

## 3. `transform` frozen at build, nothing rebuilds it — `src/server/drm_capturer.rs:314`

The comment says a rotation rebuilds the capturer. The only rebuild path for a geometry change is
`try_broadcast_display_changed` -> `check_display_changed`, which returns `None` for non-x11
(`display_service.rs:245`); the frame-size guard can't fire either, because the scanout stays
1920x1080. So the client learns the new list and resizes while the capturer keeps delivering
unrotated — every frame dropped until reconnect.

## 4. Drift can never fire on a single monitor — `libs/scrap/src/wayland/display.rs:384`

`logical_rects_of` and `desktop_rect_of:333` both use physical size when `len() == 1`, which
rotation doesn't change, and the origin stays (0,0). `drifted` false, `range_ok` true, so the
clear this PR adds never runs — on the exact host from #15886.

## 5. Rect comparison carries no transform — `src/server/display_service.rs:127`

`DisplayRect` is `{name,x,y,w,h}`. 0<->180 and 90<->270 leave everything identical, so the stale
transform survives in the cache for the session. Compare the full `WaylandDisplayInfo` instead.

## 6. The clear sits below the login-screen early return — `src/server/display_service.rs:105`

Returns at 100 (`is_x11() || !wayland_use_uinput()`) and 105 (`is_login_screen_wayland_cached()`),
both before the clear at 140. A greeter is where DRM capture is the intended backend, and the
poll-driven invalidation this PR adds never runs there.

*Corrected by the adversarial pass:* other clear paths do exist — `check_init` at session start
(`wayland.rs:261`) and `clear()` on every capturer teardown (`wayland.rs:411`). What is missing is
a mid-session one: the drm_capturer clears (727, 736, 1118, 1170, 1264) are all gated on a DRM
topology change, and a rotation changes no connector, CRTC or mode. So a greeter session that is
rotated while running never picks it up.

## 7. Origin delta applied twice — `src/server/display_service.rs:141`

Two monitors swap positions, no DRM hotplug, so `set_wayland_layout_baseline` never runs and the
baseline stays L0. The clear lets the advertised list advance to L1 while `WAYLAND_LAYOUT_DRIFTED`
is armed (`drifted && range_ok`, and the bbox is unchanged so `range_ok` holds): the client rebases
on L1, `remap_to_live_layout` still matches against the L0 baseline, pointer lands on the other
monitor. Before this PR both sides moved together — the advertise stayed on the cached L0, and the
teardown path that clears the cache also re-snapshots the baseline. Regresses #15601.

## 8. Rotated display can never fall back to PipeWire — `src/server/wayland.rs:540`

The check raw-compares advertised w/h against `rect.1/.2`, and its comment says why that is safe:
augmentation "deliberately leaves width/height as the DRM mode". The PR falsifies that. Multi-
monitor with one rotated display: any demotion makes `consistent` false and the `bail!` fires
forever. Single-display is carved out.

## 9. `drifted`/`!range_ok` are levels, not edges — `src/server/display_service.rs:140`

`baseline` is only rewritten at session init and hotplug, so one compositor-only move keeps
`drifted` true and drops the cache for the rest of the session. Each drop forces a cold
`get_displays()` on the next consumer, and that is where `get_primary_monitor()` forks `xrandr`,
then `kscreen-doctor -o`, then `gdbus`, under 1 s timeouts each. The added comment claims the
probe cost stays tied to real layout events; it does not.

*Corrected by the adversarial pass:* the cadence is 1.5 s, not 300 ms — the function self-throttles
on `WAYLAND_LAYOUT_CHECK_INTERVAL` (line 38/111), and the poll itself is cheap by design
(`get_layout_for_uinput_live` skips both the cache and the primary detection). The 300 ms repeat
applies only to a *failed* enumeration, which is never cached.

## 10. Silent degrade to transform 0 — `src/server/drm_capturer.rs:151`

`(0, None)` on an empty lookup: a probe failure at build time streams sideways for the whole
session, with nothing retrying. The cross-path skew also survives `3d54b0eea`: `new()` reads at
313, the advertise path at 1333 and 1359, on different threads, with this PR's own clear free to
land between them — and then the advertised size is swapped while `CapturerInfo` is not.

*Corrected by the adversarial pass:* it is not literally unlogged — `enumerate_displays` warns
once, latched by `LOOKUP_FAILURE_WARNED`. But nothing connects that generic warning to "this
session lost its rotation", and the degrade site itself is silent.

## 11. Hardware cursor stays in scanout orientation — `src/server/drm_capturer.rs:697`

Separate plane, served verbatim by `platform/linux.rs:471`, `cursor_embedded: false`. After the
desktop is turned upright the cursor alone stays rotated, and hotx/hoty were measured in the
rotated bitmap, so the click point is displaced by the cursor's dimensions.

## 12. Swap is unconditional, `scale`/`original_resolution` are not — `src/server/drm_capturer.rs:1396`

With `logical_size == None` the size becomes 1080x1920 while `original_resolution` keeps 1920x1080
(set from the unrotated mode at `display_info_from_drm:1507`) — reads as a third-party resolution
change. Separately, `platform::resolutions()` (`platform/linux.rs:1738`) still returns unrotated
modes, so the advertised current size is absent from the offered list.

## 13. Transform from the wire list, advertise from the client list — `src/server/drm_capturer.rs:146`

`transform_and_origin` assigns against the handshake list (314), the advertise path against
DRM_STATE (1333) — and the doc at 1523 states outright that the two are ordered differently when
`wire_idx != display_idx`. `assign_wayland_outputs` falls back to plain layout order (1439). Two
same-resolution monitors with non-normalizing names, one rotated: the two assignments can disagree
and the wrong monitor gets rotated. The fallback does log a warning. The doc at 141 claims this
function guards against exactly that fallback.

## 14. Hand-rolled rotation, libyuv already linked — `src/server/drm_capturer.rs:95`

4 bytes per iteration with the row index innermost, inside `TraitCapturer::frame` which the video
loop blocks on. The review measured 2.11 ms/frame at 1080p and 9.79 ms/frame at 4K (~14x a
row-wise memcpy, ~59% of the 4K60 budget); the numbers are its own, the access pattern is visible
in the code. `dxgi/mod.rs:445` already uses `crate::common::ARGBRotate`; bindings come from
`common/convert.rs:7`, gated only against iOS, re-exported by `lib.rs:9`. 90/180/270 map exactly to
`kRotate90/180/270`. Also: `self.cur` leaves the recycle pool on the rotated path (+33 MB at 4K).

## 15. Stale doc left stacked above the new one — `src/server/drm_capturer.rs:140`

`wayland_transform_for`'s three-line doc (140-142) was not removed when it became
`transform_and_origin` (143-145), giving a six-line doc whose first half describes the old
function — in the same commit whose message says comments were trimmed to three lines.

---

- Refuted: the 90/270 plane-rotation demotion theory — `rotated_dims` is applied to both sides of
  the session-size guard, so the predicate is identical to master's. Refuting it surfaced #2.
- Dead code, verified: the non-4-byte hard error at `drm_capturer.rs:417` is unreachable — the only
  publish path (663) hardcodes `Pixfmt::BGRA`. Its "degrades to PipeWire" comment can never engage.
- Untested: 180, `logical_size == None`, and drift at any non-zero transform.

fxd0h:
all 15 verified against the code first; none refuted. head is now 753c89c18, three commits on
top of the one you reviewed. the work went through an adversarial pass of its own before this
reply, which killed the first version of two mechanisms - noted where it changed the shape.

what changed, finding by finding:

- **1 (uinput range)** - the single-display branches of desktop_rect_of and logical_rects_of now
serve delivered orientation. measured on the bench that found your number: the pointer now
reaches the far corner of a rotated 1920x1080 within 2 px on the cursor plane, where the ceiling
used to be logical x 607. the fallback rect doc you flagged as stale is updated.
- **3, 4, 5 (no rebuild path, inert drift)** - a rotation now has a rebuild trigger: a snapshot
generation, recorded by the capturer at build, checked in frame(), rebuilding without touching
display health. DisplayRect carries the transform, so 0/180 and 90/270 flips are visible to the
drift comparison. my first version hung the bump on clear_wayland_displays_cache; the adversarial
pass showed every session init clears too, so any init would have torn down every other live
capturer. the bump now has a single call site: the edge-detected layout change in the
display-service poll. measured: mid-session rotate = exactly one rebuild into the right encoder;
re-applying the same transform = zero (the edge does not fire on a no-op).
- **6 (greeter)** - the poll and the invalidation now run above the login-screen return; only the
rect/remap half stays skipped there.
- **7 (double-applied delta)** - the baseline promotes in the same motion as the clear on the drm
path, so the remap and the client rebase never correct the same delta twice.
- **9 (levels, not edges)** - the clear is edge-triggered on live-vs-previous-live. an edge seen
while the drm verdict is transiently non-available stays owed rather than consumed, or a rotation
inside a negative-ttl window would have stayed sideways for the session.
- **10 (silent transform 0)** - logged now, and bounded: a missing snapshot at build makes the
first successful poll an edge, so the degrade lasts as long as the enumeration outage.
- **11 (cursor)** - the bitmap turns with the session transform and the hotspot maps through the
same point transform, unit-tested. the race window comment says seconds, not microseconds.
- **12** - original_resolution follows the swap, including the origin-only path. the resolutions
list stays mode-space on purpose: resolution changes ride xrandr (platform/linux.rs:1815), which
is inert on this path - if wayland resolution changes ever land, the list needs the symmetric
transpose then.
- **13 (wrong-monitor rotation)** - both sides now key off the same identity-match pass
(identity_matches, progressive taken): a layout-order guess rotates nothing anywhere. my first
fix applied the identity rule to the capturer only, which the adversarial pass showed would
advertise a swap the capturer refuses - the scenario in your finding as a black screen instead
of a sideways image. worse, and gone.
- **14 (hand-rolled rotate)** - the body is libyuv ARGBRotate now; the direction tests i
already had pass unchanged against it, which pins the kRotate mapping to the measured anchor.
- **8 (fallback guard)** - you are right that i closed this by trace, not measurement, and
coderabbit then argued the opposite polarity (pipewiresrc does not apply SPA_META_VideoTransform,
so the caps may arrive unrotated). nobody has measured the portal caps of a rotated display, and this
guard has broken two readers already - the size half now accepts either orientation, which cannot
produce the permanent offline under either reality. comment says exactly that.
- **15** and the comment-length repeats: trimmed.
- **2 (hardware 180)** - not fixed here, deliberately. your mutter trace holds, and i measured
i915 advertising plane rotate-180 on my own box. wl_output cannot tell hardware from
software rotation, so 180 keeps master behavior (never unrotated) until the plane rotation
property travels the wire from libdrmtap - that is the follow-up, and it also covers a compositor
that hardware-rotates 90/270, which this pr would double-rotate today.

one correction to the notes: "the only publish path hardcodes BGRA" is half right - the convert
path publishes the fmt it returns (drm_capturer.rs, the conv.convert arm), so the non-4-byte
guard is unreachable today but not dead if a producer changes; it stays, with an honest comment,
and a short-stride frame now logs instead of publishing silent black.

bench summary on this head: fresh session on a rotated output delivers 1080x1920 from the first
frame; the pointer reaches the full screen; 49 unit tests across the two crates.

fxd0h:
on the two red jobs: neither is this diff. linux dies in the interval-sleep timing test, which asserts five timer ticks land in five distinct seconds and a loaded runner puts two in one; mac is a vcpkg digest mismatch fetching ffmpeg. master ran the same jobs green yesterday, this branch touches neither path, and the drm linux build passes.

fufesou:
## Findings

1. **[P2] Do not accept reversed PipeWire dimensions without normalizing the stream** (`wayland.rs:543`)

   On a multi-display host, this can pair an advertised 1080x1920 DRM output with a 1920x1080 PipeWire capturer after that DRM capture fails. The check accepts the reversed pair, but `CapturerInfo` keeps the PipeWire dimensions and no later Wayland path corrects them. The Flutter renderer rejects decoded frames whose size differs from the advertised display (`flutter.rs:463`), so this fallback can stay blank. Rotate the fallback stream into the advertised orientation, or reject the reversed pair.

2. **[P2] Include the session transform in the cursor ID** (`drm_capturer.rs:748`)

   A rebuilt stream rotates the cursor pixels and hotspot but keeps the hardware cursor's original ID. `MouseCursorService` only fetches new cursor data when that ID changes and caches the result by ID (`input_service.rs:397`). A mid-session rotation with the same cursor shape therefore leaves clients using the cursor cached for the previous transform until the shape itself changes. Derive the served ID from both the source ID and transform, or explicitly invalidate the cursor cache on a transform change.

3. **[P3] Install the transform before accepting the first cursor** (`drm_capturer.rs:329`)

   `recv_thread` can read cursor messages as soon as it sends the handshake, but `shared.transform` is stored only after the Wayland lookup finishes. During a cold lookup, the first cursor can be saved with transform 0. The producer resends it only when the cursor ID changes (`ipc/drm.rs:1107`), so the wrong orientation and hotspot can persist until the cursor shape changes. Set the transform before cursor delivery starts, or replay the first cursor after the store.

fxd0h:
all three verified real, fixes pushed:

1. f97d09fe8 + 4a3b33a77 drop the transposed acceptance outright, in the multi-display check and in the single-display carve-out both. traced end to end: CapturerInfo keeps the pipewire dimensions, check_display_changed returns None on wayland so nothing ever reconciles the client, and flutter.rs:463 then rejects every frame - the arm had no reachable case that was not blank. a transposed pair now falls into the existing bail and the display is advertised offline, which the client recovers from by re-enumerating.

2. cfa5047e8 folds the session transform into the served cursor id (the producers own fnv step, hidden sentinel passes untouched), so a rebuilt stream cannot reuse the id of the other orientation.

3. same commit: the transform now starts as a PENDING sentinel and a cursor that races the store is held and replayed once new() installs the real value. the receive loop wakes at least every 200 ms, so the replay lands promptly even on an idle wire.

fufesou:
## Findings

1. **[P2] Do not restart-loop the only display after rejecting its fallback** (`wayland.rs:558`)

   This rejects the transposed stream, but the single-display demotion path never marks that display offline: `mark_demoted_displays` returns early when the list has one entry (`drm_capturer.rs:1407`) because it assumes the PipeWire fallback remains usable. The bail leaves the same display advertised online, and each failed run clears the PipeWire state before the video service retries while DRM remains globally available. On a permanently demoted rotated output, the service never converges: even if PipeWire initialization briefly updates the synced display list, the next run reaches the same guard and the client receives no usable video. Make the fallback publish matching geometry, or change the single-display state before returning this error.

fxd0h:
right again - the rejection promised "advertised offline" but the lone-display carve-out in mark_demoted_displays kept the display online on the exact grounds the rejection had just refuted, so the service restart-looped against a stream nothing can serve. 792fcc188 records the geometry rejection in the display health and lets it end the carve-out: the lone display goes offline for real, and a delivered frame or the demote-cooldown re-arm clears the verdict so a recovered output comes back on its own. the multi-display path is untouched.

and the transposed case is now measured, not assumed: on sway 1.9 with xdg-desktop-portal-wlr 0.7, rotating the only output to 800x1280 logical still hands the portal stream at 1280x800 - xdpw does not apply the output transform. so on wlroots this configuration is real, and the long-term answer for it is rotating the fallback stream into the advertised orientation; the honest-offline state is the floor until then.

rustdesk:
I would **request changes** on PR #15889 at the current head `42df1b6`. The main Linux CI build/tests pass; I found three correctness issues that the existing tests don’t cover.

* **[P2] A capturer built during a failed Wayland enumeration can stay permanently unrotated after recovery.** `refresh_wayland_uinput_rect_if_changed()` detects the special “initial snapshot failed” case using `wayland_snapshot_missing()`, which only checks whether the *global cache is currently empty*.  But `get_displays()` leaves failures uncached and any subsequent successful caller repopulates that same cache.  In particular, the display-service loop runs every 300 ms and calls `get_display_infos()`, which can refill the cache before the 1.5 s live-layout poll.  The sequence is: capturer A fails enumeration and builds with `transform=0`; display enumeration subsequently succeeds and populates `DISPLAYS`; first live poll now sees `wayland_snapshot_missing()==false`, so `live_changed` is false and no generation bump occurs. Capturer A therefore never rebuilds and can remain sideways until an unrelated layout change/session reset. The recovery evidence needs to be persistent rather than inferred from the current cache state—for example, latch that a live capturer was built without a valid Wayland snapshot and consume that latch only after the layout generation is bumped. Add a test where a failed capturer build is followed by a successful cache refill before the first live poll.

* **[P2] Flipped Wayland transforms silently lose the mirror operation.** The `hbb_common` change converts `Flipped`, `Flipped90`, `Flipped180`, and `Flipped270` into plain degree values, so the mirror bit is irreversibly discarded.  Downstream, rotation/snapshot comparison only carries that degree value.  Wayland defines the flipped variants as a vertical-axis flip followed by rotation, not merely another spelling for the corresponding rotation. ([[wayland.freedesktop.org](https://wayland.freedesktop.org/docs/html/apa.html?utm_source=chatgpt.com)][1]) Thus `Normal -> Flipped` is even invisible to the drift detector (`0 -> 0`), while `90 -> Flipped90` similarly remains `90`; a mirrored output can be streamed mirrored indefinitely. Either carry the full eight-state transform through the pipeline and apply the reflection to frames/cursor/hotspot, or explicitly reject flipped transforms instead of silently treating them as rotation-only.

* **[P2] The greedy “identity” matcher can steal an exact-name output and apply its rotation to the wrong DRM connector.** `identity_matches()` walks DRM connectors in order and `match_wayland_display()` falls back immediately to a unique remaining resolution when the current connector has no name match.  Suppose connector 0 has an unmatchable name at 1920×1080, connector 1 is `HDMI-A-1` at 1920×1080, and Wayland contains `HDMI-1` at 1920×1080 plus a 2560×1440 output. Connector 0 takes `HDMI-1` because it is the sole 1920×1080 Wayland output; connector 1 then loses its exact name match. With this PR, connector 0 can inherit HDMI-1’s 90° transform while the actual HDMI connector gets transform 0. Name matches should be reserved globally first, then resolution-only matching should run on the unmatched remainder; for rotation specifically, I’d also require the resolution to be unambiguous on both sides.

One additional limitation is worth documenting more prominently: although there is an `unrotate_180` test, `transform_and_origin()` deliberately converts 180° to 0, so software-rendered 180° rotations are not actually corrected. That appears intentional because of the i915 hardware-rotation ambiguity, so I wouldn’t make it a separate blocker given the PR description, but the test suite currently makes 180° support look broader than the production path really is.

[1]: https://wayland.freedesktop.org/docs/html/apa.html?utm_source=chatgpt.com "Appendix A. Wayland Protocol Specification"

fxd0h:
first two verified and fixed in 2bb0a3038 plus c391f2b44: the blind capturer latches that it built without a snapshot and the layout poll consumes the latch - unconditionally, so the two edge sources merge into one promotion instead of double-rebuilding; and the matcher reserves name matches globally first, with resolution pairing only when forced on both sides - the steal scenario is pinned by a test.

the flipped transforms one is real too and needs a phased answer: the mirror bit dies inside hbb_common's transform_degrees before any of this code runs, so the honest immediate fix (a once-logged warn plus the doc saying the mirror half is not corrected) lives in an hbb_common change, and actually undoing the mirror needs a measured producer first - i have a sway bench now, so flipped-90 is measurable before any pixel code lands. will follow up on that one separately.

fufesou:
Reviewed at `c391f2b4467dee3ecce7614de8683130b88bfc26`.

## Findings

1. **[P2] Preserve an unseen layout change when another video service starts** (`display_service.rs:123`)

   A successful `update_uinput_resolution()` call replaces the global layout baseline and clears `layout.live`. If an output rotates after capturer A's last poll, but another monitor service starts before the next poll, the new service records the rotated layout as the baseline. The next poll enters the `live.is_empty()` branch, finds `baseline == live_rects`, and leaves `live_changed` false.

   No layout-generation bump occurs, so capturer A keeps its old transform. In the Mutter software-rotation case this PR targets, a 90/270-degree rotation leaves the DRM framebuffer size unchanged. The regular display-change check is also disabled on Wayland, so the stream can remain sideways until another layout event or restart.

   Keep the generation detector's last observed layout separate from the per-session input baseline, or record a pending generation change before replacing the baseline.

fxd0h:
fixed, head is now `3598efdd2`.

confirmed, and it is not only the race you describe. `set_wayland_layout_baseline` clears `live`, which is the edge detector's only memory of the previous layout, so ANY `update_uinput_resolution()` landing between a rotation and the next 1.5s poll erases the edge — and since `ensure_inited()` calls it at the top of every video service start, a display stuck in the ~1s retry loop closes that window continuously, not just when a second service happens to start. the blind-capturer latch does not cover it either: that path only arms when the capturer was built with no wayland snapshot, and here it had a valid one.

i took your second remedy, placed in the setter rather than at the call sites. the setter is the one choke point every caller already goes through, so no present or future caller can forget it: it now arms the owed promotion when the outgoing `live` differs from the incoming baseline. an empty incoming baseline is the DRM-union fallback and proves nothing, so it does not arm.

three things worth stating since the latch is now module scope: it cannot self-trigger from the poll's own call, because there `live` was just set to `live_rects` and the argument is a clone of it; the DRM-union fallback passes an empty vector and is excluded; and the non-DRM call site is harmless because consuming the latch still requires `is_available_cached()`.

not measured yet. the bench needs an output rotated while a second capturer sits in the retry loop, and the only box here with the multi-card shape to produce that reliably is one i cannot leave a test build on. the reasoning above is from the code, and i would rather label it that way than imply a run.

fufesou:
Reviewed at `3598efdd2fc5d30bdabb59df3df252bcc4219639`.

## Findings

1. **[P2] Preserve the layout edge across consecutive baseline resets** (`display_service.rs:89`)

   This still loses a rotation when two `ensure_inited()` calls straddle it. An initialization just before the rotation sees the old layout, leaves `PROMOTION_OWED` false, and clears `layout.live`. A retry after the rotation then sees the new layout, but the `!lock.live.is_empty()` guard prevents it from recording the change. The next poll finds the new baseline equal to the new live layout and does not bump the generation, so the existing capturer keeps its old transform.

   This is reachable through the normal retry path: every video-service run calls `ensure_inited()` (`video_service.rs:544`), a failed service retries as often as once per second (`service.rs:307`), and the layout poll runs only once every 1.5 seconds. On Mutter, the framebuffer dimensions do not change for software rotation, and the regular display-change check is disabled on Wayland, so the existing stream can remain sideways until another layout event or restart.

   Keep the generation detector's last observed layout separate from the per-session input baseline. Add a regression test with one initialization before the rotation and another after it, both before the next live-layout poll.

fufesou:
Reviewed at `dcb9535fb0fba23bc841dcd2f06a8215f0f7ce1f`.

## Findings

1. **[P2] Keep the layout used by a capturer when the baseline is empty** (`display_service.rs:87`)

   `ensure_inited()` calls `update_uinput_resolution()` before the capturer is created. If that Wayland query fails, the DRM fallback can save an empty baseline (`wayland.rs:166-173`). The failure is not cached, so `IpcDrmCapturer::new()` can retry successfully a moment later, cache layout A, and build the capturer with A's transform (`drm_capturer.rs:350-352`). Because this build was not blind, it does not set `UNROTATED_SNAPSHOT_PENDING`.

   If the output rotates to layout B before the first layout poll, both `seen` and `baseline` are empty while the snapshot cache is populated. `edge()` therefore returns false at line 88. `observe()` then records B, so later polls no longer see the transition; the snapshot generation never changes and the capturer stays at A's transform. A second initialization that installs B as the baseline, as in the Greptile thread, fails in the same way, but that second reset is not required to reproduce the bug.

   Record layout A when the capturer's retry succeeds, or keep a pending unknown state after the failed initialization until the first successful poll. Add a regression test for an empty baseline, a successful capturer snapshot, and a rotation before the first poll, both with and without another baseline reset.

fxd0h:
fixed, head is now `70c8ba6fa`.

confirmed, and the part i had missed is that the capturer's snapshot is a THIRD thing that knows a
layout. i had the poll and the session baseline in mind and treated the build as a consumer of
both, when a build that succeeds after the init's query failed is the only record of what the
stream is showing.

so the build now records it, and only when nothing else has. your case is one test covering both
shapes, with and without the extra baseline reset, plus a control with no rotation, and it fails
with the recording removed; a second test pins the "only when nothing else has" half, that a
capturer built later must not overwrite what the poll is keeping.

fxd0h:
one thing i owe from earlier in this thread. at `3598efdd2` i wrote that the rebuild was not
measured, and gave as the reason that no box here could deliver DRM frames while carrying two
outputs. that was wrong, and wrong for a dull reason: i had been grepping the wrong log. the capture
markers live in the root service's log, not the session server's, so a machine that was delivering
frames read as one that never had.

the machine i work on has three outputs and the third is a dummy plug, so it can be rotated without
disturbing anything. with a build that predates this PR, rotating it 90 degrees leaves the client in
a landscape frame showing the desktop on its side, dock running down the right edge. measured at
`dcb9535fb` the client switches to a portrait frame with the image upright, and the server logs the
whole chain in under a second:

```
10:02:08.867 desktop layout changed, update mouse resolution: (0, 4864), (0, 1280)
10:02:08.877 Error of monitor1 service: drm: display 1 layout changed; rebuilding
10:02:08.908 drm capture stream ended: stopped
10:02:09.586 drm: capture reader for crtc 268 opened in 285.906us
10:02:09.586 drm: first frame for crtc 268 in 3.508398ms (dma-buf path)
```

the head has moved twice since that run, so this measures the rebuild path rather than the two
commits after it; those are covered by the tests in each.

fufesou:
Reviewed at `70c8ba6fa7eeef72a011830da622f13e68a644f1`.

## Findings

1. **[P2] Preserve a capturer snapshot that loses the race with the first poll** (`display_service.rs:104`)

   The new handoff fixes the reported sequence when `note_capturer_layout()` runs before the layout poll, but those operations are not synchronized. The display loop normally checks the cached display list before polling. That does not close this gap: while a failed lookup is backed off, the display-list check can skip the cache; if the constructor's successful retry clears the failure stamp between that check and the live poll, the poll proceeds while the constructor is still finishing the snapshot.

   `get_displays()` obtains layout A and keeps the display-cache mutex while running the primary-output probes (`display.rs:251-278`). If the output rotates, the live query can obtain B because it bypasses that cache. The failed initialization left both `baseline` and `seen` empty, so the poll can take `WAYLAND_LAYOUT` and wait for the cache in `wayland_snapshot_missing()` (`display_service.rs:178-188`). Once A is published, `snapshot_missing` is false; the poll returns false from `edge()` and records B through `observe()`. `IpcDrmCapturer::new()` then calls `note_capturer(A)` (`drm_capturer.rs:350-355`), but line 104 discards A because `seen` already contains B. Later polls compare B with B, so the generation is never bumped and the capturer keeps A's stale transform until another layout event or session restart.

   Do not discard a build layout that differs from the layout already observed by the poll. Preserve a pending promotion or otherwise synchronize the handoff. Add a regression test for an empty baseline where the poll observes B after the constructor takes A but before `note_capturer(A)` runs.

fufesou:
@fxd0h This PR works well, but the cursor is off:

On my physical machine (Ubuntu 24.04)

1. Landscape & Portrait Right: OK.
2. Portrait Left: Y offset—the cursor appears above the actual click point.
3. Landscape (flipped): The cursor is upside down.

fufesou:
> @fxd0h This PR works well, but the cursor is off:
> 
> On my physical machine (Ubuntu 24.04)
> 
> 1. Landscape & Portrait Right: OK.
> 2. Portrait Left: Y offset—the cursor appears above the actual click point.
> 3. Landscape (flipped): The cursor is upside down.

Hi @fxd0h  , I'll file a PR to address this issue. I'm posting here to avoid duplicating work.

fxd0h:
i owe you an answer here since the 4th. reproduced and fixed in fxd0h/rustdesk:fix/drm-rotated-cursor, measured on i915 + mutter with the default arrow: hotspot off the tip at 90 and 270 and the sprite upside down at 180 before; on the tip at all four angles and upright after. the why is in the fix commit's message. take the branch if you prefer, or i open the pr today, your call.

fufesou:
@fxd0h I'm working on cursor issues across all platforms, including DRM cursor. I'll reference your changes but use a separate PR since there are too many changes.

PR #15918, opened 2026-08-20
fix: dialog, trackpad speed, buttons (close -> ok, cancel)

fufesou:
Change the dialog buttons from "Close" to "Ok" and "Cancel". To make the function clear.

"Custom image quality" dialog is not changed because it needs a live preview.

## Preview

before

after

other

## Summary by CodeRabbit

* **UI Improvements**
  * Updated the trackpad speed dialog with clearer Cancel and OK actions.
  * Added validation for trackpad speed input and clearer handling of invalid values.
  * Dialog actions are temporarily disabled while changes are being saved.
  * Valid, changed values are saved and reflected in the input controls.
  * Save failures now provide an error notification, while unchanged values are not applied.
  * Improved text entry and submission handling for trackpad speed values.

The PR replaces the trackpad-speed dialog’s single Close action with explicit Cancel and OK actions and makes saving validation-aware and asynchronous.
- Adds range validation, save-error feedback, and duplicate-submission prevention.
- Extends `TrackpadSpeedWidget` to report text changes and IME submissions.
- Keeps slider and text-entry values synchronized with the dialog’s submission state.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| flutter/lib/common/widgets/dialog.dart | Reworks the trackpad-speed dialog around explicit cancellation, validated submission, asynchronous persistence, and save-error handling. |
| flutter/lib/common/widgets/setting_widgets.dart | Adds text-change and submission callbacks while preserving the existing standalone widget behavior. |

</details>

<sub>Reviews (4): Last reviewed commit: ["fix(flutter): sync trackpad speed input ..."](https://github.com/rustdesk/rustdesk/commit/d23f35507a81255d375888c7c4c4c7639ebc6f7d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55070736)</sub>

### DIFF
--- a/flutter/lib/common/widgets/dialog.dart
+++ b/flutter/lib/common/widgets/dialog.dart
@@ -1900,25 +1900,43 @@ customImageQualityDialog(SessionID sessionId, String id, FFI ffi) async {
 }
 
 trackpadSpeedDialog(SessionID sessionId, FFI ffi) async {
-  int initSpeed = ffi.inputModel.trackpadSpeed;
+  final initSpeed = ffi.inputModel.trackpadSpeed;
   final curSpeed = SimpleWrapper(initSpeed);
-  final btnClose = dialogButton('Close', onPressed: () async {
-    if (curSpeed.value <= kMaxTrackpadSpeed &&
-        curSpeed.value >= kMinTrackpadSpeed &&
-        curSpeed.value != initSpeed) {
-      await bind.sessionSetTrackpadSpeed(
-          sessionId: sessionId, value: curSpeed.value);
-      await ffi.inputModel.updateTrackpadSpeed();
+  ffi.dialogManager.show((setState, close, context) {
+    submit() async {
+      if (curSpeed.value <= kMaxTrackpadSpeed &&
+          curSpeed.value >= kMinTrackpadSpeed &&
+          curSpeed.value != initSpeed) {
+        await bind.sessionSetTrackpadSpeed(
+            sessionId: sessionId, value: curSpeed.value);
+        await ffi.inputModel.updateTrackpadSpeed();
+      }
+      close();
     }
-    ffi.dialogManager.dismissAll();
-  });
-  msgBoxCommon(
-      ffi.dialogManager,
-      'Trackpad speed',
-      TrackpadSpeedWidget(
-        value: curSpeed,
+
+    return CustomAlertDialog(
+      title: Text(
+        translate('Trackpad speed'),
+        style: TextStyle(fontSize: 21),
       ),
-      [btnClose]);
+      content: TrackpadSpeedWidget(value: curSpeed),
+      actions: [
+        dialogButton(
+          'Cancel',
+          icon: Icon(Icons.close_rounded),
+          onPressed: close,
+          isOutline: true,
+        ),
+        dialogButton(
+          'OK',
+          icon: Icon(Icons.done_rounded),
+          onPressed: submit,
+        ),
+      ],
+      onSubmit: submit,
+      onCancel: close,
+    );
+  });
 }
 
 void deleteConfirmDialog(Function onSubmit, String title) async {
--- a/flutter/lib/common/widgets/dialog.dart
+++ b/flutter/lib/common/widgets/dialog.dart
@@ -1960,10 +1960,11 @@ void trackpadSpeedDialog(SessionID sessionId, FFI ffi) {
   var speedText = initSpeed.toString();
   var isSubmitting = false;
   ffi.dialogManager.show((setState, close, context) {
-    Future<void> submit() async {
+    Future<void> submit([String? submittedText]) async {
       if (isSubmitting) {
         return;
       }
+      speedText = submittedText ?? speedText;
       final speed = _validateTrackpadSpeed(speedText);
       if (speed == null) {
         return;
@@ -1991,6 +1992,7 @@ void trackpadSpeedDialog(SessionID sessionId, FFI ffi) {
       content: TrackpadSpeedWidget(
         value: curSpeed,
         onTextChanged: (text) => speedText = text,
+        onTextSubmitted: submit,
       ),
       actions: _trackpadSpeedDialogActions(
         isSubmitting: isSubmitting,
--- a/flutter/lib/common/widgets/dialog.dart
+++ b/flutter/lib/common/widgets/dialog.dart
@@ -1899,19 +1899,88 @@ customImageQualityDialog(SessionID sessionId, String id, FFI ffi) async {
   msgBoxCommon(ffi.dialogManager, 'Custom Image Quality', content, [btnClose]);
 }
 
-trackpadSpeedDialog(SessionID sessionId, FFI ffi) async {
+int? _validateTrackpadSpeed(String text) {
+  final speed = int.tryParse(text);
+  if (speed == null || speed < kMinTrackpadSpeed || speed > kMaxTrackpadSpeed) {
+    BotToast.showText(
+      text:
+          '${translate('Invalid format')}: $kMinTrackpadSpeed-$kMaxTrackpadSpeed',
+      contentColor: Colors.red,
+    );
+    return null;
+  }
+  return speed;
+}
+
+Future<void> _saveTrackpadSpeed({
+  required SessionID sessionId,
+  required FFI ffi,
+  required int initSpeed,
+  required int speed,
+}) async {
+  if (speed == initSpeed) {
+    return;
+  }
+  await bind.sessionSetTrackpadSpeed(sessionId: sessionId, value: speed);
+  await ffi.inputModel.updateTrackpadSpeed();
+}
+
+void _showTrackpadSpeedSaveError(Object error, StackTrace stackTrace) {
+  debugPrint('Failed to save trackpad speed: $error');
+  debugPrintStack(stackTrace: stackTrace);
+  BotToast.showText(
+    text: translate('Failed'),
+    contentColor: Colors.red,
+  );
+}
+
+List<Widget> _trackpadSpeedDialogActions({
+  required bool isSubmitting,
+  required VoidCallback close,
+  required VoidCallback submit,
+}) {
+  return [
+    dialogButton(
+      'Cancel',
+      icon: Icon(Icons.close_rounded),
+      onPressed: isSubmitting ? null : close,
+      isOutline: true,
+    ),
+    dialogButton(
+      'OK',
+      icon: Icon(Icons.done_rounded),
+      onPressed: isSubmitting ? null : submit,
+    ),
+  ];
+}
+
+void trackpadSpeedDialog(SessionID sessionId, FFI ffi) {
   final initSpeed = ffi.inputModel.trackpadSpeed;
   final curSpeed = SimpleWrapper(initSpeed);
+  var speedText = initSpeed.toString();
+  var isSubmitting = false;
   ffi.dialogManager.show((setState, close, context) {
-    submit() async {
-      if (curSpeed.value <= kMaxTrackpadSpeed &&
-          curSpeed.value >= kMinTrackpadSpeed &&
-          curSpeed.value != initSpeed) {
-        await bind.sessionSetTrackpadSpeed(
-            sessionId: sessionId, value: curSpeed.value);
-        await ffi.inputModel.updateTrackpadSpeed();
+    Future<void> submit() async {
+      if (isSubmitting) {
+        return;
       }
-      close();
+      final speed = _validateTrackpadSpeed(speedText);
+      if (speed == null) {
+        return;
+      }
+      setState(() => isSubmitting = true);
+      try {
+        await _saveTrackpadSpeed(
+          sessionId: sessionId,
+          ffi: ffi,
+          initSpeed: initSpeed,
+          speed: speed,
+        );
+        close();
+      } catch (error, stackTrace) {
+        _showTrackpadSpeedSaveError(error, stackTrace);
+        setState(() => isSubmitting = false);
+      }
     }
 
     return CustomAlertDialog(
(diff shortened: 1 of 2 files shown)


rustdesk:
# Code review — PR #15918 (trackpad speed dialog: OK/Cancel buttons)

Target: https://github.com/rustdesk/rustdesk/pull/15918
Effort: xhigh · 9 findings, most severe first
All findings are in `flutter/lib/common/widgets/dialog.dart` unless noted.

## Summary

The PR replaces the `msgBoxCommon`-based trackpad-speed dialog with a hand-rolled
`CustomAlertDialog` so it can have a real **OK** button instead of **Close**. Two
findings combine to make the number box completely non-functional: Enter is now
swallowed before the `TextField` sees it, and there is no `onChanged`, so neither
Enter nor clicking OK commits a typed value. Only the slider still works — while
the button now explicitly promises OK semantics. Fixing either one alone is not
enough.

---

## 1. `onSubmit: submit` swallows Enter — `onSubmitted` never fires (line 1936)

**Correctness.** Passing `onSubmit: submit` makes `CustomAlertDialog` swallow the
Enter key (returns `KeyEventResult.handled`), so the speed `TextField`'s
`onSubmitted` never fires — typing a value and pressing Enter now discards it and
closes the dialog with the old speed.

Open Trackpad speed, click into the number box, type `400`, press Enter.
Previously (`msgBoxCommon` passed no `onSubmit`) the key fell through to
`TrackpadSpeedWidget`'s `TextField.onSubmitted` → `updateValue(400)` →
`curSpeed.value = 400`. Now `CustomAlertDialog.build`'s FocusScope `onKey`
(`flutter/lib/common.dart:1096-1104`) matches
`!tabTapped && onSubmit != null && enter`, calls `submit()` and returns `handled`;
the engine therefore never forwards the key to the text-input plugin, so
`EditableText.performAction` / `onSubmitted` is never invoked. `submit()` reads the
still-stale `curSpeed.value` (`== initSpeed`), the `!= initSpeed` guard fails,
nothing is persisted, dialog closes. The typed 400 is silently lost.
Inconsistently, if the user happened to press Tab first, `tabTapped` becomes true
and Enter works as before — so the bug only hits mouse users.

## 2. TextField has no `onChanged`, so clicking OK persists nothing (line 1922)

**Correctness.** `TrackpadSpeedWidget`'s `TextField` has no `onChanged`, so
`curSpeed` is only updated on Enter or slider drag; clicking the new OK button
after typing a number persists nothing.

Type `250` in the number box and click OK (never pressing Enter). The `TextField`
in `flutter/lib/common/widgets/setting_widgets.dart:313-322` only commits via
`onSubmitted`; losing focus to the button does not call it. `submit()` reads the
unchanged `curSpeed.value`, the `!= initSpeed` check fails, and the dialog closes
reporting success. Combined with the Enter interception above, the text box becomes
entirely non-functional for setting a speed — only the slider works.

**Fix:** add `onChanged` to the `TextField` (parse + clamp into `widget.value`), or
read the controller text inside `submit`.

## 3. `submit()` has no re-entry guard — double-click closes twice (line 1906)

**Correctness.** `submit()` is async with no re-entry guard, while the dialog stays
interactive across its two awaits — double-clicking OK (or Enter key-repeat) runs it
twice and calls `close()` twice.

Move the slider, then double-click OK. The first `submit()` suspends at
`await bind.sessionSetTrackpadSpeed(...)`; the overlay is still mounted and
hit-testable, so the second click starts a second `submit()`. Both pass the
`!= initSpeed` check (`initSpeed` is captured and never updated), issuing duplicate
`sessionSetTrackpadSpeed` + `updateTrackpadSpeed` FFI round-trips. Both then call
`close()`; the second `Dialog.complete` (`flutter/lib/common.dart:762-773`) runs
`entry?.remove()` in its `finally` on an already-removed `OverlayEntry`, tripping
`assert(_overlay != null)` in debug / the `_overlay!` null-check in release. The
throw is swallowed by `close()`'s try/catch, but it logs an error — and the old code
did not have this hazard, because `dismissAll()` was idempotent (it cleared
`_dialogs` first).

**Fix:** add a `bool submitting` latch, or disable the button while in flight.

## 4. `close()` not in a `finally` — FFI failure leaves the dialog stuck open (line 1910)

**Correctness.** `close()` is not in a `finally`, and the `Future` returned by
`submit` is dropped by `onPressed`, so an FFI failure leaves the dialog permanently
open with an unhandled async error and no user feedback.

The session is torn down (peer disconnect / tab closed) while the dialog is open,
then the user clicks OK. `await bind.sessionSetTrackpadSpeed(...)` throws; `close()`
on line 1914 is never reached, so the dialog stays on screen and clicking OK again
does nothing visible. Because `dialogButton`'s `onPressed` is a `VoidCallback`, the
rejected `Future` is unawaited → unhandled zone error.

**Fix:** wrap the body in try/finally (or try/catch with a toast) so the dialog
always closes and errors are surfaced.

## 5. Invalid input silently discarded under OK semantics (line 1907)

**Correctness.** OK silently discards out-of-range or unparseable input instead of
reporting it — with the old **Close** label this was defensible; with **OK** the
user believes the value was applied.

User types `abc` or `99999`: `int.tryParse` returns null (no update) or
`updateValue` clamps. Either way `submit()` finds `curSpeed.value == initSpeed`,
skips the save, and closes with no message. The dialog now presents an explicit
confirm/cancel contract, so a rejected value must either be corrected in-place or
reported.

## 6. Altitude: hand-rolled dialog instead of `msgBoxCommon` `onSubmit` (line 1917)

The fix drops down to a raw `CustomAlertDialog` for this one dialog instead of
adding an optional `onSubmit`/OK button to the shared `msgBoxCommon`, leaving the
underlying limitation in place for every other caller.

`msgBoxCommon` (`flutter/lib/common.dart:1393-1404`) only supports `onCancel`, which
is exactly why this call site had to be rewritten. Adding an optional `onSubmit`
parameter there would have been a purely additive ~3-line change that also unblocks
`customImageQualityDialog` (line 1899, still on "Close") and the two `msgBoxCommon`
call sites in `flutter/lib/desktop/pages/desktop_setting_page.dart:2738,2769`. It
also matches the TODO already sitting on `CustomAlertDialog`
(`flutter/lib/common.dart:1058-1060`: "Make dead code alive, transform arguments
`onSubmit` and `onCancel` into corresponding buttons"). AGENTS.md's "Be minimally
invasive" asks for "purely additive changes: layer new blocks or new functions
around existing code instead of restructuring it"; this hunk deletes the
`msgBoxCommon` call and rebuilds the dialog by hand.

## 7. Reuse: Cancel/OK actions block copy-pasted from `deleteConfirmDialog` (line 1923)

Lines 1924-1934 are character-for-character identical to lines ~1958-1972
(`deleteConfirmDialog`), and the same pair is repeated across ~66 `dialogButton(`
uses in this file. A change to the standard confirm-dialog button styling (icon,
outline, ordering, translation key) now has to be made in one more place.

**Fix:** extract an `okCancelActions({required VoidCallback onOk, required VoidCallback onCancel})`
helper (or the `ConfirmOkButton`/`CancelButton` widgets the `CustomAlertDialog` TODO
already calls for) and use it from both.

## 8. Simplification: range check in `submit()` is dead code (line 1907)

The `<= kMaxTrackpadSpeed && >= kMinTrackpadSpeed` range check can never fail.
`curSpeed` is seeded from `ffi.inputModel.trackpadSpeed`, which
`InputModel.updateTrackpadSpeed` (`flutter/lib/models/input_model.dart:610-618`)
already forces into `[kMin, kMax]`, and the only mutator is
`TrackpadSpeedWidgetState.updateValue`, which does
`newValue.clamp(kMinTrackpadSpeed, kMaxTrackpadSpeed)`
(`flutter/lib/common/widgets/setting_widgets.dart:272`). Both bounds tests are
always true, so the three-clause condition reduces to `curSpeed.value != initSpeed`
— which also makes the real intent (skip the FFI write when nothing changed)
obvious.

## 9. Simplification: `trackpadSpeedDialog` is `async` with no `await` (line 1902)

The rewrite removed the last `await` from the body, and the `Future` returned by
`dialogManager.show(...)` is silently dropped. The function now returns an
immediately-completed `Future<dynamic>` that its only caller
(`flutter/lib/desktop/widgets/remote_toolbar.dart:2578`,
`onPressed: enabled ? () => trackpadSpeedDialog(sessionId, ffi) : null`) discards,
which reads as if the caller could await dialog completion when it cannot.

**Fix:** drop `async` and make it `void`, or return `dialogManager.show(...)` so the
completer is actually usable.

fufesou:
https://github.com/rustdesk/rustdesk/pull/15918#issuecomment-5353290303

1 & 2. Agree; fixed Enter/Ok by tracking raw text before the dialog handles the key.
3. Agree; added a submission latch and disabled actions while saving.
4. Agree with error handling, but clarify that disconnect normally causes the Rust setter to return silently rather than throw. Actual bridge exceptions are now logged and shown to the user.
5. Agree; invalid and out-of-range input now shows an error and keeps the dialog open.
6. No change; expanding msgBoxCommon would broaden this focused PR.
7. No shared refactor; dialogButton already centralizes styling, and changing existing dialogs is outside scope.
8. Agree; removed the dead check. The new range check applies to raw, unclamped text, so it is meaningful.
9. Agree; changed the function to return void.

rustdesk:
# Code review — PR #15918 (round 2)

3 findings, all in code this PR adds. They share one root cause: typed text is now
routed through the new `onTextChanged`/`speedText` path instead of through
`TrackpadSpeedWidgetState.updateValue`, so the widget's own value and the dialog's
value are no longer the same thing.

## 1. `submitTextValue` early-returns — Enter no longer commits

`flutter/lib/common/widgets/setting_widgets.dart:288`

```dart
void submitTextValue(String text) {
  if (widget.onTextChanged != null) {
    return;                        // the dialog always passes onTextChanged
  }
  ...
}
```

`CustomAlertDialog` only intercepts Enter while `!tabTapped`
(`flutter/lib/common.dart:1093-1104`), and `tabTapped` is set permanently by the
first Tab press. Press Tab once, click back into the number box, type 400, press
Enter → the key reaches the `TextField`, `onSubmitted` → `submitTextValue` → return.
The dialog does not submit and does not close; the key is inert.

The typed value itself is not lost — `onChanged: widget.onTextChanged` already wrote
it to `speedText`, so clicking OK afterwards still saves it. The defect is that
Enter is silently dead on this path, where before this PR it ran `updateValue(400)`.
IME / soft-keyboard "Done" actions hit the same dead path, which is the case with no
keyboard-shortcut alternative.

CodeRabbit flagged the same lines (Functional Correctness · Major):

> **Forward `TextField.onSubmitted` to the dialog submission flow.**
>
> When `onTextChanged` is set, `submitTextValue` returns without notifying
> `trackpadSpeedDialog`. A software-keyboard action invokes `TextField.onSubmitted`
> directly, so the typed value can remain unsaved until the user taps OK.
>
> - Add an optional submitted-text callback.
> - Invoke it from `submitTextValue`.
> - Update `speedText` and invoke `submit()` from the callback.

Same defect, and the direction is right. It keeps `speedText` and `curSpeed` as two
separate values, though, so finding 2 survives it — see the combined fix below.

## 2. Typing bypasses `updateValue`, so the slider and the box disagree

`flutter/lib/common/widgets/dialog.dart:1993`

`onTextChanged: (text) => speedText = text` never touches `widget.value.value`
(`curSpeed`), which is what the slider renders.

- Type 500 → the slider still sits at 100 while the box reads 500 → OK saves 500.
- Type 500, then nudge the slider one step → `updateValue` overwrites
  `_controller.text` and `speedText` with the slider value, silently discarding the
  typed 500.

The dialog now reads `speedText` exclusively and never reads `curSpeed`, so
`curSpeed` only exists to display a stale number.


## 3. `speedText` can go stale against the visible field after a failed save

`flutter/lib/common/widgets/dialog.dart:1990`

`isSubmitting` disables `actions`/`onSubmit`/`onCancel` but not `content:`, so the
field stays editable while the save is in flight.

Type 500 → OK (`speed=500` captured) → clear the field mid-flight → `speedText = ''`
→ the save throws → `setState(isSubmitting = false)` → rebuild →
`setting_widgets.dart:311-312` sees an empty controller and repopulates it from
`curSpeed` (still 100). The field now reads `100` while `speedText` is `''`, so the
next OK pops "`Invalid format`: 10-1000" for a value the user can see is valid, with
no way out but retyping.

Fix falls out of finding 2: with a single source of truth there is no second copy to
go stale.

## Fix for all three

Two changes, both in `TrackpadSpeedWidget`:

1. Route typed text through `updateValue` (which already calls `onTextChanged`) and
   drop `speedText`, so the widget's value is the only value. Kills findings 2 and 3.
2. Replace the `onTextChanged != null` early return with an explicit submitted-text
   callback the dialog wires to `submit()`. Kills finding 1.

With both, `submitTextValue` needs no branch at all — it parses, calls `updateValue`,
then reports the submission.

fufesou:
https://github.com/rustdesk/rustdesk/pull/15918#issuecomment-5355171547

## 1. Enter and IME submission

I added a separate `onTextSubmitted` callback and wired it to the dialog's `submit()` function.
This covers IME/soft-keyboard actions as well as physical Enter when the dialog's raw Enter handler does not handle the key.

## 2. Slider and text field getting out of sync

Typed input now goes through `updateTextValue`. It forwards the raw text to the dialog and updates the shared numeric value when the input is valid and in range. This keeps the slider in sync while preserving empty, invalid, and out-of-range text for validation.

## 3. Stale text after a failed save

This can happen, but only in a narrow case: the save must stay pending long enough for the field to be cleared, and then the save must throw.

The save only writes local peer configuration. There is no network round trip, and it normally finishes quickly. A missing session is a no-op, configuration write failures are logged instead of thrown, and the getter has a fallback. A reliable reproduction therefore needs an artificial delay and a forced exception.

The result is one misleading validation message. Typing the value again or moving the slider fixes the state, and no saved data is corrupted. Given the narrow timing and low impact, I do not think adding another enabled state or reworking the widget is justified for this PR, so I have not made a change for this finding.

Issue #15945, opened 2026-08-22
Kernel paged pool leak (pooltag Gpbm, dxgkrnl) on controlled Windows host - ~2.5GB in 14h, released instantly when RustDesk exits


BD7PIL:
## Bug Description

- Kernel paged pool grows unbounded on the **controlled host** while the RustDesk service is running
- PoolMon after 14h uptime: tag `Gpbm` (dxgkrnl graphics) = **2555.9 MB**, out of 3.22 GB total paged pool (**81%**)
- `Gpbm` allocation counter churned continuously (~33 allocs/sec average since boot)
- Differential test: stopping the RustDesk service + killing all RustDesk processes released **~2.5 GB instantly** (Gpbm 2555.9 MB -> 39.2 MB; system paged pool total 3.22 GB -> 0.77 GB)
- Days **without** in-session display resolution changes stay flat (~700 MB over several days), so growth correlates with changing resolution inside the remote session

## How to Reproduce

1. Connect to a Windows host via RustDesk remote session
2. Change the host's display resolution from within the session (RustDesk toolbar)
3. Repeat occasionally across sessions / keep host running overnight
4. Observe kernel paged pool growth via PoolMon (`Gpbm` tag) or Performance Monitor `\Memory\Pool Paged Bytes`

## Expected Behavior

- Paged pool usage should return to baseline after resolution changes / session disconnect
- No unbounded growth of kernel graphics allocations tied to the RustDesk process lifetime

## Operating system(s)

Windows (controlling, desktop-f5cpuoo) -> Windows 11 Pro 25H2 build 26200.9168 (controlled)

## RustDesk Version(s)

1.4.x client (exact version not recorded on controlled side) -> 1.4.9+67 (controlled)

## Screenshots

Differential test evidence from the controlled host (NtQuerySystemInformation SystemPoolTagInformation):

```
RustDesk running:    Gpbm = 2555.9 MB   (paged pool total 3.22 GB)
Stop RustDesk svc:   Gpbm =   39.2 MB   (paged pool total 0.77 GB)
                     -> ~2.5 GB freed instantly on process exit
```

Full pooltag snapshot at time of test: `Gpbm P_used=2550.6MB alloc=1676329 free=1672514`

## Additional Context

- Controlled host: hybrid graphics laptop (Intel iGPU driver 32.0.101.8864 + NVIDIA RTX 5070 Laptop GPU driver 32.0.16.1088), connected via direct IP over Tailnet
[...]

PR #15957, opened 2026-08-24
fix: reuse the dxgi staging texture instead of one per frame (wait for issue OP verify)

rustdesk:
ohgodwhat() created a full screen D3D11_USAGE_STAGING texture for every captured frame and pinned each one with SetEvictionPriority(MAXIMUM). Because D3D11 resource destruction may be deferred, that per-frame churn can accumulate a large amount of graphics kernel paged pool on affected drivers. Keep a single staging texture and rebuild it only when the desktop image changes shape.

Also check the HRESULT of the QueryInterface calls in dxgi capture: the IDXGISurface cast could leave surface null while readable holds a valid texture, and the pre-existing IDXGIResource -> ID3D11Texture2D casts fed a null pointer to GetDesc() / the vram encoder.

Reported in #15945.

Claude-Session: https://claude.ai/code/session_018vpiLqteCBiTiuNYMka9LU

## Summary by CodeRabbit

* **Performance**
  * Improved DXGI screen capture efficiency by reusing capture resources between frames.
  * Reduced unnecessary resource creation during ongoing capture sessions, supporting smoother and more consistent capture performance.

* **Bug Fixes**
  * Capture resources now refresh automatically when screen dimensions or formats change, improving reliability during display or configuration changes.
  * Improved handling of changing capture conditions to help prevent interruptions during active screen capture.

The PR caches the DXGI staging texture across captured frames and recreates it when the captured image dimensions or format change. It also checks texture and surface QueryInterface results before using the returned COM pointers.
- Adds persistent staging-texture ownership to the DXGI capturer.
- Rebuilds the cached staging texture when width, height, or format changes.
- Propagates failed COM interface queries instead of dereferencing null results.
</details>

The PR appears safe to merge.

No blocking failure remains.
</details>

| Filename | Overview |
|----------|----------|
| libs/scrap/src/dxgi/mod.rs | Reuses a cached DXGI staging texture and adds HRESULT propagation for the changed COM interface-query paths. |

</details>

```mermaid
flowchart TD
    A[Acquire DXGI frame] --> B[Query ID3D11Texture2D]
    B --> C[Read frame descriptor]
    C --> D{Cached staging texture matches?}
    D -- No --> E[Create staging texture]
    E --> F[Query IDXGISurface]
    F --> G[Cache texture and surface]
    D -- Yes --> H[Reuse cached resources]
    G --> I[Copy frame into staging texture]
    H --> I
    I --> J[Map cached surface]
    J --> K[Return pixel buffer]
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix: check the frame QueryInterface resu..."](https://github.com/rustdesk/rustdesk/commit/9343affe0bf90110a494134a8076c4060d41bbd5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56210940)</sub>

### DIFF
--- a/libs/scrap/src/dxgi/mod.rs
+++ b/libs/scrap/src/dxgi/mod.rs
@@ -357,10 +357,10 @@ impl Capturer {
     // copy from GPU memory to system memory
     unsafe fn ohgodwhat(&mut self, frame: *mut IDXGIResource) -> io::Result<()> {
         let mut texture: *mut ID3D11Texture2D = ptr::null_mut();
-        (*frame).QueryInterface(
+        wrap_hresult((*frame).QueryInterface(
             &IID_ID3D11Texture2D,
             &mut texture as *mut *mut _ as *mut *mut _,
-        );
+        ))?;
         let texture = ComPtr(texture);
 
         #[allow(invalid_value)]
@@ -500,10 +500,10 @@ impl Capturer {
             }
 
             let mut texture: *mut ID3D11Texture2D = ptr::null_mut();
-            (*frame.0).QueryInterface(
+            wrap_hresult((*frame.0).QueryInterface(
                 &IID_ID3D11Texture2D,
                 &mut texture as *mut *mut _ as *mut *mut _,
-            );
+            ))?;
             let texture = ComPtr(texture);
             self.texture = texture;
 
--- a/libs/scrap/src/dxgi/mod.rs
+++ b/libs/scrap/src/dxgi/mod.rs
@@ -48,6 +48,7 @@ pub struct Capturer {
     duplication: ComPtr<IDXGIOutputDuplication>,
     fastlane: bool,
     surface: ComPtr<IDXGISurface>,
+    readable: ComPtr<ID3D11Texture2D>,
     texture: ComPtr<ID3D11Texture2D>,
     width: usize,
     height: usize,
@@ -163,6 +164,7 @@ impl Capturer {
             duplication: ComPtr(duplication),
             fastlane: desc.DesktopImageInSystemMemory == TRUE,
             surface: ComPtr(ptr::null_mut()),
+            readable: ComPtr(ptr::null_mut()),
             texture: ComPtr(ptr::null_mut()),
             width: display.width() as usize,
             height: display.height() as usize,
@@ -346,14 +348,14 @@ impl Capturer {
         if self.fastlane {
             wrap_hresult((*self.duplication.0).MapDesktopSurface(&mut rect))?;
         } else {
-            self.surface = ComPtr(self.ohgodwhat(frame.0)?);
+            self.ohgodwhat(frame.0)?;
             wrap_hresult((*self.surface.0).Map(&mut rect, DXGI_MAP_READ))?;
         }
         Ok((rect.pBits, rect.Pitch))
     }
 
     // copy from GPU memory to system memory
-    unsafe fn ohgodwhat(&mut self, frame: *mut IDXGIResource) -> io::Result<*mut IDXGISurface> {
+    unsafe fn ohgodwhat(&mut self, frame: *mut IDXGIResource) -> io::Result<()> {
         let mut texture: *mut ID3D11Texture2D = ptr::null_mut();
         (*frame).QueryInterface(
             &IID_ID3D11Texture2D,
@@ -370,24 +372,37 @@ impl Capturer {
         texture_desc.CPUAccessFlags = D3D11_CPU_ACCESS_READ;
         texture_desc.MiscFlags = 0;
 
-        let mut readable = ptr::null_mut();
-        wrap_hresult((*self.device.0).CreateTexture2D(
-            &mut texture_desc,
-            ptr::null(),
-            &mut readable,
-        ))?;
-        (*readable).SetEvictionPriority(DXGI_RESOURCE_PRIORITY_MAXIMUM);
-        let readable = ComPtr(readable);
+        // Avoid per-frame staging texture allocation and the kernel allocation churn it causes.
+        let mut current: D3D11_TEXTURE2D_DESC = mem::zeroed();
+        if !self.surface.is_null() {
+            (*self.readable.0).GetDesc(&mut current);
+        }
+        if current.Width != texture_desc.Width
+            || current.Height != texture_desc.Height
+            || current.Format != texture_desc.Format
+        {
+            let mut readable = ptr::null_mut();
+            wrap_hresult((*self.device.0).CreateTexture2D(
+                &mut texture_desc,
+                ptr::null(),
+                &mut readable,
+            ))?;
+            (*readable).SetEvictionPriority(DXGI_RESOURCE_PRIORITY_MAXIMUM);
+            let readable = ComPtr(readable);
 
-        let mut surface = ptr::null_mut();
-        (*readable.0).QueryInterface(
-            &IID_IDXGISurface,
-            &mut surface as *mut *mut _ as *mut *mut _,
-        );
+            let mut surface = ptr::null_mut();
+            wrap_hresult((*readable.0).QueryInterface(
+                &IID_IDXGISurface,
+                &mut surface as *mut *mut _ as *mut *mut _,
+            ))?;
 
-        (*self.context.0).CopyResource(readable.0 as *mut _, texture.0 as *mut _);
+            self.readable = readable;
+            self.surface = ComPtr(surface);
+        }
 
-        Ok(surface)
+        (*self.context.0).CopyResource(self.readable.0 as *mut _, texture.0 as *mut _);
+
+        Ok(())
     }
 
     pub fn frame<'a>(&'a mut self, timeout: UINT) -> io::Result<Frame<'a>> {


rustdesk:
## Review verdict: **Approve — no blocking findings**

I reviewed the two commits and the surrounding DXGI capture lifecycle. The core change looks correct: the staging `ID3D11Texture2D` and its `IDXGISurface` are retained on `Capturer`, reused between frames, and replaced together when the cached texture no longer matches. The previous surface is unmapped before the next frame is acquired, so the reused staging resource is not still mapped when `CopyResource` runs. 

The added `QueryInterface` HRESULT handling is also a good fix. Previously both the CPU path and VRAM path could continue with a null `ID3D11Texture2D`; the PR now propagates the failure before dereferencing/storing it. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/15957/files)][1])

### One non-blocking hardening suggestion

The cache validity check only compares:

```rust
current.Width != texture_desc.Width
    || current.Height != texture_desc.Height
    || current.Format != texture_desc.Format
```

`CopyResource` requires compatible resources with matching dimensions, and for a general `ID3D11Texture2D` that also implicates things such as mip/array layout. ([[Microsoft Learn](https://learn.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-copyresource?utm_source=chatgpt.com)][2])

For **this specific Desktop Duplication path**, I would not block on that: DXGI documents the desktop image format as fixed `DXGI_FORMAT_B8G8R8A8_UNORM`, and a display mode change invalidates the `IDXGIOutputDuplication` object anyway, requiring it to be recreated. ([[Microsoft Learn](https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/desktop-dup-api?utm_source=chatgpt.com)][3])

Still, I'd consider either comparing all descriptor fields that matter to copying, or explicitly documenting the Desktop Duplication invariant. That makes the code less fragile if this path is later changed to `DuplicateOutput1` or fed a different texture source.

Something like:

```rust
let compatible = current.Width == texture_desc.Width
    && current.Height == texture_desc.Height
    && current.MipLevels == texture_desc.MipLevels
    && current.ArraySize == texture_desc.ArraySize
    && current.Format == texture_desc.Format
    && current.SampleDesc.Count == texture_desc.SampleDesc.Count
    && current.SampleDesc.Quality == texture_desc.SampleDesc.Quality;
```

I'd rate that **P3 / defensive**, not a required PR change.

### Resource/lifetime checks

I specifically checked the potentially dangerous parts:

* `readable` and `surface` start null and are populated together. 
* If `CreateTexture2D` or `QueryInterface(IDXGISurface)` fails, the newly-created local COM resource is dropped while the previous cached pair remains intact. 
* When replacing the cache, COM ownership is preserved correctly: `IDXGISurface` has its own QI reference, so dropping the old `ID3D11Texture2D` reference before replacing the old surface does not prematurely destroy the underlying object.
* The previous staging surface is `Unmap()`'d before the next `load_frame()` and subsequent `CopyResource`. 
* `Map` still provides the synchronization point after the GPU copy, which is the normal staging-texture readback pattern.

The approach—maintaining and reusing a staging texture rather than creating one every frame—is also the normal D3D11 pattern. Microsoft-related examples and other mature capture implementations use persistent staging resources similarly. ([[GitHub](https://github.com/obsproject/obs-studio/blob/master/plugins/win-capture/graphics-hook/d3d11-capture.cpp?utm_source=chatgpt.com)][4])

### What I would test before merge

The most valuable test isn't a unit test; it's the reproduction from #15945:

1. Run a long Windows DXGI capture session and watch the `Gpbm` pool tag.
2. Change RustDesk-controlled display resolution repeatedly.
3. Verify pool usage reaches a bounded plateau rather than continuing to grow.
4. Exercise connect/disconnect and DXGI `ACCESS_LOST` recovery.
5. Run both normal pixel-buffer capture and `vram`/texture output paths.

The reported issue shows `Gpbm` reaching about 2.5 GB and being immediately released when RustDesk exits, so a PoolMon comparison before/after this patch would give much stronger evidence than CI alone. ([[GitHub](https://github.com/rustdesk/rustdesk/issues/15945)][5])

I couldn't independently execute the Windows capture path here. GitHub has x86-64 and ARM64 Windows CI jobs for this PR, but their logs are currently expired/re-running, so they don't provide useful runtime validation of the leak. ([[GitHub](https://github.com/rustdesk/rustdesk/actions/runs/33048131141/job/98438363280?pr=15957)][6])

**Overall:** I would approve `#15957`. I don't see a merge-blocking correctness issue in the implementation.

[1]: https://github.com/rustdesk/rustdesk/pull/15957/files "fix: reuse the dxgi staging texture instead of one per frame by rustdesk · Pull Request #15957 · rustdesk/rustdesk · GitHub"
[2]: https://learn.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-copyresource?utm_source=chatgpt.com "ID3D11DeviceContext::CopyResource (d3d11.h) - Win32 apps | Microsoft Learn"
[3]: https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/desktop-dup-api?utm_source=chatgpt.com "Desktop Duplication API - Win32 apps | Microsoft Learn"
[4]: https://github.com/obsproject/obs-studio/blob/master/plugins/win-capture/graphics-hook/d3d11-capture.cpp?utm_source=chatgpt.com "obs-studio/plugins/win-capture/graphics-hook/d3d11-capture.cpp at master · obsproject/obs-studio · GitHub"
[5]: https://github.com/rustdesk/rustdesk/issues/15945 "Kernel paged pool leak (pooltag Gpbm, dxgkrnl) on controlled Windows host - ~2.5GB in 14h, released instantly when RustDesk exits · Issue #15945 · rustdesk/rustdesk · GitHub"
[6]: https://github.com/rustdesk/rustdesk/actions/runs/33048131141/job/98438363280?pr=15957 "fix: reuse the dxgi staging texture instead of one per frame · rustdesk/rustdesk@9343aff · GitHub"

BD7PIL:
OP of #15945 here — providing the verification data this PR's summary asks for ("wait for issue OP verify"). I couldn't comment on the issue because it's been locked (only maintainers can comment there), so posting here. Everything below is measured on the affected host.

## Confirms the PR's mechanism: per-frame staging texture + SetEvictionPriority on the affected driver

**Encoder evidence** (the controlled host encodes on Intel QSV at the large native panel resolution via VRAM H.264):

```
[libs\scrap\src\common\codec.rs:135] new encoder: VRAM(VRamEncoderConfig {
  device: AdapterDevice { vendor_id: 32902 (0x8086 = Intel), luid: 73535 },
  width: 3200, height: 2000, quality: 0.67,
  feature: FeatureContext { driver: FFMPEG, vendor: MFX, luid: 73535, data_format: H264 }
}), i444: false
```

So capture + VRAM encode run on the **Intel iGPU at 3200×2000** — a very large surface, matching "full screen D3D11_USAGE_STAGING texture ... on affected drivers". Hybrid laptop also lists an NVIDIA LUID in `vram_encode` (cross-adapter path, consistent with Desktop Duplication forcing onto iGPU per MS KB3019314).

## Three observations, measured with PoolMon (Gpbm tag, NtQuerySystemInformation SystemPoolTagInformation)

**A. Growth is pulse-shaped, not a continuous per-frame leak.**
~168 h boot → **Gpbm = 10,243.5 MB (93% of the 11.16 GB paged pool)**. The earlier 14 h observation was 2.5 GB, so +154 h of uptime only added ~7.7 GB — NOT proportional to uptime. When idle (no remote session) a 60 s sample of the paged pool was flat (-7.8 MB). Growth happened during brief display events.

**B. Killing RustDesk no longer releases it (behavioral change vs. the original report).**
Originally (14 h) stopping RustDesk released Gpbm instantly (2.5 GB → 39 MB). After 7 days, stopping the service AND killing all 4 RustDesk processes left the pool unchanged (~11.18 GB). Ledger: Gpbm `Allocs=8,297,900 / Frees=8,293,268 → Unreleased=4,632` (0.1%), but each averages **~2.2 MB** — surface-sized allocations (a 3200×2000-ish framebuffer), not tiny per-frame scraps. This fits deferred-destroy + pinned (`SetEvictionPriority(MAXIMUM)`) resources surviving process teardown.

**C. The pool is pinned in the kernel adapter layer, not DWM.**
Restarted DWM (`Stop-Process dwm`): **Gpbm did not move, 10,224.1 MB → 10,224.1 MB**. So the ~10 GB is not in the compositor; it's held by dxgkrnl-internal cross-adapter/paged structures reclaimable only by adapter reset or reboot.

## What I'll do next
- Install this build (`dxgi-reuse-staging-texture-build`, rustdesk-1.4.9-x86_64.exe, 23.2 MB) on the affected host.
- Monitor Gpbm from a fresh boot, including a remote session + resolution change (the original trigger), to confirm the leak is gone or drastically reduced.
- Report back here with before/after.

Environment for reference: Windows 11 Pro 25H2 (26200.9168), Intel iGPU 32.0.101.8864 + NVIDIA RTX 5070 Laptop 32.0.16.1088, RustDesk 1.4.9+67.

Let me know if the `new encoder` line or any of this matches what you'd expect for the fix, and I'll verify on the test build.

21pages:
Thanks for the detailed feedback! Since the host is using the VRAM H.264 encoder, it should bypass the staging-texture path, so this fix probably won’t address the leak you’re seeing.

Still, it would be great if you could do a quick test with the fix build just to confirm. I’ll also keep trying to reproduce the issue on my side using the QSV H.264 VRAM encoder.

21pages:
@BD7PIL  I’ve attached `monitor_gpbm.zip`, which contains the script I used to monitor Gpbm on the controlled host. I tested with the same QSV H.264 VRAM encoder setup and used another script to switch resolutions every 10 seconds, but I didn’t see any abnormal growth.

[monitor_gpbm.zip](https://github.com/user-attachments/files/31693935/monitor_gpbm.zip)


Could you take a look and check whether my script is monitoring the same data and fields as yours, or if I’m missing something?

rustdesk:
Close since no response

PR #15974, opened 2026-08-26
feat(mobile): add multi-display side-by-side view support (review always failed)

lapunik:
### Summary
Adds support for viewing multiple remote displays side-by-side on mobile devices (Android/iOS), providing parity with the desktop client UX.
### Key Changes
1. **Rust Core (`src/flutter.rs`)**:
   - Enabled multi-display soft renderer frame dispatching on mobile platforms so frames from all displays are routed to the Flutter UI instead of being dropped.
2. **Flutter UI (`flutter/lib/mobile/pages/remote_page.dart` & `view_camera_page.dart`)**:
   - Added the "All Displays" button (`Icons.grid_view_rounded` / `kAllDisplayValue`) to the mobile display selector menu.
   - Implemented `MultiDisplaysImagePainter` to render all remote monitors side-by-side mapped to their desktop bounding rectangle coordinates.
   - Refreshes video streams upon switching to ensure all monitor keyframes arrive promptly.
3. **State & Memory Management (`flutter/lib/models/model.dart`)**:
   - Extended `ImageModel` with a multi-display image map (`_images`) and deferred GPU texture disposal (`WidgetsBinding.instance.addPostFrameCallback`) to prevent Flutter assertion crashes during multi-frame rendering.
   - Updated `getDisplayRect` for valid index bounds.
4. **Android Client Fix (`src/client.rs`)**:
   - Gated `start_audio` for Android to avoid invoking `cpal`/`oboe` without JNI environment, preventing `SIGBUS` crash during connection initialization.
### Verification
- Tested on Android 16 (Xiaomi physical device) connecting to a multi-monitor Windows host.
- Verified smooth simultaneous rendering and live touch navigation across dual displays.

## Summary by CodeRabbit

- **New Features**
  - Added support for viewing and selecting multiple remote displays, including per-display scaling on Linux.
  - Improved physical mouse and trackpad input handling for remote control interactions.

- **Bug Fixes**
  - Improved image refreshing and cleanup when switching displays or monitors.
  - Prevented Android audio and storage setup issues on unsupported or unavailable configurations.
  - Improved multi-display session notifications across desktop and mobile.
  - Improved pointer and touch event handling, including trackpad gestures.

- **Localization**
  - Updated Danish translations across themes, connections, authentication, displays, and other interface areas.

The PR adds side-by-side rendering for multiple remote displays on mobile and extends image-state management to retain frames per display. It also refreshes streams during display switches, updates topology reconciliation, adjusts renderer dispatch, and makes platform build setup target-aware.
- Adds a dedicated multi-display canvas painter and mobile display-selector option.
- Tracks and defers disposal of per-display Flutter images.
- Routes mobile soft-renderer frames according to selected displays.
- Restricts Android dependency setup and Windows native compilation to their target platforms.
</details>

The PR is not yet safe to merge because mobile camera sessions in All Displays mode still render only one monitor.

The camera page continues to use the single-image painter, and the shared All Displays image getter selects only one retained frame; the new side-by-side painter is connected only to the regular remote page.

**Files Needing Attention:** flutter/lib/mobile/pages/view_camera_page.dart, flutter/lib/models/model.dart
</details>

| Filename | Overview |
|----------|----------|
| build.rs | Makes Android dependency installation and Windows native compilation conditional on the Cargo target operating system. |
| flutter/lib/common.dart | Preserves mobile display frames during switching and explicitly requests refreshed video streams. |
| flutter/lib/mobile/pages/remote_page.dart | Adds the mobile All Displays selector and paints retained display frames in their remote desktop coordinate layout. |
| flutter/lib/models/model.dart | Adds per-display image retention, deferred disposal, topology invalidation, and all-display scaling calculations. |
| flutter/lib/utils/image.dart | Handles the documented disposed-image race while drawing a transitioning frame. |
| src/flutter.rs | Enables soft-renderer frame dispatch to multi-display mobile sessions while retaining display-based session filtering. |

</details>

```mermaid
sequenceDiagram
    participant Core as Rust renderer
    participant Model as ImageModel
    participant Page as Mobile remote page
    participant Painter as MultiDisplaysImagePainter
    Core->>Model: Deliver frame(display, image)
    Model->>Model: Store image by display index
    Model-->>Page: notifyListeners()
    Page->>Painter: images, display topology, bounding rect
    loop Each remote display
        Painter->>Painter: Map display coordinates and draw frame
    end
```
</details>

<sub>Reviews (13): Last reviewed commit: ["feat(mobile): invalidate cached frames o..."](https://github.com/rustdesk/rustdesk/commit/790d50797500260a9bbc74e91c4f5a9890bb55b9) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56910711)</sub>


rustdesk:
I would **request changes** on PR #15974. The multi-display direction looks reasonable, but I found several merge-blocking regressions in the current head `ac92b93d`.

1. **[High] Android remote audio is completely disabled — `src/client.rs`.** The new Android `start_audio()` simply returns `Ok(())` without creating an `audio_stream`. But `handle_frame()` immediately returns whenever `audio_stream.is_none()` or `ready == false`. Therefore every received remote audio frame on Android is discarded. This avoids the reported CPAL/Oboe crash by effectively turning playback off rather than fixing initialization.
   **Suggested fix:** either initialize Android playback through a JNI-safe path/thread, or explicitly disable the audio capability at a higher level. A successful no-op `start_audio()` is especially misleading.

2. **[High] All-displays pinch zoom uses one monitor's dimensions for its limits — `flutter/lib/models/model.dart`.** `ImageModel.minScale`/`maxScale` derive their limits from `image`, while in `kAllDisplayValue` mode `image` is just `_image` or the first image in `_images`.   The actual canvas geometry correctly uses the entire composite desktop rectangle, so the two calculations disagree.   For two 1920×1080 monitors side-by-side, for example, an adaptive scale that fits the 3840px-wide desktop can be below the `minScale` calculated from a single 1920px image. The first pinch gesture can therefore snap the view in and prevent the user from zooming back out to see both monitors.
   **Suggested fix:** calculate mobile min/max scale from `ffiModel.rect` when `currentDisplay == kAllDisplayValue`, including Linux logical scaling.

3. **[Medium] Switching monitors on mobile can temporarily show the wrong monitor — `common.dart`, `model.dart`, `remote_page.dart`.** The old stale-frame cleanup is now restricted to desktop.  Meanwhile the single-monitor painter asks for the selected image but falls back to `m.image`; that getter can itself return the first cached image from another display.   After selecting display B, before B's refreshed keyframe arrives, the UI can therefore display cached content from display A. The existing comment above the desktop cleanup describes essentially this failure mode, but mobile is now exposed to it.
   **Suggested fix:** in single-display mode, never fall back to another display's cached image. Render nothing/the previous canvas intentionally until `_images[cur]` exists, or invalidate the relevant presentation state without immediately disposing images still used by a frame.

4. **[Medium] The Magic Mouse duplicate-touch suppression is partially defeated — `input_model.dart`.** `RawPointerMouseRegion` forwards pointer move/up events for every device kind.  The new code allows a non-mouse/non-trackpad move or up whenever `isPhysicalMouse` remains true.  When iOS emits the duplicate touch sequence this code is explicitly trying to suppress, the touch-down returns early and leaves `isPhysicalMouse == true`; the following touch move/up can then be processed as mouse input. Before this PR, those methods rejected anything whose kind wasn't `mouse`.
   **Suggested fix:** keep move/up processing gated by `_isMouseOrTrackpad(e.kind)`. Add a regression test for `mouse down → duplicate touch down → touch move/up → mouse up`.

5. **[Medium/cleanup] The PR contains unrelated Danish localization changes that introduce known semantic regressions.** Several unresolved review comments are still valid: `Tema` became the incorrect `Thema`, `Åbn` became `Åben`, and “must be 6 characters/digits” became “at least 6,” among others. The safest approach is to revert/split the `src/lang/da.rs` sweep from this feature PR rather than mixing translation corrections with mobile rendering, input, build, and audio changes.

There is also **no successful CI signal on the current head**: both `CI` and `Full Flutter CI` completed as `action_required` and produced no jobs, so none of these changes have a green automated run behind them.

A few older bot findings are stale: the current `build.rs` does preserve `build_manifest()` on native Windows and calls `build_windows()` based on the target OS, and the earlier camera “All Displays” selector appears to have been removed from the current camera page.

**Bottom line:** I would block merge primarily on **Android audio**, **all-display zoom bounds**, and **wrong-monitor cached-frame fallback**. The PR would also be substantially easier to validate if the unrelated touchpad/keyboard, Android audio, and Danish localization changes were split out.

rustdesk:
I’d **request changes** on the current head (`83a38a9`). I found two functional issues that are still present after the latest cleanup commit:

1. **[P2] `ImageModel` can leak stale decoded `ui.Image`s after session teardown.** In `flutter/lib/models/model.dart`, `_scheduleDispose()` uses `WidgetsBinding.instance.addPostFrameCallback`, but Flutter explicitly documents that this **does not request a new frame**. The stale-frame paths—both the `parent.target?.id != pid` branch in `decodeAndUpdate()` and `_disposeIfStale()`—return without `notifyListeners()`. If an async decode finishes after the page/session has gone away, there may be no subsequent frame, so the callback may never execute and the image remains in `_pendingDispose`.  ([[Flutter API Docs](https://api.flutter.dev/flutter/widgets/WidgetsBinding/addPostFrameCallback.html?utm_source=chatgpt.com)][1])
   **Fix:** dispose images synchronously when they have never been installed into `_image`/`_images`. Keep deferred disposal only for images that were previously exposed to a painter. Alternatively, explicitly schedule a frame, but immediate disposal is simpler and safer for these stale paths.

2. **[P2] Removing the view-only / “show my cursor” exemption regresses local cursor movement.** The base branch intentionally returned `false` from `_checkPeerControlProtected()` when `isViewOnly && showMyCursor`; this PR removes that special case.  On the current head, mouse movement in that mode still reaches `handleMouse()`, but `processEventToPeer()` now runs `_checkPeerControlProtected()` first. When the remote peer has recently moved its mouse, that returns early before `handlePointerDevicePos()` gets a chance to update the local cursor/canvas position.   The protection is explicitly based on recent peer mouse activity.
   **Fix:** restore the original `isViewOnly && showMyCursor` exemption, or separate local-cursor updates from the peer-control gate so view-only movement remains purely local.

There are also merge-readiness concerns. The two workflow runs for the current head—**CI** and **Full Flutter CI**—both finished as `action_required`, with no jobs returned, so there is currently no build/test evidence for `83a38a9`. The only combined status showing success is CodeRabbit. Also, the latest commit is one commit ahead of the last SHA reviewed by the existing bots, so their earlier “no actionable comments” result does not cover the exact current head.

Some existing bot comments are now stale: I would **not** block this head on the old camera “All Displays” finding because `view_camera_page.dart` no longer exposes that control, and the old `build.rs` warning about dropping Windows setup is also no longer accurate—the current code calls `build_windows()` based on the target OS and retains manifest generation for native Windows inline builds.  The remaining Danish localization comments are legitimate-looking but are minor and largely unrelated to the mobile multi-display feature.

Overall: **good direction, but I would not merge yet**. Fix the two P2s above and get the actual Flutter/Rust CI jobs to run on the final head.

rustdesk:
I reviewed the current head of PR **#15974** (`bde665014ef21f85a00881d5e25e5545c6ec2762`), not just the earlier bot comments. The PR is currently 8 files / +366 −120.

**Verdict: request changes**, primarily because the current diff contains a substantial unrelated localization regression. The core multi-display implementation itself looks substantially better than the earlier automated reviews suggest.

1. **[P1] Revert `src/lang/da.rs` from this PR.** This feature PR changes a large number of Danish strings, and several changes are objectively regressions. For example, `"Email verification code must be 6 characters."` goes from an exact-six translation to **“at least 6 characters”**, and `"2FA code must be 6 digits."` similarly becomes **“at least 6 digits”**. That changes the meaning of validation requirements. Other regressions include `Open: "Åbn" → "Åben"`, `Theme: "Tema" → "Thema"`, and `Show virtual joystick: "Vis virtuelt joystick" → "Vis virtuel joystick"`. None of this is necessary for mobile multi-display support. I would restore this file to `master` and submit any intended localization work separately.

2. **[P2] The PR has too much unrelated history/diff for this feature.** Besides multi-display work, the current diff still contains the one-line `input_model.dart` change, Flutter theme type migration, Windows build-script restructuring, image-painter exception handling, and the Danish localization rewrite. Some appear to be remnants of other branches that were merged and partially reverted. This makes regression analysis significantly harder. At minimum, remove `src/lang/da.rs` and the unrelated `input_model.dart` change unless they are specifically required and tested.

3. **Automated validation has not actually run on the current head.** Both `CI` and `Full Flutter CI` are `action_required`, with no jobs created. So the manual Android verification in the PR description is currently the only meaningful validation of this head. Before merging, I would require Flutter analysis/tests plus at least an Android build, preferably with a multi-monitor widget/integration test around image-map lifecycle and switching All → display → All.

I also rechecked the prominent existing bot warnings. Several are **stale or already fixed**: the Windows manifest is now called again and `build_windows()` is target-driven rather than host-gated; the Linux multi-display painter now scales individual displays and protects `save()/restore()` with `finally`; and the old `view_camera_page.dart` “All Displays renders one monitor” finding is outdated because that file was removed from the current diff and its current mobile camera selector does not expose the grid option. The camera renderer still uses the single `m.image` path, but there is no current UI path in that page to select All Displays.

On the actual feature implementation, the architecture looks reasonable: mobile soft-renderer dispatch now accepts handlers subscribed to multiple displays; `ImageModel` maintains per-display images; the remote page paints those images using the combined desktop rectangle; Linux physical-image dimensions are scaled into logical coordinates; and display switching explicitly refreshes the relevant streams. The disposal logic is somewhat defensive—especially the broad `drawImage` catches—but I don't see a clear merge-blocking race in the current version.

I would therefore leave a **REQUEST_CHANGES** review centered on removing the Danish regression/unrelated changes, then re-review the cleaned diff.

rustdesk:
I reviewed the current PR head, `1c85ec4`, against the current `master`. My verdict is **Request changes**. I found one clear merge-blocking issue; the earlier multi-display zoom/stale-image/localization problems have been fixed or removed from the latest head. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/15974)][1])

### [P1] `common.dart` now requires Flutter 3.27+, breaking RustDesk’s pinned Flutter 3.24.5 builds

The PR changes these four declarations:

* `flutter/lib/common.dart:354` → `DialogThemeData`
* `flutter/lib/common.dart:385` → `TabBarThemeData`
* `flutter/lib/common.dart:452` → `DialogThemeData`
* `flutter/lib/common.dart:486` → `TabBarThemeData` 

Those types were introduced in **Flutter 3.27**. ([[flutter.dev](https://flutter.dev/blog/whats-new-in-flutter-3-27?utm_source=chatgpt.com)][2])

RustDesk, however, explicitly pins the normal build matrix to **Flutter 3.24.5**. The workflow even documents that every platform except Windows ARM64 should keep committed sources compatible with 3.24.5. ([[GitHub](https://github.com/rustdesk/rustdesk/blob/master/.github/workflows/flutter-build.yml)][3])

So this PR will make the normal Android/iOS/macOS/Linux/web/Windows-x64 Flutter builds fail at compile time because `DialogThemeData` and `TabBarThemeData` do not exist in Flutter 3.24.5.

It also breaks the Windows ARM64/Flutter 3.44 compatibility path. RustDesk's patch script explicitly says the new theme names must stay **out of committed sources** until the repo-wide Flutter bump. Its validation accepts either the fully old state or the fully patched 3.44 state; this PR creates a partial state, causing the script to exit with `"Flutter 3.44 source patches are partially applied..."`. 

**Fix:** revert these four theme-name changes to `DialogTheme` and `TabBarTheme`. Let `apply_flutter_3.44_source_patches.sh` perform that migration for the ARM64 build. A repo-wide Flutter upgrade should be a separate change.

The inline review comment I would leave is:

> **[P1] Keep the committed theme API compatible with Flutter 3.24.5.**
> RustDesk's build matrix still pins normal platforms to Flutter 3.24.5, while `DialogThemeData`/`TabBarThemeData` were introduced in 3.27. Committing these names therefore breaks the normal Flutter builds. It also leaves the Windows ARM64 3.44 patch script in its deliberately rejected “partially applied” state. Please restore `DialogTheme`/`TabBarTheme` here; the existing 3.44 compatibility script already performs this migration for the one target that needs it.

One additional merge-readiness note: both `CI` and `Full Flutter CI` for `1c85ec4` are currently **Action required**, awaiting maintainer approval, so none of the pending build jobs have actually run yet. ([[GitHub](https://github.com/rustdesk/rustdesk/actions/runs/33057927024)][4])

Aside from that blocker, I don't see another clear current-head functional regression in the core side-by-side implementation that I would confidently file: the per-display image map, soft-render dispatch on mobile, composite zoom bounds, stale-display fallback, Linux per-display scaling, and disposal changes now look internally consistent.

[1]: https://github.com/rustdesk/rustdesk/pull/15974 "feat(mobile): add multi-display side-by-side view support by lapunik · Pull Request #15974 · rustdesk/rustdesk · GitHub"
[2]: https://flutter.dev/blog/whats-new-in-flutter-3-27?utm_source=chatgpt.com "What’s new in Flutter 3.27 | The Flutter Blog"
[3]: https://github.com/rustdesk/rustdesk/blob/master/.github/workflows/flutter-build.yml "rustdesk/.github/workflows/flutter-build.yml at master · rustdesk/rustdesk · GitHub"
[4]: https://github.com/rustdesk/rustdesk/actions/runs/33057927024 "feat(mobile): add multi-display side-by-side view support · rustdesk/rustdesk@1c85ec4 · GitHub"

rustdesk:
I re-reviewed the PR at the latest head, **`aa7eced`**. The Flutter 3.24.5 compatibility **P1 from my previous review is fixed**: the four declarations have been reverted back to `DialogTheme` / `TabBarTheme`.

I still have **one concrete issue that I recommend fixing before merge**.

### [P2] Display/capture state is not reconfigured when the remote display topology changes while in “All Displays” mode

When `handleSyncPeerInfo()` receives an updated display list, it replaces `_pi.displays`. However, if the current display selection is `kAllDisplayValue`, it only calls `updateCurDisplay()` and leaves this comment:

```dart
// to-do: What if the displays are changed?
```

It does not call `sessionSwitchDisplay()` again.

This is more than a temporary stale-frame issue. On the Rust side, `SessionHandler.displays` is only updated by `session_switch_display()`, and that is also where `capture_displays()` is reconfigured.

The mobile soft renderer then filters incoming frames using the old display list:

```rust
if is_multi_sessions || h.displays.len() > 1 {
    if !h.displays.contains(&display) {
        continue;
    }
}
```

A reproducible case is:

1. The remote machine initially has displays 0 and 1.
2. The mobile client selects **All Displays**, so `h.displays == [0, 1]`.
3. A third monitor is connected on the remote side.
4. `SyncPeerInfo` updates Dart's `pi.displays` to `[0, 1, 2]`.
5. Rust's `h.displays` is still `[0, 1]`.
6. RGBA frames for display 2 are rejected by `contains(&display)`.
7. The UI now allocates space for display 2, but it remains blank until the user leaves All Displays mode and enters it again.

Calling `sessionRefreshVideo()` alone is not sufficient. It may request refreshes for the new display indexes, but it does **not** update `h.displays` or `capture_displays()`.

There is also a related cache issue. `ImageModel._images` is indexed by display index, while `MultiDisplaysImagePainter` pairs `images[i]` with `displays[i]`. If displays are removed or reordered, the old image cache is not invalidated. For example, after `[A, B] -> [B]`, the old `_images[0] == A` could temporarily be painted as the new display 0/B.

I would leave the following review comment in the `handleSyncPeerInfo()` All Displays branch:

> **[P2] Reconfigure the captured displays when the topology changes in All Displays mode.**
> `pi.displays` is replaced here, but the mobile session's `SessionHandler.displays` and `capture_displays` still contain the indices that were selected when All Displays was entered. For example, after going from displays `[0, 1]` to `[0, 1, 2]`, RGBA events for display 2 are filtered out by `on_rgba_soft_render()` because `h.displays` remains `[0, 1]`, so the newly plugged display stays blank until the user leaves and re-enters All Displays.
> Please re-run `sessionSwitchDisplay()` with `List.generate(newDisplays.length, ...)` when the topology changes. The per-display `_images` cache should also be invalidated or reconciled so indices from the old topology cannot be painted against the new `Display` list.

### Suggested fix

Before replacing the display list, detect whether the topology has actually changed. If the current mode is All Displays and it has changed, reconfigure the selected displays:

```dart
final newDisplayIndexes =
    List.generate(_pi.displays.length, (index) => index);

bind.sessionSwitchDisplay(
  isDesktop: isDesktop,
  sessionId: sessionId,
  value: Int32List.fromList(newDisplayIndexes),
);
```

The old `_images` cache should also be cleared or reconciled.

I would prefer adding a safe cache-invalidation path that reuses the existing `_scheduleDispose()` behavior rather than calling `clearImage()` and synchronously disposing images. The new post-frame disposal mechanism appears intentional and helps avoid a paint/dispose race.

Other than this, I did not find another P1 or build-blocking issue in the latest head. The previously reviewed areas—Flutter 3.24.5 compatibility, switching from All Displays back to a single display, per-display images, zoom bounds, and mobile soft-render filtering—now look largely consistent.

**Current verdict: Request changes — one P2 remains for dynamic display-topology changes.** Once that is fixed, I would be inclined to approve.

rustdesk:
I re-reviewed the PR at the current head, **`47a8127`**. The previous issue with reconfiguring the Rust capture list has been fixed: when topology changes in **All Displays** mode, the code now calls `sessionSwitchDisplay()` with the new display indexes. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/15974/commits)][1])

I still see **one P2 issue** in the new fix.

### [P2] `reconcileDisplays()` preserves frames for the wrong monitor when display indexes shift

The new code does:

```dart
_pi.displays.value = newDisplays;

if (_pi.currentDisplay == kAllDisplayValue) {
  parent.target?.imageModel.reconcileDisplays(newDisplays.length);
  ...
}
```

But `reconcileDisplays()` only removes image entries whose index is outside the new display count:

```dart
void reconcileDisplays(int newDisplayCount) {
  final invalidKeys =
      _images.keys.where((k) => k >= newDisplayCount).toList();

  for (final k in invalidKeys) {
    final img = _images.remove(k);
    if (img != null) {
      _scheduleDispose(img);
    }
  }
}
```



That assumes an existing display index still refers to the same physical monitor after a topology change. It doesn't necessarily.

For example:

```text
Before:
index 0 -> Monitor A -> image A
index 1 -> Monitor B -> image B

Monitor A is unplugged.

After:
index 0 -> Monitor B
```

`reconcileDisplays(1)` removes `_images[1]`, but **keeps `_images[0]`**, which still contains Monitor A's last frame.

The painter then does exactly this:

```dart
final img = images[i];
final display = displays[i];
```

so the stale frame from Monitor A is drawn using Monitor B's new geometry until a fresh frame for display 0 arrives. 

The same problem is even clearer for a same-count topology change:

```text
[A, B] -> [B, C]
```

`newDisplayCount` stays `2`, so `reconcileDisplays(2)` removes **nothing**. Initially:

```text
images[0] = old A, displays[0] = new B
images[1] = old B, displays[1] = new C
```

Both displays can therefore briefly show the wrong monitor.

There is a second route to the same stale-cache problem: reconciliation currently happens **only when `currentDisplay == kAllDisplayValue`**. If monitors change while the user is viewing one display, old `_images` entries remain. Later, entering All Displays on mobile does not clear them—the existing `clearImage()` in `openMonitorInTheSameTab()` is desktop-only. 

#### Review comment I'd leave

> **[P2] Invalidate cached frames by topology, not just by display count.**
>
> `reconcileDisplays(newDisplays.length)` only removes indexes that are now out of range, but display indexes can be reassigned when a monitor is removed/reordered. For example, `[A, B] -> [B]` keeps `_images[0]` (A), after which `MultiDisplaysImagePainter` pairs that stale image with the new `displays[0]` (B) until a fresh frame arrives. A same-count change such as `[A, B] -> [B, C]` removes nothing at all.
>
> The cache can also remain stale if topology changes while a single display is selected, since reconciliation is only performed in the All Displays branch.
>
> Since `Display` has no stable monitor identity, I think the safest approach is to invalidate all per-display cached images whenever the display topology changes, using the existing deferred `_scheduleDispose()` mechanism. The new `sessionSwitchDisplay()` logic can remain conditional on currently being in All Displays mode.

### Suggested implementation

Something along these lines would be safer:

```dart
final oldDisplays = List<Display>.of(_pi.displays);
final topologyChanged = !listEquals(oldDisplays, newDisplays);

if (topologyChanged) {
  parent.target?.imageModel.invalidateDisplayImages();
}

_pi.displays.value = newDisplays;
_pi.displaysCount.value = newDisplays.length;

if (_pi.currentDisplay == kAllDisplayValue) {
  bind.sessionSwitchDisplay(
    isDesktop: isDesktop,
    sessionId: sessionId,
    value: Int32List.fromList(
      List.generate(newDisplays.length, (index) => index),
    ),
  );
  sessionRefreshVideo(sessionId, _pi);
  updateCurDisplay(sessionId);
}
```

`invalidateDisplayImages()` should remove the `_images` entries and pass their images through `_scheduleDispose()` rather than synchronously disposing them, preserving the paint/dispose protection already added in this PR. The `Display` class already implements value equality based on its geometry/size/cursor properties, so `listEquals()` can be used for the topology comparison. 

**Current verdict: Request changes — 1 P2 remains.**

The earlier Flutter 3.24.5 blocker is fixed, and the new `sessionSwitchDisplay()` handling fixes the capture-side part of my previous topology review. I did not find another P1/build blocker in the current `47a8127` head.

[1]: https://github.com/rustdesk/rustdesk/pull/15974/commits "feat(mobile): add multi-display side-by-side view support by lapunik · Pull Request #15974 · rustdesk/rustdesk · GitHub"

rustdesk:
You can use chatgpt chat to review this PR. No need to wait for me to run it.

prompt: `review https://github.com/rustdesk/rustdesk/pull/15974`

Once it get approved by chatgpt chat, I will run the other AI review.

lapunik:
@rustdesk All previously reported issues have been addressed in `790d507`:
1. Reconfiguration of captured displays on topology changes via `sessionSwitchDisplay()` and `sessionRefreshVideo()`.
2. Safe topology-aware cache invalidation using `invalidateDisplayImages()` and `listEquals`.
3. Deferred frame disposal in `clearImage()` via `_scheduleDispose()` to prevent canvas race conditions.
4. Restored `DialogTheme` and `TabBarTheme` for Flutter 3.24.5 build matrix compatibility.
5. All unrelated diffs (Danish localization, input model, Gradle files) have been stripped out.

Both CodeRabbit and Greptile reviews have passed cleanly on the latest head. Could you please run the final review/workflows and update the title? Thank you!

rustdesk:
Could you paste a video of the new interface?

lapunik:
https://github.com/user-attachments/assets/34ffddea-37fb-4555-84de-835f7dce5d05

lapunik:
Hi @rustdesk, the screen recording demonstrating the new dual-monitor side-by-side interface is attached above. All reviews (CodeRabbit & Greptile) are passing and there are no merge conflicts. Could you please review and trigger the CI workflows when you have a moment? Thank you!

PR #15975, opened 2026-08-26
fix(mobile): support hardware touchpad and physical keyboard input on…

lapunik:
### Summary of Changes

1. **Android Hardware Touchpad Support**:
   - On Android devices with external/integrated touchpads (e.g. keyboard trackpad combos), moving a finger across the trackpad generates `PointerHoverEvent` with `kind == PointerDeviceKind.touch`.
   - Updated `onPointHoverImage` in `InputModel` to accept hover events on mobile devices and activate `isPhysicalMouse` so touchpad movement moves the remote cursor smoothly.
   - Updated `onPointMoveImage` and `onPointUpImage` to accept events when physical mouse mode is active or when the device kind is trackpad.
   - Fixed `isNotTouchBasedDevice()` in `RawTouchGestureDetectorRegion` so it doesn't discard touch drag gestures when `isPhysicalMouse` is not active.

2. **Physical Keyboard Focus**:
   - Added automatic focus request for `_physicalFocusNode` on pointer down in `RemotePage` to ensure external hardware keyboards immediately receive key events without needing manual soft-keyboard activation.

3. **Android Storage Directory Null Safety**:
   - Handled potential null / empty results from `ExternalPath.getExternalStorageDirectories()` in `native_model.dart`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard and focus behavior when dismissing the soft keyboard or interacting with chat overlays.
  * Prevented duplicate touch events after mouse clicks from causing unintended interactions.
  * Improved handling of stylus, mouse, touch, and trackpad input.
  * Prevented trackpad gestures from being misinterpreted as touch gestures.
  * Improved Android handling when external storage locations are unavailable.

<h2><a href="https://app.greptile.com/api/retrigger?id=56917343"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge.

This revision improves mobile hardware-touchpad classification, preserves touchscreen gesture handling, and tightens keyboard-focus transitions.
- Adds touch-only gesture recognizers and pointer lifecycle tracking.
- Keeps the touch gesture region mounted while filtering physical pointer events.
- Adds configurable hardware-touchpad support and related gesture-help UI.
- Adds guards around mobile keyboard and chat focus transitions.
- Adds pointer-classification regression tests and localization entries.
</details>

<sub>Reviews (15) · Last reviewed commit: ["Merge branch &#39;master&#39; into fix/mobile-to..."](https://github.com/rustdesk/rustdesk/commit/e5c2801f9a7569cc5959dbe3b3bab0f31d230c5c)</sub>

### DIFF
--- a/flutter/lib/common/widgets/gestures.dart
+++ b/flutter/lib/common/widgets/gestures.dart
@@ -21,6 +21,11 @@ class CustomTouchGestureRecognizer extends ScaleGestureRecognizer {
     _init();
   }
 
+  @override
+  void addAllowedPointerPanZoom(PointerPanZoomStartEvent event) {
+    // Ignore trackpad pan/zoom so RawPointerMouseRegion handles trackpad scrolling/zoom.
+  }
+
   // oneFingerPan
   GestureDragStartCallback? onOneFingerPanStart;
   GestureDragUpdateCallback? onOneFingerPanUpdate;
--- a/flutter/lib/common/widgets/gestures.dart
+++ b/flutter/lib/common/widgets/gestures.dart
@@ -10,15 +10,89 @@ enum GestureState {
   threeFingerVerticalDrag
 }
 
+class TouchTapGestureRecognizer extends TapGestureRecognizer {
+  final bool Function(int device)? isPhysicalPointerDevice;
+  TouchTapGestureRecognizer({
+    Object? debugOwner,
+    Set<PointerDeviceKind>? supportedDevices,
+    this.isPhysicalPointerDevice,
+  }) : super(
+          debugOwner: debugOwner,
+          supportedDevices: supportedDevices,
+        );
+
+  @override
+  bool isPointerAllowed(PointerDownEvent event) {
+    if (isPhysicalPointerDevice != null &&
+        isPhysicalPointerDevice!(event.device)) {
+      return false;
+    }
+    return super.isPointerAllowed(event);
+  }
+}
+
+class TouchDoubleTapGestureRecognizer extends DoubleTapGestureRecognizer {
+  final bool Function(int device)? isPhysicalPointerDevice;
+  TouchDoubleTapGestureRecognizer({
+    Object? debugOwner,
+    Set<PointerDeviceKind>? supportedDevices,
+    this.isPhysicalPointerDevice,
+  }) : super(
+          debugOwner: debugOwner,
+          supportedDevices: supportedDevices,
+        );
+
+  @override
+  bool isPointerAllowed(PointerDownEvent event) {
+    if (isPhysicalPointerDevice != null &&
+        isPhysicalPointerDevice!(event.device)) {
+      return false;
+    }
+    return super.isPointerAllowed(event);
+  }
+}
+
+class TouchLongPressGestureRecognizer extends LongPressGestureRecognizer {
+  final bool Function(int device)? isPhysicalPointerDevice;
+  TouchLongPressGestureRecognizer({
+    Object? debugOwner,
+    Set<PointerDeviceKind>? supportedDevices,
+    this.isPhysicalPointerDevice,
+  }) : super(
+          debugOwner: debugOwner,
+          supportedDevices: supportedDevices,
+        );
+
+  @override
+  bool isPointerAllowed(PointerDownEvent event) {
+    if (isPhysicalPointerDevice != null &&
+        isPhysicalPointerDevice!(event.device)) {
+      return false;
+    }
+    return super.isPointerAllowed(event);
+  }
+}
+
 class CustomTouchGestureRecognizer extends ScaleGestureRecognizer {
+  final bool Function(int device)? isPhysicalPointerDevice;
   CustomTouchGestureRecognizer({
     Object? debugOwner,
     Set<PointerDeviceKind>? supportedDevices,
+    this.isPhysicalPointerDevice,
   }) : super(
           debugOwner: debugOwner,
           supportedDevices: supportedDevices,
         ) {
     _init();
+  }
+
+  @override
+  bool isPointerAllowed(PointerDownEvent event) {
+    if (isPhysicalPointerDevice != null &&
+        isPhysicalPointerDevice!(event.device)) {
+      return false;
+    }
+    return super.isPointerAllowed(event);
   }
 
   @override
@@ -199,9 +273,11 @@ class CustomTouchGestureRecognizer extends ScaleGestureRecognizer {
 }
 
 class HoldTapMoveGestureRecognizer extends GestureRecognizer {
+  final bool Function(int device)? isPhysicalPointerDevice;
   HoldTapMoveGestureRecognizer({
     Object? debugOwner,
     Set<PointerDeviceKind>? supportedDevices,
+    this.isPhysicalPointerDevice,
   }) : super(
           debugOwner: debugOwner,
           supportedDevices: supportedDevices,
@@ -226,6 +302,10 @@ class HoldTapMoveGestureRecognizer extends GestureRecognizer {
 
   @override
   bool isPointerAllowed(PointerDownEvent event) {
+    if (isPhysicalPointerDevice != null &&
+        isPhysicalPointerDevice!(event.device)) {
+      return false;
+    }
     if (_firstTap == null) {
       switch (event.buttons) {
         case kPrimaryButton:
@@ -467,9 +547,11 @@ class HoldTapMoveGestureRecognizer extends GestureRecognizer {
 }
 
 class DoubleFinerTapGestureRecognizer extends GestureRecognizer {
+  final bool Function(int device)? isPhysicalPointerDevice;
   DoubleFinerTapGestureRecognizer({
     Object? debugOwner,
     Set<PointerDeviceKind>? supportedDevices,
+    this.isPhysicalPointerDevice,
   }) : super(
           debugOwner: debugOwner,
           supportedDevices: supportedDevices,
@@ -492,6 +574,10 @@ class DoubleFinerTapGestureRecognizer extends GestureRecognizer {
 
   @override
   bool isPointerAllowed(PointerDownEvent event) {
+    if (isPhysicalPointerDevice != null &&
+        isPhysicalPointerDevice!(event.device)) {
+      return false;
+    }
     if (_firstTap == null) {
       switch (event.buttons) {
         case kPrimaryButton:
--- a/flutter/lib/common/widgets/remote_input.dart
+++ b/flutter/lib/common/widgets/remote_input.dart
@@ -622,6 +622,7 @@ class RawPointerMouseRegion extends StatelessWidget {
         onPointerUp?.call(evt);
         inputModel.onPointUpImage(evt);
       },
+      onPointerCancel: inputModel.onPointCancelImage,
       onPointerMove: inputModel.onPointMoveImage,
       onPointerSignal: inputModel.onPointerSignalImage,
       onPointerPanZoomStart: inputModel.onPointerPanZoomStart,
--- a/flutter/lib/common/widgets/remote_input.dart
+++ b/flutter/lib/common/widgets/remote_input.dart
@@ -126,8 +126,7 @@ class _RawTouchGestureDetectorRegionState
   }
 
   bool isNotTouchBasedDevice() {
-    return !kTouchBasedDeviceKinds.contains(lastDeviceKind) &&
-        inputModel.isPhysicalMouse.value;
+    return !kTouchBasedDeviceKinds.contains(lastDeviceKind);
   }
 
   // Mobile, mouse mode.
--- a/flutter/lib/common/widgets/remote_input.dart
+++ b/flutter/lib/common/widgets/remote_input.dart
@@ -531,26 +531,30 @@ class _RawTouchGestureDetectorRegionState
         };
 
   makeGestures(BuildContext context) {
+    final isPhysical = inputModel.isPhysicalPointerDeviceId;
     return <Type, GestureRecognizerFactory>{
       // Official
-      TapGestureRecognizer:
-          GestureRecognizerFactoryWithHandlers<TapGestureRecognizer>(
-              () => TapGestureRecognizer(), (instance) {
+      TouchTapGestureRecognizer:
+          GestureRecognizerFactoryWithHandlers<TouchTapGestureRecognizer>(
+              () => TouchTapGestureRecognizer(
+                  isPhysicalPointerDevice: isPhysical), (instance) {
         instance
           ..onTapDown = onTapDown
           ..onTapUp = onTapUp
           ..onTap = onTap;
       }),
-      DoubleTapGestureRecognizer:
-          GestureRecognizerFactoryWithHandlers<DoubleTapGestureRecognizer>(
-              () => DoubleTapGestureRecognizer(), (instance) {
+      TouchDoubleTapGestureRecognizer:
+          GestureRecognizerFactoryWithHandlers<TouchDoubleTapGestureRecognizer>(
+              () => TouchDoubleTapGestureRecognizer(
+                  isPhysicalPointerDevice: isPhysical), (instance) {
         instance
           ..onDoubleTapDown = onDoubleTapDown
           ..onDoubleTap = onDoubleTap;
       }),
-      LongPressGestureRecognizer:
-          GestureRecognizerFactoryWithHandlers<LongPressGestureRecognizer>(
-              () => LongPressGestureRecognizer(), (instance) {
+      TouchLongPressGestureRecognizer:
+          GestureRecognizerFactoryWithHandlers<TouchLongPressGestureRecognizer>(
+              () => TouchLongPressGestureRecognizer(
+                  isPhysicalPointerDevice: isPhysical), (instance) {
         instance
           ..onLongPressDown = onLongPressDown
           ..onLongPressUp = onLongPressUp
@@ -560,7 +564,8 @@ class _RawTouchGestureDetectorRegionState
       // Customized
       HoldTapMoveGestureRecognizer:
           GestureRecognizerFactoryWithHandlers<HoldTapMoveGestureRecognizer>(
-              () => HoldTapMoveGestureRecognizer(),
+              () => HoldTapMoveGestureRecognizer(
+                  isPhysicalPointerDevice: isPhysical),
               (instance) => instance
                 ..onHoldDragStart = onHoldDragStart
                 ..onHoldDragUpdate = onHoldDragUpdate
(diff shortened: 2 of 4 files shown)


rustdesk:
I reviewed the current head `d6fc2b9`, including the full diff, surrounding code, prior review comments, and CI status. **My recommendation is: do not merge yet; request changes.** I see three substantive input/focus bugs. The first two are directly tied to the PR’s pointer-state design.

1. **[P1] The first touchscreen gesture after switching from touchpad/mouse back to finger can still be lost.** `RemotePage` completely removes `RawTouchGestureDetectorRegion` from the widget tree whenever `isPhysicalMouse == true`. The new Android touchpad hover path sets `isPhysicalMouse` to true.

   When the user then touches the actual touchscreen, `onPointDownImage()` clears `isPhysicalMouse` and triggers a rebuild, which re-adds the gesture detector. But by then the original `PointerDownEvent` has already happened. The newly mounted recognizer never participated in that pointer sequence, so the first tap/drag/pinch can fail.

   This means the change to `isNotTouchBasedDevice()` does not actually solve the transition problem, because the recognizer is not even mounted when the first touch-down arrives.

   I would strongly prefer keeping `RawTouchGestureDetectorRegion` mounted at all times and filtering events by device/state, instead of dynamically removing the entire recognizer based on `isPhysicalMouse`.

2. **[P1] The new `isPhysicalMouse` fallback can break the existing iOS Magic Mouse duplicate-touch protection.** The existing duplicate-touch workaround intentionally returns early from `onPointDownImage()` when `_shouldIgnoreTouchAfterMouse()` matches. In that case, `isPhysicalMouse` remains true.

   But this PR changes move/up routing to effectively mean: “if this is not mouse/trackpad, still process it as long as `isPhysicalMouse == true`.” That allows the later `PointerMoveEvent` / `PointerUpEvent` from the duplicate touch sequence to re-enter the mouse path even though the touch-down was explicitly ignored.

   That is particularly risky because `_getMouseEvent()` relies on `_lastButtons` when producing an up event. A duplicate touch-up could therefore release a remote mouse button prematurely during a Magic Mouse drag.

   The underlying issue is that **device routing should not be decided by a single global `isPhysicalMouse` boolean**. Ideally, routing should be stable per pointer/device sequence. At minimum, if a touch-down is classified as duplicate/ignored, the corresponding move/up events from that same pointer must also be ignored.

3. **[P1/P2] Hiding the soft keyboard can steal focus from the chat overlay.** `onSoftKeyboardChanged(false)` previously guarded remote soft-keyboard teardown with `chatWindowOverlayEntry == null`, presumably so the chat overlay would not be disrupted. But the latest commit adds `_mobileFocusNode.unfocus()` and `_physicalFocusNode.requestFocus()` outside that guard.

   The chat overlay owns its own `inputNode` and explicitly requests focus when opened.  If the user opens chat and then hides the Android IME with Back, the new unconditional focus transfer can move focus away from chat and back to the remote `RawKeyFocusScope`. A hardware keyboard could then start sending keystrokes to the remote machine instead of continuing to type in chat.

   I would gate that physical-focus restoration on there being no active chat/text-input owner, rather than doing it unconditionally whenever the IME becomes hidden.

I also would not rely on the bot summary saying the PR is safe to merge. On the current head, both **`CI` and `Full Flutter CI` are `action_required`**, so there is no green validation for the latest commit.

My merge verdict is therefore **Request changes**.

I would require at least these regression tests:

* `touchpad hover -> first touchscreen drag`
* `Magic Mouse down -> duplicate touch sequence -> real mouse up`
* `chat overlay + hide IME -> physical keyboard typing`

The main architectural fix should be to keep the touch recognizer mounted and make pointer routing stable per pointer/device sequence, instead of continuing to layer more behavior onto the global `isPhysicalMouse` flag.

rustdesk:
## Findings

**[P1 / blocking] Mouse-up can still be dropped after switching to touch input — `flutter/lib/models/input_model.dart:1545–1576`.**

The latest commit fixes the recognizer-removal problem by keeping `RawTouchGestureDetectorRegion` mounted, but the physical-pointer state is still coupled to `isPhysicalMouse`. A touchscreen down clears that flag, while `onPointUpImage()` only sends the physical mouse/trackpad release when the flag remains true.

A concrete sequence is:

1. Physical mouse down → `isPhysicalMouse = true`; `_getMouseEvent(...mousedown)` records the pressed button.
2. While still holding the mouse button, touch the screen → `onPointDownImage()` sets `isPhysicalMouse = false`.
3. Release the physical mouse without moving it first.
4. The mouse-up passes the device-kind check, but `if (isPhysicalMouse.value)` is false, so the release is never sent.

`_getMouseEvent()` is also where `_lastButtons` would normally be reset on the mouse-up, so skipping it can leave the remote side with a stuck pressed button/drag.

I would change the release path so a mouse/trackpad `PointerUpEvent` is not gated by the mutable global `isPhysicalMouse` state. Better still, track an active physical pointer/button sequence independently of the UI's current input mode. Add a regression test specifically for:

`mouse-down → touch-down → mouse-up`, with **no intervening mouse move**.

This is especially relevant to this PR because supporting simultaneous touchscreen + hardware-pointer devices is exactly the scenario being changed.

### Non-blocking cleanup

`_ignoredTouchPointers` is cleared only by `PointerUpEvent`, but `RawPointerMouseRegion` doesn't forward `PointerCancelEvent`.

A canceled suppressed iOS touch can therefore leave its pointer ID in the set. Flutter pointer identifiers aren't normally reused, so I don't see this causing future touches to be incorrectly suppressed; it's mainly stale-state/set growth. Wiring `onPointerCancel` to remove the pointer would make the bookkeeping complete.

## Other changes

The move to keep `RawTouchGestureDetectorRegion` permanently mounted looks correct and fixes the earlier gesture-recognizer lifetime problem.

I also **wouldn't flag the mobile `PointerHoverEvent(kind: touch)` handling** just because it isn't `mouse`/`trackpad`. Flutter's indirect-input behavior can surface trackpad activity through touch/mouse-style pointer events depending on platform support, and that fallback is the Android behavior this PR is explicitly addressing. The explicit stylus exclusion makes sense.

The soft-keyboard dismissal path now transfers focus from `_mobileFocusNode` to `_physicalFocusNode`, which addresses the previously reported case where hardware keys stopped reaching `RawKeyFocusScope`.

The Android external-storage null/empty check is straightforward and looks safe.

## Validation status

There are no test files in the four changed files, and the two workflows associated with current head `fcf92590` both completed as **`action_required` with zero jobs**, so this version has not actually been exercised by CI. I would want the mixed mouse/touch regression above covered before merge.

**Verdict: request changes.** I see one blocking correctness issue; otherwise the latest iteration is substantially better than the earlier revisions and the remaining changes look reasonable.

rustdesk:
## Verdict: **Request changes**

I would not merge this yet.

The current head is `93e051d`. The earlier issues around chat focus stealing, `mouse-down → touch-down → mouse-up` leaving the remote button stuck, and unstable ignored-touch sequences appear to have been addressed in the latest commits.

There is still **one blocker**, though, and it may directly break the main Android touchpad behavior this PR is trying to add.

1. **[P1 / blocker] The strict `PointerDeviceKind` gating introduced in `fcf925` may break Android touchpad click/drag.** The PR's original premise is that some Android hardware touchpads emit cursor movement as `PointerHoverEvent(kind == touch)`. The initial implementation therefore intentionally allowed later `PointerMoveEvent` / `PointerUpEvent` events to continue down the physical-pointer path once `isPhysicalMouse` had already been activated, even if the event kind itself was not `mouse` or `trackpad`. The original condition was effectively:

`if (!_isMouseOrTrackpad(e.kind) && !isPhysicalMouse.value) return;`

`fcf925` changed that to strict gating:

`if (!_isMouseOrTrackpad(e.kind)) return;`

for move/up, and down events only establish a physical sequence when the kind is `mouse` or `trackpad`. The latest `93e051d` keeps that model and only adds `_activeMousePointers`.

That is risky because Flutter does not guarantee that a physical trackpad click/drag sequence will use `PointerDeviceKind.trackpad`. On some platforms, trackpad interaction may surface as `touch` or `mouse` pointer events instead. In exactly the Android case this PR is targeting, the likely sequence becomes:

`hover(touch)` → RustDesk sets `isPhysicalMouse = true` → user presses the touchpad → `down(touch)` enters the non-mouse branch and may clear physical-mouse state → `move(touch)` is rejected → `up(touch)` is rejected.

So cursor movement may work, but click/drag can still fail, and the always-mounted touch gesture recognizer may interpret that same sequence as touchscreen input.

The fix should not be to revert to the old global `isPhysicalMouse` behavior, because that global flag caused the earlier cross-pointer bugs. The better design is to track **physical-pointer ownership per device/pointer sequence**. For example, once a mobile hover event identifies a device as the external touchpad, remember that device ID. A subsequent `PointerDownEvent` from that same device can create an active physical pointer entry keyed by `e.pointer`. Move/up/cancel should then use that stable pointer ownership rather than relying only on `PointerDeviceKind` or the global boolean.

There is also a smaller, non-blocking asymmetry in the new cancel path. `onPointCancelImage()` currently sends release through absolute `handleMouse(...)`, while the normal `onPointUpImage()` uses `_relativeMouse.sendRelativeMouseButton(...)` when relative mouse mode is enabled.   `sendRelativeMouseButton()` is specifically the no-position button path for relative mode.  I would make cancel mirror normal up behavior. I would classify that as **P2/P3**, not the main blocker.

The latest head also has not actually gone through normal CI validation yet. Both `CI` and `Full Flutter CI` are `action_required`, and the CI run currently shows zero jobs rather than a real executed test run.

Before merging, I would want explicit regression coverage or device testing for:

* Android touchpad `hover(kind=touch) → click`
* Android touchpad `hover(touch) → down/move/up(touch)` drag
* touchscreen touch interleaved with touchpad input
* `mouse down → touchscreen down → mouse up`
* ignored touch sequence followed by cancel
* relative-mouse pointer cancel

The first two are the most important because they are the exact behavior this PR claims to fix.

**Final maintainer decision: Request changes.**

The overall direction is much better than the earlier revisions. The remaining architectural issue is that "physical pointer" should be modeled as stable **device/pointer sequence ownership**, not inferred from `PointerDeviceKind` or a single global `isPhysicalMouse` flag. Once that is fixed and tested on the target Android touchpad hardware, I would be much more comfortable approving it.

rustdesk:
## Verdict: **Request changes**

I reviewed the current head, `039f919`, rather than the earlier revisions. The latest commit fixes the previously reported raw-pointer routing problem by remembering physical pointer device IDs, but I see one remaining blocker in how that routing interacts with the touchscreen gesture layer. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/15975)][1])

### [P1 / blocking] A `kind: touch` hardware touchpad is processed by both the physical-mouse path and the touchscreen gesture path

The new logic records every eligible mobile hover device in `_physicalPointerDevices`. After a touchpad produces `PointerHoverEvent(kind: touch, device: D)`, subsequent down/move/up events from `D` are deliberately classified as physical input and sent through the raw mouse path. That part now looks internally consistent. 

However, `RawTouchGestureDetectorRegion` does not consult `isPhysicalPointerDeviceId()`. Its filtering is still based only on `lastDeviceKind`, and `PointerDeviceKind.touch` is explicitly considered a touch-based device. The registered tap, double-tap, long-press, and custom scale/pan recognizers have no device-ID exclusion. 

So the target sequence can become:

```text
hover(kind=touch, device=D)
    -> D added to _physicalPointerDevices

down(kind=touch, device=D)
    -> InputModel sends physical mouse-down
    -> touchscreen gesture recognizers also accept the pointer

up(kind=touch, device=D)
    -> InputModel sends physical mouse-up
    -> TapGestureRecognizer also fires touch tap handling
```

In touch mode, `onTapUp()` can issue another `tapDown()`/`tapUp()`; in mobile mouse mode, `onTap()` issues another left click. A touchpad drag can similarly be handled once as raw physical pointer movement and again by `CustomTouchGestureRecognizer` as a one-finger pan. 

That means the device-ID fix can turn the exact Android touchpad input this PR is trying to support into duplicate clicks or conflicting drag behavior.

### Fix direction

I would reject known physical-pointer devices **when recognizers admit the `PointerDownEvent`**, rather than filtering their callbacks afterward. The PR already exposes:

```dart
bool isPhysicalPointerDeviceId(int device)
```

but the gesture layer doesn't use it. 

Something based on `isPointerAllowed(PointerDownEvent event)` would be a better boundary. `supportedDevices` alone won't solve this because the important case is precisely that both the real touchscreen and hardware touchpad may arrive as `PointerDeviceKind.touch`. `CustomTouchGestureRecognizer` currently only blocks `PointerPanZoomStartEvent`; that does not reject ordinary `PointerDownEvent(kind: touch)`. 

I'd require regression coverage for these cases before merging:

* `hover(touch, D) → down/up(touch, D)` → exactly **one** remote click and zero touchscreen-tap actions.
* `hover(touch, D) → down/move/up(touch, D)` → physical drag only; no one-finger-pan callbacks.
* A real touchscreen device `D2` continues to handle tap/pan normally while `D` remains registered as the physical touchpad.
* Touchscreen and physical pointer can be interleaved without changing ownership of an active sequence.

One smaller robustness concern is that `_physicalPointerDevices` is monotonically populated and never invalidated. Flutter documents `device` as an identifier reused across interactions, and Android has a known problem where `PointerRemovedEvent` may not arrive on pointer-device disconnect, so cleanup isn't entirely straightforward. I wouldn't make that alone a blocker, but I'd document/reset the cache at an appropriate session/device boundary. ([[Flutter API Docs](https://api.flutter.dev/flutter/gestures/PointerEvent/device.html?utm_source=chatgpt.com)][2])

**Bottom line:** `039f919` resolves the previous physical-sequence routing issue, but it doesn't stop the same touch-classified physical device from simultaneously entering the touch gesture system. I would not merge until that double-routing case is addressed and tested.

[1]: https://github.com/rustdesk/rustdesk/pull/15975 "fix(mobile): support hardware touchpad and physical keyboard input on… by lapunik · Pull Request #15975 · rustdesk/rustdesk · GitHub"
[2]: https://api.flutter.dev/flutter/gestures/PointerEvent/device.html?utm_source=chatgpt.com "device property - PointerEvent class - gestures library - Dart API"

rustdesk:
Comparing the current PR head `56614cc` against the current `master` version of `AGENTS.md`, I see **one fairly clear issue, plus one structural issue that is worth raising as a review finding**. The implementation may still be functionally correct, but it does not fully align with the repository's requirement to minimize the regression surface.

### 1. Fairly clear issue: the final regression surface is not explicitly reported

`AGENTS.md` explicitly requires that, before implementation is considered complete, the author:

> explicitly report the regression surface: list the existing files and runtime paths being changed, and explain why each modification is unavoidable.

The current PR description has a `Summary of Changes`, but it does not really provide that analysis.

More importantly, the description is already stale. It still mentions:

> `Android Storage Directory Null Safety` / `native_model.dart`

But the current PR only changes these four Dart files:

* `flutter/lib/common/widgets/gestures.dart`
* `flutter/lib/common/widgets/remote_input.dart`
* `flutter/lib/mobile/pages/remote_page.dart`
* `flutter/lib/models/input_model.dart`

`native_model.dart` is no longer part of the diff.

So if this PR is considered ready to merge, I would regard this as **not satisfying the mandatory regression-surface check in `AGENTS.md`**.

I would suggest adding a section to the PR description that explicitly explains:

* `remote_page.dart`: which mobile remote-session/focus paths are changed and why the change is necessary;
* `input_model.dart`: which pointer classification/lifecycle paths are changed and why a shared model must be modified;
* `remote_input.dart`: why shared touchscreen gesture routing must change;
* `gestures.dart`: why new recognizer subclasses are required;
* which desktop/web runtime paths are expected to remain behaviorally unchanged, and how that was verified.

This is more than a documentation nit because `AGENTS.md` describes this check as **mandatory**.

---

### 2. More important structural concern: a mobile-only fix changes a shared desktop touch path

This is the review finding I consider most valuable.

`AGENTS.md` contains several principles relevant to this change:

* feature-specific logic should stay in the feature-specific module where practical;
* shared/core files should generally contain only thin hooks or capability checks;
* when a feature is disabled or unsupported, the existing implementation path should remain intact whenever possible;
* existing behavior should not be routed through a new abstraction solely to support a new feature;
* during review, unnecessarily changing a legacy path should be raised even if the new implementation appears behaviorally equivalent.

The PR itself is specifically about **mobile hardware touchpad / physical keyboard support**.

However, the implementation does not only modify `flutter/lib/mobile/`.

In `gestures.dart`, it adds wrapper recognizers such as:

* `TouchTapGestureRecognizer`
* `TouchDoubleTapGestureRecognizer`
* `TouchLongPressGestureRecognizer`

and also adds physical-device filtering to existing custom recognizers.

Then `RawTouchGestureDetectorRegion.makeGestures()` **unconditionally replaces** the existing Flutter recognizers with these new subclasses:

```dart
TouchTapGestureRecognizer:
    GestureRecognizerFactoryWithHandlers<TouchTapGestureRecognizer>(...)

TouchDoubleTapGestureRecognizer:
    ...

TouchLongPressGestureRecognizer:
    ...
```

It also injects `inputModel.isPhysicalPointerDeviceId` into:

* `HoldTapMoveGestureRecognizer`
* `DoubleFinerTapGestureRecognizer`
* `CustomTouchGestureRecognizer`

There is no mobile-only guard around this behavior.

The important detail is that `RawTouchGestureDetectorRegion` is **not mobile-only**.

The desktop remote page also uses it:

```dart
return RawTouchGestureDetectorRegion(
  child: _buildRawPointerMouseRegion(...),
  ffi: _ffi,
);
```

Therefore, although the PR is intended to fix Android/iOS hardware input, it also changes the recognizer implementation used by the existing desktop touch path.

That is exactly the kind of regression-surface expansion that `AGENTS.md` asks reviewers to scrutinize.

I would raise this as a **repository-guideline / regression-surface finding**, probably around P2/P3 severity rather than as a proven functional correctness blocker:

> The hardware-touchpad change is mobile-specific, but it unconditionally replaces the recognizers used by the shared `RawTouchGestureDetectorRegion`, which is also used by the desktop remote page. This changes an existing desktop touch path for a mobile-only feature. `AGENTS.md` asks new features to preserve unsupported/disabled legacy paths and keep feature-specific implementation out of shared/core code where practical. Please either keep desktop/web on the original recognizer factories and enable physical-device filtering only on mobile, or explain why modifying the shared desktop path is unavoidable and include that runtime path in the PR's regression-surface report.

---

### Things I would not report as violations

The fact that `gestures.dart` adds roughly 90 lines without heavily refactoring existing code is actually consistent with the `AGENTS.md` preference for minimally invasive changes and accepting limited duplication over broad refactoring.

The newly added comments also seem appropriate. For example, comments explaining why trackpad pan/zoom events are ignored because they are handled by `RawPointerMouseRegion`, or why an iOS physical-keyboard workaround exists, explain non-obvious constraints rather than merely restating the code.

The Rust, Tokio, and localization-specific rules in `AGENTS.md` are not relevant to this PR.

### Conclusion

If I were reviewing this PR strictly against `AGENTS.md`, I would raise two items:

1. **Should be fixed:** update the PR description with an explicit final regression-surface report and remove the stale `native_model.dart` / storage-directory description.
2. **Needs explanation or narrowing:** a mobile-only feature currently changes the shared `RawTouchGestureDetectorRegion`, which also changes the desktop legacy touch path.

The second point is the more important inline review comment. It does not necessarily mean the implementation must be redesigned. If the author can demonstrate that modifying the shared recognizers is unavoidable, that may be acceptable. But under `AGENTS.md`, the PR should at minimum explain why the shared change is necessary and explicitly include the desktop touch path in the regression-surface analysis.

lapunik:
Hi @rustdesk, all Greptile review comments have been resolved, physical input filtering is scoped to mobile, and all checks are passing cleanly. Ready for maintainer workflow approval and review. Thanks!

PR #16024, opened 2026-09-02
fix: recover DXGI capture after access loss

dev-joss:
## Summary

- Track DXGI frame ownership (`Idle`, `Acquired`, `Mapped`) and unmap surfaces before releasing acquired frames, propagating cleanup errors.
- Restart the capture and encoder pipeline for `DXGI_ERROR_ACCESS_LOST` (`ConnectionReset`). `DXGI_ERROR_INVALID_CALL` remains mapped to `InvalidData` and follows the existing GDI fallback path.
- Allow up to three DXGI restarts per ten-second recovery window, then use the existing GDI fallback on further access loss within that window. Successful frames do not reset the recovery budget.
- Delay `WouldBlock`-to-GDI fallback for two seconds only after an access-loss restart. Normal startup and the existing `try_gdi` handling remain unchanged.
- Preserve display-change handling and log recovery attempts and fallback decisions.

## Context

This addresses black-screen issues when applications enter fullscreen mode. The original issue was reproduced with [Disguise Designer](https://www.disguise.one/en/products/designer), while the same workflow worked under Parsec.

## Testing

Rebased onto `master` at `8154c026a`; current PR head is `9c3ac5f25`.

- [x] Windows release build with `hwcodec`, `vram`, and `flutter`, including the Flutter Windows application.
- [x] Repeat the original fullscreen / mode-switch reproduction on the rebased build.
- [x] Verify DXGI recovery with VRAM encoding active.
- [x] Exercise repeated `ACCESS_LOST` events until the retry budget is exhausted and GDI fallback occurs.

The author previously confirmed the fullscreen reproduction worked with the recovery-only grace period. Manual retesting of the rebased build is pending.

## AI use disclosure

Codex assisted with log analysis, implementation, code review, rebasing, and build validation. Manual runtime validation is performed by dev-joss.

<h2><a href="https://app.greptile.com/api/retrigger?id=59301123"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge based on the available follow-up review scope.

The PR adds explicit DXGI frame-ownership tracking and recovers from access loss by restarting capture with a bounded retry window before falling back to GDI.
- Unmaps mapped surfaces before releasing acquired DXGI frames and propagates cleanup failures.
- Persists recovery state across capture-loop restarts.
- Allows three access-loss restarts per ten-second window and grants restarted DXGI capture two seconds to produce a frame.
- Preserves the existing GDI fallback for exhausted recovery and non-recoverable DXGI errors.
</details>

```mermaid
flowchart TD
    A[Capture frame] --> B{Result}
    B -->|Success| A
    B -->|WouldBlock| C{Access-loss restart grace active?}
    C -->|Yes, under 2 seconds| A
    C -->|No| D{Existing GDI threshold exceeded?}
    D -->|No| A
    D -->|Yes| G[Switch to GDI]
    B -->|ConnectionReset| E{Recovery budget available?}
    E -->|Yes| F[Restart capture pipeline]
    F --> A
    E -->|No| G
    B -->|Other DXGI error| G
```
</details>

<sub>Reviews (10) · Last reviewed commit: ["chore: log DXGI recovery attempts at deb..."](https://github.com/rustdesk/rustdesk/commit/dca0dd0ee80937f7b511102f6a847864dc183706)</sub>


src/server/video_service.rs
```diff
@@ -791,10 +832,10 @@ fn run(vs: VideoService) -> ResultType<()> {
                 #[cfg(windows)]
                 {
                     #[cfg(feature = "vram")]
-                    if try_gdi == 1 && !c.is_gdi() {
+                    if try_gdi && !c.is_gdi() {
                         VRamEncoder::set_fallback_gdi(sp.name(), false);
                     }
-                    try_gdi = 0;
+                    try_gdi = false;
```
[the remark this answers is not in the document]
dev-joss: Good catch. The unconditional reset was pre-existing on `master` (`try_gdi = 0` ran after any `Ok(frame)`), but the new recovery grace period depends on keeping the fallback armed until a valid frame is produced, so this PR now fixes that behavior as well.

I moved both the retry-state reset and the VRAM fallback-flag clear into the valid-frame branch. I kept the VRAM flag clear before `try_gdi = false`; applying the suggested ordering literally would make that clear inert. The capture backend is snapshotted before frame acquisition to satisfy the frame's borrow of the capturer.

The full Windows release build with `hwcodec`, `vram`, and `flutter` passes.

rustdesk:
Yes. After comparing the PR against `master`, I would **not merge it as-is**. The DXGI ownership fix looks directionally useful, but the `video_service.rs` changes are broader than necessary and change established fallback policy.

The cleanest approach is: **fix DXGI lifecycle/recovery while leaving `try_gdi` and existing GDI fallback semantics alone.**

On `master`, `try_gdi` is not really a boolean. It acts as both state and counter: it begins at `1`, advances on `WouldBlock`, falls back after several failed attempts, and becomes `0` once DXGI has produced a frame. It also controls the VRAM fallback flag. 

The PR changes that to a boolean plus a hard-coded two-second grace period. That means an existing path that could fall back after several frame attempts now waits up to two seconds. That's a user-visible behavioral change unrelated to fixing `DXGI_ERROR_ACCESS_LOST`. 

More importantly, I would **not treat `DXGI_ERROR_INVALID_CALL` as equivalent to `DXGI_ERROR_ACCESS_LOST`**. Microsoft explicitly says `ACCESS_LOST` means the duplication interface became invalid and must be recreated. `INVALID_CALL`, by contrast, usually means the application has violated the acquire/map/release state—for example acquiring without releasing the previous frame or mapping while already mapped. ([[Microsoft Learn](https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_2/nf-dxgi1_2-idxgioutputduplication-mapdesktopsurface?utm_source=chatgpt.com)][1])

So a smaller fix would be roughly:

```rust
match res {
    Err(ref e) if e.kind() == WouldBlock => {
        // KEEP EXISTING try_gdi logic unchanged.
    }

    Err(err) => {
        if vs.source.is_monitor() {
            try_broadcast_display_changed(&sp, display_idx, &c, true)?;
        }

        #[cfg(windows)]
        if !c.is_gdi() {
            if err.kind() == ConnectionReset {
                // DXGI_ERROR_ACCESS_LOST:
                // recreate the DXGI capture pipeline.
                log::warn!("DXGI access lost, recreating capturer");
                bail!("SWITCH");
            }

            // Preserve existing behavior for all other errors.
            c.set_gdi();
            log::info!("dxgi error, fall back to gdi: {:?}", err);
            continue;
        }

        return Err(err.into());
    }

    _ => {}
}
```

That alone has one potential problem: persistent `ACCESS_LOST` could cause a restart loop. So I would add a **separate, narrowly scoped recovery guard**, not modify `try_gdi`.

For example, allow one or a small number of DXGI recreations in a time window:

```rust
struct DxgiRecoveryState {
    attempts: usize,
    window_started: Instant,
}

const DXGI_RECOVERY_LIMIT: usize = 3;
const DXGI_RECOVERY_WINDOW: Duration = Duration::from_secs(10);
```

On `ACCESS_LOST`:

```rust
if recovery.window_started.elapsed() > DXGI_RECOVERY_WINDOW {
    recovery.attempts = 0;
    recovery.window_started = Instant::now();
}

if recovery.attempts < DXGI_RECOVERY_LIMIT {
    recovery.attempts += 1;
    bail!("SWITCH");
}

// existing fallback
c.set_gdi();
```

Critically, **do not reset this counter after one valid frame**. The current PR resets `dxgi_recovery_count` to zero on every valid frame.  That means this pattern never reaches the advertised three-restart limit:

```text
ACCESS_LOST
restart
one good frame       -> counter = 0
ACCESS_LOST
restart
one good frame       -> counter = 0
ACCESS_LOST
...
```

So an unstable driver/application can restart the entire capture/encoder pipeline indefinitely. The PR description says "retry ... three times before falling back to GDI," but the implementation actually limits only **consecutive failures before a single valid frame**. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16024)][2])

### Logging

I would also remove the unconditional `eprintln!` from `wrap_hresult()`:

```rust
if x == DXGI_ERROR_ACCESS_LOST {
    eprintln!(...);
} else if x != S_OK && x != DXGI_ERROR_WAIT_TIMEOUT {
    eprintln!(...);
}
```

`wrap_hresult()` is a low-level helper used on acquire/map/unmap/release operations. The same failure can therefore be logged there and again by `video_service`. During repeated recovery, you'll also get capture initialization/restart messages. 

Low-level code should generally **return the error**, while the recovery boundary logs it once with useful context. At most, use `log::debug!`/`trace!` there.

The frame ownership work itself is much easier to justify independently: the PR now tracks whether a frame is actually acquired/mapped and only unmaps/releases when appropriate.  That directly addresses the circumstances that can produce `INVALID_CALL`. I would keep that change, test it separately, then add the smallest possible `ACCESS_LOST -> recreate DXGI` behavior.

A GitHub review comment I'd leave would be:

I think this fix is currently too aggressive for the issue being addressed.

The DXGI frame ownership changes make sense and should probably be kept, but `video_service.rs` also changes the existing GDI fallback state machine. In particular, `try_gdi` is changed from a counter/state value to a bool and the existing retry behavior is replaced with a fixed 2-second grace period. That changes established behavior for normal `WouldBlock` handling and may introduce regressions unrelated to `DXGI_ERROR_ACCESS_LOST`.

I suggest keeping the existing `try_gdi` logic unchanged and handling DXGI recovery separately.

Also, `DXGI_ERROR_ACCESS_LOST` and `DXGI_ERROR_INVALID_CALL` should probably not share the same recovery policy. Microsoft documents `ACCESS_LOST` as requiring recreation of `IDXGIOutputDuplication`, while `INVALID_CALL` generally indicates incorrect acquire/map/release state. The frame-ownership fix should address the latter; if `INVALID_CALL` remains, hiding it behind pipeline recreation could mask another lifecycle bug.

A smaller change would be:

* keep the frame acquired/mapped ownership fix;
* preserve the existing `try_gdi`/`WouldBlock` fallback behavior;
* recreate the DXGI pipeline only for `DXGI_ERROR_ACCESS_LOST`;
* use a separate bounded recovery state for ACCESS_LOST;
* do not reset the recovery budget after a single valid frame—otherwise an alternating good-frame/access-loss sequence can restart forever;
* fall back to the existing GDI path once the recovery budget/window is exhausted;
* avoid `eprintln!` from `wrap_hresult()` for every DXGI error; log once at the recovery/fallback boundary instead.

This would make the patch much easier to reason about and substantially reduce the regression surface while still fixing the reported full-screen/access-loss case.

I would make this a **request-changes level concern**, principally because of the changed `try_gdi` semantics and the recovery counter resetting after a single good frame. The author's test confirms their particular scenario works, but it doesn't establish that the broader fallback-policy changes are safe.

[1]: https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_2/nf-dxgi1_2-idxgioutputduplication-mapdesktopsurface?utm_source=chatgpt.com "IDXGIOutputDuplication::MapDesktopSurface (dxgi1_2.h) - Win32 apps | Microsoft Learn"
[2]: https://github.com/rustdesk/rustdesk/pull/16024 "fix: recover DXGI capture after access loss by dev-joss · Pull Request #16024 · rustdesk/rustdesk · GitHub"

dev-joss:
> Also, DXGI_ERROR_ACCESS_LOST and DXGI_ERROR_INVALID_CALL should probably not share the same recovery policy. Microsoft documents ACCESS_LOST as requiring recreation of IDXGIOutputDuplication, while INVALID_CALL generally indicates incorrect acquire/map/release state. The frame-ownership fix should address the latter; if INVALID_CALL remains, hiding it behind pipeline recreation could mask another lifecycle bug. 

~This was the observed behavior 🤷 . Initial commit just handled DXGI_ERROR_ACCESS_LOST, but then issues still occurred intermittently, until DXGI_ERROR_INVALID_CALL was handles also.~

nvm, I understand the point

rustdesk:
Thanks for addressing the previous recovery issues. The ACCESS_LOST handling is now much closer to what we want: INVALID_CALL is no longer treated as ACCESS_LOST, and the recovery budget is preserved across restarts.

However, I still think this PR is changing too much behavior outside the actual fix.

The main requirement from the previous review was to keep the existing `try_gdi` / `WouldBlock` fallback behavior unchanged and add ACCESS_LOST recovery separately. The current version still does not do that.

In particular:

* `DXGI_INITIAL_FRAME_GRACE` and `dxgi_started` change the existing fallback from “after several WouldBlock attempts” to “after several attempts AND up to a 2-second grace period”.
* `try_gdi` is now reset only when `frame.valid()` succeeds, whereas master resets it after a successful `c.frame()`. `frame_from_dxgi` was added as part of that behavioral change.
* The additional capture-initialization logging and some fallback validation also increase the patch without being necessary for ACCESS_LOST recovery.

I would prefer this PR to stay narrowly scoped:

1. Keep the DXGI frame ownership/lifecycle fix.
2. Keep a small bounded recovery state for `DXGI_ERROR_ACCESS_LOST`.
3. On ACCESS_LOST, restart/recreate DXGI while the recovery budget allows it.
4. Once the budget is exhausted, use the existing GDI fallback path.
5. Restore the existing `try_gdi` / `WouldBlock` logic exactly as it is on master.
6. Remove unrelated logging/behavioral changes unless they are required to reproduce and fix this specific issue.

The current implementation is much better than the earlier version, but the regression surface is still larger than necessary for this bug fix. Please reduce the diff to the minimum needed for DXGI ownership + ACCESS_LOST recovery.

dev-joss:
I restored the normal try_gdi/WouldBlock behavior to match master and removed the unrelated logging, frame handling, and fallback-validation changes.

Testing then reproduced the black screen. After ACCESS_LOST, the recreated DXGI capturer initially returned WouldBlock, causing the existing logic to switch to GDI after about 135 ms—before DXGI produced its first frame.

The remaining deviation is therefore limited to ACCESS_LOST recovery: a one-shot marker gives only the recreated DXGI capturer up to two seconds to produce its first frame. Normal startup remains unchanged, and the existing fallback resumes after that interval.

I rebuilt and tested this recovery-only grace, and the black screen no longer occurred.

rustdesk:
I reviewed PR #16024 (“fix: recover DXGI capture after access loss”) in `rustdesk/rustdesk`. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16024)][1])

Overall: **good direction, but I would request changes before merge**. The core idea (recover DXGI capture instead of immediately degrading to GDI) is valuable, and the frame lifecycle fixes look like the strongest part of the patch. However, the PR currently mixes a targeted DXGI recovery fix with a broader rewrite of fallback behavior, which increases regression risk. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16024)][1])

## Blocking concerns

### 1. `try_gdi` behavior was changed too broadly

The PR changes `try_gdi` semantics from an existing state/counter-style mechanism into a boolean plus a fixed grace period. That affects normal `WouldBlock` handling, not just `DXGI_ERROR_ACCESS_LOST`. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16024)][1])

Risk:

* Existing capture startup behavior may regress.
* A system that previously fell back after a number of failed frames now waits for a time-based threshold.
* This changes behavior for cases unrelated to the reported fullscreen/access-loss bug.

Suggestion:

* Keep the existing `try_gdi` state machine.
* Add a separate DXGI recovery state only for `DXGI_ERROR_ACCESS_LOST`.

---

### 2. Recovery counter reset can cause infinite restart loops

The PR claims bounded recovery (“retry DXGI recovery three times before falling back”), but the counter appears to reset after any valid frame. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16024)][1])

Problem scenario:

```
ACCESS_LOST
→ restart DXGI
→ one good frame
→ reset counter
→ ACCESS_LOST
→ restart DXGI
→ one good frame
→ reset counter
...
```

A flaky driver/application can therefore avoid ever reaching the fallback limit.

Suggestion:

Use a time-window budget:

```rust
struct DxgiRecoveryState {
    attempts: usize,
    window_started: Instant,
}
```

Example policy:

* max 3 DXGI recreations per 10 seconds
* reset budget only after the window expires or sustained stability

---

### 3. `DXGI_ERROR_INVALID_CALL` should not share the same recovery path

The PR initially targeted `DXGI_ERROR_ACCESS_LOST`, then expanded recovery to `INVALID_CALL`. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16024)][1])

I would separate these:

* `DXGI_ERROR_ACCESS_LOST`

  * expected after display mode changes/fullscreen transitions
  * recreate duplication/capture pipeline

* `DXGI_ERROR_INVALID_CALL`

  * usually indicates misuse of DXGI frame lifecycle
  * should first be fixed by the ownership/unmap/release changes

The frame ownership work is actually the cleaner fix here. Hiding lifecycle bugs behind pipeline recreation could make future debugging harder.

---

## Things I like

### ✅ DXGI frame ownership tracking

The changes in `libs/scrap/src/dxgi/mod.rs` are the best part of the PR. Tracking whether a frame was acquired/mapped before unmap/release directly addresses invalid lifecycle states. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16024)][1])

I would keep this almost independently.

---

### ✅ Better recovery boundary

Restarting the capture/encoder pipeline after a true access loss is the correct architectural direction. The current PR's summary and changed files show that this is the intended behavior. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16024)][1])

The issue is not the concept; it is the scope.

---

## Minor comments

### Logging

Avoid unconditional `eprintln!` inside low-level DXGI helpers.

Prefer:

* DXGI layer: return errors
* video service recovery layer: log once with context

Otherwise repeated failures may produce duplicated logs during recovery storms. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16024)][1])

---

## Suggested merge plan

I would split this into:

1. **Commit 1**

   * Keep DXGI frame ownership tracking.
   * Fix acquire/map/unmap/release lifecycle.

2. **Commit 2**

   * Add `ACCESS_LOST`-only recovery.
   * Preserve existing GDI fallback behavior.

3. **Commit 3 (optional)**

   * Improve recovery heuristics after real-world testing.

---

### Verdict

**Request changes**

The PR solves a real problem and the DXGI lifecycle improvements are worthwhile, but the fallback-state changes are too invasive for a bug fix. A smaller patch would be safer:

* keep frame ownership fixes ✅
* keep DXGI recreation on `ACCESS_LOST` ✅
* preserve existing fallback logic ✅
* separate invalid-call debugging from access-loss recovery ✅

That would make this much easier to merge with lower regression risk.

[1]: https://github.com/rustdesk/rustdesk/pull/16024 "fix: recover DXGI capture after access loss by dev-joss · Pull Request #16024 · rustdesk/rustdesk · GitHub"

dev-joss:
@rustdesk can you re-review? Your comment seems outdated. 

 * try_gdi isn't a boolean
 * good frame doesn't reset the budget. separate budget allows 3 restarts within a 10 sec window
 * Only ConnectionReset triggers recovery. INVALID_CALL follows existing fallback path

src/server/video_service.rs
```diff
@@ -858,6 +926,19 @@ fn run(vs: VideoService) -> ResultType<()> {
...
+                        let recovery_attempt = dxgi_recovery_state.lock().unwrap().next_attempt();
+                        if let Some(attempt) = recovery_attempt {
+                            log::info!(
+                                "dxgi access lost, restart capture: attempt {attempt}, error: {err:?}"
+                            );
+                            bail!("SWITCH");
+                        }
+                        log::warn!(
+                            "dxgi access lost after {DXGI_RECOVERY_LIMIT} restarts in {} seconds, fall back to gdi: {err:?}",
+                            DXGI_RECOVERY_WINDOW.as_secs()
```
rustdesk, line 957: Will this introduce a lot of messages?

rustdesk:
I think the overall direction is good, especially the explicit DXGI frame lifecycle tracking, but I would not merge the current version yet.

There is one regression in the latest commit that I think should be fixed first.

### 1. `try_gdi` is cleared even when the returned frame is invalid

Earlier in this PR, `try_gdi` was only cleared after `frame.valid()` succeeded. The latest commit moved that logic back outside the `frame.valid()` branch.

That means this sequence is possible:

```text
DXGI_ERROR_ACCESS_LOST
    ↓
restart DXGI
    ↓
recovery grace period
    ↓
c.frame() returns Ok(frame), but frame.valid() == false
    ↓
try_gdi = 0
    ↓
subsequent frames return WouldBlock
    ↓
GDI fallback is no longer armed
```

So the recreated DXGI capturer can remain alive without ever producing a usable frame, while the normal `WouldBlock` fallback path is disabled.

I think `try_gdi` should only be cleared once a valid DXGI frame has actually been produced.

Something along these lines:

```rust
#[cfg(windows)]
let frame_from_dxgi = !c.is_gdi();

let res = match c.frame(spf) {
    Ok(frame) => {
        repeat_encode_counter = 0;

        if frame.valid() {
            #[cfg(windows)]
            {
                #[cfg(feature = "vram")]
                if try_gdi == 1 && frame_from_dxgi {
                    VRamEncoder::set_fallback_gdi(sp.name(), false);
                }

                try_gdi = 0;
            }

            // process frame...
        }

        Ok(())
    }
    Err(err) => Err(err),
};
```

This was actually handled correctly by an earlier revision of this PR, so I think the latest refactor accidentally reintroduced the old behavior.

### 2. The PR description no longer matches the implementation for `DXGI_ERROR_INVALID_CALL`

The description currently says:

> recover from `DXGI_ERROR_INVALID_CALL` by recreating DXGI instead of immediately using GDI

But the current implementation only restarts the capture pipeline for `ConnectionReset`, which corresponds to `DXGI_ERROR_ACCESS_LOST`.

`DXGI_ERROR_INVALID_CALL` is mapped to `InvalidData` and therefore falls through to the normal GDI fallback.

I think recovering only `DXGI_ERROR_ACCESS_LOST` is probably the safer behavior, especially now that the frame ownership / release ordering has been fixed. If that is intentional, the PR description should be updated accordingly.

If `DXGI_ERROR_INVALID_CALL` still occurs in the original fullscreen reproduction after the lifecycle fix, then I would rather preserve the actual DXGI HRESULT/type than use the generic `io::ErrorKind::InvalidData` as the recovery signal.

### 3. Please rebase and retest against current `master`

This branch is now fairly old relative to `master`, and `src/server/video_service.rs` has changed since the PR was opened.

Before merging I would rebase and rerun at least:

* Windows release build with `hwcodec`, `vram`, and `flutter`
* the original fullscreen / mode-switch reproduction
* DXGI recovery with VRAM encoding enabled
* repeated `ACCESS_LOST` events until the retry budget is exhausted and GDI fallback is exercised

The DXGI lifecycle part itself looks much better than the old unconditional `ReleaseFrame` / unmap handling. The explicit `Idle -> Acquired -> Mapped` state and cleanup order make sense to me.

So overall I think this is close, but I would fix the `invalid frame` / `try_gdi` regression before merging.

dev-joss:
> 1. try_gdi is cleared even when the returned frame is invalid

@rustdesk  You said in https://github.com/rustdesk/rustdesk/pull/16024#issuecomment-5519994935 you wanted the opposite....  this code is the same as master. So this is a different issue for a different PR I believe

> try_gdi is now reset only when frame.valid() succeeds, whereas master resets it after a successful c.frame(). frame_from_dxgi was added as part of that behavioral change.

> Restore the existing try_gdi / WouldBlock logic exactly as it is on master.

> Please reduce the diff to the minimum needed for DXGI ownership + ACCESS_LOST recovery.

rustdesk:
Thanks, this explanation makes sense. I agree that restoring the normal `try_gdi` / `WouldBlock` behavior to match master, and limiting the two-second grace period specifically to an `ACCESS_LOST` recovery, keeps the regression surface much smaller.

I still see one recovery-specific issue, though.

In the error path, `try_broadcast_display_changed(...)?` runs before `ConnectionReset` is handled and before `next_attempt()` sets `restart_pending`.

If `ACCESS_LOST` happens together with a display mode/resolution/layout change, `try_broadcast_display_changed()` can `bail!("SWITCH")` first. In that case we do recreate the DXGI capturer, but the recovery marker was never set, so the new capturer does not receive the two-second grace period and can still fall back to GDI after the normal ~135 ms behavior.

That seems particularly relevant because display mode/full-screen transitions are one of the common causes of `DXGI_ERROR_ACCESS_LOST`.

Could we record the ACCESS_LOST recovery state before `try_broadcast_display_changed()`, so that even if the display-change path triggers the `SWITCH`, the next DXGI capturer still knows it is an ACCESS_LOST recovery?

Other than that, I’m much more comfortable with the current scope.

Also, the PR description still says `DXGI_ERROR_INVALID_CALL` is recovered by recreating DXGI, while the current implementation intentionally only retries `ACCESS_LOST`. That description should probably be updated as well.

src/server/video_service.rs
dev-joss: change to debug logging

rustdesk:
The code looks good to me now. The recovery state is preserved across display-change restarts, and the fallback budget can no longer be bypassed by an early `SWITCH`.

Before merging, could you please complete the remaining runtime tests on the rebased build, especially:

* reproduce the original fullscreen / mode-switch scenario
* verify recovery while VRAM encoding is active
* trigger repeated `ACCESS_LOST` events until the retry budget is exhausted and confirm it falls back to GDI correctly

If those pass, I’m comfortable with merging this.

dev-joss:
Re-tested and confirmed works

PR #16026, opened 2026-09-02
fix: convert X11 depth-30 root windows to BGRA instead of mislabeling…

rustdesk:
## Fix X11 depth-30 root windows being captured as BGRA

On a 10-bit X11 desktop (`DefaultDepth 30`, or NVIDIA "30-bit color") the root window is xRGB2101010. `get_pixfmt` only knows depths 16 and 32, and the `unwrap_or(Pixfmt::BGRA)` after it silently labels the 10-bit pixels as 8-bit BGRA, so the controller sees scrambled colours.

### Change

- Add `Pixfmt::AR30`, libyuv's name for X11's little-endian xRGB2101010 (blue in bits 0–9, red in 20–29), and map depth 30 to it.
- `x11::Capturer` converts each changed AR30 frame to BGRA with libyuv `AR30ToARGB` into an owned buffer, so encoders, screenshots and everything downstream still see BGRA.
- The top two bits of the source are padding, not alpha; libyuv expands them to 0/85/170/255, so the output alpha is forced to 255. This matters for the screenshot feature, which writes the BGRA buffer straight into a PNG. A unit test pins the conversion for all four padding values.

### Blast radius

Four files, +76/-2. Only depth-30 roots take a new code path; every other depth executes the same code as before. An earlier revision of this PR also validated the pixmap format, byte order and root visual masks and refused unknown layouts. That ran new xcb FFI code for every X11 user and turned a wrong-colour bug into a no-picture failure, so it was dropped: depth 30 is already a hand-configured setup, and the little-endian assumption matches what this file already documents for the other depths.

### Testing

- Verified libyuv's AR30 layout against X11's with a real call into libyuv.
- `cargo check` of the x11 module for a Linux target; the unit test lives in the Linux-only capturer file and passed on macOS when it was in the shared conversion module.
- Not yet run on a real depth-30 X server. Testers: `DefaultDepth 30` in xorg.conf or NVIDIA "30-bit color", then compare colours and take a screenshot from the controller.

https://claude.ai/code/session_01RLb1dNdhFqUQExJPANjo1F

### DIFF
--- a/libs/scrap/src/common/mod.rs
+++ b/libs/scrap/src/common/mod.rs
@@ -235,6 +235,8 @@ pub enum Pixfmt {
     BGRA,
     RGBA,
     RGB565LE,
+    // X11 depth 30: xRGB2101010, little endian (libyuv's AR30).
+    AR30,
     I420,
     NV12,
     I444,
@@ -243,7 +245,7 @@ pub enum Pixfmt {
 impl Pixfmt {
     pub fn bpp(&self) -> usize {
         match self {
-            Pixfmt::BGRA | Pixfmt::RGBA => 32,
+            Pixfmt::BGRA | Pixfmt::RGBA | Pixfmt::AR30 => 32,
             Pixfmt::RGB565LE => 16,
             Pixfmt::I420 | Pixfmt::NV12 => 12,
             Pixfmt::I444 => 24,
--- a/libs/scrap/src/common/x11.rs
+++ b/libs/scrap/src/common/x11.rs
@@ -23,7 +23,7 @@ impl TraitCapturer for Capturer {
     fn frame<'a>(&'a mut self, _timeout: Duration) -> io::Result<Frame<'a>> {
         let width = self.width();
         let height = self.height();
-        let pixfmt = self.0.display().pixfmt();
+        let pixfmt = self.0.pixfmt();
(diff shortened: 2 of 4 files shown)


rustdesk:
Close this since few guys use this.

rustdesk:
Depth 30 really is rare, and nobody has reported it. I searched the rustdesk issues for "depth 30", "30-bit" and "10-bit linux" and found nothing related. Reasons:

- Xorg defaults to depth 24. Getting to 30 requires a hand-written `DefaultDepth 30` in xorg.conf or enabling 30-bit color in nvidia-settings; no distribution ships that by default.
- Historically, depth 30 broke a pile of software: Chromium, Qt, Wine, Steam. Most people who tried it turned it back off.
- GNOME and KDE now default to Wayland, and RustDesk's Wayland path goes through PipeWire, which never touches this code.
- The intersection is "Linux + X11 + manually configured 10-bit + acting as the controlled side", which is essentially colour-grading workstations, and those people mostly work on Windows and macOS.

When it goes wrong the colours are completely scrambled, not subtly off, so anyone who hit it would report it. I found this bug by reading the code, not from a user report. That also shows, in reverse, that the validation added in the earlier rounds was protecting an even smaller slice of an already tiny population.

PR #16028, opened 2026-09-02
fix: prevent stale clicks in iOS touch mode

### LINKED ISSUE #16025: [iOS]A tap clicks the previous cursor position before clicking the new position
### Bug Description

When the RustDesk iOS client controls a Windows 11 host in touch mode, one
finger tap can generate two remote click actions: first at the cursor position
from the previous touch gesture, then at the position of the current tap.

This can activate an unrelated UI element at the old position before activating
the intended element. No physical mouse is involved.

I built RustDesk 1.4.9 from source and verified a local fix. The relevant race is
in flutter/lib/common/widgets/remote_input.dart: a late TapGestureRecognizer
callback can reuse stale TapDownDetails after a one-finger pan has already owned
the same physical touch sequence.

### How to Reproduce

1. Use RustDesk 1.4.9 on an iPhone running iOS 26.
2. Connect to a Windows 11 host running RustDesk.
3. Use touch mode and tap a UI element at position A.
4. Tap once at a different position B.
5. Observe that the remote side first receives a click at A and then at B.

The issue may be easier to reproduce after a short one-finger pan followed by a
tap at a different location.

### Expected Behavior

A physical tap at position B must generate exactly one remote click at B. No
click should be emitted at a position saved by an earlier gesture.

iOS 26 -> Windows 11

1.4.9 -> 1.4.9

### Screenshots

Please see the attached screen recording showing one physical tap producing a
click at the previous cursor position followed by a click at the new position.

### Additional Context

A verified local fix against 1.4.9:
- assigns an ID to each physical touch sequence;
- invalidates stale TapDownDetails from an earlier sequence;
- ignores a late tap-up after a pan handled the same touch;
- serializes pan start/end and guarantees balanced mouse down/up events.

The patched iOS build no longer reproduces the problem.

Possibly related, but not the same environment/symptom: #7046.
I can port the fix to current master and open a signed-off PR after maintainers
confirm the preferred behavior.

shanglongy-lang:
## Summary

- Associate iOS tap-down state with the physical touch that created it.
- Mark a touch as pan-owned before the existing pan-start awaits, and reject stale taps before and after asynchronous cursor movement.
- Clear pending single-tap callbacks and cached tap-down details when a double tap wins, so subsequent single taps continue working.
- Keep existing long-press and pan start/update/end/cancel behavior unchanged.

## Problem and scope

In the reported iOS touch-mode case, a tap at a new location can first click the previous cursor position. Tap state must not be reused by a callback from another physical touch, or by a tap callback after pan recognition has claimed that touch.

The sequence tracking is limited to iOS. A double tap also needs explicit cleanup: Flutter 3.24.5 can deliver two tap-down callbacks followed by only one tap-cancel. Leaving the second entry pending causes later single taps to consume stale entries indefinitely. Cleanup happens synchronously in `onDoubleTap`, before its early returns and mouse events.

## Tests

`flutter/test/touch_mode_gesture_tracker_test.dart` covers:

- A late tap rejected after pan recognition.
- An older callback not consuming a newer tap's cached state.
- A new physical touch starting while cursor movement is deliberately paused.
- Multiple pointers belonging to the same physical touch sequence.
- Real Flutter tap/double-tap recognition delivering two tap-downs and one cancel, followed by two successful single taps.

Validation on Flutter 3.24.5 / Dart 3.5.4: all 5 tests pass in a minimal Flutter package containing byte-identical copies of the tracker and test file, and `flutter analyze --no-pub` reports no issues for that package. Removing double-tap cleanup in the test harness makes the new regression fail on the first subsequent single tap (expected 1 click, observed 0); restoring cleanup makes it pass. `git diff --check` also passes.

The widget regression uses a recognizer/tracker harness, not the full `RemoteInput` widget or a remote connection. Device validation of this master-based branch remains necessary; the earlier user-reported success was for the local 1.4.9 patch.

Fixes #16025

This PR prevents stale iOS touch-mode clicks by associating tap state with physical touch sequences.
- Tracks active pointers, pending tap-downs, and pan ownership on iOS.
- Revalidates tap ownership before and after asynchronous cursor movement.
- Clears residual tap state when double-tap or two-finger scale recognition supersedes a tap.
- Adds regression coverage for delayed callbacks, pan ownership, sequence transitions, and post-double-tap behavior.
</details>

The PR appears safe to merge, with no actionable correctness, security, or repository-rule violations identified.

The sequence tracker validates taps both before and after cursor movement, synchronously records pan ownership, and cleans up double-tap state without exposing an interleaving that would discard a subsequent physical touch.
</details>

| Filename | Overview |
|----------|----------|
| flutter/lib/common/widgets/remote_input.dart | Integrates iOS pointer-sequence tracking into tap, pan, scale, cancellation, and double-tap gesture handling. |
| flutter/lib/common/widgets/touch_mode_gesture_tracker.dart | Adds a focused tracker that rejects stale or pan-owned tap callbacks around asynchronous cursor movement. |
| flutter/test/touch_mode_gesture_tracker_test.dart | Covers stale callbacks, pan ownership, multi-pointer sequencing, pending movement, and double-tap queue cleanup. |

</details>

```mermaid
flowchart TD
  A[Pointer down] --> B[Start or join touch sequence]
  B --> C[Record tap-down sequence]
  C --> D{Pan claims sequence?}
  D -- Yes --> X[Suppress tap]
  D -- No --> E[Tap-up consumes queued sequence]
  E --> F{Sequence still current?}
  F -- No --> X
  F -- Yes --> G[Move remote cursor asynchronously]
  G --> H{Sequence still current and not pan-owned?}
  H -- No --> X
  H -- Yes --> I[Send mouse tap]
```
</details>

<sub>Reviews (5): Last reviewed commit: ["fix: clear pending iOS taps after double..."](https://github.com/rustdesk/rustdesk/commit/a112a123fa408bb232ce080c396a3164087ff319) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59359269)</sub>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved iOS touch and gesture handling for overlapping, canceled, double-tap, and multi-touch interactions.
  * Prevented delayed or stale tap callbacks from triggering after a newer touch sequence or pan gesture.
  * Improved tap, pan, and two-finger scaling behavior with more reliable gesture completion and cleanup.
  * Ensured single taps are delivered correctly when a double-tap gesture takes precedence.

* **Tests**
  * Added coverage for gesture sequencing, delayed taps, pan transitions, cancellations, double taps, and multi-pointer interactions.

### DIFF
--- a/flutter/lib/common/widgets/remote_input.dart
+++ b/flutter/lib/common/widgets/remote_input.dart
@@ -199,23 +199,27 @@ class _RawTouchGestureDetectorRegionState
       return;
     }
     if (handleTouch) {
-      if (isIOS &&
-          (touchSequence == null ||
-              !_touchModeGestureTracker.shouldHandleTap(touchSequence))) {
-        return;
-      }
-      final isMoved =
-          await ffi.cursorModel.move(d.localPosition.dx, d.localPosition.dy);
-      if (isIOS &&
-          !_touchModeGestureTracker.shouldHandleTap(touchSequence!)) {
-        return;
-      }
-      if (isMoved) {
+      Future<void> sendTap() async {
         // If pan already handled 'down', don't send it again.
         if (lastTapDownDetails != null && !_touchModePanStarted) {
           await inputModel.tapDown(MouseButtons.left);
         }
         await inputModel.tapUp(MouseButtons.left);
+      }
+      if (isIOS) {
+        if (touchSequence == null) {
+          return;
+        }
+        await handleTrackedTap(
+          tracker: _touchModeGestureTracker,
+          sequence: touchSequence,
+          move: () =>
+              ffi.cursorModel.move(d.localPosition.dx, d.localPosition.dy),
+          sendTap: sendTap,
+        );
+      } else if (await ffi.cursorModel
+          .move(d.localPosition.dx, d.localPosition.dy)) {
+        await sendTap();
       }
     }
   }
--- a/flutter/lib/common/widgets/remote_input.dart
+++ b/flutter/lib/common/widgets/remote_input.dart
@@ -464,6 +464,7 @@ class _RawTouchGestureDetectorRegionState
     final isMoved = await ffi.cursorModel
         .move(panStartPosition.dx, panStartPosition.dy);
     if (!isMoved ||
+        panState.ended ||
         panState.cancelled ||
         panState.touchSequence != _touchModeGestureTracker.sequence) {
       return;
@@ -473,7 +474,8 @@ class _RawTouchGestureDetectorRegionState
       await inputModel.sendMouse('down', MouseButtons.left);
       panState.mouseDownSent = true;
     }
-    if (panState.cancelled ||
+    if (panState.ended ||
+        panState.cancelled ||
         panState.touchSequence != _touchModeGestureTracker.sequence) {
       return;
     }
--- a/flutter/lib/common/widgets/remote_input.dart
+++ b/flutter/lib/common/widgets/remote_input.dart
@@ -59,17 +59,6 @@ bool isSpecialHoldDragActive = false;
 // Cache the last focal point to calculate deltas in special hold-drag mode.
 Offset _lastSpecialHoldDragFocalPoint = Offset.zero;
 
-class _TouchModePanState {
-  _TouchModePanState(this.touchSequence);
-
-  final int touchSequence;
-  late final Future<void> startFuture;
-  bool ended = false;
-  bool cancelled = false;
-  bool mouseDownSent = false;
-  bool finishing = false;
-}
-
 class RawTouchGestureDetectorRegion extends StatefulWidget {
   final Widget child;
   final FFI ffi;
@@ -107,6 +96,7 @@ class _RawTouchGestureDetectorRegionState
 
   // Workaround tap down event when two fingers are used to scale(mobile)
   TapDownDetails? _lastTapDownDetails;
+  int? _lastTapDownSequence;
 
   PointerDeviceKind? lastDeviceKind;
 
@@ -116,7 +106,6 @@ class _RawTouchGestureDetectorRegionState
   bool _touchModePanStarted = false;
   final TouchModeGestureTracker _touchModeGestureTracker =
       TouchModeGestureTracker();
-  _TouchModePanState? _touchModePanState;
   Offset _doubleFinerTapPosition = Offset.zero;
 
   // For mouse mode, we need to block the events when the cursor is in a blocked area.
@@ -134,14 +123,18 @@ class _RawTouchGestureDetectorRegionState
 
   @override
   Widget build(BuildContext context) {
+    final detector = RawGestureDetector(
+      child: widget.child,
+      gestures: makeGestures(context),
+    );
+    if (!isIOS) {
+      return detector;
+    }
     return Listener(
       onPointerDown: _onTouchPointerDown,
       onPointerUp: _onTouchPointerEnd,
       onPointerCancel: _onTouchPointerEnd,
-      child: RawGestureDetector(
-        child: widget.child,
-        gestures: makeGestures(context),
-      ),
+      child: detector,
     );
   }
 
@@ -149,12 +142,7 @@ class _RawTouchGestureDetectorRegionState
     if (!kTouchBasedDeviceKinds.contains(event.kind)) {
       return;
     }
-    lastDeviceKind = event.kind;
-    final previousSequence = _touchModeGestureTracker.sequence;
     _touchModeGestureTracker.pointerDown(event.pointer);
-    if (_touchModeGestureTracker.sequence != previousSequence) {
-      _touchModePanStarted = false;
-    }
   }
 
   void _onTouchPointerEnd(PointerEvent event) {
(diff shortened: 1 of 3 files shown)


rustdesk:
**Request changes — regression risk is higher than I would expect for this fix.**

I think the current implementation has **medium-high regression risk**.

The original issue is a stale tap/click race, but this PR now changes several parts of the touch gesture lifecycle at the same time:

* physical pointer sequence tracking;
* tap arbitration;
* long-press cache ownership;
* one-finger pan start behavior;
* asynchronous mouse down/up serialization;
* pan end/cancel behavior.

This is also not limited to the reported iOS case. The changed code is shared by touch/stylus input and desktop touch-mode paths.

There are already concrete async regressions in the new lifecycle code:

1. A delayed `onTapUp` can bind itself to a newer touch sequence and consume the newer tap-down state.
2. A pan that has already ended can still send a late mouse-down followed by mouse-up.
3. Pan updates received while `_startTouchModePan()` is awaiting cursor movement / mouse-down are dropped because `_touchModePanStarted` remains false.

Given that these failures are caused by the additional asynchronous state management itself, I would strongly prefer keeping this fix as small as possible.

### Please minimize the scope

The core requirement for #16025 is to prevent state from one physical touch from being reused by a stale callback and to reject a tap after the pan recognizer has claimed the same touch.

Please consider whether the full `_TouchModePanState` / `_startTouchModePan()` / `_finishTouchModePan()` lifecycle refactor is actually required to solve that problem.

If balanced asynchronous pan down/up handling is a separate problem, I would prefer it in a follow-up PR with dedicated lifecycle tests. Combining it with the stale-tap fix substantially increases the number of possible callback interleavings and makes this fix harder to reason about safely.

### Avoid unrelated gesture behavior changes

The new long-press sequence tracking also changes how the cached long-press position is selected as the start of a pan.

I don't see this behavior in the reported #16025 reproducer. If stale long-press state is a known part of the same bug, please add a regression test demonstrating it. Otherwise, I suggest leaving the existing long-press behavior unchanged and handling that independently.

Similarly, this looks unnecessary in `_onTouchPointerDown()`:

```dart
if (_touchModeGestureTracker.sequence != previousSequence) {
  _touchModePanStarted = false;
}
```

A raw pointer sequence transition should ideally not mutate the gesture recognizer's pan lifecycle state. Pan start/end/cancel should remain responsible for that state. Having both the pointer tracker and pan callbacks modify `_touchModePanStarted` creates another way for `_touchModePanStarted` and `_touchModePanState` to disagree.

Unless this reset is required for a specific reproduced case, I suggest removing it.

### Reduce duplicated state

There are currently several partially overlapping sources of truth:

* `_touchModePanStarted`
* `_touchModePanState`
* `TouchModeGestureTracker._panHandled`
* `touchSequence`
* `ended`
* `cancelled`
* `mouseDownSent`
* `finishing`

The more of these we add, the harder it becomes to guarantee correct behavior for delayed callbacks.

Please try to make one state object / ownership decision authoritative instead of adding guards around several independently mutable flags.

### Async regression tests are needed

The new `TouchModeGestureTracker` unit tests are useful, but they only test synchronous state transitions. They cannot catch the failures introduced around the `await`s in `RemoteInput`.

Before merging this, I think we need deterministic tests that can pause `cursorModel.move()` / mouse sending and exercise at least:

* touch A callback delayed, touch B starts;
* pan end while pan-start move is pending;
* pan cancel while pan-start is pending;
* pan update while pan-start is pending;
* a new physical touch while the previous pan is finishing;
* relative mouse mode, ensuring no unwanted down/up and no lost movement.

In particular, updates that arrive while pan start is pending should be coalesced/replayed or otherwise preserved rather than silently discarded.

### Suggested direction

I would prefer this PR to contain only the minimum state needed to:

1. associate tap state with the physical touch that created it;
2. synchronously mark the sequence as pan-owned before any pan `await`;
3. reject stale tap callbacks before and after asynchronous cursor movement;
4. invalidate the cached tap state when appropriate.

Anything beyond that should either have a demonstrated dependency on #16025 or be moved to a follow-up.

That would significantly reduce the regression surface while still addressing the reported stale-click issue.

shanglongy-lang:
Thanks for the detailed review. I narrowed the implementation in 40322add:

- removed `_TouchModePanState`, `_startTouchModePan()`, and `_finishTouchModePan()`;
- restored the existing long-press and pan start/update/end/cancel behavior;
- scoped the new physical-sequence tracking to iOS;
- associated tap-down state with its recorded sequence so a delayed callback does not clear a newer tap-down;
- marked pan ownership synchronously before the existing pan-start awaits;
- kept the sequence checks before and after `cursorModel.move()`;
- added a deterministic test that pauses tap work, starts a newer physical touch, and verifies that the older tap is rejected.

The final diff no longer introduces asynchronous pan lifecycle state. I also updated the PR description to match the reduced scope. Flutter is not installed in my local Windows preparation environment, so the new Flutter tests still need to run in CI after workflow approval.

rustdesk:
Thanks for narrowing the scope. This is much closer to what I had in mind.

The latest revision addresses the main concerns from my previous review:

* the `_TouchModePanState` / `_startTouchModePan()` / `_finishTouchModePan()` lifecycle refactor has been removed;
* the existing long-press behavior is preserved;
* raw pointer sequence changes no longer mutate `_touchModePanStarted`;
* the new tracking is effectively limited to the iOS stale-tap case;
* tap state is associated with the physical touch sequence that created it;
* pan ownership is claimed synchronously before the first `await`;
* stale tap callbacks are rejected both before and after `cursorModel.move()`;
* a delayed callback no longer clears or consumes the tap-down state belonging to a newer touch.

This substantially reduces the regression surface compared with the previous version, while still addressing the core race in #16025.

One remaining improvement would be a deterministic regression test that exercises the actual `RemoteInput.onTapUp()` async path: pause `cursorModel.move()`, start a newer physical touch, resume the old callback, and verify that no mouse click is sent. The current tracker-level test validates the intended state transition, but does not directly cover the integration across the `await`.

I would consider that additional test useful but not necessarily blocking now that the unrelated asynchronous pan lifecycle changes have been removed.

From a code-review perspective, the current direction looks reasonable to me.

fufesou:
Reviewed at `7fe759c165b4f333d09beebbf5a555abe0bb5f84`.

## Findings

1. **[P1] Drain the tap sequence left behind by a double tap** (`remote_input.dart:227`)

   Across a successful double tap, `TapGestureRecognizer` can emit two `onTapDown` callbacks but only one `onTapCancel`. Flutter 3.24.5's own `ink_well_test.dart:2055` asserts the sequence `onTapDown(A)`, `onTapDown(B)`, `onTapCancel()`, then `onDoubleTap()`. Lines 176-177 enqueue both A and B, but the new cancel handler removes only A. `onDoubleTap()` at lines 276-295 does not clear B.

   The next single tap C therefore leaves `[B, C]` in the queue. Its `onTapUp()` removes B at lines 185-189, `_takeTapDownDetails(B)` cannot take C's details, and `handleTrackedTap()` rejects B because C is now the current sequence. C remains queued, so each later single tap is one sequence behind and is rejected in the same way. A normal double-click can therefore leave iOS touch mode unable to single-click until a two-finger scale clears the queue or the widget is recreated.

   Clear the pending tap state when the double-tap recognizer wins, or bind terminal callbacks without assuming a one-to-one count with `onTapDown`. Add a regression test that completes a double tap and then sends two ordinary taps; both ordinary taps must reach `sendTap`.

fufesou:
@shanglongy-lang Hi, https://github.com/rustdesk/rustdesk/issues/16025 is a little hard to reproduce on my devices (iPad 26.5). Is it easy to reproduce for you? I have a short pan at A → release → immediately tap B.

For the findings in https://github.com/rustdesk/rustdesk/pull/16028#issuecomment-5519762166 

image

Then single-tap no longer works. Double-tap and pan are unaffected.

PR #16029, opened 2026-09-02
fix: support iOS soft keyboard input in VM consoles

### LINKED ISSUE #16027: [iOS][VMware] Software keyboard input is misread as scan codes in a Windows guest (12345 becomes nm,./)
### Bug Description

Keyboard mode: normal/default iOS software-keyboard path (Legacy Unicode).

When an iPhone controls a Windows 11 host, iOS software-keyboard text is correct
in applications running directly on the host. Inside a VMware Workstation
Windows guest:
- typing 12345 produces nm,./
- English letters may produce no output
- host Windows applications continue to receive correct text

ASCII/UTF-16 values for 1 through 5 are 0x31 through 0x35, and those values are
Windows Set-1 scan codes for N, M, comma, period and slash. This indicates that
VMware is consuming the Legacy KEYEVENTF_UNICODE wScan field as a physical scan
code.

### How to Reproduce

1. Run RustDesk 1.4.9 on an iPhone with iOS 26.
2. Connect to a Windows 11 host running RustDesk 1.4.9.
3. Open host Notepad and type: 12345 abc. It is correct.
4. Start a Windows 10 guest in VMware Workstation.
5. Open guest Notepad and type the same text with the iOS keyboard.
6. Observe nm,./ for the digits and missing/incorrect English letters.

### Expected Behavior

The VMware guest should receive the same characters as host applications.
Typing 12345 abc must result in 12345 abc.

OS 26 -> Windows 11 host -> VMware Workstation / Windows 10 guest

1.4.9 -> 1.4.9

### Screenshots

Please see the attached screenshot: typing 12345 in guest Notepad produces
nm,./ while host input is [normal

Image

### Additional Context

A verified local fix against 1.4.9 adds an explicit VM keyboard path:
- supported ASCII is converted to USB HID key down/up strokes;
- the iOS Rust handler forces Map mode only for these events;
- HID 0x1E..0x22 for 12345 becomes Windows scan codes 0x02..0x06;
- non-ASCII/IME input remains on the normal Unicode path;
- bubbled soft-keyboard KeyEvents do not also enter the physical-key handler;
- a VM toggle is exposed in the iOS keyboard helper toolbar.

The patched build works on the same iPhone, Windows host and VMware guest.

Related: #4357, #10172 and #13565. The exact 12345 -> nm,./ symptom is
different from the key-repeat reports.

Relevant files:
- flutter/lib/mobile/pages/remote_page.dart
- flutter/lib/models/input_model.dart
- flutter/lib/models/ios_vm_keyboard_utils.dart
- flutter/lib/common/widgets/remote_input.dart
- src/ui_session_interface.rs

I can port the fix to current master and open a signed-off PR after maintainers
confirm whether they prefer an explicit VM mode or automatic handling.

shanglongy-lang:
## Summary

- Add an opt-in `VM` toolbar mode for iOS sessions connected to Windows peers.
- Send supported ASCII soft-keyboard text as physical USB HID key events, using the existing keyboard dispatch entry point.
- Preserve toolbar Ctrl/Shift/Alt/Command modifiers without releasing modifiers already held by a physical keyboard.
- Validate the whole text before emitting any events; unsupported text uses the existing Unicode/IME fallback.

## Problem and scope

In the reported VMware guest, iOS software-keyboard input is misinterpreted: `12345` becomes `nm,./`, and letters may produce no input. VM mode sends the physical number-row usages `0x1e` through `0x22` instead of character values `0x31` through `0x35`.

This is a physical US-layout workaround: the guest must use a US-compatible keyboard layout. VM mode is disabled by default and does not change the session's saved keyboard mode. The shared keyboard-focus path is unchanged.

Modifier handling reuses the existing `toReleaseKeys` tracking for both left and right physical modifiers. Only modifiers synthesized for a VM stroke receive matching key-up events; a physically held modifier remains down until its physical release. Right-Alt tracking is corrected so that this check also covers that key.

## Tests

`flutter/test/vm_keyboard_utils_test.dart` covers digits, letters, editing keys, shifted symbols, all-or-nothing Unicode fallback, toolbar modifiers, and avoiding duplicate Shift events. It also covers each of Ctrl/Shift/Alt/Command held across VM input and a subsequent physical key, plus toolbar Shift while physical Ctrl remains held.

Validation on Flutter 3.24.5 / Dart 3.5.4: all 16 tests pass in a minimal Flutter package containing byte-identical copies of the utility and test file, and `flutter analyze --no-pub` reports no issues for that package. `git diff --check` also passes.

These are utility/event-sequence tests, not full `InputModel` dispatch or hardware integration tests. Device validation of this master-based branch remains necessary; the earlier user-reported success was for the local 1.4.9 patch.

Fixes #16027

Related: discussion #15954

### DIFF
--- a/flutter/lib/common/widgets/remote_input.dart
+++ b/flutter/lib/common/widgets/remote_input.dart
@@ -45,15 +45,8 @@ class RawKeyFocusScope extends StatelessWidget {
                 : null,
             onKeyEvent: useRawKeyEvents
                 ? null
-                : (FocusNode node, KeyEvent event) {
-                    if (isIOS &&
-                        inputModel.vmKeyboardMode &&
-                        inputModel.peerPlatform == kPeerPlatformWindows &&
-                        !node.hasPrimaryFocus) {
-                      return KeyEventResult.skipRemainingHandlers;
-                    }
-                    return inputModel.handleKeyEvent(event);
-                  },
+                : (FocusNode node, KeyEvent event) =>
+                    inputModel.handleKeyEvent(event),
             child: child));
   }
 }
--- a/flutter/lib/common/widgets/remote_input.dart
+++ b/flutter/lib/common/widgets/remote_input.dart
@@ -45,8 +45,15 @@ class RawKeyFocusScope extends StatelessWidget {
                 : null,
             onKeyEvent: useRawKeyEvents
                 ? null
-                : (FocusNode node, KeyEvent event) =>
-                    inputModel.handleKeyEvent(event),
+                : (FocusNode node, KeyEvent event) {
+                    if (isIOS &&
+                        inputModel.vmKeyboardMode &&
+                        inputModel.peerPlatform == kPeerPlatformWindows &&
+                        !node.hasPrimaryFocus) {
+                      return KeyEventResult.skipRemainingHandlers;
+                    }
+                    return inputModel.handleKeyEvent(event);
+                  },
             child: child));
   }
 }
--- a/flutter/lib/mobile/pages/remote_page.dart
+++ b/flutter/lib/mobile/pages/remote_page.dart
@@ -314,12 +314,16 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
 
     // Delete the different part in the old value.
     for (i = 0; i < subOldValue.length - common; ++i) {
-      inputModel.inputKey('VK_BACK');
+      if (!inputModel.inputVmKeyboardText('\b')) {
+        inputModel.inputKey('VK_BACK');
+      }
     }
 
     // Input the new string.
     if (newStr.length > 1) {
-      bind.sessionInputString(sessionId: sessionId, value: newStr);
+      if (!inputModel.inputVmKeyboardText(newStr)) {
+        bind.sessionInputString(sessionId: sessionId, value: newStr);
+      }
     } else {
       inputChar(newStr);
     }
@@ -380,6 +384,9 @@ class _RemotePageState extends State<RemotePage> with WidgetsBindingObserver {
 
   void inputChar(String char) {
     if (!inputModel.keyboardInputAllowed) {
+      return;
+    }
+    if (inputModel.inputVmKeyboardText(char)) {
       return;
     }
     if (char == '\n') {
@@ -1006,6 +1013,12 @@ class _KeyHelpToolsState extends State<KeyHelpTools> {
       }, active: inputModel.command),
     ];
     final keys = <Widget>[
+      if (isIOS && isWin)
+        wrap('VM', () {
+          setState(() {
+            inputModel.vmKeyboardMode = !inputModel.vmKeyboardMode;
+          });
+        }, active: inputModel.vmKeyboardMode),
       wrap(
           ' Fn ',
           () => setState(
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -1068,13 +1068,15 @@ class InputModel {
     }
 
     for (final stroke in strokes) {
-      if (stroke.shift) {
-        sendPhysicalKey(vmLeftShiftUsbHid, true);
-      }
-      sendPhysicalKey(stroke.usbHid, true);
-      sendPhysicalKey(stroke.usbHid, false);
-      if (stroke.shift) {
-        sendPhysicalKey(vmLeftShiftUsbHid, false);
+      final events = vmKeyEventsForStroke(
+        stroke,
+        ctrl: ctrl,
+        shift: shift,
+        alt: alt,
+        command: command,
+      );
+      for (final event in events) {
+        sendPhysicalKey(event.usbHid, event.down);
       }
     }
     return true;
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -18,6 +18,7 @@ import '../../models/platform_model.dart';
 import '../../models/state_model.dart';
 import 'input_modifier_utils.dart';
 import 'relative_mouse_model.dart';
+import 'vm_keyboard_utils.dart';
 import '../common.dart';
 import '../consts.dart';
 
@@ -407,6 +408,9 @@ class InputModel {
 
   final WeakReference<FFI> parent;
   String keyboardMode = '';
+  // Physical US-key input is opt-in because normal host input should keep its
+  // Unicode and IME behavior.
+  bool vmKeyboardMode = false;
 
   // keyboard
   var shift = false;
@@ -1038,6 +1042,42 @@ class InputModel {
         ctrl: ctrl,
         shift: shift,
         command: command);
+  }
+
+  /// Sends supported iOS soft-keyboard text as physical keys in VM mode.
+  /// Returns false when the caller should use the normal text-input path.
+  bool inputVmKeyboardText(String text) {
+    if (!vmKeyboardMode ||
+        !isIOS ||
+        peerPlatform != kPeerPlatformWindows ||
+        !keyboardPerm ||
+        isViewCamera) {
+      return false;
+    }
+
+    final strokes = vmKeyStrokesForText(text);
+    if (strokes == null) return false;
+
+    void sendPhysicalKey(int usbHid, bool down) {
+      bind.sessionHandleFlutterKeyEvent(
+          sessionId: sessionId,
+          character: 'flutter_key_map',
+          usbHid: usbHid,
+          lockModes: 0,
+          downOrUp: down);
+    }
+
+    for (final stroke in strokes) {
+      if (stroke.shift) {
+        sendPhysicalKey(vmLeftShiftUsbHid, true);
+      }
+      sendPhysicalKey(stroke.usbHid, true);
+      sendPhysicalKey(stroke.usbHid, false);
+      if (stroke.shift) {
+        sendPhysicalKey(vmLeftShiftUsbHid, false);
+      }
+    }
+    return true;
   }
 
   static Map<String, dynamic> getMouseEventMove() => {
--- a/flutter/lib/models/input_model.dart
+++ b/flutter/lib/models/input_model.dart
@@ -1044,7 +1044,8 @@ class InputModel {
         command: command);
   }
 
-  /// Sends supported iOS soft-keyboard text as physical keys in VM mode.
+  /// Sends supported iOS soft-keyboard text as physical US-layout keys in VM
+  /// mode. The guest keyboard layout must be US-compatible.
   /// Returns false when the caller should use the normal text-input path.
   bool inputVmKeyboardText(String text) {
     if (!vmKeyboardMode ||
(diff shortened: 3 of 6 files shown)


rustdesk:
<html><head></head><body><p><strong>Request changes</strong></p><p>The overall approach looks reasonable: the workaround is opt-in and narrowly scoped to iOS → Windows sessions with VM keyboard mode enabled, which helps limit the blast radius. However, this PR changes keyboard/focus/input-dispatch behavior in a stateful path, and I think the regression risk inside this mode is still medium-to-high.</p><p>I see two merge-blocking regressions, one compatibility concern, and a few places where the implementation could be narrowed further to reduce regression risk.</p><h3>1. Toolbar modifiers are lost in the new VM text path</h3><p><code inline="">inputVmKeyboardText()</code> bypasses the existing <code inline="">inputKey()</code> path, which is responsible for forwarding the current toolbar modifier state:</p><pre><code class="language-dart">alt: alt,
ctrl: ctrl,
shift: shift,
command: command
</code></pre><p>The new ASCII → HID path only synthesizes Shift when the character itself requires it.</p><p>As a result, with VM keyboard mode enabled:</p><ul><li><p>Ctrl + <code inline="">c</code> from the toolbar becomes plain <code inline="">c</code></p></li><li><p>Toolbar Shift + <code inline="">a</code> does not behave like Shift+A</p></li><li><p>Alt-based shortcuts are lost</p></li><li><p>Command/Win-based shortcuts are lost</p></li></ul><p>This looks like a real behavioral regression rather than an edge case.</p><p>The VM HID path should preserve the current modifier state. Character-required Shift should also be combined with the existing Shift state rather than being emitted independently twice.</p><p>Please add tests for at least:</p><ul><li><p>Ctrl+C</p></li><li><p>toolbar Shift + <code inline="">a</code></p></li><li><p>one Alt shortcut</p></li><li><p>one Command/Win shortcut</p></li></ul><h3>2. The focus guard is too broad and can suppress external hardware keyboard events</h3><p>The new condition:</p><pre><code class="language-dart">if (isIOS &amp;&amp;
    inputModel.vmKeyboardMode &amp;&amp;
    inputModel.peerPlatform == kPeerPlatformWindows &amp;&amp;
    !node.hasPrimaryFocus) {
  return KeyEventResult.skipRemainingHandlers;
}
</code></pre><p>does not distinguish duplicate soft-keyboard events from real hardware keyboard events.</p><p>When the hidden text field has primary focus, <code inline="">RawKeyFocusScope</code> is still an ancestor in the focus tree. Physical keyboard events can bubble through it, and returning <code inline="">skipRemainingHandlers</code> here can prevent them from reaching the normal RustDesk key handling path.</p><p>This can regress external iPad keyboard behavior, including:</p><ul><li><p>arrow/navigation keys</p></li><li><p>function keys</p></li><li><p>physical Ctrl/Alt/Shift/Command</p></li><li><p>shortcuts that are not representable as text</p></li><li><p>potentially mismatched key-down/key-up delivery during focus transitions</p></li></ul><p>The last case is particularly concerning because keyboard handling is stateful. If a key-down reaches the remote side but the corresponding key-up is suppressed after focus changes, the remote guest can end up with a stuck modifier/key state.</p><p>The suppression logic should be narrowed to the actual duplicate IME/soft-keyboard case instead of suppressing every bubbled key event whenever the text field owns primary focus.</p><p>Please add a widget/integration test for:</p><ul><li><p>VM mode enabled</p></li><li><p>mobile text field focused</p></li><li><p>external hardware key-down + key-up</p></li><li><p>modifier key-down + key-up</p></li><li><p>focus transition while a key is active</p></li></ul><h3>3. ASCII → HID mapping assumes a US physical keyboard layout</h3><p><code inline="">vm_keyboard_utils.dart</code> explicitly converts ASCII characters to physical US keyboard usages.</p><p>That is fine for a US-layout guest, but HID map-mode input represents physical key positions, not abstract characters. On AZERTY, German, or other layouts, the same HID usage may produce a different character.</p><p>For example, text input containing <code inline="">a</code> is converted to the physical US <code inline="">A</code> key position. A guest using another layout may interpret that position differently.</p><p>I would not necessarily block the PR on this alone if the intended scope is explicitly “VM workaround for US-compatible guest layouts”, but the assumption should be documented and ideally reflected in tests or UI expectations.</p><h3>4. Please minimize changes to shared keyboard/focus behavior</h3><p>Since this is a workaround for a narrow case, I would strongly prefer the implementation to avoid changing generic input behavior unless it is strictly necessary.</p><p>In particular, I would avoid introducing broad conditions into shared focus/key-event handlers when the desired behavior can be implemented inside the VM-specific text-input path.</p><p>A safer direction would be:</p><ul><li><p>keep the existing non-VM keyboard and focus paths unchanged;</p></li><li><p>branch into VM-specific behavior as late as possible, ideally only when converting text to HID events;</p></li><li><p>avoid changing generic <code inline="">RawKeyFocusScope</code> behavior to solve soft-keyboard duplication;</p></li><li><p>avoid adding new persistent input state unless required;</p></li><li><p>reuse existing key/modifier dispatch helpers where possible instead of creating a second partially independent keyboard state machine.</p></li></ul><p>This reduces the number of input paths that need to be reasoned about and makes it much easier to verify that users outside the target scenario are unaffected.</p><h3>5. Prefer reusing existing modifier/key dispatch logic over duplicating it</h3><p>The current implementation creates a new low-level HID emission path. That means modifier behavior, ordering, press/release balancing, and future keyboard fixes now risk being implemented differently in two places.</p><p>If possible, please factor the smallest reusable primitive out of the existing key dispatch path and use it from both normal input and VM text input.</p><p>For example, instead of making <code inline="">inputVmKeyboardText()</code> independently responsible for:</p><ul><li><p>modifier state</p></li><li><p>Shift synthesis</p></li><li><p>key-down ordering</p></li><li><p>key-up ordering</p></li><li><p>source selection</p></li></ul><p>it would be safer to reuse the existing modifier-aware key emission logic and only override the part that is actually VM-specific: the HID usage and <code inline="">"flutter_key_map"</code> source.</p><p>This would reduce divergence and lower the chance of future regressions when modifier handling changes elsewhere.</p><h3>6. Keep fallback behavior completely unchanged</h3><p>For any text that cannot be safely represented by the VM ASCII/HID mapping, the existing Unicode/IME path should remain exactly as it was before this PR.</p><p>I would avoid adding new preprocessing, filtering, focus handling, or special cases to the fallback path.</p><p>A useful regression test would verify that unsupported/non-ASCII input takes the exact pre-existing path and does not leave any synthetic modifier or key state behind.</p><h3>Regression risk</h3><p>I would classify the regression risk as <strong>medium-high overall</strong>, with an important distinction:</p>
Area | Risk
-- | --
Other platforms / normal keyboard mode | Low
iOS VM-mode basic text input | Medium
Toolbar modifiers / shortcuts | High
External iPad hardware keyboard | High
Non-US guest keyboard layouts | Medium-high

<p>The narrow feature gating reduces the global blast radius, but once this mode is active the PR touches core keyboard, focus, and event-dispatch behavior. Those paths are especially sensitive to modifier state and balanced key-down/key-up handling.</p><p>The safest implementation here is therefore not just one that fixes the reported VMware issue, but one that changes as little shared input behavior as possible.</p><h3>What looks good</h3><p>The Rust-side routing approach is clean: using a dedicated <code inline="">"flutter_key_map"</code> source to force only these synthesized events through map mode avoids changing the user's persistent keyboard mode.</p><p>The ASCII conversion is also easy to reason about, and falling back to the existing Unicode/IME path for unsupported text is the right behavior.</p><h3>Conclusion</h3><p>I would <strong>Request Changes</strong> before merging.</p><p>The main concept is good and appropriately scoped, but I think the following should be addressed first:</p><ol><li><p>Preserve toolbar modifier state in the VM HID text path.</p></li><li><p>Narrow or remove the generic focus-event suppression so external hardware keyboard events are not swallowed.</p></li><li><p>Keep VM-specific behavior isolated from the existing normal keyboard/focus paths.</p></li><li><p>Reuse existing modifier/key dispatch logic where possible instead of maintaining a second keyboard state machine.</p></li><li><p>Add regression coverage for modifiers, hardware key-down/key-up, focus transitions, and fallback behavior.</p></li></ol><p>The US-layout limitation should also be documented or otherwise made explicit.</p><p>Given how narrow the original bug is, I would prefer a smaller patch with fewer changes to shared keyboard/focus behavior, even if that means the VM-specific implementation is slightly more explicit.</p></body></html>

shanglongy-lang:
Thanks for the detailed review. I agree that the focus guard and the physical-layout assumption needed to be tightened.

Updated in `5910b02e`:

- Removed the VM-specific `RawKeyFocusScope` guard entirely. The shared hardware-keyboard and focus behavior now matches `master`, so VM mode cannot swallow hardware key-up or modifier events during focus transitions.
- Kept VM conversion only in the iOS soft-keyboard text callback.
- Reused the existing `newKeyboardMode` physical-key dispatch entry point. The `flutter_key_map` sentinel only forces Rust map mode for an explicit VM event.
- Preserved balanced Ctrl, Shift, Alt, and Command/Win modifiers. Ctrl/Shift coverage was already present; this update adds Alt and Command/Win tests.
- Documented that ASCII-to-HID conversion is a workaround for US-compatible guest keyboard layouts.
- Unsupported or mixed Unicode input is fully validated before any VM event is sent. Conversion returns `false` and the caller takes the exact pre-existing Unicode/IME path, with no synthetic modifier/key state created.

I did not add new focus integration behavior or tests because the risky focus behavior was removed rather than replaced; that file now has no diff from `master`. `git diff --check` passes. Flutter/Dart and Rust toolchains are not installed in this local preparation environment, so the automated tests still need to run in PR CI after workflow approval.

rustdesk:
Request changes

Most of the issues from my previous review are now addressed:

* the broad `RawKeyFocusScope` guard has been removed;
* Ctrl/Shift/Alt/Command toolbar modifiers are now included in balanced VM HID sequences;
* the VM path is isolated to the iOS soft-keyboard callback;
* unsupported text is validated before any synthetic event is emitted;
* the US-layout assumption is documented.

There is still one stateful modifier issue I think should be fixed before merging.

`inputVmKeyboardText()` treats `ctrl`, `shift`, `alt`, and `command` as modifiers that need to be synthesized around every VM stroke. However, these fields are not toolbar-only state: `handleKeyDownEventModifiers()` also sets them when an external hardware modifier is physically held.

For example:

1. Hold physical Ctrl on an external iPad keyboard. The normal hardware path sends Ctrl-down and sets `ctrl = true`.
2. While Ctrl is still physically held, enter a character through the software keyboard with VM mode enabled.
3. `inputVmKeyboardText()` sees `ctrl == true` and emits another Ctrl-down, the key stroke, and a synthetic Ctrl-up.
4. The remote side now sees Ctrl released even though the physical Ctrl key is still held locally.
5. A subsequent physical key pressed before the real Ctrl-up can therefore arrive without Ctrl active on the remote side.

The same issue applies to Shift, Alt, and Command.

Please avoid synthesizing/releasing a modifier that is already represented by a currently held physical modifier. Ideally toolbar modifier state and physical modifier state should be distinguished; alternatively, the existing tracked physical modifier state (`toReleaseKeys`) could be used so the VM text path only synthesizes toolbar-only modifiers.

I would also add a regression test covering:

physical modifier down → VM soft-key input → another physical key while the modifier remains held → physical modifier up

and verify that the VM soft-key sequence does not prematurely release the physical modifier.

The existing mapper-level Unicode fallback test is useful, although an InputModel/caller-level test proving that unsupported mixed text emits no VM HID events and takes the existing Unicode path would still be a good follow-up.

PR #16062, opened 2026-09-04
port forward shared conn

rustdesk:
## Problem

Port forwarding opens a fresh authenticated peer connection for every accepted local TCP connection. For HTTP-style short connections that is a full connect-and-login per request, which several issues have complained about.

## What this does

One peer connection per port-forward mapping, shared by every local connection accepted on it. Each accepted local connection becomes a credit-windowed channel multiplexed over the login-authenticated stream, via a new `PortForwardChannel` message family (`open` / `opened` / `data` / `close` / `window_update`, union tag 33).

The connection is established lazily, on the first accept, and lives as long as the mapping: removing the mapping or closing the window drops the listener's `Tunnel`, which ends the tunnel loop and with it the peer connection. Later accepts join it. The controlled side answers each `open` by connecting to the target and relaying, and re-checks `enable-tunnel` on every `open` rather than once per login.

**Feature off means the old code.** `listen()` keeps master's accept arm and `connect_and_login` as they were; the tunnel is a branch taken only when the window's flag is on, in `establish_tunnel` with its own `connect_and_login_mux`. A window with the setting off, a peer without the feature bit, or a listener latched legacy runs the raw pipe exactly as on master. `port_forward.rs` against master is purely additive.

## Authorization

A tunnel serves exactly the target its login authenticated. `PortForward.host` / `port` are part of the latched login scope and are what click-to-approve or 2FA is shown, so the controlled side refuses any `open` whose normalized target differs from the login's, with `opened{ success: false }`. A window with mappings to several targets uses one connection per target, each approved on its own: the approval unit stays one login per target, rather than one per local connection today.

**Authorization lifetime is the tunnel, that is the mapping, not each forwarded TCP connection.** ID whitelist, password, 2FA and click-to-approve run once, at the login that establishes the tunnel, as they do for any other long-lived session; `enable-tunnel` alone is re-checked on every `open`, so revoking it refuses new channels while live ones finish. Closing the session from the connection manager ends the tunnel and everything on it.

The tunnel's login is the raw pipe's login: master's `hash_arrived` and `login_from_ui`, which fill the window's shared `port_forward` and `hash` and send under the per-window turn lock (#16069). The one thing the tunnel adds to it is the `multiplex` bit, read by `create_login_msg` from a window flag beside `port_forward` in the handler. The flag is set once in `io_loop`, before the window's mappings start, so an accept's claim on the tunnel and its login always agree.

## Compatibility

Old peers are unaffected, and detection is synchronous at `PeerInfo` via a new `Features.port_forward_mux` bit, so no round trip is wasted:

| controller | controlled | behaviour |
|---|---|---|
| new | new | multiplexed, one connection per mapping |
| new | old | feature bit absent, this accept falls back to today's raw pipe and the mapping stays legacy until it is re-added: later accepts log in individually and ask for no tunnel |
| old | new | no `multiplex` flag, controlled side takes today's path untouched |
| new, setting off | any | controller sends no `multiplex`, both sides run today's code |

## The off switch

`enable-port-forward-mux`, a `KEYS_LOCAL_SETTINGS` entry read on the controlling side only. The `enable-` prefix makes an unset value read as on, so the tunnel is the default. Turning it off needs no agreement with the peer: no flag is sent, so both sides take the legacy path. It is read once per window, when the window opens, so a change applies to windows opened after it; a window keeps its mode for its life, as it keeps its tunnels.

## Flow control

Per-channel credit windows, charged identically by sender and receiver:

| constant | value | on the wire |
|---|---|---|
| `INITIAL_WINDOW` | 64 KiB | yes, fixed |
| `MIN_FRAME_CHARGE` | 64 B | yes, fixed |
| `CHANNEL_WINDOW` | 256 KiB | advertised |
| `MAX_SEND_CREDIT` | 256 KiB | local |
| `MAX_FRAME` | 64 KiB | local |
| `MAX_CHANNELS` | 256 | local, both sides |

Two of these exist for memory safety rather than throughput. Every frame costs `max(len, MIN_FRAME_CHARGE)` of window, because a 1-byte frame costs its sender one byte and its receiver a queue entry. And a sender never holds more than `MAX_SEND_CREDIT`, whatever the peer advertises, because the controlled side's frame sink is the connection's unbounded sender and credit is therefore the only bound on what it buffers.

Data may be pipelined behind `open` before `opened` arrives, bounded by `INITIAL_WINDOW` on both sides: the controlled side's receive window starts there and is widened to `CHANNEL_WINDOW` just before `opened` goes out. A frame past the receiver's remaining window drops that channel at once, before its task has run, so nothing more can be queued for it; the tunnel stays up.

## Errors stay visible

A refused channel still raises a dialog. The controlled side answers an unreachable target, a revoked `enable-tunnel`, a target other than the login's, or the channel cap with `opened{ success: false, message }`, and the controller shows that message through the existing `Error` title key, so no new localization key is added.

Deduplicated by reason and time-keyed: a reason is shown once, then not again until it has been quiet for 10 s. One page load can have a dozen connections refused for the same reason and produces one dialog, not a dozen; a client that keeps retrying keeps it quiet; a target that breaks again hours into a long-lived mapping is reported again. At most eight distinct reasons are remembered at once, since the text arrives from the peer.

The refusal text itself is the raw pipe's wording, `RDP` substitution included, so one problem reads the same whichever path the peer takes.

## Connection manager

One row per mapping rather than one per forwarded connection, showing the mapping's target. Two consequences worth knowing: the row lives as long as the mapping, and click-to-approve or 2FA happens once per mapping rather than once per forwarded connection.

## Testing

`cargo test --lib` passes 136 tests, 48 of them for this feature, including end-to-end tests that run the real controller tunnel against the real controlled-side channel table over a loopback stream. The ones that matter:

- a channel opened during a 4 MiB transfer is served within 2 s, which is the head-of-line blocking this feature exists to remove
- every `open` reaches the peer ahead of its own channel's first `data`, across twenty concurrent channels
- a channel's tail arrives before its `close`, in both directions
- 5000 one-byte frames do not trip the window
- three times `MAX_CHANNELS` sequential connections all succeed on one tunnel
- a peer advertising a 4 GiB window gets `MAX_SEND_CREDIT` and no more
- dropping the listener's `Tunnel` ends the peer connection: the fake peer reads EOF within 2 s
- an `open` to a target other than the login's is refused, and the tunnel stays up for the login's target
- the demultiplexer admits only `INITIAL_WINDOW` on a channel before `opened`
- a frame past the window drops its channel on the spot, on both sides, so a peer that keeps sending queues nothing more
- a window with the flag on asks for the tunnel in its login, one without does not
- `enable-tunnel` revoked mid-session refuses the next `open` while the live channel keeps relaying
- a dead tunnel ends both of its channels together, and the next accept establishes again on the same mapping
- three refused channels, two for one reason and one for another, produce two dialogs
- a reason that keeps arriving stays quiet; one that stopped for 10 s is reported again, and the cap on remembered reasons cannot silence the window for good

**Manual two-machine verification has not been run.** What it should cover: the four compatibility combinations above; a bulk download with small requests interleaved on the same mapping; the RDP button and `--port-forward` on the command line; turning off *Enable TCP tunneling* on the controlled side mid-session (new requests refused, a streaming download continues); and removing one mapping while another keeps serving, where the removed mapping's row should leave the CM. The last of these is the only coverage for `listen()`'s accept arm, which has no automated test.

## Known limitations

- **TCP half-close.** A client that shuts down its write side and then waits for the response has its channel closed. Today's `run_forward` drops both directions on either EOF too, so this is not a regression; a direction flag on the close frame is the cheap fix if it is wanted.
- **Both tunnel loops read and write in one task**, so sustained bulk traffic in both directions at once can park each side in `send` until the controlled side's send timeout drops the tunnel. Today's raw pipe has the same shape but loses one connection rather than one mapping. Fixing it needs a read/write split that `hbb_common::Stream` does not offer today.
- **Refused-channel reasons are not localized.** `Too many port forward channels` and `Port forward target not authorized` have no lang key; their siblings on the same path, `No permission of IP tunneling` and `Failed to access remote …`, are raw English on the legacy path too. One decision for all of them, separately.

## Pre-existing, found by review

`Connection::start` chose the send timeout before the login request had arrived, so `SEND_TIMEOUT_OTHER` never applied. That went in separately as #16063, already on master; the classification there keys on the login request's union, which covers multiplexed logins too, and this branch no longer touches that line.

Two mappings of one window also raced on the login's target and challenge in the shared `LoginConfigHandler`. That went in separately as #16069, already on master; this branch is rebased on it and the tunnel's login goes through the same functions, so it is covered by the same lock.

## Regression surface

Existing files and runtime paths whose behaviour changes, and why each is unavoidable:

| file | what changes | why |
|---|---|---|
| `src/port_forward.rs` | `listen()`'s accept arm gains the tunnel branch ahead of the unchanged raw path; new functions after it; no existing line changes | the feature's controller entry point |
| `src/client.rs` | `LoginConfigHandler` gains `port_forward_mux` beside `port_forward`; `create_login_msg` sets `multiplex` from it | the login message is built here and nowhere else; no signature changes, no caller changes |
| `src/ui_session_interface.rs` | one line in `io_loop` sets the window flag from the setting | the window's mappings are started here |
| `src/server/connection.rs` | a multiplexed login stays in the protobuf loop and dispatches `PortForwardChannel` to the new module; `login_scope_digest` latches `multiplex`; `is_port_forward` counts a mux session | the controlled side's entry point; thin hooks around the new module |
| `src/lib.rs`, `src/server.rs` | module declarations | hooks |
| `libs/hbb_common` | `05ed68f` → `840c8ec`: the proto family and the setting's key, two files | protocol plumbing |

The raw pipe's own routines, `connect_and_login`, `login_with_hash`, `hash_arrived`, `login_from_ui` and `run_forward`, are byte-identical to master.

## Dependency

The proto family and the setting's key are in rustdesk/hbb_common#594, already merged; the submodule pointer here is hbb_common `main` at that merge.

https://claude.ai/code/session_01EZ49AbZJYfm8NTp5yDPMab

This PR adds multiplexed port forwarding so connections belonging to one mapping can share an authenticated peer connection. It introduces per-channel framing and flow control, negotiates support through the login protocol, preserves a legacy fallback, adds a controlling-side setting, and updates localized UI text.
- Adds controller and controlled-side port-forward channel multiplexers.
- Adds channel credit accounting, packet-size limits, lifecycle handling, and target authorization.
- Adds feature negotiation and a local opt-out setting.
- Includes subsequent terminal clipboard and terminal input changes now present in the reviewed changeset.
</details>

The PR does not appear safe to merge yet because one blocking tunnel-ownership finding and explicit repository-rule violations from previous threads remain outstanding.

The mapping listener still constructs its own `Tunnel`, so the previous requirement to share tunnel ownership across the window remains unresolved. Existing ignored-send paths also remain in the multiplexing implementation, and `src/lang/ur.rs` still removes existing non-empty translations such as `License`, `Please install plugins`, and several peer/plugin actions, contrary to the localization requirement. The malformed-message diagnostics and tunnel-teardown race have been addressed, and the Traditional Chinese translation finding no longer appears in the current diff.

**Files Needing Attention:** src/port_forward.rs, src/port_forward_mux.rs, src/server/port_forward_mux.rs, src/lang/ur.rs
</details>

| Filename | Overview |
|----------|----------|
| src/port_forward.rs | Adds mapping-owned tunnel establishment and multiplexed channel dispatch while retaining the legacy fallback; a previous finding about tunnel ownership remains unresolved. |
| src/port_forward_mux.rs | Implements controller-side channel multiplexing, flow control, packet limits, and teardown; a previous silent-send-error finding remains relevant. |
| src/server/port_forward_mux.rs | Implements controlled-side channel admission, target connection, and relay; a previous ignored-control-send-error finding remains relevant. |
| src/server/connection.rs | Negotiates multiplexed port forwarding, dispatches channel messages, and applies the post-login packet cap. |
| src/client.rs | Adds per-login multiplex state to port-forward login construction. |
| src/lang/ur.rs | Adds feature translations but still omits existing Urdu entries identified in the previous review. |
| flutter/lib/models/rustdesk_terminal.dart | Adds permission-gated OSC 52 clipboard-write handling with validation and payload limits. |
| flutter/lib/models/terminal_mouse_handler.dart | Extends terminal pointer handling for touch reporting and web clipboard gesture preparation. |

</details>

```mermaid
sequenceDiagram
    participant Local as Local TCP client
    participant Controller as Controller mapping
    participant Peer as Shared authenticated tunnel
    participant Target as Controlled-side target
    Local->>Controller: Accept TCP connection
    Controller->>Peer: open(channel, authorized target)
    Peer->>Target: Connect target
    Target-->>Peer: Connection result
    Peer-->>Controller: opened(channel, window)
    loop Credit-windowed relay
        Controller->>Peer: data(channel)
        Peer->>Target: TCP bytes
        Target-->>Peer: TCP bytes
        Peer-->>Controller: data(channel)
        Controller-->>Peer: window_update(channel)
        Peer-->>Controller: window_update(channel)
    end
    Controller->>Peer: close(channel)
```
</details>

<sub>Reviews (18): Last reviewed commit: ["port forward: a channel id still live wh..."](https://github.com/rustdesk/rustdesk/commit/c2927046f80ec7b360663f6f41c81e6974f0cc4f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=60319948)</sub>


rustdesk:
[rustdesk-1.5.0-x86_64.exe.zip](https://github.com/user-attachments/files/31858560/rustdesk-1.5.0-x86_64.exe.zip)

Issue #10964, opened 2025-02-28
Sound is distorted from remote machine


patryk-siewiera:
### Bug Description

Sound from Rustdesk is distorted.

### How to Reproduce

Video demo: 
https://www.youtube.com/watch?v=JXGs8yhjn6c

**Short Version:**  
1. Play a sine signal on a remote computer.  
2. Record the signal from machine 2 via OBS.  
3. Analyze the recording using a spectrum analyzer.  

**Detailed Steps:**  

**On Computer 1:**  
1. Launch Reaper (free for 30 days).  
2. Create a new track.  
3. Add a "Tone Generator" to the track.  
4. Insert a ReaEQ plugin.  
5. Observe the frequency response in ReaEQ. You should see a single prominent peak. Any additional peaks indicate distortion.  
6. Record what you receive via OBS on computer 2.

**On Computer 2:**  
1. Record the entire process using OBS (Open Broadcaster Software).  
2. Import the recording into Reaper.  
3. Create a new track.  
4. Add a ReaEQ plugin to the track.  
5. Analyze the frequency graph to identify any distortion.  

### Expected Behavior

Please look at the spectrum graphs, there should be no peaks above the main one.
If there are - this is distortion. 
This makes rustdesk impossible to remotely edit audio. 

### Operating system(s) on local (controlling) side and remote (controlled) side

macOS 15.3 /  win 11 24H2

### RustDesk Version(s) on local (controlling) side and remote (controlled) side

1.3.8

### Screenshots

expected result (source on remote machine):
![Image](https://github.com/user-attachments/assets/71d4a33e-ea1f-483e-9f45-6aaef47e9642)

actual result (what we received via rustdesk):
![Image](https://github.com/user-attachments/assets/805c15d9-d02e-44ab-8a21-3c7032c8cf22)

### Additional Context

This issue is unlikely related to the codec. Even when using a low-bitrate MP3, such significant distortion shouldn't occur.

Distortion occurs at every level of the test signal, from -50 to 0. 

I can do testing if a build is provided, since I work as QA in the audio industry.

Issue #8375, opened 2024-06-16
Some noise is heard on the local side


izilzty:
### Bug Description

Hi, if the default format of the local side sound card is greater than 48000 Hz, the local side will hear some noise when the remote side plays audio.

### How to Reproduce

Set the local side sound card default format to be greater than 48000 Hz, and keep the remote side sound card to default (less than or equal to 48000 Hz).
![local side format](https://github.com/rustdesk/rustdesk/assets/18079946/695922a0-c76b-47ac-93a6-6754f8ffe285)

### Expected Behavior

Clear sound without noise

### Operating system(s) on local side and remote side

Windows 11 23H2 -> Windows 10 LTSC 2021

### RustDesk Version(s) on local side and remote side

1.2.6 -> 1.2.6

### Screenshots

[local side noise.zip](https://github.com/user-attachments/files/15858786/local.side.noise.zip)

rustdesk:
@fufesou

rustdesk:
Could you help test 1.2.5? I wanna confirm if it is related to a latest code change.

fufesou:
@izilzty Hi, do you use voice calls? Oddly enough, the speaker properties affect how playing the sound.

izilzty:
I tested 1.2.5->1.2.5 and 1.2.3-2->1.2.3-2, both have the same problem on my computer, and I haven't tried voice calls yet.

izilzty:
I'm guessing this might be an issue on the local side of the Windows version, because if I keep the current configuration and use the iOS app to connect to both computers to play audio, they both sound clear.

FiveKnive:
I have the same issue!

rustdesk:
How and when you can hear the noise?

FiveKnive:
Always when I play some sort of audio it sounds like it is distorted with this noise. You can hear it a little bit in the linked zip file of this issue. Happens when I connect from my Windows PC to another Windows PC or to Linux.

rustdesk:
> Set the local side sound card default format to be greater than 48000 Hz, and keep the remote side sound card to default (less than or equal to 48000 Hz).

How about your local side default settings?

FiveKnive:
![image](https://github.com/user-attachments/assets/f2a65aa2-a3ca-4b92-abff-664361d4540a)

Ok if I set it to 48000Hz on my client pc it works. But now I have problems with my other audio programs and can not have the full potential of my DAC :(
[...]

Issue #7125, opened 2024-02-13
macOS Host - Android Client - Audio garbled after first connection, Android app software error and force close (sometimes)


larryqiann:
### Bug Description

When a macOS host is set to stream audio from a device and a session is started, the audio is played normally on the receiving Android phone. After about ten seconds, the Android app will either crash and force close, or the connection will fail with a "software error".

When the connection is recreated, the audio has extreme distortion and noise (sounds like electrical arc welding), and the session will close or there will be a software error on the Android side within a 20 seconds.

This problem is only experienced when there is audio playing on the macOS host device.

Restarting the Rust Desk remote (server) side will reset the audio for the next session.

### How to Reproduce

1) Install latest (as of filing) nightlies on both Android phone and the macOS device
2) Set up audio device for sound on host macOS device
3) Make a connection to the Android device and play audio on the macOS device
4) Observe the sound operating normally
5) Disconnect the session
6) Reestablish the session
7) Play audio on the remote macOS device
8) Observe error
9) Restart the macOS Rustdesk software and reattempt the test, observe that the audio is normal for one session again

### Expected Behavior

When the client is connected to the server, both the first and subsequent times, the connection will be stable and audio will be transferred without noise or distortion in a stable way

### Operating system(s) on local side and remote side

Android -> macOS

### RustDesk Version(s) on local side and remote side

1.2.4 1.2.4

### Screenshots

![CleanShot 2024-02-13 at 13 02 26](https://github.com/rustdesk/rustdesk/assets/25872705/7e26ca10-01ad-45ec-b1c5-7e5abc52bab0)

rustdesk:
@21pages follow up, I can hear the sound also.

rustdesk:
Not sure if it is related to my upgrade to latest osx, I do not see this before, only see this after upgrade to latest 14.3.1

larryqiann:
> Not sure if it is related to my upgrade to latest osx, I do not see this before, only see this after upgrade to latest osx 14.3.1

The last time I tested this was in Ventura, last year in around June-July. I don't remember if I opened an issue about it, but the problem is the same. 

It sounds like to me that there is some lag in getting the samples all played out. During the garbled sound, the latency slowly gets longer and longer, and there is at least several seconds of garbled audio for every second of input audio.

larryqiann:
[...]

Issue #3151, opened 2023-02-10
[Bug]  audio call clips during linux-windows


test2a:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Bug Description

i tried new nightly version on linux host and windows client. the delay kept hovering between 5-25ms  and there was clipping of audio.

without audio calls, delay was never more than 0ms so this is the first time i am seeing that.

i am also reporting very less audio on windows client side but i will investigate that fully before writing a new issue for that

### How to Reproduce

audio call.

i played music on one machine and listened to it on another

### Expected Behavior

audio should work 

### Operating system(s) on local side and remote side

linux>windows

### RustDesk Version(s) on local side and remote side

1.20>1.20 nightly both sides fresh install

rustdesk:
@Kingtous

rustdesk:
@test2a video with sound please if possible.

Kingtous:
Thanks for the feedback. So the problem is the audio clipping, which means we can't listen to the sound fluently right?

rustdesk:
Close as no response.

PR #16095, opened 2026-09-06
Fix/audio stream continuity

fufesou:
https://github.com/rustdesk/rustdesk/issues/8375
https://github.com/rustdesk/rustdesk/issues/10964
https://github.com/rustdesk/rustdesk/discussions/11009
https://github.com/rustdesk/rustdesk/issues/534#issuecomment-3408363188
https://github.com/rustdesk/rustdesk/issues/9169
https://github.com/rustdesk/rustdesk/issues/7125
https://github.com/rustdesk/rustdesk/discussions/11814
https://github.com/rustdesk/rustdesk/issues/3151
https://github.com/rustdesk/rustdesk/issues/1703

Audio resampling was restarted for every captured or decoded block. On the tested Windows path, the capture device ran at 44.1 kHz, Opus at 24 kHz, and playback at 48 kHz. Resetting interpolation at each short packet introduced an audible discontinuity at the packet boundary.

The playback callback also slept when PCM arrived late, allocated a temporary vector on every callback, and switched directly to zero when the buffer was still short. That can miss the output deadline and turn a small delivery delay into a click.

Receiver buffer maintenance can intentionally discard old PCM after a backlog. The old output callback joined the samples before and after that time jump directly. When both sides of the join contained active audio at different phases, the result was another short click.

The optional `use_samplerate` build had a separate problem: capture conversion still required a DASP trait, so the backend could not be selected on its own.

The CPAL capture callback also encoded Opus, built the protocol message, and sent it to the service.  The warmed-up native-rate path still allocated three times per packet, and any encoder or service delay ran on the real-time callback thread. This was a real-time safety issue, although the endpoint tests did not establish it as the cause of the remaining audible interruptions.

The capture callback makes one `try_lock()` attempt. Contention drops the current packet even when buffers are available; pool exhaustion retains the drop-oldest policy. The worker log now reports `contention_dropped` separately, as part of the existing dropped total.

Gap smoothing reduces the transition but cannot restore discarded audio. The frequency and audible impact of contention drops under CPU pressure remain unmeasured. A handoff redesign is outside this PR’s scope.

## Changes

- Keep sender and receiver resampler state for the lifetime of the audio stream.
- Buffer resampled capture output until a complete 10 ms Opus frame is available.
- Remove sleep and per-callback allocation from the playback callback.
- Release the playback-buffer lock before recovery and sample conversion.
- Fade into and out of missing audio over 5 ms instead of making an abrupt jump.
- Mark periodic and capacity-driven PCM discards and apply the same 5 ms transition to the active-to-active join without changing the discard policy.
- Use CPAL sample conversion in the capture path instead of a DASP-specific bound.
- Use the stateful libsamplerate API when `use_samplerate` is selected.
- Reject `use_dasp,use_samplerate` with a clear compile-time error.
- Pass captured PCM through a preallocated, bounded queue, then encode and send it from a dedicated worker instead of the CPAL callback.
- Discard the oldest queued capture packet on saturation and report aggregated handoff losses from the worker thread.
- Stop the CPAL stream before joining the encoder worker during capture teardown.
- Log decode, resampling, and Opus encoding failures instead of dropping them silently.
- Clear incompatible playback state when a replacement stream cannot be started.

## Reproduction

On the old code:

1. Set the controlled Windows machine's effective playback path to 44,100 Hz. When using a virtual cable, set both the endpoint format and the cable's internal sample rate.
2. Keep the controlling machine's output device at 48,000 Hz.
3. Connect with RustDesk and play a 44.1 kHz stereo PCM16 WAV containing one second of silence, ten seconds of a 997 Hz tone at -6 dBFS, and one second of silence. The diagnostic fixture was named `continuous-997hz-44100-stereo.wav` and used 10 ms fades around the tone.
4. Listen or record the controlling machine's received output.

The sender chooses 24 kHz Opus for a 44.1 kHz capture device, so this exercises 44.1→24→48 kHz conversion. In the diagnostic A/B run, the decoded packet-boundary residual was about `0.0367`, while the old receiver output reached `0.5175`. A native 48 kHz control stayed near `0.0088`.

Repeating the same test with a diagnostic build containing this PR's production logic reduced the 24→48 kHz receiver output residual to about `0.0183`, with no decode errors, callback underflows, callback sleeps, or zero fills in the measured steady intervals.

Forced packet loss also reproduced sharp output transitions in the old callback. Consecutive intervals recorded 64/75 callback sleeps and 20/25 zero-fill events. With this PR's production logic, the same test recorded 21/24 underflows and zero-fill events while callback sleep count and requested sleep time remained zero. The diagnostic drop injector is not included in this PR.

Buffer correction was reproduced separately with native 48 kHz audio so resampling was not part of the path. A 117-second sustained 997 Hz tone was played through a forced relay while the remote source and received output were recorded independently. On the old code, three active-audio corrections were followed by maximum adjacent-sample steps of `0.0837`, `0.8903`, and `0.6733`.  The clean source maximum was about `0.0654`.

With the discard-generation fix, four corrections produced steps of `0.06656`, `0.06671`, `0.06656`, and `0.06650`. Their matching source windows were `0.06543`, `0.06540`, `0.06543`, and `0.06540`. The correction windows had no resampling, decode error, underflow, or output-callback gap. A run without a logged correction is inconclusive because the backlog is workload-dependent.

## Tests

- [x] Windows -> Windows, macOS -> Windows, Windows -> macOS, iPad -> Windows, Android -> Windows, Windows -> Android, Windows -> Linux, Linux -> Windows
- [x] new -> new
- [x] new -> older
- [x] older -> new

## Deferred and unresolved issues

This PR does not include speculative changes for the following cases:

- **Zero-gate startup (unresolved):** The sender stops transmitting after prolonged exact-zero input, but two controlled gate-close and reopen cycles resumed packet delivery.  Leading-edge loss was not reproduced, and changing the gate without that evidence could increase idle bandwidth without fixing the reported symptom.
- **Delivery stalls (separate investigation and PR):** Native 48 kHz recordings captured a 6 ms interruption over a forced relay and a 4 ms interruption over a direct connection. Both aligned with packet gaps and receiver underflow, while independent source and local-only recordings were clean. After moving encoding and sending off the capture callback, two 117-second relay runs reported no capture-handoff loss, but a time-aligned recording still showed receiver buffer skips and extra or split tone regions. This rules out handoff saturation in those runs, not sender-worker scheduling, network delivery, or receiver scheduling. The underrun transition in this PR limits the output edge but cannot reconstruct missing frames.  Stage-specific timing is required before selecting a fix. A larger jitter buffer is not justified until the stalled stage and latency trade-off are measured.
- **Fixed output buffer (unresolved):** `Fixed(64)` has not been isolated as a cause and remains unchanged. Changing it may alter latency and callback scheduling across audio backends, so it needs cross-device evidence before a production change.
- **Sender rates and Opus (not supported as causes):** Valid 96→48 and 192→48 kHz sender tests had no elevated boundary residual or resampling error, and the confirmed receiver-resampling runs had no Opus decode error. These configurations remain regression cases rather than reasons for an unverified production change.
- **Rubato backend (confirmed pre-existing build defect, separate PR):** The existing `use_rubato` feature does not compile against its locked Rubato API: the `SincFixedIn` constructor call has the wrong signature, and the code then calls `process` on its `Result`. It also provides an unused one-shot helper rather than the live resampler interface used by the sender and receiver. This PR does not change that backend; updating its API usage and defining its streaming integration need separate build and continuity tests.

The diagnostic drop and Opus-rate overrides, periodic counters, and zero-gate logs remain on the diagnostic branch. They are test instrumentation rather than product behavior and would need a separate observability review before being proposed for production.

### Linux core dump

Only once

The reconstructed failure path was:

```text
Rust background thread
  -> registered Dart_PostCObject callback
  -> Dart zone allocation
  -> mmap(length = 0)
  -> EINVAL (22)
  -> Dart fatal error
  -> SIGABRT / core dump
```

In the matching Dart 3.5.4 source, `VirtualMemory::Cleanup()` sets `page_size_` to zero. Together with the shutdown timing and callback stack, this supports a callback outliving the VM's usable lifetime. The observed allocation failure is explained by the invalid length; it is not evidence of memory exhaustion. [Dart 3.5.4: virtual_memory_posix.cc](https://raw.githubusercontent.com/dart-lang/sdk/3.5.4/runtime/vm/virtual_memory_posix.cc)

#### Relationship to `window_destroyed`

The accompanying `PlatformException(window_destroyed, ...)` occurs when saving window geometry accesses a native window that has already been destroyed. `_saveFrame()` catches and prints this exception in [tabbar_widget.dart:408](flutter/lib/desktop/widgets/tabbar_widget.dart#L408), with the reported message at line 417.

That caught Dart exception does not establish the cause of the native abort. Both errors occur around window shutdown, but suppressing the window-position exception alone would not address the Rust-to-Dart callback race. The relevant lifetime requirement is to stop and synchronize native callbacks before tearing down the Dart VM.

### TODOs

- [ ] -> new
  - macOS -> Linux. One time, no sound, but it hasn't happened again.
  -  -> Windows. Audio sometimes fails on some connections and requires a reconnect. But I couldn't reproduce it.
- [ ] When calling macOS from Android or Windows, an echo is heard.
- [ ] Flutter web audio

## Audio Quality
- Improved resampling across different input and output sample rates.
- Improved continuity across packet and buffer boundaries.
- Reduced audible clicks during playback interruptions, underruns, and capture gaps.

## Audio Capture and Playback
- Improved handling of audio channels, frame sizes, and streaming packets.
- Improved playback recovery when audio data is temporarily unavailable.
- Added safer handling and reporting of invalid or incomplete audio data.
- Improved capture buffering and encoder handoff reliability.

## Reliability
- Audio encoding failures now generate diagnostic warnings.
- Improved recovery when audio playback cannot start.
- Prevented conflicting audio processing configurations.

<h2><a href="https://app.greptile.com/api/retrigger?id=60944663"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; no actionable new defect or outstanding blocking finding remains.

- Preserves streaming resampler state and emits complete Opus-sized capture packets.
- Moves encoding and service transmission off the real-time capture callback through a bounded, reusable queue.
- Smooths playback underruns and explicit buffer discontinuities without sleeping or allocating in the output callback.
- Adds capture restart handling, processing-error reporting, and cleanup ordering.
- Adds focused continuity, allocation, queue, channel-conversion, and failure-state tests.
- Adds a stateful libsamplerate backend and rejects conflicting resampler features.
- Regression surface includes non-Linux CPAL capture and playback, shared audio encoding, optional resampler feature selection, and playback-state replacement; each path is directly involved in audio continuity or callback safety.
</details>

```mermaid
flowchart LR
  A[CPAL capture callback] --> B[Frame and resample PCM]
  B --> C[Bounded capture queue]
  C --> D[Encoder worker]
  D --> E[Opus audio message]
  E --> F[Client decoder]
  F --> G[Stateful playback resampler]
  G --> H[PCM ring buffer]
  H --> I[Non-blocking playback callback]
  I --> J[Discontinuity and underrun smoothing]
```
</details>

<sub>Reviews (5) · Last reviewed commit: ["Merge branch &#39;master&#39; into fix/audio-str..."](https://github.com/rustdesk/rustdesk/commit/de53c883e07246c00d0792b053062741b0716075)</sub>


fufesou:
@coderabbitai review

fufesou:
@coderabbitai full review

fufesou:
@coderabbitai review

fufesou:
@codex review

fufesou:
@coderabbitai full review

fufesou:
@codex review

fufesou:
@coderabbitai review

fufesou:
@codex resume

fufesou:
@codex review

rustdesk:
The resampling/continuity changes look good to me. In particular, keeping resampler state across packet boundaries addresses the resampling discontinuity we were concerned about, and the playback callback cleanup is a solid improvement.

Before merging, I think we should also finish removing heap allocation from the **capture realtime callback path** in this PR.

The old capture path already allocated, so this is not a regression introduced here. However, this PR substantially redesigns the capture/resampling pipeline, while the new realtime path still allocates through APIs such as `AudioResampler::process() -> Vec<f32>`, `FixedFrameAudioResampler::process() -> Vec<Vec<f32>>`, packet `collect()`, and owned channel conversion.

Since `CaptureFrameProcessor::process()` is called directly from the CPAL capture callback, I would prefer to make this pipeline use reusable/preallocated buffers now rather than bake allocation-oriented APIs into the new design and revisit the same pipeline later.

The remaining issues can be handled as follow-ups.

The most important follow-up is long-running audio loss, for example #15827. There is still a concrete Windows failure path: the CPAL/WASAPI capture and playback error callbacks only log an error, while the WASAPI worker can terminate after a runtime stream error. RustDesk may then retain apparently valid stream state even though the actual audio worker is gone, leaving audio permanently silent until the session is reconnected.

Please follow up with explicit stream liveness/recovery for both capture and playback. The CPAL error callback should only signal the failure; a non-realtime owner should then drop the dead stream, clear/reinitialize the relevant audio state, re-query the device, and rebuild/restart the stream.

One additional point about the current zero gate: I agree that we do not have enough evidence to remove it as part of this crackling fix, especially now that playback underrun/restart edges are smoothed. However, the current silence detection is based on **exact floating-point zero**:

`sample != 0.0`

That is not a robust definition of silence. Very small non-zero values from an audio backend, dithering, conversion noise, or the device noise floor can continuously reset `AUDIO_ZERO_COUNT`, even though the signal is effectively silent. On such devices the gate may therefore never close at all.

I would keep this separate from the current PR, but if we retain the zero gate, please follow up by measuring the actual capture noise floor and replacing exact-zero detection with an effective-silence detector, preferably block-level peak/RMS detection with hysteresis rather than a single arbitrary `f32` epsilon. We should avoid choosing a threshold without measurements because we also do not want to suppress genuinely quiet audio.

Other separate investigations:

* whether `Fixed(64)` should remain the desktop output buffer strategy;
* receiver backlog/buffer correction;
* the observed delivery stalls.

So my recommendation is:

1. finish the capture realtime allocation cleanup in this PR;
2. merge the resampling/continuity fix;
3. address CPAL stream-death recovery / #15827 next;
4. separately clean up the zero-gate silence detector based on measured noise-floor data.

fufesou:
@coderabbitai full review

fufesou:
@codex review

fufesou:
@coderabbitai review

rustdesk:
I reviewed the current head of PR #16095, including the September 9 merge from `master`. Overall, the redesign is substantially better: stateful resampling, preallocated capture framing, bounded/drop-oldest capture buffering, encoding off the CPAL callback, and click suppression around discontinuities all look sensible. The PR currently spans 26 commits and touches the playback/capture pipeline fairly deeply. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16095/files)][1])

### Findings

**[P2] Restart capture after a `CaptureFrameProcessor` failure instead of retrying forever**

`src/server/audio_service.rs:498-503`

The capture callback currently does this:

```rust
frame.process(convert_input_samples(data), |frame| {
    if let Err(error) = processor.process(frame) {
        log::error!("Failed to process captured audio frame: {error:#}");
    }
});
```

`processor` contains the persistent resampler state, so after `process()` returns an error, the same processor is used again on every subsequent 10 ms frame. There is no transition into the existing service restart path. 

This is particularly problematic with the optional libsamplerate backend, because `src_process()` can return a backend error and that error propagates through `FixedFrameAudioResampler`. 

The result can be:

* capture remains silent/broken indefinitely;
* the callback can emit an error roughly every 10 ms;
* formatting/logging happens on the real-time audio callback, making recovery conditions even worse.

There is already a `restart()` mechanism, and the service loop replaces the stream when `RESTARTING` is set. 

I'd change the callback to latch the first processing failure and request a restart exactly once. For example, conceptually:

```rust
let mut failed = false;

move |data: &[T], _: &InputCallbackInfo| {
    if failed {
        return;
    }

    frame.process(convert_input_samples(data), |frame| {
        if let Err(error) = processor.process(frame) {
            failed = true;
            log::error!("Failed to process captured audio frame: {error:#}");
            restart();
        }
    });
}
```

I'd preferably keep even the error formatting/logging off the RT thread eventually, but restarting once is the important correctness fix.

This issue was also identified in an earlier automated review, and importantly, it is **still present in the current head**, so it hasn't become stale. ([[github.com](https://github.com/rustdesk/rustdesk/pull/16095)][2])

---

**[P2 / real-time concern] The playback callback is still capable of blocking on a mutex**

`src/client/audio_playback.rs:63-70`

The new playback writer eliminates the old sleep and callback allocation, which is a significant improvement, but it still does:

```rust
let mut buffer = self.audio_buffer.lock().unwrap();
...
buffer.pop_slice(...);
```

inside the CPAL callback. 

The producer takes that same mutex while pushing decoded PCM:

```rust
let mut lock = self.0.lock().unwrap();
...
lock.push_slice_overwrite(buffer);
```

and `try_shrink()` can also hold it while discarding samples. 

So the commit titled `fix(audio): keep playback callback nonblocking` doesn't make the callback strictly nonblocking. It removes some much worse operations, but priority inversion between the producer and CPAL callback is still possible. The callback also locks `ready` on every invocation—and twice on the first invocation. 

I would prefer either:

1. split the ring buffer into a single-producer/single-consumer pair so playback doesn't need the mutex; or
2. use `try_lock()` from the callback and treat contention as an underrun. The newly added fade/recovery mechanism is actually well suited to making that graceful.

`ready` should just be an `AtomicBool`.

This particular locking pattern existed in `master` as well, so I **wouldn't call it a regression introduced by this PR**. The PR has made it much better by removing `sleep`, temporary `Vec` allocation, and a longer critical section. But because this PR's purpose is specifically audio continuity/RT behavior, I'd still address it or explicitly leave a follow-up. The old implementation indeed held the same ring mutex and could even sleep from inside the callback. 

---

**[P3] The allocation test installs a global allocator for the entire library test binary**

`src/audio_allocation_tests.rs:36-37`

```rust
#[global_allocator]
static ALLOCATOR: CountingAllocator = CountingAllocator;
```

Because this module is compiled into the crate's test binary, the allocator intercepts allocations from unrelated unit tests as well, not just these audio tests. Every allocation additionally touches the thread-local accounting logic. 

It isn't a production issue, but it can distort timings and gives an ostensibly local audio test a crate-wide side effect. I'd move these allocation assertions into an isolated integration/test binary if practical.

### Things that look good

I didn't find a correctness issue in the new streaming resampler. It preserves interpolation position and buffered samples across calls, pre-reserves input state for capture, and the fixed-frame layer retains incomplete output between packets. The libsamplerate implementation also writes into the caller's preallocated output buffer rather than allocating a fresh result per capture packet. 

The capture queue design is also a good improvement: the CPAL producer uses `try_lock()`, drops on contention rather than waiting, uses a bounded pool, evicts the oldest queued packet under backlog, tracks sequence numbers, and moves Opus encoding/service transmission to a worker thread. 

The playback recovery logic is reasonable as well: underruns and explicit buffer discards transition over 5 ms instead of introducing abrupt zero/nonzero discontinuities. 

### Verdict

I'd **request changes for the capture error/restart case** before merging. The playback mutex is the other thing I'd want documented as a follow-up if it isn't fixed here. I don't see a reason to block on the allocator-test nit.

So my review would roughly be:

**1 blocking issue, 1 RT-performance concern, 1 test nit. Otherwise the architecture looks substantially improved over `master`.**

[1]: https://github.com/rustdesk/rustdesk/pull/16095/files "Fix/audio stream continuity by fufesou · Pull Request #16095 · rustdesk/rustdesk · GitHub"
[2]: https://github.com/rustdesk/rustdesk/pull/16095 "Fix/audio stream continuity by fufesou · Pull Request #16095 · rustdesk/rustdesk · GitHub"

fufesou:
https://github.com/rustdesk/rustdesk/pull/16095#issuecomment-5603522969

**Finding 1: Restart capture after a processing error**

Added a recovery guard in `23a202c98`.

After the first processing error, the callback logs it and stops further processing. The service thread recreates the stream; teardown and joining the encoder worker stay outside the audio callback. Each stream owns its error flag, so a late callback from an old stream cannot restart its replacement.

I still haven't reproduced a persistent processing failure with valid capture input. The code comments explicitly describe this as defensive recovery.

One regression test injects a processing error and checks that the remaining frames and later callbacks do not retry the processor, and that the replacement error state is independent. The existing allocation test now also exercises the guard during normal processing.

The macOS library audio tests passed with both backends: 36 with the default backend and 34 with `use_samplerate`. No hardware audio test was performed.

**Finding 3: Global allocator in the library tests**

Kept the existing allocation checks in the library tests.

The allocator wraps the whole library test binary, so other tests also incur the bookkeeping check. The counter is thread-local and only enabled during `assert_no_allocations`; the module is excluded from production builds.

The earlier validation confirmed that 1,000 allocations on another thread were excluded, while an allocation on the measured thread was detected. I haven't found a timing or test-isolation problem that would justify moving these tests into a separate binary.

rustdesk:
I reviewed the current PR head, **`8d4abfa`**, including the September 10 follow-up fixes. The PR is currently back in **Draft**. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16095)][1])

### Remaining finding: capture can lose 10 ms on mutex contention without actual backlog

I would raise one remaining RT/continuity concern in `src/server/audio_service/audio_capture_queue.rs`.

`CapturePcmSender::submit()` uses `try_lock()` on the shared `CapturePcmBuffers` mutex and immediately counts the current packet as dropped whenever the worker happens to hold that mutex:

```rust
let mut buffers = match self.handoff.buffers.try_lock() {
    Ok(buffers) => buffers,
    Err(error) => {
        self.handoff.dropped.fetch_add(1, Ordering::Relaxed);
        ...
        return;
    }
};
```

The encoder worker takes the same mutex when popping and recycling buffers. Consequently, contention is treated exactly like capacity exhaustion even though the queue may still have room. 

This isn't only theoretical behavior—the new test deliberately pauses the worker while it owns that mutex and asserts that the callback drops one packet. In the `Recycle` case there is already a buffer returned to `available`, so the packet would fit if the producer could access the handoff; it is lost solely because the lock is temporarily owned. 

For ordinary scheduling this critical section is very small, so the observed loss rate may indeed be zero—as the PR's 117-second tests reported. But a worker can be descheduled while holding the mutex, turning a scheduler event into a full **10 ms capture discontinuity**. That's particularly relevant because eliminating these discontinuities is the purpose of this PR. The PR itself correctly notes that those test runs rule out saturation only for those runs, not sender-worker scheduling. ([[GitHub](https://github.com/rustdesk/rustdesk/pull/16095)][1])

I'd probably comment something along these lines conceptually: **keep the RT callback nonblocking, but avoid using mutex ownership itself as a packet-loss condition**. A bounded lock-free handoff using preallocated buffers—e.g. free/ready queues—or another design where producer and consumer don't contend on one mutex would preserve both requirements. If that's too large a change for this PR, I'd at least stress-test this path under CPU contention and make the accepted loss behavior explicit.

**Severity:** I'd call this **P2/P3** rather than an unconditional blocker. It is a real source of avoidable loss, but the measured runs haven't demonstrated it occurring in practice.

### Previous blockers are fixed

I would **not repeat the September 9 findings** in a new review. The processing-error problem has been addressed: `CaptureErrorHandler::process_frame()` now latches the failure, stops subsequent processing, and lets the service thread recreate the stream. 

Likewise, playback no longer blocks on the PCM mutex: it performs one `try_lock()`, retains queued PCM on contention, and uses atomics for readiness/error reporting. That's the appropriate resolution to the earlier RT callback concern. 

The streaming resampler implementation also still looks sound to me: it preserves fractional position and incomplete output across calls, while the fixed-frame wrapper packetizes output without throwing away state. I don't see a new correctness problem there. 

### One follow-up I wouldn't charge to this PR

`CaptureErrorHandler::handle()` only requests recreation for `StreamInterrupted`; `DeviceNotAvailable` falls through to trace logging. CPAL explicitly defines `DeviceNotAvailable` as the device having disappeared, which normally warrants rebuilding the stream. 

However, **current `master` already has the same behavior**, so I wouldn't make that a PR #16095 regression or block this PR on it. It deserves a separate recovery fix. 

### Verdict

At `8d4abfa`, I no longer see the earlier merge-blocking correctness bug. My review would be **approve / changes recommended**, with the capture-handoff contention loss as the one meaningful remaining concern. If RustDesk wants a strict "RT callback must never wait, occasional contention loss is acceptable" policy, this is probably mergeable after the current CI rerun; if this PR is meant to eliminate avoidable capture discontinuities under scheduler pressure too, I would fix the handoff before merging. The visible x86_64 Linux job is currently succeeding; [GitHub](https://github.com/rustdesk/rustdesk/pull/16095/checks) still shows jobs re-running/loading on the draft head. ([GitHub][2])

[1]: https://github.com/rustdesk/rustdesk/pull/16095 "Fix/audio stream continuity by fufesou · Pull Request #16095 · rustdesk/rustdesk · GitHub"
[2]: https://github.com/rustdesk/rustdesk/pull/16095/checks "Fix/audio stream continuity by fufesou · Pull Request #16095 · rustdesk/rustdesk · GitHub"

fufesou:
https://github.com/rustdesk/rustdesk/pull/16095#issuecomment-5612018455

The contention case is real. A temporary test using the current Rust implementation confirmed that submission drops a packet when the worker holds the mutex, even with an empty ready queue and all ten buffers available.

For this PR, we are keeping the single `try_lock()` attempt so the capture callback does not wait for a suspended worker. I have expanded the source comment to state that this can reject the current packet without backlog. The existing drop-oldest policy still applies when the buffer pool is exhausted, and the callback can acquire the lock.

I also added `contention_dropped` to the existing worker-side loss report. It counts normal mutex contention and is included in the existing `dropped` total. This helps distinguish contention from other drops without changing the handoff policy. The existing paused-worker and saturation tests now check that distinction.

A handoff redesign belongs in a separate PR. We previously reproduced the old Crossbeam queues waiting for a worker paused between reserving and publishing or releasing a slot. A replacement needs to pass those progress checks as well as preserve buffer ownership, packet order, and drop-oldest behavior; using a queue described as lock-free is not enough by itself.

The two reported 117-second runs had no handoff loss, but they do not establish how often contention loss occurs under CPU pressure. The current smoothing reduces the transition after a sequence gap; it cannot restore the discarded audio. This PR therefore does not claim to eliminate all capture discontinuities caused by scheduling.

PR #16119, opened 2026-09-08
server: end a connection a newer one from the same session has replaced

rustdesk:
When a controller's link dies without a close reaching the peer - the only case that matters, since anything graceful already closes cleanly - the connection it abandoned keeps running here until the inactivity timeout. If the controller reconnects in the meantime, and it does, two authorized connections for one session overlap, and everything the stale one costs is charged to the session that replaced it.

The cost is not only cosmetic. It stays a video subscriber, and the capture loop waits for every subscriber to fetch the current frame before capturing the next, for up to three seconds; a connection whose own send is blocked never fetches, so frames for the live session arrive at that rate. It stays a viewer the rate control averages in, holding the shared target down. And the connection manager lists it: its own same-peer dedup only drops entries already marked disconnected, which this one is not.

So end it when the replacement authorizes. Nothing is sent to the peer - the link it would go over is the one that just failed, and the peer that owned this connection has already moved on - and the screen is not locked, because the session is continuing rather than ending. A connection alongside it that is not remote control is untouched: a file transfer or port forward sharing the session is expected.

Claude-Session: https://claude.ai/code/session_019aokqJuhjvB3kijXtAg5Ns

## Summary by CodeRabbit

- **Bug Fixes**
  - Connection displacement now occurs after login completes, preventing interruptions during authentication.
  - Existing connections close quietly without sending an unnecessary response.
  - Reconnecting through a newer connection no longer interrupts the active session or unnecessarily locks the screen.
  - Concurrent reconnects are handled consistently.
  - Port-forward connections are no longer displaced by newer connections.

- **Documentation**
  - Clarified connection displacement behavior in the IPC documentation.

<h2><a href="https://app.greptile.com/api/retrigger?id=61530555"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a><a href="https://app.greptile.com/rustdesk/-/pull-requests/rustdesk/rustdesk/16119"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptileDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewInGreptile.svg?v=1">View in Greptile</picture></a>Confidence Score: 4/5</h2>

The PR is not yet safe to merge because concurrent authorization can still leave two remote-control connections active for one session.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Asrc%2Fserver%2Fconnection.rs%3A6686%0AWhen%20an%20older%20connection%20is%20delayed%20before%20registration%2C%20a%20later-created%20connection%20can%20register%2C%20subscribe%2C%20and%20scan%20%60AUTHED_CONNS%60%20before%20the%20older%20connection%20appears.%20The%20delayed%20connection%20then%20registers%20and%20subscribes%2C%20but%20this%20comparison%20prevents%20it%20from%20displacing%20the%20active%20higher-ID%20connection.%20Because%20IDs%20are%20assigned%20when%20connections%20are%20accepted%20rather%20than%20when%20takeover%20completes%2C%20both%20remote-control%20connections%20remain%20active%20and%20preserve%20the%20duplicate%20subscriber%20and%20rate-control%20problem.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=rustdesk%2Frustdesk&pr=16119&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>Findings</h2>

1. P1&nbsp;**Authorization order leaves duplicates** <a href="https://github.com/rustdesk/rustdesk/pull/16119#discussion_r3964703492">▶</a>

- Restricts displacement to older remote-control connections with the same session key.
- Leaves file-transfer, port-forward, camera, and terminal connections untouched.
- Handles displacement in the authenticated-connection channel and logs failed sends.
- Regression surface: `src/ipc.rs` adds the internal signal, while the remote authorization and Windows delayed-session paths in `src/server/connection.rs` invoke displacement after subscription; these paths are necessary to avoid interrupting a working session before takeover completes.
- A remaining authorization-order race can still leave duplicate remote-control connections active.
</details>

<details><summary><h3>Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant A as Older-created connection A
    participant B as Later-created connection B
    participant L as AUTHED_CONNS
    A->>A: Authentication delayed
    B->>L: Register higher connection ID
    B->>B: Subscribe services
    B->>L: Scan for lower registered IDs
    Note over B,L: A is not registered yet
    A->>L: Register lower connection ID
    A->>A: Subscribe services
    A->>L: Scan for IDs lower than A
    Note over A,B: B has a higher ID, so neither is ended
```
</details>

### DIFF
--- a/src/ipc.rs
+++ b/src/ipc.rs
@@ -355,6 +355,10 @@ pub enum Data {
     MouseMoveTime(i64),
     Authorize,
     Close,
+    // A newer authorized connection has taken this one's place. Ends the superseded connection
+    // like `Close`, but sends the peer nothing: the peer has already moved on, and the link it
+    // would be sent over is the very one that stopped working.
+    Displaced,
     #[cfg(windows)]
     SAS,
     UserSid(Option<u32>),
--- a/src/server/connection.rs
+++ b/src/server/connection.rs
@@ -1018,6 +1018,15 @@ impl Connection {
                 },
                 Some(data) = rx_from_authed.recv() => {
                     match data {
+                        // A newer connection from this session has taken over. Nothing is sent
+                        // to the peer and the screen is not locked: the session continues, on
+                        // the connection that replaced this one.
+                        ipc::Data::Displaced => {
+                            conn.chat_unanswered = false; // seen
+                            conn.file_transferred = false; //seen
+                            conn.on_close("displaced by a newer connection", false).await;
+                            break;
+                        }
                         #[cfg(all(target_os = "windows", feature = "flutter"))]
                         ipc::Data::PrinterData(data) => {
                             if Self::permission(keys::OPTION_ENABLE_REMOTE_PRINTER, &conn.control_permissions) {
@@ -2065,6 +2074,7 @@ impl Connection {
         } else if sub_service {
             if !wait_session_id_confirm {
                 self.try_sub_monitor_services();
+                raii::AuthedConnID::displace_others(self.inner.id(), &self.session_key());
             }
         }
         true
@@ -3850,6 +3860,10 @@ impl Connection {
                                 self.try_sub_camera_displays();
                             } else if !self.terminal {
                                 self.try_sub_monitor_services();
+                                raii::AuthedConnID::displace_others(
+                                    self.inner.id(),
+                                    &self.session_key(),
+                                );
                             }
                         }
                     }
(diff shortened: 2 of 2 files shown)


rustdesk:
Closing this. Review turned up three problems in the displacement itself, and the benefit behind it is smaller than I had assumed.

The ordering the design rests on does not hold. Connection ids come from `get_new_id()` at accept (`server.rs:207`), but displacement is decided at authorize, and the CM prompt sits between them. Two connections both waiting at that prompt authorize in whatever order the operator clicks, so the one that sweeps first may not see the other yet, and the one that sweeps second may have the lower id and skip it. `conn_id <` stops them from ending each other, but the result is that neither is displaced.

`SessionKey` takes `session_id` straight from the login request (`connection.rs:5569`), and clients older than the field leave it at 0. Two remote control windows from such a peer collapse into one key, and the second login would end the first — with no close reason sent, so the first window just freezes.

The `SelectedSid` call site guards on `!self.terminal` and the predicate only checks the kind of the connection being ended, never the kind of the one doing it. Nothing but the message scope table, three thousand lines away, keeps a non-remote connection out of that path today.

Against that, the case where a stale connection costs the most is the one displacement cannot reach: with privacy mode on, the stale connection still holds it, so the reconnect is refused in `check_privacy_mode_on` (`connection.rs:2805`) and never authorizes to sweep at all. And a displaced connection only acts on the message when its select loop next yields, which on the dead link this targets can be a `SEND_TIMEOUT_VIDEO` of 12s (`connection.rs:397`) — by which point the send it is parked in has timed out on its own anyway.

The part that actually mattered here was the screen lock: the timeout at `SEC30` calls `on_close("Timeout", true)` with no check for whether the session is still being controlled. That is now #NEW as a single condition on that gate, which needs none of the ordering, none of the identity assumptions and ends no connections.

Issue #44, opened 2021-05-25
Request: standalone binaries in .tar.gz format for releases


fluffynuts:
I'm trying to build rustdesk on Gentoo. My rust-fu is weak, so I don't really understand the message that is coming back or how to resolve it (well, I _kinda_ understand it, but I don't know how to resolve it):

```
warning: build failed, waiting for other jobs to finish...
error[E0277]: the trait bound `std::vec::Vec<u8>: std::convert::From<[u8; 32]>` is not satisfied
   --> libs/hbb_common/src/config.rs:485:52
```
(this is from the 1.1.5 tag, after cloning the repo locally)

However, for now, I'd just like to try out rustdesk between my windows and linux machines; you provide great packages for a few distros, and I'd like to ask if it would be possible to provide a vanilla .tar.gz of the binaries that I could unpack on my system to give rustdesk a go?

Ideally, I'd like to build from source, but before I chase this much further, I'd like to see if rustdesk gives me what I'm looking for: a viable alternative to anydesk

Thanks for your time.

fluffynuts:
neat 😕

PR #16160, opened 2026-09-11
Use constant-QP rate control for Linux VAAPI encoding

mmdmcy:
## Summary

On Linux systems where the VAAPI driver exposes only constant-QP rate control,
RustDesk currently initializes the VAAPI encoder with CBR. On the Intel iHD
setup tested for this change, that causes h264_vaapi initialization to fail and
forces a software encoding fallback.

This change makes the RustDesk VAAPI integration select the existing constant-
quality rate-control mode. The corresponding FFmpeg VAAPI option mapping is
provided by https://github.com/rustdesk-org/hwcodec/pull/44.

## Changes

- Select constant-quality rate control for Linux VAAPI encoders.
- Leave the existing rate-control behavior for other encoder backends unchanged.
- Keep the change limited to encoder selection; no UI or configuration changes
  are included.

## Validation

Tested on an Intel HD Graphics 530 using the Intel i915/iHD VAAPI stack.

- The CBR probe fails because the driver supports only CQP.
- The CQP probe succeeds with QP 23.
- The resulting RustDesk session no longer falls back to software AV1 encoding
  and became substantially smoother.
- The changed Rust source passes rustfmt and the patch passes whitespace checks.

## Dependencies

This change depends on https://github.com/rustdesk-org/hwcodec/pull/44 for the
VAAPI-specific rc_mode=CQP and QP option handling.

It is intentionally narrower than the broader image-quality work in
https://github.com/rustdesk/rustdesk/pull/14294: this change fixes capability
selection for Linux VAAPI hosts.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved VAAPI hardware video encoding quality control with quality-based QP settings.
  - VAAPI configurations now use constant-quality rate control for more consistent output across supported presets.
  - Quality changes now correctly recreate the encoder when live adjustment is unavailable, limited to supported preset quality changes. Custom quality changes no longer trigger an unnecessary encoder switch.

<h2><a href="https://app.greptile.com/api/retrigger?id=62992800"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not yet safe to merge because VAAPI custom-quality QoS changes can still leave the encoder using its original QP.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Alibs%2Fscrap%2Fsrc%2Fcommon%2Fhwcodec.rs%3A86-88%0AVAAPI%20now%20derives%20a%20fixed%20QP%20from%20%60config.quality%60%20only%20when%20the%20encoder%20is%20constructed.%20When%20a%20custom-quality%20QoS%20ratio%20later%20changes%2C%20VAAPI%20cannot%20update%20quality%20in%20place%2C%20but%20%60check_qos%60%20also%20skips%20encoder%20recreation%20for%20custom%20quality.%20The%20session%20therefore%20keeps%20its%20initial%20QP%2C%20silently%20ignoring%20adaptive%20quality%20reductions%20and%20recoveries.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=rustdesk%2Frustdesk&pr=16160&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>Findings</h2>

1. P1&nbsp;**VAAPI QP becomes stale** <a href="https://github.com/rustdesk/rustdesk/pull/16160#discussion_r3987696296">▶</a>

`````markdown
### Issue 1
libs/scrap/src/common/hwcodec.rs:86-88
VAAPI now derives a fixed QP from `config.quality` only when the encoder is constructed. When a custom-quality QoS ratio later changes, VAAPI cannot update quality in place, but `check_qos` also skips encoder recreation for custom quality. The session therefore keeps its initial QP, silently ignoring adaptive quality reductions and recoveries.

---

`````

</details>

- Selects `RC_CQ` and a calculated QP for VAAPI encoders while preserving existing rate control for other backends.
- Adds tests for preset ordering, monotonicity, bounds, and invalid quality values.
- The latest revision again excludes custom-quality changes from encoder recreation, leaving the previous stale-QP finding outstanding.
</details>

```mermaid
flowchart TD
    A[Create hardware encoder] --> B{VAAPI backend?}
    B -->|Yes| C[Select constant-quality rate control]
    C --> D[Convert quality ratio to bounded QP]
    B -->|No| E[Use existing backend rate control]
    F[QoS quality changes] --> G{Encoder supports live quality update?}
    G -->|Yes| H[Update encoder in place]
    G -->|No, preset| I[Recreate encoder]
    G -->|No, custom| J[Keep existing encoder and stale QP]
```
</details>

<sub>Reviews (4) · Last reviewed commit: ["fix(vaapi): preserve encoder on custom q..."](https://github.com/rustdesk/rustdesk/commit/742fafe26072681d986ce54d53fec49ca560bfe9)</sub>

### DIFF
--- a/libs/scrap/src/common/hwcodec.rs
+++ b/libs/scrap/src/common/hwcodec.rs
@@ -1,5 +1,7 @@
 use crate::{
-    codec::{base_bitrate, codec_thread_num, enable_hwcodec_option, EncoderApi, EncoderCfg},
+    codec::{
+        base_bitrate, codec_thread_num, enable_hwcodec_option, EncoderApi, EncoderCfg, BR_BALANCED,
+    },
     convert::*,
     CodecFormat, EncodeInput, ImageFormat, ImageRgb, Pixfmt, HW_STRIDE_ALIGN,
 };
@@ -81,7 +83,11 @@ impl EncoderApi for HwRamEncoder {
                     gop,
                     quality: DEFAULT_HW_QUALITY,
                     rc,
-                    q: -1,
+                    q: if config.name.contains("vaapi") {
+                        Self::vaapi_qp(config.quality)
+                    } else {
+                        -1
+                    },
                     thread_count: codec_thread_num(16) as _, // ffmpeg's thread_count is used for cpu
                 };
                 let format = match Encoder::format_from_name(config.name.clone()) {
@@ -249,6 +255,18 @@ impl HwRamEncoder {
             return RC_CQ;
         }
         RC_CBR
+    }
+
+    fn vaapi_qp(ratio: f32) -> i32 {
+        const DEFAULT_QP: i32 = 23;
+        if !ratio.is_finite() || ratio <= 0.0 {
+            return DEFAULT_QP;
+        }
+        // Keep the balanced preset at QP 23. Each doubling of the quality
+        // ratio lowers QP by six; this is a quality curve, not a bitrate cap.
+        (DEFAULT_QP as f32 - 6.0 * (ratio / BR_BALANCED).log2())
+            .round()
+            .clamp(1.0, 51.0) as i32
     }
 
     pub fn bitrate(name: &str, width: usize, height: usize, ratio: f32) -> u32 {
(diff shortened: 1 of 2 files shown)


21pages:
Assuming rustdesk-org/hwcodec#44 is included, the encoding path now selects CQP. Two integration gaps remain:

1. `check_available_hwcodec()` still passes `RC_CBR` with `kbs = 1000` and does not call `HwRamEncoder::rate_control()`. CQP-only drivers will therefore still fail the availability probe. Please make VAAPI probing use CQP too, while preserving the probe settings for other backends.

2. `HwRamEncoder::new()` still passes `q = -1`, which rustdesk-org/hwcodec#44 maps to QP 23. Image quality currently only affects the calculated bitrate, so all quality presets will use the same QP—even after recreating the encoder. Please map the requested image quality to QP and apply it when quality changes.

Please also update the pinned `hwcodec` revision when integrating #44.

For validation, could you test fresh hardware detection with the cached result cleared, and verify that switching quality presets changes the QP?

mmdmcy:
Follow-up applied and pushed.

- Kept 21pages' quality-to-QP implementation and original commit authorship.
- Added an mmdmcy fix for the remaining QoS edge case: because VAAPI selects QP only at encoder construction, custom quality changes now take the encoder-recreation path too instead of silently retaining the old QP.
- The matching hwcodec changes are now in PR 44, including CQP capability probing and QP-zero normalization.
- The hwcodec dependency remains pinned to the current upstream revision until PR 44 merges. Updating the lockfile before that would point RustDesk at an unmerged revision; I will update it as the next integration commit once the dependency is available upstream.

This branch is therefore intentionally ordered behind rustdesk-org/hwcodec#44.

21pages:
Pushed the follow-up here:
https://github.com/rustdesk/rustdesk/commit/742fafe26072681d986ce54d53fec49ca560bfe9

The custom-quality exception is intentional. VAAPI returns `false` from `support_changing_quality()`, so `in_vbr_state()` is false while VAAPI is active and `adjust_ratio()` returns without performing adaptive quality changes. The review's concern about silently ignoring automatic quality reductions and recoveries therefore does not apply to this path.

We want manual preset changes to recreate the encoder with the new QP, but custom-quality changes to keep the current codec. Removing `!latest_quality().is_custom()` introduces codec restarts for custom ratio changes, including changes that map to the same QP. This commit restores that condition.

A custom quality value is still used when the encoder is created. Changing it during a session intentionally leaves the current QP in place until the encoder is next created for another reason.

The dependency update can remain ordered after https://github.com/rustdesk-org/hwcodec/pull/44, as discussed.

PR #16167, opened 2026-09-12
Share the refresh limit between video queue recovery paths

21pages:
When the client video queue fills, each overwritten frame can request a refresh, causing the controlled device to repeatedly recreate its encoder. FPS control can also request a refresh independently when frames accumulate.

This PR gives both queue recovery paths one shared per-display limit:

- At most one recovery refresh every 10 seconds.
- At most 20 recovery refreshes for the lifetime of each video thread, shared between queue overflow and FPS control.
- Manual refresh continues to work during the cooldown and after the automatic recovery limit is reached.

Queue-overflow warnings include the codec, queue length, decode FPS estimate and discard state. RAM decoder creation parameters are also logged.

Software HEVC decoding retains its existing thread-count policy. The underlying FFmpeg HEVC/WPP deadlock is addressed separately by the FFmpeg 7.1.1 upgrade in #16169. This PR bounds queue recovery requests; it does not resolve a decoder deadlock in an affected FFmpeg build.

### Validation

Four deterministic tests passed on macOS arm64 using the existing local FFmpeg 7.1.1 installation. They exercise the actual receive, FPS-control and manual-refresh message paths, covering:

- The first queue overflow requests exactly one refresh.
- Repeated overflow and FPS control cannot bypass each other's cooldown, in either trigger order.
- Refresh is suppressed at 9.999 seconds and allowed at 10 seconds.
- Different displays have independent limits.
- Both automatic recovery paths share the 20-request cap, while manual refresh still reaches the peer after that cap.

```sh
cargo test --locked --lib \
  --features flutter,hwcodec,screencapturekit,unix-file-copy-paste \
  client::io_loop::refresh_tests -- --nocapture
```

`git diff --check` passed. The new test file passes rustfmt; the modified existing files introduce no new formatting issues. This validation covers the local test build and recovery logic, not a new cross-platform live-session test.

### Regression surface

- `src/client/io_loop.rs`: queue overflow and FPS control reuse the existing per-display timestamp and retry counter, preventing independent encoder restarts. The shared cooldown expires at exactly 10 seconds.
- `libs/scrap/src/common/hwcodec.rs`: adds decoder-configuration logging; the final diff leaves decoder thread-count selection unchanged.
- `src/client/io_loop/refresh_tests.rs`: adds tests only.

The existing decoder-error recovery path in `src/client.rs` is outside these two queue recovery paths and remains unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability for H.265 software video decoding, reducing the risk of playback deadlocks.
  - Reduced unnecessary display refreshes when video queues are full, helping prevent repeated disruptions during playback.

<h2><a href="https://app.greptile.com/api/retrigger?id=63339666"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR appears safe to merge, with the non-blocking caveat that the new queue-full timing behavior should receive focused regression coverage.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Asrc%2Fclient%2Fio_loop.rs%3A1411-1418%0AThe%20new%20queue-full%20timing%20branch%20was%20not%20exercised%20during%20validation.%20Without%20focused%20coverage%20for%20the%20first%20overflow%2C%20suppression%20during%20the%20ten-second%20window%2C%20and%20another%20refresh%20after%20the%20window%2C%20future%20changes%20could%20silently%20disable%20recovery%20or%20restore%20the%20repeated%20encoder-recreation%20loop%20this%20change%20is%20meant%20to%20prevent.%0A%0ANote%3A%20If%20this%20suggestion%20doesn't%20match%20your%20team's%20coding%20style%2C%20reply%20to%20this%20and%20let%20me%20know.%20I'll%20remember%20it%20for%20next%20time!%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=rustdesk%2Frustdesk&pr=16167&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=6"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=6">Fix All in Claude Code</picture></a>Findings</h2>

1. P2&nbsp;**Rate limit lacks coverage** <a href="https://github.com/rustdesk/rustdesk/pull/16167#discussion_r3994672903">▶</a>

`````markdown
### Issue 1
src/client/io_loop.rs:1411-1418
The new queue-full timing branch was not exercised during validation. Without focused coverage for the first overflow, suppression during the ten-second window, and another refresh after the window, future changes could silently disable recovery or restore the repeated encoder-recreation loop this change is meant to prevent.

Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!

---

`````

</details>

- `libs/scrap/src/common/hwcodec.rs` changes the existing RAM decoder construction path only for software H.265 decoding; this shared-path change is necessary to disable the affected FFmpeg threading mode while retaining existing thread counts for hardware decoding and other codecs.
- `src/client/io_loop.rs` changes the existing queue-overflow recovery path by tracking the last queue-full refresh per display and allowing another request after ten seconds; this is necessary to bound encoder restarts while preserving eventual recovery.
- Decoder construction and queue state are now logged to aid diagnosis.
- The queue-full timing behavior still needs focused regression coverage.
</details>

```mermaid
flowchart TD
    A[Incoming video frame] --> B{Keyframe?}
    B -- Yes --> C[Send directly to decoder]
    B -- No --> D{Queue full?}
    D -- No --> E[Queue frame and notify decoder]
    D -- Yes --> F[Overwrite oldest queued frame]
    F --> G{No prior queue-full refresh or 10 seconds elapsed?}
    G -- Yes --> H[Record timestamp and request display refresh]
    G -- No --> I[Suppress repeated refresh]
    H --> J[Controlled device recreates encoder and emits keyframe]
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["Work around software HEVC slice-thread s..."](https://github.com/rustdesk/rustdesk/commit/6a2bc71a8263de54a76c819835b0bb5b28452c23)</sub>


rustdesk:
I reviewed PR #16167. The overall direction is reasonable, and the FFmpeg 7.1 HEVC WPP/slice-thread deadlock diagnosis looks credible. RustDesk currently bundles FFmpeg 7.1, while FFmpeg 7.1.1 explicitly backported `lavc/hevcdec: unbreak WPP/progress2 code`. The upstream fix specifically addresses races and deadlocks in HEVC slice threading.

However, I would not merge this PR as-is. I see two issues that are more important than the existing automated review comment about missing test coverage.

### 1. The queue-full limiter does not cover all automatic refresh paths

The PR adds a new per-display timer:

```rust
last_queue_full_refresh
```

which limits the queue-full path to one `refresh_video()` call every 10 seconds.

However, `fps_control()` already has a separate refresh mechanism using:

```rust
last_refresh_instant
refresh_times
```

and it also calls `refresh_video()` when the video queue is backed up.

This means the two paths can still independently trigger encoder recreation. For example:

```text
t = 0.0s
queue becomes full
-> queue-full path calls refresh_video()

t = 0~1s
fps_control() runs
-> refresh_times == 0
-> another refresh_video() may happen immediately

t = 10s
queue-full limiter expires
-> another refresh

t = 10~11s
fps_control() limiter may also expire
-> another refresh
```

So this change rate-limits one call site, but it does not actually rate-limit automatic refreshes as a whole.

That matters because the problem described by this PR is specifically repeated `refresh_video()` requests causing the controlled side to repeatedly recreate the encoder.

There is another inconsistency: the existing `fps_control()` path has:

```rust
refresh_times < 20
```

while the new queue-full path bypasses this limit completely. Under a persistent decoder stall or overload, it can therefore continue requesting a refresh every 10 seconds indefinitely.

**Requested change:**

> Please share the automatic refresh rate limit with the existing FPS-control path.
>
> `last_queue_full_refresh` only throttles this particular call site, while `fps_control()` has an independent `last_refresh_instant` and can also call `refresh_video()` when the queue is backed up. In particular, its first refresh (`refresh_times == 0`) can happen immediately after a queue-full refresh, so we can still recreate the remote encoder twice in quick succession. The two 10-second timers can also continue independently afterwards.
>
> Please use a single per-display automatic-refresh gate for both paths. Manual refresh should remain unaffected.
>
> We should also decide whether the existing `refresh_times < 20` safety cap should apply to queue-full recovery instead of allowing queue-full refreshes indefinitely.

I consider this a real issue that should be fixed before merge.

### 2. Forcing all software HEVC decoding to one thread may introduce a significant regression

The current hwcodec FFmpeg wrapper configures decoding roughly as:

```cpp
c_->thread_count =
    device_type_ != AV_HWDEVICE_TYPE_NONE ? 1 : thread_count_;
c_->thread_type = FF_THREAD_SLICE;
```

So the relationship between the deadlock and slice threading is credible, and using one thread avoids the problematic multithreaded path.

The concern is the scope of the workaround.

This PR changes every software H.265 decoder to:

```rust
if format == CodecFormat::H265
    && info.hwdevice == AV_HWDEVICE_TYPE_NONE
{
    1
}
```

Software H.265 is not merely an exceptional fallback in RustDesk. `HwRamDecoder::try_get()` exposes software H.265 as a normal decoding capability, and `supported_decodings()` advertises that capability to the remote side.

H.265 is also preferred by automatic codec negotiation when available.

Therefore, on machines without usable hardware decoding, or where hardware decoding is disabled, this can become:

```text
advertise H.265 normally
        ↓
remote selects H.265
        ↓
controller decodes HEVC using one CPU thread
        ↓
decoder may fail to sustain 1080p60 / 1440p / 4K
        ↓
video queue grows
        ↓
queue-full recovery starts firing
```

The PR validation shows that single-thread decoding successfully completed 300 rounds / 21,300 frames, which is useful for proving that the deadlock disappears, but it does not prove that decoding remains fast enough for real-time remote desktop workloads.

Since RustDesk already maintains its own FFmpeg vcpkg port, and the exact upstream fix is available in FFmpeg 7.1.1, I would strongly prefer fixing the dependency instead of shipping a global single-thread policy.

**Requested change:**

> Please quantify the software HEVC performance regression before making this the release workaround.
>
> RustDesk advertises software H.265 as a normal decoding capability, and H.265 can be selected by automatic codec negotiation. Forcing every software HEVC decoder to one thread may therefore affect normal sessions, not just the deadlock case. A replay completing successfully does not show that it can keep up in real time.
>
> Since our bundled FFmpeg is 7.1 and the exact WPP/progress2 fix is already included in FFmpeg 7.1.1, I would strongly prefer validating a 7.1.1 bump, or backporting the upstream fix into our existing FFmpeg port, before shipping a global single-thread workaround.
>
> If we need the temporary workaround first, please at least benchmark software HEVC decode throughput before/after at representative 1080p60 and higher-resolution workloads.

I consider this more important than the current automated-review comment about unit-test coverage.

### Test coverage

The existing comment about testing the 10-second rate limiter is valid, but the test should cover the interaction between both automatic refresh mechanisms rather than testing the new timer in isolation.

At minimum, I would test:

* First queue overflow causes exactly one automatic refresh.
* Repeated queue overflow within 10 seconds does not cause another refresh.
* `fps_control()` cannot issue a second automatic refresh during the same cooldown.
* Another recovery refresh is allowed after the cooldown expires.
* Different displays have independent cooldown state.
* Manual refresh is unaffected.

### CI

The normal CI workflow passed.

The Full Flutter CI failure appears unrelated to this PR: the `i686-pc-windows-msvc` job failed while installing the Rust toolchain, before the RustDesk build step ran.

### Conclusion

**Request changes.**

I agree with the HEVC deadlock diagnosis and with adding protection against repeated queue-full refreshes, but I would require two changes before merging:

1. Use one shared per-display limiter for all automatic `refresh_video()` paths.
2. Prefer upgrading to FFmpeg 7.1.1 or backporting the upstream HEVC fix. If the single-thread workaround must ship temporarily, first quantify its real-time decoding performance impact.

Given that RustDesk already maintains `res/vcpkg/ffmpeg`, fixing the FFmpeg bug itself is cleaner than permanently changing the decoder policy to `thread_count = 1`, and it avoids leaving a workaround behind after the dependency is eventually upgraded.

PR #16172, opened 2026-09-12
fix(audio): throttled debug logs

fufesou:
## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced repetitive audio playback diagnostics by limiting repeated buffer-contention messages to once every five seconds.
  * Audio contention information now remains available while accumulating between reports, improving diagnostic accuracy.
  * Lowered routine audio capture and Linux audio availability messages to less prominent diagnostic levels, keeping logs clearer while preserving detailed troubleshooting information.
  * Audio playback diagnostic reporting is now tracked independently for separate playback sessions.

<h2><a href="https://app.greptile.com/api/retrigger?id=63388128"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge with no outstanding actionable findings.

- Throttles audio buffer-discard diagnostics to once every five seconds.
- Uses a per-playback throttle so contention counts accumulate until emitted and remain isolated between playback instances.
- Lowers routine PulseAudio and capture-stat diagnostics from debug to trace.
- Adds regression coverage for accumulated contention counts and independent playback throttles.
- Regression surface is limited to diagnostic logging in the existing audio playback and capture paths; audio decoding, buffering, and capture behavior are unchanged.
</details>

<sub>Reviews (2) · Last reviewed commit: ["fix(audio): preserve contention counts a..."](https://github.com/rustdesk/rustdesk/commit/26712cb05dee9f806b0f31f5e6e335152ae8d14d)</sub>

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -2250,7 +2250,9 @@ impl AudioBuffer {
         let occupied = lock.occupied_len();
         drop(lock);
         if let Some((discarded, generation)) = discard {
-            log::debug!(
+            hbb_common::throttled_log!(
+                audio_playback::AUDIO_PLAYBACK_LOG_INTERVAL,
+                debug,
                 "Audio buffer capacity discard: samples={discarded}, generation={generation}"
             );
         }
@@ -2441,7 +2443,7 @@ impl AudioHandler {
         }
         #[cfg(target_os = "linux")]
         if self.simple.is_none() {
-            log::debug!("PulseAudio simple binding does not exists");
+            log::trace!("PulseAudio simple binding does not exists");
             return;
         }
         self.audio_decoder.as_mut().map(|(d, buffer)| {
--- a/src/client/audio_playback.rs
+++ b/src/client/audio_playback.rs
@@ -6,6 +6,8 @@ use std::sync::{
 };
 
 pub(super) const UNDERRUN_DECLICK_MS: usize = 5;
+pub(super) const AUDIO_PLAYBACK_LOG_INTERVAL: std::time::Duration =
+    std::time::Duration::from_secs(5);
 const MILLISECONDS_PER_SECOND: usize = 1_000;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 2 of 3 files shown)


rustdesk:
## Review

The overall direction is good, and the two `debug -> trace` changes look appropriate. I found one blocking issue in the playback contention throttling.

### Blocking: `contentions` is already an aggregated metric, but `throttled_log!` counts report invocations

Current code:

```rust
let contentions = self.contentions.swap(0, Ordering::Relaxed);
if contentions != 0 {
    hbb_common::throttled_log!(
        AUDIO_PLAYBACK_LOG_INTERVAL,
        debug,
        "Audio playback PCM buffer contention: callbacks={contentions}"
    );
}
```

`contentions` is incremented from the playback callback whenever `try_lock()` fails, so it already represents the real number of contended callbacks.

However, `report_errors()` runs once per received audio frame, and it clears `contentions` with `swap(0)` before the throttle decides whether to emit the log.

This means callback counts from suppressed reports are permanently lost.

There is also a semantic mismatch with `throttled_log!`: its `(xN)` suffix counts how many times the log site was invoked, not how many playback callbacks contended.

For example, an emitted line could become:

```text
Audio playback PCM buffer contention: callbacks=2 (x37)
```

where:

* `callbacks=2` is only the latest batch of contended callbacks;
* `(x37)` is the number of non-zero `report_errors()` invocations suppressed by the throttle.

Those are different metrics, so the resulting diagnostic is misleading.

Please aggregate the actual callback count across the reporting interval and only clear it when a log is actually emitted.

One clean option would be to extend `LogThrottle` with something like `due_n(n)`, where `due()` is equivalent to `due_n(1)`, so this site can accumulate arbitrary counts instead of invocation counts.

Alternatively, keep the timing gate locally in `AudioPlaybackStatus`, but the important part is that suppressed callback counts must not be cleared.

A regression test for this behavior would also be useful.

### Other changes

The buffer-capacity discard throttling looks fine. Each macro invocation corresponds to one discard event, so `(xN)` has a meaningful interpretation there, and the throttle runs after the PCM buffer lock has already been released.

The two log-level changes also look good:

* `PulseAudio simple binding does not exists` can fire for every incoming audio frame, while the actual playback initialization failure is already logged at error level.
* Capture handoff queue statistics are routine diagnostics, while real capture loss still remains at `warn!`.

I do not see a new allocation, locking, or real-time regression in the playback callback path. The callback still only does `try_lock()` plus an atomic increment on contention.

CI and Full Flutter CI are both passing.

**Verdict:** Request changes for the contention-counter throttling. After that is fixed, this looks good to merge.

fufesou:
@coderabbitai review

PR #16178, opened 2026-09-12
fix(linux): resolve Wayland reconnect failure with "Failed to get capturer display info"

WinterSold1er:
### Description

Fixes an issue on Linux Wayland where connecting once and disconnecting causes subsequent connections to fail with:
`Failed to get capturer display info`

This occurs primarily when running RustDesk in non-server / standalone desktop mode.

### Steps to Reproduce
1. Launch RustDesk on Linux Wayland (without system `--server` background service, i.e., standalone client mode).
2. Connect to the machine from another client. The connection succeeds and screen capture works.
3. Disconnect the session.
4. Attempt to connect to the machine again from the client.
5. Connection fails with error: `Failed to get capturer display info`.

### Root Cause Analysis
1. **Stale Session Reuse in Non-Server Mode**: In `libs/scrap/src/wayland/pipewire.rs`, `try_close_session()` previously only closed the session if `is_server_running() && rdp_info.is_support_restore_token`. In non-server / standalone mode (`!is_server_running()`), `*RDP_SESSION_INFO` was never cleared on disconnect. Consequently, subsequent connections reused the stale portal/pipewire session handle, which had already become invalid after disconnection.
2. **Premature `PIPEWIRE_INITIALIZED` Flag**: In `src/server/wayland.rs`, `check_init()` set `*PIPEWIRE_INITIALIZED = true` before the loop initializing display capturers (`Capturer::new(display)?`). When capturer initialization failed (due to the stale session handle), the function returned early with an error, leaving `CAP_DISPLAY_INFO` empty while `PIPEWIRE_INITIALIZED` remained latched to `true`. On subsequent attempts, `check_init()` saw `*PIPEWIRE_INITIALIZED == true`, returned `Ok(())` prematurely without initializing any capturers, and caused `get_displays_and_primary()` to bail with `"Failed to get capturer display info"`.
3. **Missing Penetration Cleanup on Disconnect**: `wayland::clear()` cleared `CAP_DISPLAY_INFO` and reset `PIPEWIRE_INITIALIZED = false`, but did not penetrate to clean up the underlying PipeWire portal session (`scrap::wayland::pipewire::close_session()`).

### Solution
1. **Allow Session Teardown in Non-Server Mode**: Updated `try_close_session()` in `libs/scrap/src/wayland/pipewire.rs` to close the session if `!is_server_running() || rdp_info.is_support_restore_token`.
2. **Deferred Initialization Flag & Error Cleanup**: In `src/server/wayland.rs`, only set `*PIPEWIRE_INITIALIZED = true` after all capturers are successfully initialized. If capturer creation fails in the loop, cleanly destroy already-allocated raw pointers, reset `*PIPEWIRE_INITIALIZED = false`, and call `scrap::wayland::pipewire::close_session()`. Additionally, reset `*PIPEWIRE_INITIALIZED = false` if `CAP_DISPLAY_INFO` is empty in `get_displays_and_primary()`.
3. **Penetrate Underlying Session Teardown**: In `src/server/wayland.rs`, call `scrap::wayland::pipewire::close_session()` inside `wayland::clear()` so that whenever wayland capturers are cleared on disconnect/teardown, the underlying PipeWire portal session is also cleanly closed.

### Verification & Testing
- Verified with `rustfmt --edition 2021` ensuring consistent code style.
- Verified minimal diff footprint adhering to RustDesk's codebase conventions.
- Tested repeated connect-disconnect-reconnect lifecycle on Wayland desktop sessions. Reconnection now successfully initializes a fresh PipeWire capture session without getting stuck on stale session handles or empty display maps.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Wayland/PipeWire session cleanup when the server is unavailable or a connection is interrupted.
  - Fixed initialization handling so partially created display capture sessions are cleaned up when setup fails.
  - Prevented failed or empty display discovery from leaving sessions marked as active.
  - Improved resource cleanup when clearing a PipeWire session.

<h2><a href="https://app.greptile.com/api/retrigger?id=63430348"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; the changed Wayland initialization and teardown paths consistently discard stale state without an established regression.

- Broadens session cleanup after the last standalone remote connection disconnects.
- Marks PipeWire initialized only after every display capturer has been created.
- Frees partially initialized capturers and closes the portal session after initialization failure.
- Resets inconsistent empty-map state so later initialization attempts can retry.
- Extends the existing last-display and server teardown paths to clear the underlying PipeWire session.
- Regression surface: `libs/scrap/src/wayland/pipewire.rs` changes only the session-retention decision at final remote disconnect; `src/server/wayland.rs` changes Wayland capture initialization failure and teardown paths, both necessary to prevent stale-session reuse.
</details>

```mermaid
flowchart TD
  A[Remote connection requests Wayland displays] --> B[check_init]
  B --> C{Capture map already populated?}
  C -->|Yes| G[Return display information]
  C -->|No| D[Create capturer for each display]
  D -->|All succeed| E[Set PIPEWIRE_INITIALIZED true]
  E --> G
  D -->|Any fails| F[Free created capturers<br/>clear map and flag<br/>close portal session]
  F --> H[Return initialization error]
  G --> I[Connection disconnects]
  I --> J{Last active display?}
  J -->|No| K[Keep shared capture state]
  J -->|Yes| L[Clear capturers and flag<br/>close portal session]
  L --> M[Next connection creates fresh session]
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(linux): cleanup Wayland PipeWire ses..."](https://github.com/rustdesk/rustdesk/commit/984a2a5c006d635e80b31539889efe9b65f9f49a)</sub>

### DIFF
--- a/libs/scrap/src/wayland/pipewire.rs
+++ b/libs/scrap/src/wayland/pipewire.rs
@@ -84,8 +84,9 @@ pub fn try_close_session() {
     let mut rdp_info = RDP_SESSION_INFO.lock().unwrap();
     let mut close = false;
     if let Some(rdp_info) = &*rdp_info {
-        // If is server running and restore token is supported, there's no need to keep the session.
-        if is_server_running() && rdp_info.is_support_restore_token {
+        // If restore token is supported, there's no need to keep the session.
+        // In non-server mode, the session does not survive disconnect, so it must also be closed.
+        if !is_server_running() || rdp_info.is_support_restore_token {
             close = true;
         }
     }
--- a/src/server/wayland.rs
+++ b/src/server/wayland.rs
@@ -524,7 +524,6 @@ pub(super) async fn check_init() -> ResultType<()> {
                 }
                 log::debug!("Attempting to fix logical size with try_fix_logical_size()");
                 try_fix_logical_size(&mut all);
-                *PIPEWIRE_INITIALIZED.write().unwrap() = true;
                 let num = all.len();
                 let primary = super::display_service::get_primary_2(&all);
                 let mut displays = super::display_service::update_sync_displays(&all);
@@ -550,8 +549,23 @@ pub(super) async fn check_init() -> ResultType<()> {
                 for (idx, display) in all.into_iter().enumerate() {
                     // No `with_context` here: the peer is shown `format!("{}", err)`, which
                     // renders only the outermost layer, and the mapped reason is the inner one.
-                    let capturer = Box::into_raw(Box::new(Capturer::new(display)?));
-                    let capturer = CapturerPtr(capturer);
+                    let capturer = match Capturer::new(display) {
+                        Ok(c) => CapturerPtr(Box::into_raw(Box::new(c))),
+                        Err(e) => {
+                            for (_, addr) in lock.iter() {
+                                let cap_display_info: *mut CapDisplayInfo = *addr as _;
+                                unsafe {
+                                    let _box_capturer =
+                                        Box::from_raw((*cap_display_info).capturer.0);
+                                    let _box_cap_display_info = Box::from_raw(cap_display_info);
+                                }
+                            }
+                            lock.clear();
+                            *PIPEWIRE_INITIALIZED.write().unwrap() = false;
+                            scrap::wayland::pipewire::close_session();
+                            return Err(e.into());
+                        }
+                    };
 
                     let cap_display_info = Box::into_raw(Box::new(CapDisplayInfo {
                         rects: rects.clone(),
@@ -564,6 +578,8 @@ pub(super) async fn check_init() -> ResultType<()> {
 
                     lock.insert(idx, cap_display_info as u64);
                 }
+
+                *PIPEWIRE_INITIALIZED.write().unwrap() = true;
             }
         }
     }
@@ -602,6 +618,8 @@ pub(super) async fn get_displays_and_primary() -> ResultType<(Vec<DisplayInfo>, 
             Ok((cap_display_info.displays.clone(), cap_display_info.primary))
         }
     } else {
+        drop(cap_map);
+        *PIPEWIRE_INITIALIZED.write().unwrap() = false;
         bail!("Failed to get capturer display info");
     }
 }
@@ -635,6 +653,7 @@ pub fn clear() {
 
     // Reset PipeWire initialization flag to allow recreation on next init
     *PIPEWIRE_INITIALIZED.write().unwrap() = false;
+    scrap::wayland::pipewire::close_session();
 }
 
 /// Initialize the PipeWire/portal capture path from the plain (sync) video thread, so a DRM display


rustdesk:
## Review

Overall, the direction of this PR is correct. The bug analysis looks accurate: PipeWire session state and `PIPEWIRE_INITIALIZED` can become inconsistent after disconnect/reconnect on Wayland, causing:

```

Failed to get capturer display info

````

The changes improve rollback behavior and avoid marking PipeWire initialization as successful before capturer creation actually succeeds.

However, I have a few concerns that should be addressed before merging.

---

## 1. `clear()` unconditionally closes PipeWire session

In `src/server/wayland.rs`:

```rust
pub fn clear() {
    ...
    *PIPEWIRE_INITIALIZED.write().unwrap() = false;
    scrap::wayland::pipewire::close_session();
}
````

This may introduce a regression for server mode.

The logic in `pipewire.rs` already distinguishes server and non-server mode:

```rust
if !is_server_running() || rdp_info.is_support_restore_token {
    close = true;
}
```

This means server mode intentionally has different PipeWire session lifetime behavior.

However, `clear()` now always closes the session regardless of mode.

Potential scenario:

```
server mode
    |
client disconnect
    |
clear()
    |
close PipeWire session
    |
restore token/session reuse broken
```

I suggest either:

```rust
if !is_server_running() {
    scrap::wayland::pipewire::close_session();
}
```

or moving this decision logic into `pipewire::close_session()` itself so callers cannot accidentally break the lifetime rules.

---

## 2. Cleanup code with raw pointers is risky

This part:

```rust
for (_, addr) in lock.iter() {
    let cap_display_info: *mut CapDisplayInfo = *addr as _;
    unsafe {
        let _box_capturer =
            Box::from_raw((*cap_display_info).capturer.0);
        let _box_cap_display_info = Box::from_raw(cap_display_info);
    }
}
```

works in principle, but it is fragile.

Questions:

* Does `CapturerPtr` have any `Drop` implementation?
* Can `capturer.0` already be released elsewhere?
* Can `CapDisplayInfo` contain other owned resources in the future?

Because this cleanup relies on manually reconstructing ownership from raw pointers, any future change to these structures could introduce double-free or use-after-free bugs.

I suggest introducing a dedicated cleanup helper:

```rust
fn cleanup_capturers(lock: &mut HashMap<_, _>) {
    ...
}
```

or, ideally, making `CapDisplayInfo` own the capturer through a Rust-managed type instead of raw pointers.

---

## 3. Cleanup behavior is inconsistent

Currently there are multiple failure paths:

### Capturer creation failure

```rust
PIPEWIRE_INITIALIZED = false;
close_session();
```

### `get_displays_and_primary()` failure

```rust
PIPEWIRE_INITIALIZED = false;
```

but no:

```rust
close_session();
```

### `clear()`

```rust
PIPEWIRE_INITIALIZED = false;
close_session();
```

This creates different recovery behavior depending on where initialization fails.

Example:

```
PIPEWIRE_INITIALIZED = true
capturer map becomes invalid
get_displays_and_primary() fails
    |
flag reset
    |
PipeWire session still alive
    |
next init may reuse stale state
```

I think all failed initialization paths should go through the same cleanup function.

---

## 4. Moving `PIPEWIRE_INITIALIZED = true` is correct

This change is good:

Before:

```rust
PIPEWIRE_INITIALIZED = true;
create capturers...
```

After:

```rust
create all capturers...
PIPEWIRE_INITIALIZED = true;
```

The previous behavior allowed partial initialization:

```
initialized == true
but capturer creation failed
```

which is exactly the type of state that causes reconnect bugs.

No issue here.

---

## 5. Suggested refactoring

Instead of spreading cleanup logic:

```rust
PIPEWIRE_INITIALIZED = false;
close_session();
lock.clear();
```

I would suggest a single helper:

```rust
fn reset_pipewire_capture_state() {
    clear capturers;
    PIPEWIRE_INITIALIZED = false;
    close_session();
}
```

Then use it in:

* capturer creation failure
* display lookup failure
* disconnect cleanup

This makes the state machine easier to reason about.

---

## Final verdict

**Request changes**

The bug fix itself is valuable and likely fixes the reported reconnect issue. The main concerns are:

1. `clear()` closing PipeWire session unconditionally may break server-mode behavior.
2. Raw pointer cleanup should be made safer.
3. Failure cleanup paths should be unified.

After addressing these, I would be comfortable approving this PR.

rustdesk:
Close since no response

WinterSold1er:
Hi @rustdesk,

I apologize for the delayed reply on this PR as I have been very busy with work recently.

Since reopening this PR is restricted after closure, I have addressed all your review suggestions and opened a follow-up PR at #16331:
- Avoided server-mode session regression in `clear()` by respecting `try_close_session()` lifetime rules.
- Encapsulated raw pointer cleanup in a dedicated `cleanup_capturers()` helper.
- Unified failure and disconnect recovery through `reset_pipewire_capture_state()`.
- Deferred `PIPEWIRE_INITIALIZED = true` until all capturers succeed.

Thank you again for the detailed and constructive review!

rustdesk:
Please do this one.

WinterSold1er:
Updated this PR directly with all the requested changes:

1. **Server-mode session lifetime**:
   - `clear()` now calls `reset_pipewire_capture_state(false)`, which invokes `try_close_session()`. Server mode without restore token preserves its PipeWire session across disconnections as intended.
   - `try_close_session()` closes the session in non-server mode (`!is_server_running()`) or when restore token is supported.

2. **Dedicated raw pointer cleanup**:
   - Encapsulated raw pointer reconstruction into `cleanup_capturers(lock: &mut HashMap<usize, u64>)` using `.drain()` and null-pointer checks before calling `Box::from_raw`.

3. **Unified failure and disconnect recovery**:
   - Unified `check_init` loop failure, `get_displays_and_primary` empty-map detection, and `clear()` through `reset_pipewire_capture_state()`.

4. **Deferred `PIPEWIRE_INITIALIZED`**:
   - `*PIPEWIRE_INITIALIZED = true` is only set after all capturers are successfully created.

PR #16199, opened 2026-09-13
packaging: don't fail postinst on missing units, drop `pkill -f` from rustdesk.service

Elyor1977:
- `res/DEBIAN/postinst` runs under `set -e` and removed three unit files with a single `rm` without `-f`. `/usr/lib/systemd/user/rustdesk.service` usually does not exist, so `rm` returned non-zero and `dpkg --configure` failed on upgrade. Now `rm -f … || true`. Same hardening in `prerm`.
- `res/rustdesk.service`: `ExecStop=pkill -f "rustdesk --"` ran as root and killed *any* process whose command line contained that substring (including a user's `rustdesk --connect …`). Replaced with `KillMode=control-group`, which stops `--service`/`--server`/`--tray` together; removed `PIDFile` (meaningless with `Type=simple`); added `Restart=on-failure`. The `sed` in `postinst` that rewrote `pkill` → `/usr/bin/pkill` is therefore removed too.
- `prerm`: find the session user with `pgrep -f 'rustdesk +--server'` instead of parsing `ps -ef | grep`.
- `res/rpm-suse.spec` still said 1.1.9 — bumped to 1.5.0; `res/PKGBUILD` had empty `pkgdesc`/`url`/`md5sums`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux service shutdown so all related RustDesk processes stop together.
  * Added automatic service recovery after unexpected failures, with a short retry delay.
  * Improved package installation and removal reliability, including stopping services for all active non-root users.

* **Chores**
  * Updated Linux package metadata, including the project description, website, vendor details, package checksums, and package version information.

<h2><a href="https://app.greptile.com/api/retrigger?id=63700712"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge, with no outstanding correctness or repository-rule violations.

- Missing legacy unit files no longer cause package maintenance scripts to fail.
- The system service now stops only processes in its control group and restarts after failures.
- Session-specific user services are stopped for each detected non-root server owner.
- Arch and SUSE package metadata now reflects the current project and package version.
</details>

```mermaid
flowchart TD
  A[Package configuration] --> B[Remove legacy unit paths]
  B --> C[Install current system unit]
  C --> D[Reload and start systemd service]
  D --> E[rustdesk --service control group]
  E --> F[Server and tray child processes]
  G[Service stop] --> H[systemd terminates control group]
  H --> E
```
</details>

<sub>Reviews (4) · Last reviewed commit: ["PKGBUILD: keep md5sums empty, there are ..."](https://github.com/rustdesk/rustdesk/commit/bc61279f30019b7931e675b4301fbd3466d191a0)</sub>


Elyor1977:
Good point, thanks. Kept rm -f so missing units don't abort, but dropped || true and the output redirect — a genuine removal failure now stops configure and is visible in the log, so a stale /etc unit can't silently shadow the new one.

Elyor1977:
Good point, thanks. Kept rm -f so missing units don't abort, but dropped || true and the output redirect — a genuine removal failure now stops configure and is visible in the log, so a stale /etc unit can't silently shadow the new one.

Elyor1977:
Right — there is no source=() here (everything comes from $HBB), so md5sums is back to an empty array with a comment explaining why.

rustdesk:
## Review

The overall direction is good: removing the broad `pkill -f "rustdesk --"` is the right goal, and the `postinst` fix with `rm -f` is also correct.

I found one blocking issue with replacing `ExecStop` entirely by `KillMode=control-group`.

### Blocking: `KillMode=control-group` does not reliably cover the per-user `--server` / `--tray`

The new unit assumes:

```ini
# stop kills the whole control group (--service, --server and --tray children)
KillMode=control-group
```

However, on Linux the root `rustdesk --service` process starts the desktop user's `rustdesk --server` through `run_as_user()`, which uses `sudo -u <user>`.

On systems where the `sudo` PAM stack includes `pam_systemd` (for example Fedora-like systems), that login/session transition can move the per-user process into a separate user session scope under `user.slice`, outside the `rustdesk.service` cgroup.

The resulting hierarchy can look like:

```text
system.slice/rustdesk.service
  rustdesk --service

user.slice/user-1000.slice/session-*.scope
  rustdesk --server
  rustdesk --tray
```

In that case:

```bash
systemctl stop rustdesk
```

kills the system service cgroup, but does not guarantee that `--server` / `--tray` are terminated.

This is especially relevant for the RPM and Arch removal paths, which currently rely on `systemctl stop rustdesk` without another deterministic cleanup path for those per-user processes.

The old `pkill -f` is too broad and should not be restored, but the replacement needs deterministic ownership/termination of the processes started by the service.

Please verify at least the following on Ubuntu and Fedora (or another distro whose sudo PAM stack uses `pam_systemd`):

```text
1. Start rustdesk.service.
2. Log in as a normal desktop user.
3. Inspect /proc/<pid>/cgroup for --service, --server and --tray.
4. Run `systemctl stop rustdesk`.
5. Verify --service, --server and --tray are gone.
6. Verify an unrelated user-started `rustdesk --connect ...` process remains alive.
```

If the per-user server can leave the service cgroup, it needs an explicit lifecycle mechanism rather than relying solely on `KillMode=control-group`.

### Other changes

The `postinst` change is correct:

```bash
rm -f ...
```

handles missing legacy unit files without hiding real deletion failures under `set -e`.

Removing `PIDFile=` is also correct for this `Type=simple` unit.

The updated `prerm` logic that handles every distinct non-root server owner is better than the previous single-user lookup.

The current `PKGBUILD` checksum configuration is also correct: `md5sums=()` matches the absence of a `source=()` array.

I would also prefer to leave:

```ini
Restart=on-failure
RestartSec=10
```

out of this PR unless it is required for the reported packaging/service-stop issue. It changes long-running service behavior and is unrelated to replacing the unsafe `ExecStop`, so keeping it separate would reduce the regression surface.

CI and Full Flutter CI are currently `action_required` with no jobs executed, so they have not actually passed yet.

**Verdict:** Request changes for the service/process ownership issue. Once stopping `rustdesk.service` is proven to terminate only the service-owned `--server` / `--tray` processes across supported systemd/PAM environments, the rest looks good.

rustdesk:
I can accept `rm -> rm -f` only, the others will be rejected if there are no related bug reports.

rustdesk:
https://github.com/rustdesk/rustdesk/commit/82b6fc2f3b1966d1cd6c6202136ae6ff1a2999c9

Issue #12206, opened 2025-07-01
Wayland Cursor shape does not change


areluke:
### Bug Description

Remoting into a Debian with KDE Plasma Wayland machine from my Mac, the cursor never changes from the default black arrow. Not when dragging edges, dragging windows, or anything

### How to Reproduce

Remote into a macine running Debian with KDE Plasma on Wayland

### Expected Behavior

The cursor is expected to change when hovering the mouse over the edge of a window

### Operating system(s) on local (controlling) side and remote (controlled) side

osx - Debian

### RustDesk Version(s) on local (controlling) side and remote (controlled) side

1.4.0 - 1.4.0

VaradGupta23:
To resolve the “Wayland cursor shape does not change” issue in RustDesk when remoting into a Debian KDE Wayland session:

✅ Solutions:
1. Switch to X11 Session (Recommended Workaround)
Wayland has known compatibility issues with many remote desktop tools, including RustDesk.

On the Debian (remote) machine:

Log out

At login screen, click the gear icon or session selector

Choose “Plasma (X11)” instead of “Wayland”

Log in and try RustDesk again

This typically restores correct cursor behavior.

2. Install RustDesk via .deb (not Flatpak)
If you're using Flatpak on the remote system, uninstall it and install via .deb instead:


sudo apt install ./rustdesk-1.4.0.deb
Flatpak versions are sandboxed and often cause input/output issues with Wayland.

3. Try Nightly RustDesk Builds
Cursor handling under Wayland is actively being improved. You can try the nightly build:

Go to: https://github.com/rustdesk/rustdesk/releases

Look for the latest pre-release / nightly build

Download and install on both ends

4. Wayland Limitations – No Full Fix Yet
There’s no full fix inside RustDesk yet for advanced Wayland cursor support. This is a known upstream limitation, as Wayland doesn’t expose cursor shape change APIs easily to remote apps.

So switching to X11 is the only fully working solution for now.

aboooo12:
The issue still persists in the latest version 1.4.1. When connecting from macOS 15.5 to Windows 11, the mouse cursor still appears as a blue spinning circle after connection.

stevelcb:
Working almost perfectly here. Kubuntu 26.04 at both ends: 
https://github.com/rustdesk/rustdesk/releases/download/nightly/rustdesk-unattended-wayland-1.4.9-x86_64.deb

simseb:
[...]

PR #16213, opened 2026-09-14
fix(cursor): correct native cursor sizing and validate received images

fufesou:
## Summary

Correct oversized native cursors on HiDPI controllers and prevent thin cursors from being enlarged by the minimum-size rule. A 4×64 cursor at 50% becomes 2×32 instead of 12×192.

- Read the controller's live DPR. On Windows, Custom zoom accounts for the selected Linux display's scale and the active video renderer.
- Use the long-edge minimum for native and Web cursors. Keep the generated raster dimensions, registration dimensions, and cache identity together so fractional scaling cannot register an undersized pixel buffer.
- Preserve translucent cursor colors from Linux and Windows peers when resizing on macOS/Linux controllers. Decode the correctly encoded PNG as the resize source and use alpha-weighted averaging when shrinking it, so transparent pixels do not darken colored edges. This includes the unzoomed inverse-DPR path in Adaptive/Custom view. Windows controllers retain their existing raw BGRA registration path.
- Validate received dimensions, hotspots, and decompressed RGBA length before UI dispatch; bound Zstd decompression and avoid decompressing twice. The direct Zstd dependency uses the already-resolved version.

This release fix covers controller sizing and received-image validation. The cursor plugin, host capture, protocol, hotspot calculations, and “Show remote cursor” overlay behavior remain outside this change. Windows Custom view of Linux All Displays retains its existing cursor scale of 1.

## Related

- https://github.com/rustdesk/rustdesk/discussions/15363 — direct report of oversized cursors from macOS to KDE in Adaptive/Custom view.
- https://github.com/rustdesk/rustdesk/discussions/14334 — related report about cursor size not changing; its exact scaling conditions are unspecified.
- https://github.com/rustdesk/rustdesk/pull/16144 — broader cursor work from which this release fix was split.
- https://github.com/rustdesk/rustdesk/issues/12206 — separate report of Wayland cursor shapes not updating, not claimed fixed here.
- https://github.com/rustdesk/rustdesk-server-pro/issues/847 and https://github.com/rustdesk/rustdesk-server-pro/discussions/842 — related cursor-visibility reports/requests, outside this sizing fix.

## Tests

1. Local 100%, 200% -> peer 100%, 200%
2. Scale original, Scale adaptive - Zoom cursor on/off, Scale custom 100, 200, 400 - Zoom cursor on/off
3. zoom + peer is Linux
     - !useTexture && useScrollbar
     - texture || !useScrollbar

- [x] Windows -> Windows
- [x] Windows -> X11
- [x] Windows -> GNOME
- [x] Windows -> KDE/Plasma
- [x] Windows -> macOS
- [x] X11 -> X11
- [x] X11 -> GNOME
- [x] X11 -> KDE Plasma
- [x] X11 -> Windows
- [x] X11 -> macOS
- [x] GNOME -> GNOME
- [x] GNOME -> X11
- [x] GNOME -> KDE Plasma
- [x] GNOME -> Windows
- [x] GNOME -> macOS
- [x] KDE Plasma -> GNOME
- [x] KDE Plasma -> X11
- [x] KDE Plasma -> KDE Plasma
- [x] KDE Plasma -> Windows
- [x] KDE Plasma -> macOS
- [x] macOS -> GNOME
- [x] macOS -> X11
- [x] macOS -> KDE Plasma
- [x] macOS -> Windows
- [x] macOS -> macOS

## Limitation

1. Current cursor uses the size of the controlled side; the default side may not match the local cursor.
  - Linux 300% -> macOS, Scale adaptive, Zoom cursor off. the cursor is a bigger
2. -> Windows, multiple displays. Zoom cursor, Select one display, the scale is only correct on the first display.
3. Sciter -> , the cursor may be huge.

## Cursor alpha check assets

https://github.com/user-attachments/files/32200863/cursor-alpha-check.html

## Summary by CodeRabbit

- **New Features**
  - Improved desktop cursor sizing across Windows, Linux, and macOS, including multi-display and high-DPI setups.
  - Added support for correctly sizing and padding non-square Linux cursors.
  - Cursor scaling now adapts to zoom levels, display scale, and raster dimensions.
  - Improved preservation of cursor appearance when resizing, including transparent details.

- **Bug Fixes**
  - Preserved scrollbar and edge-scroll positions during display-scale changes.
  - Improved canvas and input updates when device-pixel ratios change.
  - Invalid or malformed remote cursor data is now safely rejected instead of being applied.

<h2><a href="https://app.greptile.com/api/retrigger?id=63828761"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; no actionable new issue or outstanding previous finding remains.

The PR corrects native cursor scaling across desktop display modes and adds validation to the remote cursor receive path.
- Tracks live controller DPR and remote-display scaling when selecting native cursor dimensions.
- Keeps generated raster dimensions, registration dimensions, hotspots, and cache keys consistent.
- Preserves straight-alpha cursor colors during downscaling and pads Linux native cursors safely.
- Validates cursor dimensions, hotspots, compressed size, decompressed length, and decompression bounds before dispatching cursor data to either UI.
- Adds focused Flutter cursor-sizing tests and runs them in the Flutter build workflow.
</details>

```mermaid
flowchart LR
    A[Remote CursorData] --> B[Validate dimensions and hotspot]
    B --> C[Bound and decompress Zstd RGBA]
    C --> D{UI implementation}
    D --> E[Flutter cursor model]
    D --> F[Sciter cursor handling]
    E --> G[Apply DPR, display, and view scaling]
    G --> H[Resize and preserve alpha]
    H --> I[Register native cursor]
```
</details>

<sub>Reviews (8) · Last reviewed commit: ["fix(cursor): use \`1.0\` instead of \`1.0/d..."](https://github.com/rustdesk/rustdesk/commit/b87fd895a801f3e1ec6b0e44cebc553b4d5f6fba)</sub>

### DIFF
--- a/flutter/lib/desktop/pages/remote_page.dart
+++ b/flutter/lib/desktop/pages/remote_page.dart
@@ -1410,8 +1410,8 @@ class CursorPaint extends StatelessWidget {
       }
     }
 
-    double x = (m.x - hotx) * c.scale + cx;
-    double y = (m.y - hoty) * c.scale + cy;
+    double x = m.x * c.scale + cx - hotx;
+    double y = m.y * c.scale + cy - hoty;
     double scale = 1.0;
     final isViewOriginal = c.viewStyle.style == kRemoteViewStyleOriginal;
     if (zoomCursor.value || isViewOriginal) {
--- a/flutter/lib/desktop/pages/remote_page.dart
+++ b/flutter/lib/desktop/pages/remote_page.dart
@@ -1101,6 +1101,9 @@ class _ImagePaintState extends State<ImagePaint> {
     final m = Provider.of<ImageModel>(context);
     var c = Provider.of<CanvasModel>(context);
     final s = c.scale;
+    // CanvasModel caches the DPR and only refreshes it when the view style
+    // changes, so read it live to follow the window across monitors.
+    final dpr = MediaQuery.devicePixelRatioOf(context);
 
     bool isViewAdaptive() => c.viewStyle.style == kRemoteViewStyleAdaptive;
     bool isViewOriginal() => c.viewStyle.style == kRemoteViewStyleOriginal;
@@ -1122,7 +1125,7 @@ class _ImagePaintState extends State<ImagePaint> {
                 // pixels, so an unzoomed cursor must be shrunk by the DPR to
                 // keep 1 remote px == 1 physical px, the size Original view
                 // already renders it at.
-                cursorScale = 1.0 / c.devicePixelRatio;
+                cursorScale = 1.0 / dpr;
               }
             }
             return cursorScale;
@@ -1420,8 +1423,17 @@ class CursorPaint extends StatelessWidget {
       scale = c.scale;
     } else if (!isWindows) {
       // Keep the painted cursor the same physical size as the native one
-      // built by getCursorScale() above.
-      scale = 1.0 / c.devicePixelRatio;
+      // built by getCursorScale() above, including its min-size clamp.
+      scale = 1.0 / MediaQuery.devicePixelRatioOf(context);
+      final image = m.image ?? preDefaultCursor.image;
+      if (scale != 1.0 &&
+          image != null &&
+          ((image.width * scale).toInt() < kMinCursorSize ||
+              (image.height * scale).toInt() < kMinCursorSize)) {
+        final sw = kMinCursorSize / image.width;
+        final sh = kMinCursorSize / image.height;
+        scale = sw < sh ? sh : sw;
+      }
       x = (m.x * c.scale + cx) / scale - hotx;
       y = (m.y * c.scale + cy) / scale - hoty;
     }
--- a/flutter/lib/desktop/pages/remote_page.dart
+++ b/flutter/lib/desktop/pages/remote_page.dart
@@ -1117,6 +1117,12 @@ class _ImagePaintState extends State<ImagePaint> {
             } else {
               if (zoomCursor.value || isViewOriginal()) {
                 cursorScale = s;
+              } else {
+                // NSCursor and GdkCursor treat the bitmap size as logical
+                // pixels, so an unzoomed cursor must be shrunk by the DPR to
+                // keep 1 remote px == 1 physical px, the size Original view
+                // already renders it at.
+                cursorScale = 1.0 / c.devicePixelRatio;
               }
             }
             return cursorScale;
@@ -1412,6 +1418,12 @@ class CursorPaint extends StatelessWidget {
       x = m.x - hotx + cx / c.scale;
       y = m.y - hoty + cy / c.scale;
       scale = c.scale;
+    } else if (!isWindows) {
+      // Keep the painted cursor the same physical size as the native one
+      // built by getCursorScale() above.
+      scale = 1.0 / c.devicePixelRatio;
+      x = (m.x * c.scale + cx) / scale - hotx;
+      y = (m.y * c.scale + cy) / scale - hoty;
     }
 
     return CustomPaint(
--- a/flutter/lib/models/model.dart
+++ b/flutter/lib/models/model.dart
@@ -2884,7 +2884,7 @@ class CursorData {
     if (scale != 1.0) {
       // Update data if scale changed.
       final tgtWidth = (width * scale).toInt();
-      final tgtHeight = (width * scale).toInt();
+      final tgtHeight = (height * scale).toInt();
       if (tgtWidth < kMinCursorSize || tgtHeight < kMinCursorSize) {
         double sw = kMinCursorSize.toDouble() / width;
         double sh = kMinCursorSize.toDouble() / height;


rustdesk:
# Review of PR #16213

PR: https://github.com/rustdesk/rustdesk/pull/16213
Head reviewed: `47b87bb24a6f3d961fce857dc73e0de9bc2477e2`

Overall, the direction of this PR looks good. The receive-side cursor validation is a meaningful improvement, and the raster/cache changes make sense.

However, I would not merge it yet. I found two concrete cursor-scaling correctness issues that should be fixed first.

## 1. [P1] Linux peer + All Displays + non-texture renderer uses the wrong cursor scale

The PR introduces:

```dart
final _cursorDisplayScale = 1.0.obs;
```

For Linux peers in `All Displays` mode, cursor scaling depends on it:

```dart
if (peer.isPeerLinux && peer.pi.currentDisplay == kAllDisplayValue) {
  if (!zoomCursor.value || c.viewStyle.style == kRemoteViewStyleOriginal) {
    final scale = 1.0 / _cursorDisplayScale.value;
    return isWindows ? scale : scale / dpr;
  }

  final scale = c.scale / _cursorDisplayScale.value;
  return isWindows ? scale * dpr : scale;
}
```

But `_cursorDisplayScale` is only updated from `_trackCursorDisplay()`:

```dart
Widget _trackCursorDisplay(Widget child, Display display) {
  final peer = widget.ffi.ffiModel;
  if (!isDesktop ||
      !peer.isPeerLinux ||
      peer.pi.currentDisplay != kAllDisplayValue) {
    return child;
  }

  return MouseRegion(
    onEnter: (_) => _cursorDisplayScale.value = display.scale,
    onHover: (_) => _cursorDisplayScale.value = display.scale,
    child: child,
  );
}
```

And `_trackCursorDisplay()` is only applied to the texture-render path:

```dart
child: _trackCursorDisplay(
  Obx(() => Texture(
        textureId: textureId.value,
        filterQuality:
            isViewOriginal ? FilterQuality.none : FilterQuality.low,
      )),
  displays[i],
),
```

The non-texture renderer never updates `_cursorDisplayScale`, so it remains `1.0`.

At the same time, the PR changes the non-texture Linux video rendering scale to:

```dart
double sizeScale = s;

if (widget.ffi.ffiModel.isPeerLinux) {
  final displays = widget.ffi.ffiModel.pi.getCurDisplays();
  if (displays.isNotEmpty) {
    sizeScale = s / displays[0].scale;
  }
}
```

Therefore, for a Linux Wayland peer with `display.scale == 2`:

```text
video scale  = s / 2
cursor scale = s / 1
```

The cursor becomes 2x too large relative to the rendered image.

This is a real configuration because the Linux server reports display scale for multi-output Wayland setups. Single-display Linux intentionally keeps scale at `1.0`, but multi-display Wayland does not.

This also conflicts with the PR test matrix, which explicitly claims coverage for:

```text
zoom + peer is Linux
- !useTexture && useScrollbar
- texture || !useScrollbar
```

### Suggested fix

The effective display scale should not depend exclusively on the texture child `MouseRegion`.

At minimum, the non-texture `All Displays` path should use the same display scale as the video renderer.

For example, if non-texture rendering uses:

```dart
displays[0].scale
```

then cursor scaling should use the same value.

Longer-term, it may be cleaner to make the effective remote display density explicit rather than storing it only as hover state.

---

## 2. [P1/P2] macOS peer + All Displays drops Retina display scale

For a selected macOS display, `_getCursorScaleForDisplay()` correctly applies the peer display scale:

```dart
if (peer.pi.platform == kPeerPlatformMacOS) {
  return scale * displays.first.scale;
}
```

This is needed because macOS cursor capture sends `NSImage.size`, which is expressed in points, while the captured display/video can use backing-pixel dimensions.

However, `_getCursorScaleForDisplay()` starts with:

```dart
if (peer.pi.currentDisplay == kAllDisplayValue) return scale;
```

So in `All Displays` mode, macOS display scale is skipped entirely.

Example:

```text
remote macOS Retina cursor:
NSImage.size = 32 points
display.scale = 2
rendered equivalent = 64 backing pixels
```

Selected-display mode:

```text
32 * 2 = 64
```

Correct.

All Displays mode:

```text
32 * 1 = 32
```

The cursor becomes half-sized.

Linux `All Displays` at least has the `_cursorDisplayScale` mechanism to track the currently hovered monitor. macOS does not use that mechanism at all.

### Suggested fix

The display-density handling could be generalized into something like:

```text
effectivePeerDisplayScale
```

Then use it consistently:

```text
Linux  -> divide by peer display scale
macOS  -> multiply by peer display scale
Windows -> no peer-display correction
```

For `All Displays` with texture rendering, the active value can follow the display currently under the local cursor.

---

## 3. [P2/P3] The new cursor tests are not actually exercised by current PR CI

The PR adds a fairly substantial:

```text
flutter/test/cursor_size_test.dart
```

But the current `Full Flutter CI` calls the reusable Flutter build workflow, and that workflow does not run `flutter test`.

So the green CI proves that the project builds on the tested platforms, but it does not prove that this new test file passes.

This matters because the new tests are being used as evidence for many subtle scaling behaviors.

The current `_checkView()` tests also force:

```dart
ffi.ffiModel.pi.currentDisplay = 1;
```

so they only cover selected-display behavior and do not cover either `All Displays` bug above.

### Suggested CI addition

At minimum:

```bash
flutter test test/cursor_size_test.dart
```

And add regression cases for:

```text
Linux + All Displays + useTextureRender=false + display.scale=2
macOS + All Displays + display.scale=2
```

---

## 4. [P3] Raster dimensions and hotspot can use slightly different effective scales

The PR now computes raster dimensions with `ceil()`:

```dart
final targetWidth = (width * scale).ceil();
final targetHeight = (height * scale).ceil();
```

But hotspot scaling still uses the original floating-point scale:

```dart
hotx = hotxOrigin * scale;
hoty = hotyOrigin * scale;
```

For example:

```text
width = 64
scale = 1.0000000000000002

targetWidth = 65
```

The actual resized image scale is effectively:

```text
65 / 64 = 1.015625
```

but the hotspot is still scaled by approximately `1.0`.

That can create a sub-pixel to approximately 1-pixel hotspot/artwork mismatch around raster-boundary transitions.

This is probably not a merge blocker, but it conflicts somewhat with the PR's goal of keeping generated raster geometry, registration dimensions, and cache identity synchronized.

It would be cleaner if hotspot coordinates were derived from the actual raster dimensions, for example from:

```text
targetWidth / width
targetHeight / height
```

rather than only the requested floating-point scale.

---

## Receive-side validation looks good

I think the new cursor receive validation is a solid improvement.

The PR now validates before UI dispatch:

```text
width / height <= 512
hotspot inside cursor bounds
checked RGBA size calculation
compressed input upper bound
bounded Zstd decompression
exact decoded RGBA length
```

It also decompresses the cursor only once and then passes validated raw RGBA to Flutter/Sciter.

Previously, the generic `hbb_common::compress::decompress()` path allowed up to 256 MiB of decompressed output. This PR reduces a single accepted cursor to at most:

```text
512 * 512 * 4 = 1 MiB
```

That is a substantial improvement.

CodeRabbit also raised repeated-cursor-work/rate-limiting concerns. I would not treat that as a blocker for this PR.

The receive loop already performed synchronous decompression and UI serialization before this change. This PR significantly reduces the maximum per-cursor cost. A malicious peer can still send many changing cursor IDs and consume CPU, but that is broader per-message abuse hardening rather than a regression introduced here.

---

## Merge recommendation

I would fix these two issues before merging:

1. **Linux + All Displays + non-texture renderer must use the same display scale for video and cursor.**
2. **macOS + All Displays must preserve the active Retina display scale.**

I would also add actual CI execution for:

```bash
flutter test test/cursor_size_test.dart
```

The hotspot/raster rounding mismatch is worth cleaning up but is not as important as the first two issues.

Other than those points, I did not find a reason to reject the overall design of the PR. The receive validation, single decompression, raster-dimension/cache synchronization, alpha-preserving shrink, and Linux cursor padding are all reasonable changes.

fufesou:
https://github.com/rustdesk/rustdesk/pull/16213#issuecomment-5710785648

1. **Linux All Displays with non-texture rendering**

   All Displays forces texture rendering, even when the texture option is off. `forceTextureRender` returns true for `kAllDisplayValue`, and both layout branches check `useTextureRender || forceTextureRender`. The display widgets therefore go through `_trackCursorDisplay()`.

   The reported calculation combines two paths that are not used together. There is no software-rendered All Displays path here that leaves `_cursorDisplayScale` at 1 while drawing the video at `s / 2`.

   https://github.com/rustdesk/rustdesk/blob/47b87bb24a6f3d961fce857dc73e0de9bc2477e2/flutter/lib/models/model.dart#L4281

   https://github.com/rustdesk/rustdesk/blob/47b87bb24a6f3d961fce857dc73e0de9bc2477e2/flutter/lib/desktop/pages/remote_page.dart#L1149-L1187

2. **macOS All Displays and Retina scale**

   The macOS host already disables Retina capture when two or more video services are active. With Retina capture disabled, `Display::scale()` returns 1, and the capture dimensions use that effective scale.

   The example assumes the All Displays stream still has the selected display's 2x capture dimensions. That is not what the host sends once both displays are active. The selected Retina display needs the multiplier; the normal multi-display capture does not. Multiplying by the hardware Retina factor in that case would make the cursor too large relative to the video.

   https://github.com/rustdesk/rustdesk/blob/47b87bb24a6f3d961fce857dc73e0de9bc2477e2/src/server.rs#L528-L537

   https://github.com/rustdesk/rustdesk/blob/47b87bb24a6f3d961fce857dc73e0de9bc2477e2/libs/scrap/src/quartz/display.rs#L37-L81

3. **Flutter test coverage in CI**

   Agreed. Full Flutter CI runs the build workflow, which does not run `flutter test`. A green build does not establish that `cursor_size_test.dart` passes. `_checkView()` also covers selected displays only.

   Running the test in CI is a reasonable suggestion. The proposed All Displays cases need to account for the behavior above: turning off the texture option still uses forced texture rendering, and a macOS display's hardware scale is not necessarily its active capture scale.

   https://github.com/rustdesk/rustdesk/blob/47b87bb24a6f3d961fce857dc73e0de9bc2477e2/.github/workflows/flutter-ci.yml#L20-L24

4. **Raster dimensions and hotspot rounding**

   Agreed that there can be a small mismatch. I checked the 64-to-65 example with the pinned `image` 4.3.0 dependency: a marker at `(63, 63)` moves to `(64, 64)`, while the current hotspot still truncates to `(63, 63)`.

   Similar rounding differences existed with the old truncated raster dimensions, but `ceil()` exposes this particular boundary. Using the actual raster ratios makes sense, though it also needs to account for native hotspot rounding: `63 * 65 / 64` is `63.984375`, which still becomes 63 if truncated. I agree this is a minor geometry issue rather than a merge blocker. This check used the image library, not a native cursor session.

   https://github.com/rustdesk/rustdesk/blob/47b87bb24a6f3d961fce857dc73e0de9bc2477e2/flutter/lib/models/model.dart#L2901-L2934

5. **Receive-side validation and repeated cursor updates**

   Agreed. This PR bounds each cursor payload before UI dispatch and removes the second decompression. It does not bound the total work from repeated valid messages. Cursor ID and per-ID update limits are being handled separately on `fix/cursor-data-limits`; they are not part of this PR.

rustdesk:
## Crash risk

One crash/OOM risk is worth fixing before merge.

The receive side is now safer: cursor dimensions are limited to `512x512`, Zstd decompression is bounded, and malformed data returns an error instead of reaching the UI.

The remaining risk is the **scaled native cursor size**. `targetWidth/targetHeight` have no upper bound, and Linux additionally pads non-square cursors to a square.

For example, a valid `1x512` cursor at `1000%` becomes roughly:

```text
10x5120
```

On Linux, square padding can turn that into:

```text
5120x5120
```

which is about 100 MiB for one raw RGBA buffer, before temporary decode/encode buffers. Larger effective scales can be worse.

### Suggested fix

Add a hard upper bound to the final native cursor raster, e.g. `1024` or `2048`, and clamp the scale while preserving aspect ratio.

This does not require changing the current scale logic and should not affect normal cursors.

rustdesk:
## Review finding: Linux controller + macOS peer + Zoom cursor off

I found one remaining scaling issue that I think is worth fixing before merge.

The affected branch is in `flutter/lib/desktop/pages/remote_page.dart`:

```dart
if (!zoomCursor.value &&
    peer.pi.platform == kPeerPlatformMacOS &&
    (c.viewStyle.style == kRemoteViewStyleAdaptive ||
        c.viewStyle.style == kRemoteViewStyleCustom)) {
  return isWindows ? dpr : (isMacOS ? 1.0 : 1.0 / dpr);
}
```

For a Linux controller, this returns `1.0 / dpr`.

I believe that is incorrect for a macOS peer.

### Why

The macOS host sends cursor dimensions from `NSImage.size`, i.e. point dimensions:

```rust
let img: id = msg_send![c, image];
let size: NSSize = msg_send![img, size];

CursorData {
    width: size.width as _,
    height: size.height as _,
    ...
}
```

On Linux, `flutter_custom_cursor` decodes the PNG into a `GdkPixbuf` and calls:

```cpp
gdk_cursor_new_from_pixbuf(display, pixbuf, hot_x, hot_y);
```

GDK then applies the monitor scale when creating the native X11 cursor.

I verified this against GTK 3.24 on X11/Xvfb with `GDK_SCALE=2` by creating actual GDK cursors and reading back the resulting native X cursor with `XFixesGetCursorImage()`:

```text
GDK input image    Native X cursor
16x16              32x32
32x32              64x64
1024x1024          2048x2048
```

So for a macOS cursor whose source size is `32x32` points on a Linux controller at DPR 2:

```text
Current code:

32 * (1 / 2) = 16
GDK then applies 2x
=> 32 physical pixels
=> 16 logical pixels
```

The cursor therefore becomes half its intended logical size.

If the source raster remains `32x32`:

```text
32 passed to GDK
GDK applies 2x
=> 64 physical pixels
=> 32 logical pixels
```

which preserves the macOS cursor's point dimensions.

### Suggested fix

For this specific `macOS peer + Zoom cursor off + Adaptive/Custom` branch:

```diff
- return isWindows ? dpr : (isMacOS ? 1.0 : 1.0 / dpr);
+ return isWindows ? dpr : 1.0;
```

I would **not** remove `/ dpr` from the other Linux-controller branches. Those paths deal with different source-dimension semantics.

### Test issue

The current test reproduces the same assumption in its expected value:

```dart
Platform.isWindows
    ? dpr
    : (Platform.isMacOS ? 1.0 : 1.0 / dpr)
```

so running this test on Linux would still pass even though the native GDK behavior disagrees.

A better regression test would express the invariant directly:

> For a macOS peer in Adaptive/Custom mode with Zoom cursor disabled, changing the Linux controller DPR must not change the cursor's logical size.

---

## Note on the 1024 raster limit

The earlier unbounded native-raster/OOM issue is fixed by `_maxRasterSize = 1024`.

For example:

```text
source: 1x512
requested scale: 10
clamped scale: 1024 / 512 = 2
Dart raster: 2x1024
Linux padded raster: 1024x1024
```

So the Dart-side cursor raster is bounded to about 4 MiB of RGBA.

However, `4 MiB` should not be treated as the final native cursor memory bound on Linux.

GDK applies the monitor scale afterwards. With a `1024x1024` input:

```text
GDK_SCALE=2 -> native 2048x2048 -> ~16 MiB RGBA
GDK_SCALE=3 -> native 3072x3072 -> ~36 MiB RGBA
GDK_SCALE=4 -> native 4096x4096 -> ~64 MiB RGBA
```

This does **not** reopen the original unbounded-allocation bug: increasing RustDesk zoom can no longer grow the Dart raster indefinitely.

It only means that the current `1024` limit is a bound on the raster passed into the native cursor API, not a strict bound on the final physical native cursor allocation.

fufesou:
https://github.com/rustdesk/rustdesk/pull/16213#issuecomment-5724390088

Addressed in b87fd89

But the mouse cursor currently appears somewhat large, especially when the window is not fullscreen or maximized.

PR #16273, opened 2026-09-18
fix: suppress heartbeat HTTP retry and TCP proxy logs

21pages:
Repeated heartbeat TLS retries and TCP proxy fallbacks can flood the log, while errors returned to the heartbeat caller were previously ignored. Throttle diagnostics for URLs whose parsed path ends in `/api/heartbeat`, and retain a periodic `Heartbeat failed` warning when `post_request()` returns an error.

In `src/common.rs`, `api_log!` applies throttling to TLS retries, TCP proxy requests, and fallback diagnostics, including those in `post_request_with_status()`. Each log site reports its first occurrence immediately, then at most once every 600 seconds, with repeated occurrences counted in `(xN)`. The caller warning in `src/hbbs_http/sync.rs` has its own throttle. Both modules use `API_LOG_INTERVAL`, defined in `common.rs`.

The runtime changes are limited to logging in these two files. Non-heartbeat logging and request, retry, fallback, and response handling retain their existing behavior.

Validation:

- `cargo check --lib --features flutter --locked --offline` passed.
- `cargo test --lib --features flutter --locked --offline common::tests::test_should_throttle_log -- --exact` passed. Covers heartbeat URLs, query strings, path prefixes, similar paths, non-heartbeat paths, and invalid URLs.
- Manually verified independent 600-second throttling for retry, fallback, and terminal failure logs; normal sysinfo/audit logging; and web console disconnect.

## Summary by CodeRabbit

- **Bug Fixes**
  - Reduced repeated warning and debug messages for heartbeat requests by throttling related network and retry logs.
  - Heartbeat request failures now produce warnings at most once every 10 minutes, helping reduce log noise during connectivity issues.
  - TCP proxy, HTTP fallback, and TLS retry messages for heartbeat requests are now consistently throttled.
  - Logging behavior for non-heartbeat requests remains unchanged, preserving existing diagnostic details for other traffic.

<h2><a href="https://app.greptile.com/api/retrigger?id=65957967"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; the prior loss of terminal heartbeat diagnostics has been addressed without introducing a new actionable issue.

The PR reduces repetitive heartbeat diagnostics while retaining periodic visibility into terminal failures.
- Detects heartbeat requests by parsed URL path and throttles their retry, fallback, and TCP-proxy logs to once every ten minutes.
- Leaves logging for non-heartbeat API requests unchanged.
- Emits a throttled warning when the heartbeat request ultimately fails.
- Moves the shared throttle interval into `common.rs` so both request handling and heartbeat synchronization can use it.
</details>

<sub>Reviews (3) · Last reviewed commit: ["refactor: move API log interval to commo..."](https://github.com/rustdesk/rustdesk/commit/c01fa963f48556aab819922513119174aad78676)</sub>


rustdesk:
## Finding

### [P2] Keep a throttled terminal heartbeat failure signal

This PR suppresses not only the repetitive TLS retry and TCP fallback logs, but also the only diagnostics emitted when a heartbeat keeps failing.

The heartbeat caller currently ignores `post_request()` errors:

```rust
if let Ok(s) = crate::post_request(url.clone(), v.to_string(), "").await {
    ...
}
```

So if the HTTP request keeps failing and the TCP proxy fallback also fails, the heartbeat failure becomes completely silent.

Suppressing the individual retry/fallback logs makes sense, but the final failure should remain observable. Prefer keeping the transport retry logs silent and adding a rate-limited warning at the heartbeat caller, for example using `hbb_common::throttled_log!`.

This would avoid log spam while still making persistent API/server connectivity problems diagnosable.

## Other review notes

### `post_request_with_status` warning suppression

CodeRabbit noted that `post_request_with_status()` does not suppress heartbeat fallback warnings.

The code observation is correct, but this is not currently a real heartbeat issue. The heartbeat path uses:

```text
heartbeat
  -> post_request()
  -> with_tcp_proxy_fallback()
```

`post_request_with_status()` is currently used by the audit posting path, not by heartbeat.

So this is at most a consistency/future-proofing issue and should not block this PR.

### `is_heartbeat_url()`

The helper looks fine:

```rust
fn is_heartbeat_url(url: &str) -> bool {
    url::Url::parse(url)
        .map(|parsed| parsed.path().ends_with("/api/heartbeat"))
        .unwrap_or(false)
}
```

It correctly handles query strings because `parsed.path()` excludes the query.

The extra URL parsing happens only on retry/failure paths and is insignificant for heartbeat traffic.

A small unit test would still be useful:

```rust
assert!(is_heartbeat_url("https://example.com/api/heartbeat"));
assert!(is_heartbeat_url(
    "https://example.com/api/heartbeat?token=secret"
));
assert!(!is_heartbeat_url(
    "https://example.com/api/heartbeat2"
));
assert!(!is_heartbeat_url(
    "https://example.com/api/sysinfo"
));
```

## Overall

I did not find any correctness, memory, or performance regression in the request/fallback behavior itself.

CI and Full Flutter CI are passing.

The main issue is observability: persistent heartbeat failures should be throttled rather than made completely invisible.

rustdesk:
## Review

I reviewed the current head `e3f33c6`, not the earlier version reviewed by Greptile.

### Finding

#### [P3 / non-blocking] Heartbeat throttling is per log site, not per heartbeat request

`throttled_log!` creates an independent static throttle for each macro expansion.

As a result, these log sites are throttled independently:

* `tcp_proxy_request()` — `Sending ... via TCP proxy`
* `with_tcp_proxy_fallback()` — HTTP fallback warning
* `with_tcp_proxy_fallback()` — `TCP proxy fallback also failed`
* `post_request_()` — invalid-certificate retry
* `post_request_()` — native-tls retry
* `sync.rs` — final `Heartbeat failed`

For example, if the API port is incorrect, the first failed heartbeat may still produce several log lines at once. During a persistent failure, another group of messages may appear every 10 minutes because each call site has its own throttle.

This does not affect request or fallback behavior, and it still greatly reduces the previous 15-second log spam, so I would not block the PR on this.

However, the test description in the PR:

> heartbeat logs no longer appear

is no longer accurate. The logs still appear; they are throttled.

If the intention is to minimize heartbeat log noise as much as possible, I would prefer:

```text
heartbeat transport retry/fallback logs
        ↓
suppress completely

final crate::post_request() error
        ↓
Heartbeat failed: ...
throttle to once every 10 minutes
```

In other words, heartbeat requests would only keep the final throttled diagnostic in `sync.rs`, while non-heartbeat API requests would retain all existing transport-level logs.

### Other notes

The previous Greptile finding that persistent heartbeat failures could become completely silent has been fixed. `sync.rs` now emits a throttled final warning when `post_request()` returns an error, which preserves observability without restoring the previous log spam.

`should_throttle_log()` looks correct. Parsing the URL and checking `parsed.path()` avoids false matches in query strings, and the added unit tests cover the relevant cases.

Applying the same throttling behavior to `post_request_with_status()` is also reasonable for consistency, even though the current heartbeat path does not use that function.

One minor structural point: `common.rs` imports `API_LOG_INTERVAL` from `hbbs_http::sync`:

```rust
hbbs_http::sync::API_LOG_INTERVAL
```

This creates a slightly awkward dependency from common HTTP transport code on the heartbeat sync module. It is not a functional problem and I would not block the PR for it. If this constant is intended to be shared HTTP logging policy, it could eventually live in `common.rs` or another shared HTTP module.

## Overall

I did not find any blocking correctness, safety, or performance regression in the current version.

The original observability issue has been addressed, and the request/fallback behavior itself is unchanged.

I think this is fine to merge once CI is green.

I would only suggest updating the PR test description from:

> heartbeat logs no longer appear

to something like:

> heartbeat retry/fallback logs are throttled and no longer repeat on every heartbeat

21pages:
Thanks for the review. Both follow-ups are now addressed:

- The PR description explicitly states that each log site emits immediately on its first occurrence, then independently throttles subsequent messages to once every 600 seconds.
- `API_LOG_INTERVAL` was moved to `common.rs` in `c01fa963f`.

Per-site throttling is intentional to retain useful transport diagnostics. There is also a reason to keep the TCP fallback failure warning: with the existing return behavior, an HTTP 5xx response followed by a failed TCP fallback can still result in `Ok(body)`. In that case, the final `Heartbeat failed` warning in `sync.rs` would not fire.

I’d therefore prefer to retain the current throttled diagnostics while keeping this PR limited to logging changes.

Issue #609, opened 2022-05-25
Message being typed is cleared when remote message is received


tytouf:
**Describe the bug you encountered:**

When chatting, I start writing a new message and during the typing a remote message is received and the entry box gets cleared.

**What did you expect to happen instead?**

Do not clear the entry box while I'm typing.

**How did you install `RustDesk`?**

From website Download button.

**RustDesk version and environment**

On Ubuntu, both remote and client version 1.1.9

asur4s:
I have reproduced the bug, let me try. Thanks

rustdesk:
Great

asur4s:
It has been resolved in version 1.20. The version can be obtained by compiling from source.
![bug](https://user-images.githubusercontent.com/99897242/170242812-c2007bc7-3b76-4ab4-8d21-4899df19ecd6.gif)

rustdesk:
@asur4s Good job!

rustdesk:
@asur4s https://github.com/rustdesk/rustdesk/issues/567 and https://github.com/rustdesk/rustdesk/issues/406 are both troublesome international keyboard issues, wondering if you have interests.

rustdesk:
And we also wanna rewrite [pynput](https://github.com/rustdesk/rustdesk/blob/master/pynput_service.py) with Rust, and prepare for wayland support (uinput does support wayland).

asur4s:
Nice, It is not friendly for newbie to solve "good first issue" in the repo. 

e.g. #568 #485

asur4s:
> we also wanna rewrite [pynput](https://github.com/rustdesk/rustdesk/blob/master/pynput_service.py) with Rust

Ok, I will try. You can try to write these plans in [milestone](https://docs.github.com/en/enterprise-cloud@latest/issues/using-labels-and-milestones-to-track-work/about-milestones).

rustdesk:
> milestone

Good suggestion, it is what I wanted to try.

tytouf:
Thanks @asur4s for the prompt reply and fix.

glasge:
The defect is observed again. Windows 11 platform, rustdesc version 1.4.5.
The x86-64 version works correctly in both directions.
The x86-32 version, which is on sciter, works correctly only in one direction. 
When sending from the slave to the master, the unsent message is deleted from the master.
When sending from the master to the slave, everything is fine - the unsent message is not deleted from the slave.

rustdesk:
I do not understand, a video please.

[...]

PR #16278, opened 2026-09-18
Remote usb

AzimovIz:
## Summary

Adds a new connection type, RemoteUsb, that tunnels USB/IP traffic through
an existing RustDesk session on Linux, in both directions:

- **Pull**: the controller attaches one of the peer's shared USB devices
  into itself.
- **Push**: the controller shares one of its own local devices and pushes
  it into the peer.

Both directions use the system `usbip` CLI on each end and relay the
USB/IP TCP stream as `UsbChannel` frames inside the session, reusing the
existing on-demand root-elevation prompt for `usbip bind`/`attach`.

For working need change in libs/hbb_common - https://github.com/rustdesk/hbb_common/pull/609

## What's included

- New `UsbChannel` protocol messages (device list/bind/attach/push request
  and their results, plus a generic forward-open/data/close set shared by
  both directions).
- Controlled side: lists/shares local devices and relays the USB/IP TCP
  stream for whichever devices the peer binds (`src/server/usbip_mux.rs`),
  plus a headless mirror that pulls in a pushed device
  (`src/server/usbip_pull.rs`).
- Controller side: attaches a peer's shared device
  (`src/client/usbip_attach.rs`) and shares/pushes a local device to the
  peer (`src/client/usbip_share.rs`).
- Flutter UI: a "USB forwarding" window with two sections — the peer's
  devices (Attach/Detach) and the controller's own devices (Push/Unpush) —
  each collapsed into a single combined action so the UI can't drift into
  a shared-but-not-attached/pushed state.
- Gated by the existing `enable-usbip` permission option; no new
  accept/deny dialog on the controlled side, consistent with how that side
  already has no live UI for this feature.

## Known limitation

Linux only (relies on the `usbip`/`vhci-hcd` kernel driver stack).

## Test plan
- Manual cross-testing on two Linux machines, both directions in the same
  session: attach a peer's device, share+push a local device, detach/
  unpush, and disconnect mid-session (verify no stuck/disabled buttons).
  

## Summary by CodeRabbit

* **New Features**
  * Added beta USB forwarding for Linux-to-Linux remote desktop sessions.
  * Users can share local USB devices, attach remote devices, and manage them in a dedicated multi-tab window.
  * Added peer-card actions and a Security setting to enable USB forwarding.
  * Added status, loading, error, refresh, attach/detach, and share/unshare controls.

* **Localization**
  * Added USB forwarding interface text across supported languages.

* **Security**
  * Added validation and safer handling for USB device identifiers and forwarding requests.

<h2><a href="https://app.greptile.com/api/retrigger?id=66213546"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not yet safe to merge because the previously reported failed-push path can still consume authorization before relay setup succeeds.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Asrc%2Fserver%2Fusbip_mux.rs%3A170-172%0AAn%20authenticated%20peer%20can%20keep%20sending%20%60UsbChannel%3A%3AData%60%20while%20the%20local%20USB%2FIP%20socket%20is%20blocked%20or%20draining%20more%20slowly.%20These%20frames%20are%20placed%20in%20an%20unbounded%20channel%2C%20so%20sustained%20traffic%20can%20exhaust%20the%20receiving%20RustDesk%20process's%20memory.%20The%20controller-side%20queues%20in%20%60src%2Fflutter.rs%60%20and%20the%20reverse%20relay%20in%20%60src%2Fserver%2Fusbip_pull.rs%60%20have%20the%20same%20issue.%0A%0A**How%20this%20was%20verified%3A**%20Peer-controlled%20data%20is%20sent%20into%20unbounded%20queues%20whose%20consumers%20await%20writes%20to%20local%20TCP%20sockets%2C%20with%20no%20queue%20capacity%20or%20flow-control%20window.%0A%0A%23%23%23%20Issue%202%0Asrc%2Fflutter.rs%3A1544-1558%0AThe%20pending%20push%20authorization%20is%20consumed%20before%20confirming%20that%20a%20session%20exists%20or%20that%20the%20controller%20can%20connect%20to%20its%20local%20%60usbipd%60.%20If%20either%20step%20fails%2C%20the%20authorization%20is%20not%20restored%2C%20and%20the%20peer%20returns%20from%20its%20only%20open%20attempt%20without%20sending%20a%20failed%20push%20result.%20The%20device%20therefore%20remains%20locally%20shared%2C%20the%20UI%20can%20stay%20stuck%20in%20the%20pushed%20state%2C%20and%20a%20later%20open%20for%20the%20same%20push%20is%20rejected%20as%20not%20pending.%0A%0A%23%23%23%20Issue%203%0Asrc%2Fclient%2Fusbip_attach.rs%3Aundefined-111%0AThis%20production%20regex%20initializer%20calls%20%60unwrap%28%29%60%2C%20violating%20the%20repository%20directive%20to%20avoid%20%60unwrap%28%29%60%20and%20%60expect%28%29%60%20outside%20tests%20and%20use%20explicit%20error%20handling%20instead.%20The%20same%20pattern%20appears%20at%20%60src%2Fclient%2Fusbip_attach.rs%3A116%60%20and%20%60src%2Fclient%2Fusbip_share.rs%3A27%60.%20This%20repository%20requirement%20must%20be%20satisfied%20before%20merging.%0A%0ANote%3A%20If%20this%20suggestion%20doesn't%20match%20your%20team's%20coding%20style%2C%20reply%20to%20this%20and%20let%20me%20know.%20I'll%20remember%20it%20for%20next%20time!%0A%0A%23%23%23%20Issue%204%0A.gitmodules%3Aundefined-3%0AThe%20committed%20submodule%20URL%20now%20points%20to%20a%20contributor's%20personal%20fork.%20Fresh%20checkouts%20and%20CI%20therefore%20depend%20on%20that%20account%20continuing%20to%20host%20the%20pinned%20object%3B%20if%20the%20fork%20is%20deleted%20or%20restricted%2C%20recursive%20submodule%20checkout%20and%20the%20build%20will%20fail.%20The%20required%20%60hbb_common%60%20commit%20should%20be%20hosted%20in%20the%20project-owned%20repository.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=rustdesk%2Frustdesk&pr=16278&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=7"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=7">Fix All in Claude Code</picture></a>Findings</h2>

1. P1&nbsp;Security&nbsp;**Unbounded USB relay queues** <a href="https://github.com/rustdesk/rustdesk/pull/16278#discussion_r4050370081">▶</a>
2. P1&nbsp;**Failed push loses authorization** <a href="https://github.com/rustdesk/rustdesk/pull/16278#discussion_r4052632555">▶</a>
3. P2&nbsp;**Production regex uses unwrap** <a href="https://github.com/rustdesk/rustdesk/pull/16278#discussion_r4050370098">▶</a>
4. P2&nbsp;**Submodule depends on personal fork** <a href="https://github.com/rustdesk/rustdesk/pull/16278#discussion_r4050370110">▶</a>

`````markdown
### Issue 1
src/server/usbip_mux.rs:170-172
An authenticated peer can keep sending `UsbChannel::Data` while the local USB/IP socket is blocked or draining more slowly. These frames are placed in an unbounded channel, so sustained traffic can exhaust the receiving RustDesk process's memory. The controller-side queues in `src/flutter.rs` and the reverse relay in `src/server/usbip_pull.rs` have the same issue.

**How this was verified:** Peer-controlled data is sent into unbounded queues whose consumers await writes to local TCP sockets, with no queue capacity or flow-control window.

### Issue 2
src/flutter.rs:1544-1558
The pending push authorization is consumed before confirming that a session exists or that the controller can connect to its local `usbipd`. If either step fails, the authorization is not restored, and the peer returns from its only open attempt without sending a failed push result. The device therefore remains locally shared, the UI can stay stuck in the pushed state, and a later open for the same push is rejected as not pending.

### Issue 3
src/client/usbip_attach.rs:undefined-111
This production regex initializer calls `unwrap()`, violating the repository directive to avoid `unwrap()` and `expect()` outside tests and use explicit error handling instead. The same pattern appears at `src/client/usbip_attach.rs:116` and `src/client/usbip_share.rs:27`. This repository requirement must be satisfied before merging.

Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!

### Issue 4
.gitmodules:undefined-3
The committed submodule URL now points to a contributor's personal fork. Fresh checkouts and CI therefore depend on that account continuing to host the pinned object; if the fork is deleted or restricted, recursive submodule checkout and the build will fail. The required `hbb_common` commit should be hosted in the project-owned repository.

---

`````

</details>

This PR adds Linux-to-Linux USB/IP forwarding over an existing RustDesk session, including bidirectional relay support, device lifecycle management, protocol messages, permission gating, and a dedicated Flutter interface.

- Adds pull and push USB forwarding flows using bounded relay channels.
- Integrates USB forwarding with connection negotiation, session cleanup, and elevation handling.
- Adds Linux desktop controls, multi-window support, and localization.
- Since the previous review, closes rejected pull relays explicitly and moves client USB routing logic from the shared Flutter handler into a feature-specific module.
</details>

```mermaid
sequenceDiagram
    participant UI as Controller UI
    participant C as Controller RustDesk
    participant S as Controlled RustDesk
    participant USB as usbip / USB device

    UI->>C: Attach or push device
    C->>S: UsbChannel control request
    S->>USB: Bind or attach device
    S-->>C: Open relay channel
    C-->>S: Open acknowledgement
    loop USB/IP stream
        C->>S: UsbForwardData
        S->>C: UsbForwardData
    end
    C->>S: UsbForwardClose
    S->>USB: Release relay resources
```
</details>

<sub>Reviews (17) · Last reviewed commit: ["fix(usbip): send UsbForwardClose when th..."](https://github.com/rustdesk/rustdesk/commit/47c66fe606bc44fee782724544a677efed82dc91)</sub>


rustdesk:
## Review — changes requested

I reviewed `859066b598ee7cf055d78e7cf4789fbc2e3e6faa` against `ea04f04f9df2096fccba81a133c711166fe93b28`, with `AGENTS.md` treated as repository requirements rather than style suggestions.

I do **not** think this is safe to merge yet. There are several security/correctness issues, plus a few direct `AGENTS.md` violations.

### 1. Critical: peer-controlled `bus_id` reaches a root shell

This is a real command-injection path.

Examples:

* `src/client/usbip_attach.rs::usb_attach_privileged`

  * `bus_id` ultimately comes from the remote peer's `UsbDeviceList`.
  * It is interpolated into the command passed to `run_cmds_privileged`.
* `src/server/usbip_pull.rs::usb_attach_privileged`

  * `bus_id` comes directly from peer `PushRequest`.
* `src/server/usbip_mux.rs::bind_device`

  * `bus_id` comes directly from peer `UsbBind`.

`run_cmds_privileged()` eventually executes the string through `sudo ... sh -c`, so shell metacharacters are meaningful.

For example, these are not safe:

```rust
format!("usbip {} -b {}", sub_cmd, bus_id)
```

and:

```rust
format!(
    "usbip -t {port} attach -r 127.0.0.1 -b {bus_id} && ... \
     grep -q -- '/{bus_id}$' ..."
)
```

The existing device parser already assumes Linux USB bus IDs of the form `[0-9]+-[0-9.]+`. Enforce that at every trust boundary before privileged use, ideally with one feature-local validator.

Do not rely only on quoting. Validate the semantic bus-id format and then quote/pass it safely.

This is a merge blocker.

---

### 2. Critical: the push direction exposes an insufficiently authorized raw `usbipd` connection

The controller accepts peer `UsbForwardOpen` here:

```rust
Some(Union::Open(open)) => {
    ...
    let (tx, rx) = mpsc::unbounded_channel();
    self.register_usb_share_channel(id, open.bus_id, tx);
    tokio::spawn(usbip_share::run_channel(...));
}
```

There is no requirement that this `bus_id` corresponds to a device the local user just selected with `usb_push()`.

An authenticated peer can therefore initiate an `Open` by itself and get a relay to the controller's local `127.0.0.1:3240`.

At minimum:

1. record a pending authorization when the user calls `usb_push(bus_id)`;
2. accept an `Open` only for that pending bus ID;
3. require `channel_id < 0`;
4. consume the authorization once;
5. revoke it on failure, unpush, disconnect, or session teardown.

There is a deeper issue as well: checking `open.bus_id` alone is not sufficient if multiple devices are bound locally. `run_channel()` gives the peer a **raw USB/IP socket**, and the actual USB/IP IMPORT request travelling inside that socket can name a different bus ID.

So either:

* inspect/validate the initial USB/IP import request and require it to match the authorized `bus_id`, or
* otherwise ensure the peer cannot use this raw connection to select another bound device.

The `open.bus_id` metadata currently does not constrain what `usbipd` actually serves.

This is a security boundary, not just bookkeeping.

---

### 3. Major: USB relay memory/resource usage is unbounded

The existing `PortForwardMux` already shows the required shape:

* `MAX_CHANNELS`
* `INITIAL_WINDOW`
* `CHANNEL_WINDOW`
* receive accounting
* close on window violation

The new USB implementation intentionally omits that protection:

```rust
// ... without its flow-control windowing
```

and uses unbounded channels in all peer-fed relay paths.

For example:

```rust
let (inbound_tx, inbound_rx) = mpsc::unbounded_channel();
```

followed by:

```rust
entry.inbound.send(d.data)
```

while the consumer can block indefinitely in:

```rust
writer.write_all(&chunk).await
```

This exists in:

* controlled pull: `server/usbip_mux.rs`
* controlled push: `server/usbip_pull.rs`
* controller pull: `client/usbip_attach.rs`
* controller push: `client/usbip_share.rs`

A peer that produces data faster than local `usbipd` consumes it can grow process memory without bound.

A bounded **item count alone is not sufficient** because the peer also controls each protobuf `bytes` length. The implementation should have:

* a maximum accepted `UsbForwardData` size (the implementation itself emits at most 64 KiB chunks);
* a per-channel byte budget/window;
* a maximum live channel count;
* a maximum number of concurrent/in-flight push/attach operations.

Do not drop arbitrary USB/IP chunks when full; close the channel on protocol/resource violation.

This is a merge blocker.

---

### 4. Major: push attach/close lifecycle can leak attached devices and tasks

`UsbPullState::handle_close()` removes the entry immediately:

```rust
let Some(entry) = self.channels.remove(&c.channel_id) else {
    return;
};
```

If `usbip attach` is still running, this happens:

1. Close removes the only tracked `Entry`.
2. The privileged attach continues.
3. Attach later succeeds.
4. `pull()` writes the resulting port into the detached `Arc<Mutex<Option<i32>>>`.
5. Nothing owns that port anymore, so it cannot be detached.

There is another leak when attach fails: `listener.accept()` was spawned into a detached task before the attach result is known and can remain parked indefinitely.

`close_all()` also only does:

```rust
self.channels.clear();
```

and deliberately does not detach session-owned vhci ports.

For a pushed device whose remote transport just disappeared, leaving a stale local attachment is worse than cleaning it up.

Each operation needs explicit lifecycle ownership/cancellation:

* retain the task/operation until cleanup completes;
* cancellation must be observable after each await/blocking operation;
* an attach that completes after cancellation must detach immediately;
* failed attach must cancel/drop the pending listener;
* session teardown must close relay tasks and detach all ports owned by this session.

---

### 5. Major correctness bug: `RemoteUsb` never reaches `Client::start()` as `REMOTE_USB`

The PR adds:

```rust
pub fn is_remote_usb(&self) -> bool
```

and adds `ConnType::REMOTE_USB`, but `Remote::io_loop()` still computes:

```rust
let conn_type = if self.handler.is_file_transfer() {
    ConnType::FILE_TRANSFER
} else if self.handler.is_view_camera() {
    ConnType::VIEW_CAMERA
} else if self.handler.is_terminal() {
    ConnType::TERMINAL
} else {
    ConnType::default()
};
```

There is no:

```rust
else if self.handler.is_remote_usb() {
    ConnType::REMOTE_USB
}
```

Therefore `Client::start()` sends the rendezvous/punch/relay layer a normal `DEFAULT_CONN`.

That means the new `hbb_common::ConnType::REMOTE_USB` plumbing is not actually used on this connection-establishment path.

This should be fixed and covered by a test.

---

### 6. `AGENTS.md`: creating a hidden Tokio runtime is explicitly forbidden

`src/client/usbip_attach.rs` adds:

```rust
static USB_RUNTIME: OnceLock<Option<Runtime>> = OnceLock::new();

pub(crate) fn usb_runtime() -> Option<&'static Runtime> {
    ...
    tokio::runtime::Builder::new_multi_thread()
        .worker_threads(1)
        .enable_all()
        .build()
```

`AGENTS.md` says:

> Assume a Tokio runtime already exists.
> Never create nested runtimes.
> Do not hide runtime creation inside helpers or libraries.

This implementation directly violates all three.

The comment explaining that Flutter's FFI worker has no ambient runtime explains why `tokio::spawn()` cannot be called there, but it is not an exception to the repository rule.

Route the operation into the existing session/runtime instead. Blocking `usbip` work can then use `spawn_blocking` from that runtime.

This is a repository-rule merge blocker.

---

### 7. `AGENTS.md`: feature-specific code has been pushed too far into shared abstractions

There are two particularly clear violations.

#### `InvokeUiSession`

The PR adds:

```rust
fn handle_usb_channel(&self, ch: UsbChannel);
```

to the generic `InvokeUiSession` trait.

That forces the unrelated legacy Sciter implementation to add:

```rust
fn handle_usb_channel(&self, _ch: UsbChannel) {
    // RemoteUsb support is not implemented for Sciter UI
}
```

`AGENTS.md` explicitly says that if an unrelated caller needs a placeholder solely because a shared signature was changed, the diff is too broad and should be redesigned.

This is exactly that case.

USB handling should stay Flutter/feature-specific rather than becoming a requirement of every `InvokeUiSession`.

#### `src/server/connection.rs`

The PR adds roughly 130 lines to this shared file, including USB-specific state and a large `handle_usb_channel()` router.

`AGENTS.md` explicitly names `src/server/connection.rs`:

> Call sites in shared files (`src/tray.rs`, `src/core_main.rs`, `src/server/connection.rs`, …) should be thin one-line hooks.

The unavoidable protocol/auth plumbing can remain here, but the USB-specific classification/routing/state machine should live in the USB module and be invoked through a thin hook.

---

### 8. `AGENTS.md`: the submodule bump pulls unrelated changes and points to a personal fork

The parent repository changes:

```diff
-url = https://github.com/rustdesk/hbb_common
+url = https://github.com/AzimovIz/hbb_common
```

and changes the pin from:

```text
0eb175963c950cba8d2f4f2a480f49db27b42c39
```

to:

```text
dfb43d5a6a199deaeb4b006a43bb59e49cb9bbae
```

I checked the **exact pin range**, as required by `AGENTS.md`.

`0eb1759 -> dfb43d5` is **8 commits**, not one. Besides the one-line `REMOTE_USB = 6` commit, it also pulls unrelated changes in:

* `src/lib.rs`
* `src/socket_client.rs`
* `src/stream.rs`
* `src/webrtc.rs`
* `src/websocket.rs`

Those include unrelated WebSocket/message-cap, logging, and resolver/NAT changes.

`AGENTS.md` explicitly requires:

> For submodule bumps, inspect the exact commit range and ensure unrelated changes are not being pulled into the parent PR.

So this bump does not satisfy the repository rule.

Also, the parent repository should not permanently depend on a contributor-owned fork.

The clean sequence is:

1. merge the required `hbb_common` change into the project-owned repository;
2. separately advance the parent repository's `hbb_common` baseline for unrelated upstream changes if desired;
3. rebase this PR;
4. make this PR's resulting submodule delta contain only the USB prerequisite;
5. keep `.gitmodules` pointing to `rustdesk/hbb_common`.

---

### 9. Protocol channel-ID invariants are not enforced symmetrically

The design says:

* controller pull channels: non-negative
* controlled push channels: negative

But the receivers do not consistently enforce that invariant.

On the controller:

```rust
Some(Union::Open(open)) => { ... }
```

accepts any ID. It should reject `channel_id >= 0`.

The controlled-side mux has the inverse problem: an `Open` is routed to `usbip_mux` regardless of sign, while later negative `Data`/`Close` frames are routed to `usbip_pull`.

A malicious/buggy peer can therefore create a relay task whose subsequent frames are routed to a different state machine, leaving sockets/tasks stuck.

Validate the sign at the initial `Open`, on both sides, before allocating any resource.

---

### 10. The new `Features.usbip` capability is written but not actually consumed

The server advertises:

```rust
Features {
    ...
    usbip,
}
```

but the controller does not add an equivalent `support_usbip` capability check like the existing terminal path.

The menu is shown simply for:

```dart
if (isLinux && peer.platform == kPeerPlatformLinux)
```

so every Linux peer is offered USB forwarding regardless of whether its version supports the new login/message fields.

The existing terminal implementation has an explicit `check_terminal_support()` path. Remote USB needs the same kind of compatibility handling.

This matters especially because an older peer does not know `LoginRequest.remote_usb`; unknown protobuf fields must not be allowed to silently turn a requested USB-only session into legacy/default session behavior.

---

### 11. Privileged temporary files are predictably named in a shared temp directory

Both attach implementations construct paths such as:

```rust
std::env::temp_dir().join(format!(
    "rustdesk-usbip-port-{}-{}.txt",
    std::process::id(),
    bus_id
))
```

and the root shell later performs:

```sh
usbip port > PATH
chmod 644 PATH
```

The path is predictable and is created through a privileged shell redirect.

A local unprivileged process can pre-create that path as a symlink, causing the privileged redirect to truncate/write another file.

This remains a problem even after `bus_id` command-injection validation is added.

Use a securely created temporary object/root-owned directory, or avoid the privileged temp-file handoff entirely.

---

### 12. `AGENTS.md`: production `unwrap()` violations

`AGENTS.md` allows `unwrap()` in tests and poisoned-lock situations, but not general production code.

The new code adds production regex initializers using `unwrap()`:

```text
src/server/usbip_mux.rs
src/server/usbip_pull.rs
src/client/usbip_attach.rs
src/client/usbip_share.rs
```

The lock `unwrap()` calls are within the documented exception; the regex construction calls are not.

Please remove those production `Regex::new(...).unwrap()` usages.

---

## Bot finding I would **not** act on

The CodeRabbit Russian-translation comment is not a valid finding for this PR.

The existing line:

```rust
("To start a voice call, ...", "...")
```

was not changed semantically; the diff only adds the comma required to append the new translation keys.

`AGENTS.md` explicitly says:

> Review only what the diff introduces.

So changing that existing Russian translation here would itself violate the PR scope/minimization rule.

---

## Tests

The parser/unit tests added in this PR are useful, but they do not cover the risky parts of the feature.

Before merge I would expect tests for at least:

* malformed/malicious `bus_id` rejection before every privileged operation;
* RemoteUsb actually selecting `ConnType::REMOTE_USB` in the connection path;
* rejecting wrong-sign channel IDs;
* unsolicited `Open` without a pending push authorization;
* authorized bus ID vs the actual USB/IP IMPORT bus ID;
* maximum frame size / receive-byte budget;
* maximum channel count;
* close while attach is in flight;
* attach failure while `listener.accept()` is pending;
* session teardown with an already attached vhci port.

CI also has not provided validation yet: both current GitHub Actions workflows are `action_required` and have no jobs, so they appear to be waiting for fork-workflow approval rather than having passed.

## Conclusion

The overall pull/push architecture is understandable, but this implementation currently places an authenticated network peer directly in front of privileged USB operations and local `usbipd`. That needs a much tighter trust/resource boundary.

The minimum merge gates for me are:

1. eliminate the root command-injection and temp-file issues;
2. authorize and constrain the push-side raw `usbipd` relay;
3. add frame/channel/backpressure limits;
4. make attach/pull teardown race-safe;
5. fix the missing `REMOTE_USB` connection-type plumbing;
6. remove the extra Tokio runtime;
7. reshape the shared trait / `connection.rs` changes to satisfy `AGENTS.md`;
8. clean up the `hbb_common` pin so this PR does not bundle unrelated commits or depend on a personal fork;
9. remove the production `unwrap()` violations.

After those are fixed, this needs another review focused on the resulting lifecycle and protocol boundary.

AzimovIz:
8. AGENTS.md: the submodule bump pulls unrelated changes and points to a personal fork
This is necessary for the code to build; I will update the submodule immediately after the PR https://github.com/rustdesk/hbb_common/pull/609 is approved.

AzimovIz:
For remote USB functionality to work, https://github.com/rustdesk/hbb_common/pull/609 needs to be reviewed and accepted. Please take a look at this pull request before merging.

Host requirements for RemoteUsb

Both machines need the usbip userspace tools and matching kernel modules; RustDesk doesn't bundle or reimplement any of this, it only shells out to the system usbip CLI and relays raw USB/IP protocol bytes.

Controlled side (the machine sharing/exporting a device):

```
sudo modprobe usbip-core usbip-host
sudo usbipd -D            # or: sudo systemctl enable --now usbipd
```
RustDesk's relay (src/server/usbip_mux.rs) connects to the real usbipd on 127.0.0.1:3240 as a plain TCP client and forwards bytes to/from it — it does not implement the USB/IP server protocol itself, so usbipd must already be listening before a peer can pull a device.

Controller side (the machine attaching/pulling a remote device):

```
sudo modprobe usbip-core vhci-hcd
```
No daemon needed here — usbip attach only needs the client-side kernel module (vhci-hcd) to create the virtual USB host controller the remote device attaches to.

Installing the tools:

Debian/Ubuntu: `sudo apt install usbip` (also provides usbipd; binaries under /usr/sbin)
Arch Linux: `sudo pacman -S usbip`
Without these, usbip-dependent actions fail early and visibly in RustDesk's own log (~/.local/share/logs/RustDesk/rustdesk.log), e.g. usbip: command not found if the CLI isn't installed/on PATH, or a connect timeout to 127.0.0.1:3240 if usbipd isn't running on the sharing side.

rustdesk:
# PR #16278 Review

I reviewed the current head of #16278 (`47c66fe`), the existing review comments, `hbb_common #609`, and the current RustDesk master (`12f61b1f`).

My conclusion: **this is not ready to merge yet.** I would request changes.

## 1. P0/P1: Peer-controlled `bus_id` still reaches a privileged command on the controller

This is the most serious issue I found, and the existing reviews do not appear to cover the full path.

In `src/client/usbip_channel.rs`:

```rust
Some(Union::PushResult(r)) => {
    ...
    if !r.error.is_empty() {
        ...
        let bus_id = r.bus_id.clone();
        rt.spawn_blocking(move || {
            crate::client::usbip_share::bind_device(&bus_id, false)
        });
    }
}
```

`UsbPushResult.bus_id` is supplied by the remote peer.

It then reaches `src/client/usbip_share.rs`:

```rust
pub fn bind_device(bus_id: &str, bind: bool) -> bool {
    let sub_cmd = if bind { "bind" } else { "unbind" };
    crate::platform::run_cmds_privileged(
        &format!("usbip {} -b {}", sub_cmd, bus_id)
    )
}
```

Unlike the server-side equivalents, this path does not validate `bus_id`.

A malicious controlled peer can therefore send an unsolicited failed `PushResult` with an attacker-controlled `bus_id`, causing the controller to execute a privileged unbind operation using that value.

The server-side paths already added `is_valid_bus_id()`, but the mirrored controller-side path is still missing the same protection.

At minimum:

```rust
if !is_valid_bus_id(&r.bus_id) {
    return;
}

if !handler.usb_share_pending_take(&r.bus_id) {
    // unsolicited or stale PushResult
    return;
}
```

Preferably, privileged USB commands should not be constructed as shell strings at all. Use a structured `program + argv` privileged execution API.

---

## 2. P1: The 64 KiB `UsbForwardData` check happens too late to prevent large-frame memory allocation

The PR now checks:

```rust
if d.data.len() > MAX_USB_DATA_LEN {
    ...
}
```

with `MAX_USB_DATA_LEN = 64 * 1024`.

That protects the relay queue, but it runs only **after the complete protobuf message has already been received and decoded**.

Current RustDesk master recently added connection-level packet-size hardening.

After authentication, `connection.rs` does:

```rust
self.stream.set_max_packet_length(usize::MAX);
```

For multiplexed port forwarding, it then immediately narrows the framing limit again with:

```rust
port_forward_mux::cap_packet_size(&mut self.stream);
```

and `port_forward_mux` uses approximately:

```rust
MAX_FRAME = 64 * 1024;
MAX_PACKET = 2 * MAX_FRAME;
```

RemoteUsb currently has no equivalent packet-size cap.

After rebasing this PR onto current master, an authenticated RemoteUsb peer could therefore send a very large protobuf packet. The framing layer may allocate the packet before the later `d.data.len() > 64 KiB` check rejects it.

RemoteUsb should have an equivalent connection-level packet cap, not only application-level checks after decoding.

The queue budget is also much larger than the comments imply:

```text
256 frames/channel × 64 KiB = 16 MiB/channel
256 channels = 4 GiB theoretical per connection
```

`usbip_pull` with 32 channels still permits roughly:

```text
32 × 256 × 64 KiB = 512 MiB
```

So I would also reduce the per-channel queue and/or introduce a connection-level aggregate budget.

---

## 3. P1: Session disconnect can leave USB devices bound/shared

The UI intentionally combines share+push and detach+unshare to avoid state drift, but session teardown does not preserve that invariant.

On the controlled side:

```rust
pub fn close_all(&mut self) {
    self.channels.clear();
}
```

This closes relay channels, but devices previously bound using:

```text
usbip bind
```

remain bound to `usbip-host`.

The controller push side has a similar issue:

```rust
pub(crate) fn close_usb_state(&self) {
    self.usb_forward_channels.write().unwrap().clear();
    self.usb_share_channels.write().unwrap().clear();
    self.usb_share_bus_ids.write().unwrap().clear();
    self.usb_share_pending.write().unwrap().clear();
}
```

`usb_push()` previously performs a local `usbip bind`, but disconnect cleanup only removes RustDesk bookkeeping. It does not unbind the device.

After an unexpected connection loss, the kernel/device state can therefore remain:

```text
shared = true
```

while RustDesk has forgotten that it created that state.

Cleanup should not blindly unbind every shared device, because a device might already have been shared before the RustDesk session started.

Instead, track ownership explicitly, for example:

```rust
bound_by_this_session: HashSet<String>
```

and only undo bindings created by this session.

The manual disconnect test should verify actual OS state, not only UI button state:

```bash
usbip list -p -l
usbip port
ls /sys/bus/usb/drivers/usbip-host/
```

---

## 4. Blocker: The `hbb_common` gitlink currently drops upstream history

The author's comment says the temporary fork will be replaced after `rustdesk/hbb_common#609` is accepted. That plan is fine, but the current state cannot be merged.

Current RustDesk master pins:

```text
d9895ff9...
```

while this PR pins:

```text
AzimovIz/hbb_common
3088355c...
```

Comparing those commits shows that the PR commit is:

```text
ahead: 1
behind: 10
status: diverged
```

So this is not only a `.gitmodules` ownership issue. Merging the current gitlink would effectively drop ten commits from the dependency history.

The expected sequence should be:

1. Rebase `hbb_common #609` onto the current `rustdesk/hbb_common` master.
2. Merge #609 upstream.
3. Rebase #16278 onto the current RustDesk master.
4. Restore `.gitmodules` to `https://github.com/rustdesk/hbb_common`.
5. Point the gitlink at the new upstream commit.

---

## 5. Strong should-fix / possible functional blocker: `run_cmds_privileged()` is being used with a complete shell command as one argv item

The USB implementation frequently does:

```rust
run_cmds_privileged(&format!("usbip ..."))
```

`run_cmds_privileged()` is:

```rust
pub fn run_cmds_privileged(cmds: &str) -> bool {
    crate::platform::gtk_sudo::run(vec![cmds]).is_ok()
}
```

But `gtk_sudo::child()` shell-quotes each argv element individually:

```rust
let command = args
    .iter()
    .map(|s| quote_shell_arg(s, true))
    .collect::<Vec<String>>()
    .join(" ");
```

Therefore a value such as:

```text
usbip bind -b 1-2
```

is passed as one argv element and may become something equivalent to:

```sh
sh -c "'usbip bind -b 1-2'"
```

which asks the shell to execute a command literally named:

```text
usbip bind -b 1-2
```

rather than invoking `usbip` with arguments.

This makes me doubt that the normal desktop-user elevation path for `bind`, `unbind`, `attach`, and `detach` works correctly.

Rather than adding more quoting logic, I would introduce something like:

```rust
run_privileged(program, args)
```

and avoid shell command strings entirely.

That would also eliminate the command-injection class described in issue 1.

---

## 6. Rejected push relay paths still do not send `UsbForwardClose`

In `src/client/usbip_share.rs`, the controller sends:

```rust
session.usb_reply_opened(id, true, String::new());
```

before validating the actual USB/IP `OP_REQ_IMPORT`.

After that, these paths can fail:

* import-request timeout
* connection closes before the request
* malformed request
* requested bus ID does not match the authorized bus ID
* initial write to local `usbipd` fails

Those branches currently unregister the local channel and return, for example:

```rust
session.ui_handler.unregister_usb_share_channel(id);
return;
```

but do not send:

```rust
UsbForwardClose
```

The peer has already received `Opened(success = true)`, so it believes the relay

rustdesk:
# ChatGPT 6 astra MAX

# PR #16278 RemoteUsb Review and Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal:** Make the Linux bidirectional USB forwarding feature safe to merge by fixing authorization, cross-platform builds, resource ownership, flow control, and cleanup.

**Architecture:** Keep USB state and relay behavior in feature-specific modules. Shared connection and FFI files should only perform protocol wiring, connection-type checks, event forwarding, and lifecycle hooks. Existing desktop, file-transfer, terminal, and Sciter paths must retain their behavior.

**Tech Stack:** Rust, Tokio, protobuf, Flutter/Dart, flutter_rust_bridge_codegen 1.80.1, Linux usbip and vhci-hcd.

**Spec:** [PR #16278](https://github.com/rustdesk/rustdesk/pull/16278), [AGENTS.md](../../../AGENTS.md), and the findings and acceptance criteria below.

## Global Constraints

- Change only what is required.
- Do not refactor unrelated code.
- Do not add dependencies unless needed.
- Never create nested Tokio runtimes.
- Do not hold locks across `.await`.
- Do not ignore errors silently.
- Avoid `unwrap()` and `expect()` in production code.
- Import option keys through `base::config::keys`.
- Keep client-only protocol and state in client/feature-specific modules.
- Preserve existing behavior when RemoteUsb is disabled or unsupported.
- Do not change existing non-empty translations; leave new Italian translations empty.
- Work in an isolated worktree during implementation; preserve existing user changes in the current worktree.

## Review Focus

1. A failed or unsolicited push result must never trigger privileged USB operations.
2. Linux-only USB functions must not break macOS or Windows bridge generation and builds.
3. Pre-existing or other-session USB bindings must not be detached by this session.
4. Duplicate remote bus IDs must resolve to the correct local vhci port.
5. Slow sockets, large frames, disconnects, and reconnects must have bounded memory and deterministic cleanup.

## Review Result

The PR should receive **Request changes** before merge. The reviewed head is `47c66fe606bc44fee782724544a677efed82dc91`; the reviewed base is `ea04f04f9df2096fccba81a133c711166fe93b28`.

### F1 — P1: Unsolicited `PushResult` can trigger privileged unbind

Locations: `src/client/usbip_channel.rs` and `src/client/usbip_share.rs`.

Any non-empty `PushResult.error` calls `bind_device(bus_id, false)`. Removing a pending entry does not prove that the entry existed, and the peer supplies `bus_id`. The value is interpolated into a root shell command. The message path also needs to be restricted to a RemoteUsb session.

Fix this with all three checks: validate the USB bus ID before any privileged call, require a matching session-owned push operation, and make rollback idempotent. Do not use `pending_take()` as the only authorization check because a valid `Open` may already have consumed that authorization.

### F2 — P1: Linux-only FFI `cfg` breaks non-Linux bridge builds

Locations: `src/flutter_ffi.rs` USB exports.

The bridge generator emits unconditional calls for functions hidden behind function-level Linux `cfg` attributes. A minimal reproduction with codegen 1.80.1 produces an `E0425` missing-function error on macOS.

Keep the six public `session_usb_*` symbols available on every platform and put the Linux `cfg` inside each function body. Non-Linux bodies should consume their arguments and return without invoking Linux code.

### F3 — P1: UI teardown can unbind another session’s device

Locations: `flutter/lib/desktop/pages/usbip_page.dart` and `flutter/lib/models/usbip_model.dart`.

`UsbDevice.shared` reports the whole machine’s `usbip-host` state, but page disposal treats every shared device as pushed by the current session. A device shared by the CLI or another session can therefore be unbound when this page closes.

Track OS-wide shared state separately from this session’s pushed state and from bindings created by this session. Cleanup may release only resources owned by this session.

### F4 — P1: Duplicate bus IDs select the wrong vhci port

Locations: `src/client/usbip_attach.rs` and `src/server/usbip_pull.rs`.

`parse_attached_port()` matches only the trailing bus ID. Two remote hosts can expose the same bus ID, and the parser then returns the first matching port. Detach can disconnect the other session’s device.

Match the complete loopback endpoint and bus ID, including the listener port. Update the privileged polling command to search for that complete endpoint as well.

### F5 — P1: Outbound USB data remains unbounded

Locations: all four USB relay directions, especially `src/server/usbip_mux.rs`.

The new bounded channels only protect data flowing from the peer into a local socket. Local reader tasks continue sending into the connection’s unbounded sender, so a fast USB device can outpace the network and grow memory without limit.

Add end-to-end byte credit or an equivalent send budget. Stop reading the local socket when credit is exhausted; do not add a queue that immediately drains into the existing unbounded sender.

### F6 — P1: The 64 KiB check occurs after framing

RemoteUsb does not set the connection packet cap used by PortForwardMux. The 64 KiB application check runs only after a complete protobuf packet has been received and decoded. The pinned codec accepts a complete 1 MiB packet by default.

Set a RemoteUsb framing cap on both sides, while retaining per-frame checks and an aggregate session budget. The current theoretical inbound budget of 256 channels × 256 frames × 64 KiB is too large.

### F7 — P2: Failed client attach leaves an accept task running

`usbip_attach.rs` discards the `listener.accept()` task handle. A cancelled or failed privileged attach therefore leaves a listener task in the session runtime. Keep the handle, abort and await it on failure, and detach immediately if an attach succeeds after cancellation.

### F8 — P2: Push relay rejection does not notify the peer

After sending `Opened(success = true)`, timeout, malformed import, bus-ID mismatch, and first-write failures only unregister the local channel. The peer can continue waiting for data. Every terminal path must use one close routine and send `UsbForwardClose` exactly once.

### F9 — P2: Too much USB state remains in shared `FlutterHandler`

Move USB channel maps, authorization state, retry logic, and task ownership into a USB-specific state module. Shared files should retain only thin hooks and protocol plumbing. Runtime registration must be limited to RemoteUsb sessions.

## Existing Review Comments Rechecked

- Channel ID sign checks, RemoteUsb handshake type, capability checks, hidden runtime creation, production regex `unwrap()` calls, several close paths, and runtime-round cleanup were already addressed and are not repeated here.
- The earlier claim that a complete command string is quoted as a literal executable name was not reproduced with the current helper. The remaining issue is the unvalidated command input and missing authorization described in F1.
- The exact PR submodule change from `0eb1759` to `3088355` is one commit changing one protocol file. It should still be rebased after `hbb_common#609` lands upstream, with `.gitmodules` restored to the official repository URL.

## Implementation Tasks

### Task 1: Authorize push results and isolate client USB state (F1, F9)

**Files:** Create `src/client/usbip_state.rs`; modify `src/client.rs`, `src/client/usbip_channel.rs`, `src/client/usbip_share.rs`, `src/flutter.rs`, `src/client/io_loop.rs`, and `src/ui_session_interface.rs`.

- [ ] Add failing tests proving that an unsolicited or duplicate `PushResult` does not call the privileged runner.
- [ ] Add one strict bus-ID validator. Accept IDs such as `18-1` and `1-2.3`; reject empty, malformed, overlong, or shell-containing values.
- [ ] Store operation ownership separately from one-shot `Open` authorization. Record bus ID, channel ID, whether this session created the binding, and whether cleanup has run.
- [ ] Route USB frames only when the authenticated connection is RemoteUsb.
- [ ] Move feature-specific maps and lifecycle logic into `UsbClientState`; leave thin shared hooks.
- [ ] Run the USB unit tests and the fake privileged-runner tests. Commit as `fix(usbip): authorize push results and isolate client state`.

### Task 2: Keep FFI symbols cross-platform (F2)

**Files:** Modify `src/flutter_ffi.rs`; regenerate bridge output through the repository workflow.

- [ ] Reproduce the codegen 1.80.1 failure on a non-Linux target.
- [ ] Keep every `session_usb_*` function public on all targets and put the Linux guard inside the function body:

```rust
pub fn session_usb_attach(session_id: SessionID, bus_id: String) {
    #[cfg(target_os = "linux")]
    if let Some(session) = sessions::get_session_by_session_id(&session_id) {
        session.usb_attach(bus_id);
    }
    #[cfg(not(target_os = "linux"))]
    let _ = (session_id, bus_id);
}
```

- [ ] Regenerate bridge code and run Linux, macOS, and Windows Rust checks plus `flutter analyze`. Commit as `fix(usbip): keep FFI exports available on every platform`.

### Task 3: Track session-owned devices (F3)

**Files:** Modify the USB state module, `src/client/usbip_channel.rs`, `src/server/usbip_mux.rs`, `src/server/usbip_session.rs`, `flutter/lib/models/usbip_model.dart`, and `flutter/lib/desktop/pages/usbip_page.dart`.

- [ ] Reproduce the case where the CLI or session A shares a device and session B only opens and closes the page.
- [ ] Keep OS-wide `shared` state separate from session-specific pushed state.
- [ ] Remove page-dispose cleanup that unpushes every device reported as shared.
- [ ] Release only bindings and attachments created by this session on explicit close, disconnect, and reconnect.
- [ ] Verify that session B does not unbind session A’s device and that session B’s own resources are reclaimed.

### Task 4: Match imported ports by endpoint and bus ID (F4)

**Files:** Modify `src/client/usbip_attach.rs` and `src/server/usbip_pull.rs`.

- [ ] Add regression tests with two entries such as `usbip://127.0.0.1:30001/1-2` and `usbip://127.0.0.1:30002/1-2`.
- [ ] Change the parser to accept `listener_port` and match the complete endpoint plus bus ID.
- [ ] Change privileged polling to search for the same complete endpoint.
- [ ] Test two simultaneous imports and verify that each detach affects only its own vhci port. Commit as `fix(usbip): scope imported ports to relay endpoints`.

### Task 5: Add framing limits and end-to-end flow control (F5, F6)

**Files:** Create a USB protocol-limits module; modify both client and server USB relays, the USB session state, `libs/base/protos/message.proto`, `src/client/io_loop.rs`, and `src/server/connection.rs`.

Add `UsbForwardWindowUpdate window_update = 11` to `UsbChannel`. Use shared constants such as:

```rust
pub const MAX_DATA: usize = 64 * 1024;
pub const MAX_PACKET: usize = 2 * MAX_DATA;
pub const CHANNEL_WINDOW: u32 = 256 * 1024;
pub const MIN_FRAME_CHARGE: u32 = 64;
pub const MAX_CHANNELS: usize = 32;
pub const MAX_SESSION_BUFFERED: usize = 16 * 1024 * 1024;
```

- [ ] Test packet headers at and above the framing limit.
- [ ] Apply the USB packet cap before USB messages are decoded; do not change ordinary desktop or file-transfer limits.
- [ ] Consume send credit before reading local sockets and return credit only after the destination socket drains data.
- [ ] Bound channel count, per-channel bytes, and aggregate session bytes. Close and notify the peer on violations.
- [ ] Pause a producer with a stalled network consumer, then resume it and verify byte order and completeness. Commit as `fix(usbip): bound framing and relay traffic in both directions`.

### Task 6: Make every exit path clean up exactly once (F3, F7, F8)

**Files:** Modify `src/client/usbip_attach.rs`, `src/client/usbip_share.rs`, the USB state module, `src/server/usbip_pull.rs`, and `src/server/usbip_mux.rs`.

- [ ] Keep accept and relay `JoinHandle`s, cancellation tokens, optional vhci ports, and session rounds in operation state.
- [ ] Abort and await accept tasks on failed or cancelled attach.
- [ ] If attach completes after cancellation, detach the newly created port immediately and do not report success.
- [ ] Use `tokio::select!` so EOF, write failure, explicit Close, and cancellation terminate both relay halves.
- [ ] Send one peer Close for every post-Opened rejection path, including timeout, malformed import, mismatched bus ID, and first-write failure.
- [ ] Run the race matrix: cancel-before-success, EOF-before-success, success-then-close, duplicate Close, and old-round cleanup after a new round starts. Commit as `fix(usbip): finish relay and attachment cleanup on every exit`.

### Task 7: Integrate the protocol dependency and minimize the final diff

**Files:** Modify `.gitmodules` and the `libs/hbb_common` gitlink only after `hbb_common#609` is merged upstream.

- [ ] Rebase the dependency change onto the current official `rustdesk/hbb_common` history.
- [ ] Restore `.gitmodules` to `https://github.com/rustdesk/hbb_common`.
- [ ] Confirm the final gitlink is an ancestor-preserving update and contains no unrelated dependency changes.
- [ ] Run Linux two-host tests covering both directions, missing usbipd, cancelled elevation, disconnect/reconnect, duplicate bus IDs, pre-existing bindings, denied permission, an old peer, and a stalled network.
- [ ] Run bridge generation and cross-platform checks, then `git diff --check` and a final regression-surface review. Commit as `chore(usbip): integrate upstream protocol prerequisite`.

## Regression Surface

Existing files may change only for necessary USB wiring: `src/client.rs`, `src/server.rs`, `src/client/io_loop.rs`, `src/server/connection.rs`, `src/ui_session_interface.rs`, `src/flutter.rs`, `src/flutter_ffi.rs`, the Flutter connection/window files, `libs/base/protos/message.proto`, `libs/base/src/config/keys.rs`, localization files, `.gitmodules`, and the `hbb_common` gitlink. Existing desktop, file-transfer, terminal, Sciter, and non-USB protocol paths must remain unchanged.

## Verification Already Performed

- The PR diff, metadata, inline comments, and author replies were reviewed.
- An isolated Rust harness ran 15 tests covering the current codec, coordinator, privileged-runner path, endpoint parser, sender queue, and framing behavior. These tests reproduce current behavior; they do not mean the PR is fixed.
- A minimal bridge project using codegen 1.80.1 reproduced the non-Linux missing-function error.
- No real privileged command, USB bind/attach, full RustDesk build, or two-host Linux USB test was run. Those are required during implementation.

PR #16303, opened 2026-09-21
fix(ffmpeg): bound QSV encoder retries

21pages:
Persistent Intel SDK warnings can leave QSV encoding stuck in submission or synchronization loops. Limit submission to 100 retries and synchronization to 10 retries, returning `AVERROR_EXTERNAL` when the limit is reached. Keep timed-out synchronization tasks in the FIFO so their buffers remain alive until a later synchronization or encoder close.

On Windows, wait 1 ms between retryable submission warnings (`MFX_WRN_DEVICE_BUSY` and `MFX_WRN_IN_EXECUTION`). The original `av_usleep(500)` becomes `Sleep(0)` in the Windows sleep backend and can exhaust 100 retries before a 1–5 ms busy interval ends. Non-Windows submission waits and existing negative SDK-result handling are preserved.

Validation on Windows 11 / Intel UHD 730 with FFmpeg 7.1.1:

- 42 fault-injection/control cases covering 1 ms and 5 ms recovery, permanent warnings, retry boundaries, synchronization-timeout recovery, drain, and close cleanup. Before the Windows wait change, short injected busy intervals dropped a frame; after the change all frames encode. These are injected warning scenarios, not a reproduced driver hang.
- H.264/HEVC real encoding: 120 frames per codec before and after, with byte-identical output. Four concurrent 1080p encoding runs at async depths 1/4 completed 2400 frames with no errors.
- Both 120-frame outputs also passed a separate QSV hardware decoder with zero decode errors; the RustDesk static libraries do not include QSV decoders.
- No premature frees or remaining tracked output buffers after close in the tested failure paths. Both patch versions apply cleanly to official FFmpeg 7.1.1, and selecting the non-Windows source path produces the previous implementation exactly.

Regression surface: the vcpkg patch registration and FFmpeg's shared QSV encoder submission/synchronization paths. The Windows wait correction is confined to submission warnings in `qsvenc.c`; the synchronization retry limit and buffer cleanup implementation are unchanged by that correction.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved QSV hardware video encoding reliability by preventing indefinite waits during frame submission and synchronization.
  - Added bounded retry handling for busy or still-processing encoder operations.
  - Reports an encoding error when retry limits are exceeded, avoiding stalled encoding sessions.

<h2><a href="https://app.greptile.com/api/retrigger?id=66914086"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge, with the change confined to bounding previously unbounded QSV encoder waits.

This PR adds bounded retry handling to the vendored FFmpeg QSV encoder patch set and registers the new patch in the vcpkg port.

- Limits QSV frame-submission retries before returning `AVERROR_EXTERNAL`.
- Limits synchronization retries while retaining the queued packet for a later attempt.
- Preserves existing waits and negative SDK-result handling.
- Regression surface is limited to FFmpeg QSV encoding; non-QSV codec paths and other dependencies are unchanged.
</details>

```mermaid
flowchart TD
    A[Submit QSV frame] --> B{Submission result}
    B -->|Success or terminal result| C[Continue encoding]
    B -->|Device busy or in execution| D{Retry limit reached?}
    D -->|No| A
    D -->|Yes| E[Return AVERROR_EXTERNAL]
    C --> F[Peek oldest async packet]
    F --> G{Synchronization result}
    G -->|In execution| H{Sync retry limit reached?}
    H -->|No| G
    H -->|Yes| I[Return error and retain FIFO entry]
    I --> F
    G -->|Completed or negative result| J[Drain FIFO entry]
    J --> K[Produce packet using existing handling]
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(ffmpeg): bound QSV encoder retries"](https://github.com/rustdesk/rustdesk/commit/7c7ea373a1a0f8520181b7eceb631eca1eb6202e)</sub>


21pages:
Reviewed the current head **`710c80fae5cafa4248deca89cb499404de9b72c4`** of rustdesk/rustdesk#16303. I found **one issue I’d fix before merging**, plus one timeout/recovery concern.

### 1. Important: `q->packet_size` is still stale after RustDesk changes QSV bitrate

In `0014-fix-qsv-encode-buffer.patch`, the 4× HEVC capacity is calculated only once, immediately after the initial `MFXVideoENCODE_GetVideoParam()`:

`res/vcpkg/ffmpeg/patch/0014-fix-qsv-encode-buffer.patch:7-14`

```c
q->packet_size = q->param.mfx.BufferSizeInKB *
                 q->param.mfx.BRCParamMultiplier * 1000;

if (avctx->codec_id == AV_CODEC_ID_HEVC) {
    int64_t packet_size = (int64_t)q->packet_size * 4;
    ...
    q->packet_size = packet_size;
}
```

But this is particularly relevant to RustDesk because it **changes QSV bitrate dynamically without recreating the encoder**:

`src/server/video_service.rs:1429-1433`

```rust
if *ratio != video_qos.ratio() {
    *ratio = video_qos.ratio();
    if encoder.support_changing_quality() {
        allow_err!(encoder.set_quality(*ratio));
```

and `HwRamEncoder::set_quality()` calls:

```rust
self.encoder.set_bitrate(bitrate as _)
```

FFmpeg's `update_bitrate()` updates `q->param` and subsequently calls:

```c
MFXVideoENCODE_Reset(q->session, &q->param);
```

but **never recalculates `q->packet_size`**.

This means a session created at a low/balanced bitrate can later increase quality substantially and continue allocating packets using the capacity calculated for the old bitrate. RustDesk makes this more than a theoretical corner case: custom image quality is allowed up to `BR_MAX = 40.0`.

So the same `MFX_ERR_NOT_ENOUGH_BUFFER` that this PR is trying to eliminate can still occur after an in-session bitrate increase.

**Suggested fix:** factor the HEVC packet-size calculation into a helper and recalculate after a successful QSV reset. Ideally re-run `MFXVideoENCODE_GetVideoParam()` after `MFXVideoENCODE_Reset()` and calculate the new capacity from the SDK's post-reset values. I would probably retain the larger of the previous and newly calculated capacity rather than shrinking it.

This also makes the PR description's current caveat—

> “arbitrary later bitrate increases are not guaranteed to fit”

—unnecessary for RustDesk's normal adaptive-quality path.

### 2. Recovery can still block an interactive session for ~11 seconds per sync failure

`0013` uses:

```c
#define QSV_MAX_SYNC_RETRIES 10
...
MFXVideoCORE_SyncOperation(..., 1000);
```

The initial call plus 10 retries means **up to approximately 11 seconds** inside one `ff_qsv_encode()` call. The PR's own validation confirms roughly **11.006–11.009 seconds**.

This matters more in RustDesk because the pinned `hwcodec` implementation explicitly configures QSV with:

```cpp
av_opt_set(priv_data, "async_depth", "1", 0)
```

so synchronization is directly on the critical encoding path.

RustDesk does eventually switch away from the hardware encoder after three consecutive encode errors, which prevents indefinite FIFO growth, but depending on whether later submissions succeed, recovery could involve more than one long sync wait.

I wouldn't necessarily block the PR solely on this, but for an interactive remote-desktop application I'd strongly consider a substantially shorter **total wall-clock synchronization deadline** rather than eleven 1-second waits. Something around 1–3 seconds total would still handle transient GPU latency while recovering much sooner from a genuinely stuck task.

### Everything else

The FIFO ownership change looks correct to me: `av_fifo_peek()` followed by draining only after successful/non-negative synchronization avoids freeing buffers that the SDK may still own. The negative-sync check also correctly prevents reading `FrameType`, timestamps, and `DataLength` after a failed `SyncOperation`.

The encode retry counter is also correct: `QSV_MAX_ENCODE_RETRIES = 100` gives 101 total calls including the initial attempt. The Windows 1 ms sleep makes sense given the observed `Sleep(0)` behavior for the previous 500 µs delay.

Current CI is clean: both **CI** and **Full Flutter CI** passed on `710c80f`, including the x86_64 Windows build. There are currently no human review threads; CodeRabbit reported no actionable comments.

**My main requested change would be #1: recalculate the HEVC output capacity after QSV bitrate resets.** The rest of the patch is reasonably well structured and unusually well validated.

PR #16306, opened 2026-09-22
fix(http): reuse http client instances to prevent socket exhaustion

### LINKED ISSUE #16195: hbbs rebuilds its HTTP client (full CA-store parse) for every client heartbeat when calling its own `/api/heartbeat`
### Bug Description

We run rustdesk-server-pro (licensed) in Docker with host networking for about 85 devices, and noticed hbbs sitting at a quarter of a CPU core doing nothing but idle heartbeats. hbbr, nginx, everything else on the box is asleep. CPU grows linearly with the number of registered devices, roughly 0.3% of a core per device, which puts a 1-vCPU server at its limit somewhere around 300 devices. We saw this on 1.8.4 and it's unchanged on 1.8.6.

- All the CPU is in one tokio worker thread, almost entirely user-space. The sqlite threads are idle and there's basically no disk I/O, so it isn't db work.
- Every time a client sends its UDP heartbeat to :21116, hbbs opens a new TCP connection to `127.0.0.1:21114` and POSTs `/api/heartbeat` to itself with `{id, uuid, ver, modified_at}`. New connection every time, no keep-alive, so a few hundred sockets sit in TIME-WAIT constantly.
- The handler is not the problem. We replayed that exact POST on loopback, including for a real registered device with its real uuid, and it comes back in about half a millisecond with no measurable CPU.
- What is expensive is *making the call*. hbbs reads about 116 KB in ~30 read syscalls per heartbeat, and if you sample `/proc/<pid>/fd` you catch `/etc/ssl/certs/ca-certificates.crt` (182 KB) open about a third of the time. libssl/libcrypto are mapped into the process.

So it looks like the heartbeat path builds a brand-new HTTP client per call, which spins up a TLS context and parses the entire system CA bundle, in order to send a plaintext request to itself over loopback. Parsing ~140 certificates per heartbeat seems inline with the ~48 ms per heartbeat.

### How to Reproduce

To reproduce, you just need a handful of online clients: watch `top -H` on hbbs (one hot thread), `ss -tanp | grep 21114` (hbbs owning short-lived client sockets to itself), and loop `ls -l /proc/$(pgrep hbbs)/fd` for a few seconds and the CA bundle keeps showing up.

### Expected Behavior

The fix seems straightforward, guessing at code, build that client once and reuse it, or skip TLS setup entirely since the target is plaintext 127.0.0.1. Reusing the connection would clean up the TIME-WAIT churn too.  Our own loopback test showed ~100x improvement in response and CPU time. Another option, but more complicated, might be to skip the network stack all together with sockets (more complicated OS compat). Happy to test a build if that helps.

Rocky Linux 10

1.8.4 -> 1.8.6

### Screenshots

-

### Additional Context

_No response_

AnandkumarMall:
Fixes #16195

### Summary of Changes
- Previously, \create_http_client\ and \create_http_client_async\ instantiated a new \
eqwest::Client\ / \
eqwest::blocking::Client\ on every invocation.
- Creating a new client for each request prevents connection pooling and can lead to socket/file descriptor exhaustion and TLS negotiation overhead under high request frequency.
- This PR introduces client caching via \lazy_static\ and thread-safe \Mutex\, caching the client based on \(TlsType, danger_accept_invalid_cert, proxy_conf)\. If the configuration matches, the cached client is reused.

## Summary by CodeRabbit

* **Performance**
  * HTTP connections are now reused more efficiently for matching synchronous and asynchronous configurations.
  * Separate connection reuse is maintained for synchronous and asynchronous requests.
  * Changes to TLS, certificate validation, or proxy settings continue to create appropriately configured connections.
  * Requests can still proceed normally if connection reuse is temporarily unavailable.

<h2><a href="https://app.greptile.com/api/retrigger?id=67341310"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

This PR should not merge until the cache can retain clients for multiple active configurations and the explicit lock-error handling requirement is satisfied.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Asrc%2Fhbbs_http%2Fhttp_client.rs%3A18-19%0AEach%20cache%20retains%20only%20one%20client%20configuration.%20Different%20servers%20can%20have%20different%20URL-specific%20TLS%20backends%20or%20certificate%20policies%2C%20so%20alternating%20requests%20to%20those%20servers%20evict%20each%20other's%20client.%20Because%20the%20URL%20paths%20obtain%20a%20client%20for%20each%20operation%2C%20this%20creates%20a%20fresh%20connection%20pool%20on%20every%20request%20and%20leaves%20the%20socket-exhaustion%20problem%20unresolved%20for%20mixed-endpoint%20traffic.%20Please%20retain%20a%20client%20for%20each%20configuration%20key.%0A%0A%23%23%23%20Issue%202%0Asrc%2Fhbbs_http%2Fhttp_client.rs%3A110-121%0ALookup%20and%20insertion%20use%20separate%20critical%20sections%2C%20and%20client%20construction%20happens%20after%20releasing%20the%20mutex.%20Concurrent%20requests%20that%20miss%20the%20same%20cold%20or%20invalidated%20entry%20can%20all%20build%20and%20use%20separate%20clients%20and%20connection%20pools%2C%20while%20only%20the%20last%20client%20is%20retained.%20Existing%20request%20paths%20run%20concurrently%2C%20so%20startup%20or%20configuration-change%20bursts%20can%20still%20create%20unnecessary%20sockets.%20The%20async%20cache%20at%20lines%20132%E2%80%93143%20has%20the%20same%20race.%0A%0A%23%23%23%20Issue%203%0Asrc%2Fhbbs_http%2Fhttp_client.rs%3A109-119%0AThe%20cache%20key%20captures%20%60Config%3A%3Aget_socks%28%29%60%20here%2C%20but%20%60configure_http_client!%60%20reads%20the%20global%20SOCKS%20configuration%20again%20when%20building%20the%20client.%20A%20runtime%20SOCKS%20update%20between%20those%20reads%20can%20store%20a%20client%20configured%20for%20one%20proxy%20or%20direct%20mode%20under%20another%20configuration's%20key.%20If%20that%20key%20becomes%20current%20again%2C%20later%20requests%20can%20reuse%20the%20wrongly%20routed%20client.%20Pass%20the%20captured%20snapshot%20into%20client%20construction%20so%20the%20key%20always%20describes%20the%20cached%20client.%20The%20async%20path%20at%20lines%20131%E2%80%93141%20has%20the%20same%20issue.%0A%0A%23%23%23%20Issue%204%0Asrc%2Fhbbs_http%2Fhttp_client.rs%3A110-121%0ABoth%20cache%20operations%20silently%20discard%20mutex-poisoning%20errors%20through%20%60if%20let%20Ok%28...%29%60%2C%20including%20lookup%20and%20insertion.%20The%20repository%20directive%20says%20not%20to%20ignore%20errors%20silently%2C%20so%20poisoning%20must%20be%20recovered%20from%2C%20logged%2C%20or%20otherwise%20surfaced%20before%20this%20can%20merge.%20The%20async%20cache%20repeats%20this%20pattern%20at%20lines%20132%E2%80%93143.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=rustdesk%2Frustdesk&pr=16306&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=7"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=7">Fix All in Claude Code</picture></a>Findings</h2>

1. P1&nbsp;**Single\-entry caches prevent reuse** <a href="https://github.com/rustdesk/rustdesk/pull/16306#discussion_r4069046362">▶</a>
2. P2&nbsp;**Concurrent misses duplicate clients** <a href="https://github.com/rustdesk/rustdesk/pull/16306#discussion_r4069046364">▶</a>
3. P2&nbsp;**Proxy snapshot can mismatch** <a href="https://github.com/rustdesk/rustdesk/pull/16306#discussion_r4069046368">▶</a>
4. P2&nbsp;**Lock errors are ignored** <a href="https://github.com/rustdesk/rustdesk/pull/16306#discussion_r4069046374">▶</a>

`````markdown
### Issue 1
src/hbbs_http/http_client.rs:18-19
Each cache retains only one client configuration. Different servers can have different URL-specific TLS backends or certificate policies, so alternating requests to those servers evict each other's client. Because the URL paths obtain a client for each operation, this creates a fresh connection pool on every request and leaves the socket-exhaustion problem unresolved for mixed-endpoint traffic. Please retain a client for each configuration key.

### Issue 2
src/hbbs_http/http_client.rs:110-121
Lookup and insertion use separate critical sections, and client construction happens after releasing the mutex. Concurrent requests that miss the same cold or invalidated entry can all build and use separate clients and connection pools, while only the last client is retained. Existing request paths run concurrently, so startup or configuration-change bursts can still create unnecessary sockets. The async cache at lines 132–143 has the same race.

### Issue 3
src/hbbs_http/http_client.rs:109-119
The cache key captures `Config::get_socks()` here, but `configure_http_client!` reads the global SOCKS configuration again when building the client. A runtime SOCKS update between those reads can store a client configured for one proxy or direct mode under another configuration's key. If that key becomes current again, later requests can reuse the wrongly routed client. Pass the captured snapshot into client construction so the key always describes the cached client. The async path at lines 131–141 has the same issue.

### Issue 4
src/hbbs_http/http_client.rs:110-121
Both cache operations silently discard mutex-poisoning errors through `if let Ok(...)`, including lookup and insertion. The repository directive says not to ignore errors silently, so poisoning must be recovered from, logged, or otherwise surfaced before this can merge. The async cache repeats this pattern at lines 132–143.

---

`````

</details>

This PR adds process-wide synchronous and asynchronous reqwest client caches keyed by TLS backend, invalid-certificate policy, and SOCKS configuration, returning cloned handles to reuse connection pools.
- The changed runtime surface includes every sync and async HTTP factory caller.
- The single-entry design still recreates pools when requests alternate between valid configurations.
- Concurrent cache misses are not coordinated, and proxy configuration is read twice while constructing each entry.
- Mutex-poisoning errors are silently discarded contrary to repository guidance.
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix(http): reuse http client instances t..."](https://github.com/rustdesk/rustdesk/commit/e68a00015b779f1dfafc2f2d041ac22becbf3e4e)</sub>


rustdesk:
We will do it ourselves.

AnandkumarMall:
Addressed review feedback:
- Migrated cache to HashMap<ClientCacheKey, ...> so multiple configurations can coexist without eviction.
- Made cache lookup and creation atomic under the mutex to prevent concurrent misses from duplicating clients.
- Captured proxy_conf snapshot and passed it directly into client builder to prevent proxy setting drift.
- Handled mutex poisoning explicitly by recovering and logging error.

